PR #15, opened 2017-01-12
Make tasks markdown checkable.

vic:
Also added a bit of instructions on creating a new issue when someone picks a task. The idea is we can use README to keep low hanging fruit available for people. Another option (maybe later, but much less visible)  would be to create issues for each task and tag them as unasigned, by complexity and such. 

By now I guess it's a good idea to have them (while they are few) discoverable on the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,46 +346,47 @@ Pull requests welcome, no copyright assignment required. This project is under t
 
 ## Help Needed
 
-There's lots to do!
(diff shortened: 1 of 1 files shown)


Wilfred:
Looks good, thanks. 

Agreed, I imagine we'll reach a point where we just use GitHub issues for good first bugs. Still, at the moment the purpose of the readme is to make contributing seem straightforward and interesting :)

PR #24, opened 2017-01-12
Change welcome message to 'Remacs'

mrwacky42:
I haven't even gotten to the Rust part yet..

Not sure how deep you want to go replacing `Emacs` with `Remacs`, but this is a start.

### DIFF
--- a/lisp/startup.el
+++ b/lisp/startup.el
@@ -1459,9 +1459,9 @@ If this is nil, no message will be displayed."
 (defconst fancy-startup-text
   `((:face (variable-pitch font-lock-comment-face)
      "Welcome to "
-     :link ("GNU Emacs"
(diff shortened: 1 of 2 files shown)


lisp/startup.el
```diff
@@ -1918,7 +1918,7 @@ splash screen in another window."
       ;; The convention for this piece of code is that
       ;; each piece of output starts with one or two newlines
       ;; and does not end with any newlines.
-      (insert (if startup "Welcome to GNU Emacs" "This is GNU Emacs"))
+      (insert (if startup "Welcome to GNU Remacs" "This is GNU Remacs"))
```
Wilfred: Remacs is not a GNU project, so we shouldn't say GNU here.

mrwacky42: I thought you might say that.
*fixed*

Wilfred:
Marvellous, thank you :)

PR #31, opened 2017-01-12
Fix LispMisc type.

jeandudey:
There are a couple of errors when compiling for a 32 bits architecture.
This commit solves these errors.

PR #32, opened 2017-01-12
Run on both 32 and 64 bit architectures

mjs2600:


mjs2600:
I think this addresses the issue with #31.

PR #35, opened 2017-01-13
Properly define the emacs integer types.

crlf0710:
Hi, according to lisp.h, this should be the correct definition of the emacs integer types. (Auto generated from build.rs)

With or without the patch, remacs still crashes during the bootstrapping phrase on my Windows machine. Will investigate more about this later.


crlf0710:
Currently on Windows/Mingw64 the build terminates the same output as #38 .

Wilfred:
Thanks :)

PR #39, opened 2017-01-13
Ports the atom function to Rust

mfeckie:
Converts the `atom` function to Rust.

This required the use of `CONSP`, so I made it public in `cons.rs`, not sure if this is a reasonable thing to do.

Look forward to feedback

### DIFF
--- a/rust_src/src/atom.rs
+++ b/rust_src/src/atom.rs
@@ -1,0 +1,20 @@
+use std::os::raw::c_char;
+use std::ptr;
+
+extern crate libc;
+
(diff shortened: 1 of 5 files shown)


mfeckie:
Post rebasing to match master, I encountered this issue

```
   Compiling remacs v0.1.0 (file:///home/mfeckie/rust/remacs/rust_src)
error: the crate `libc` is compiled with the panic strategy `abort` which is incompatible with this crate's strategy of `unwind`
```

So I had to bump the libc version to resolve.

PR #46, opened 2017-01-13
Remove MSDOS code.

jeandudey:
This is a work in progress, but it can be reviewed.


Wilfred:
There are some DOS references in INSTALL that we can remove too.

jeandudey:
@Wilfred this now can be merged, i think removing dos code from emacs lisp requires a bit more of work and testing and is more easy to make another PR.

Wilfred:
Groovy. Agreed that smaller PRs are easier to work with.

PR #48, opened 2017-01-14
Correct path for `make check`

mfeckie:
The current setup doesn't correctly run the tests and the Travis file causes normal exit even if tests fail.

This just sets the correct path and ensure that `make clean` will fail if tests fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
   - ./configure --without-makeinfo --with-xpm=no --with-gif=no
   - make -j 3
 
-  - make check || echo 'make check failed'
+  - make check
 
 notifications:
   fast_finish: true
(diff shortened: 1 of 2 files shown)


mfeckie:
Damn, so, this works on my local build, but fails on Travis.... 

Example error .

```
(error "Can't find library /home/travis/build/wilfred/remacs/lisp/progmodes/elisp-mode.el")
```

Not sure why that would be.  Hopefully someone cleverer than me will look and know :wink:

Wilfred:
Yep, I deliberately wrote the Travis file like that. I wanted to run 'make check' on travis, even though some of them are failing. Once we get the ERT tests working (presumably we're missing some dependencies) then we can require 'make check' to work.

mfeckie:
👍

PR #49, opened 2017-01-14
Use LispObject as a newtype of EmacsInt. 

crlf0710:
This make it possible to add our own methods.

I've added a few helper functions. Haven't added corresponding tests and change the subrs to actually use them though.

### DIFF
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -10,7 +10,6 @@ use std::os::raw::c_char;
 #[cfg(test)]
 use std::cmp::max;
 use std::mem;
-use std::ptr;
 use std::ops::Deref;
 
 use marker::{LispMarker, marker_position};
@@ -54,8 +53,6 @@ pub const EMACS_LISP_FLOAT_SIZE: EmacsInt = 8;
 #[repr(C)]
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
 pub struct LispObject(EmacsInt);
-// TODO: set CHECK_LISP_OBJECT_TYPE and use a struct here, as it would
-// give us stronger guarantees from the type checker.
 
 extern "C" {
     pub fn wrong_type_argument(predicate: LispObject, value: LispObject) -> LispObject;
@@ -143,13 +140,13 @@ pub enum LispType {
 
 impl LispObject {
     #[allow(unused_unsafe)]
-    pub unsafe fn get_type_unchecked(self) -> LispType {
+    pub fn get_type(self) -> LispType {
         let res = (self.to_raw() & !VALMASK) as u8;
         unsafe { mem::transmute(res) }
     }
 
     #[inline]
-    pub unsafe fn get_untaggedptr_unchecked(self) -> * mut libc::c_void {
+    pub fn get_untaggedptr(self) -> * mut libc::c_void {
         (self.to_raw() & VALMASK) as libc::intptr_t as * mut libc::c_void
     }
 }
@@ -223,11 +220,11 @@ fn test_lisp_misc_any_size() {
 
 impl LispObject {
     pub fn is_misc(self) -> bool {
-        unsafe { self.get_type_unchecked() == LispType::Lisp_Misc }
+        self.get_type() == LispType::Lisp_Misc
     }
 
     pub unsafe fn to_misc_unchecked(self) -> LispMiscRef {
-        LispMiscRef::new(mem::transmute(self.get_untaggedptr_unchecked()))
+        LispMiscRef::new(mem::transmute(self.get_untaggedptr()))
     }
 
     pub unsafe fn get_misc_type_unchecked(self) -> LispMiscType {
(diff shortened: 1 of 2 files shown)


crlf0710:
Hi, I'm actually trying to make the code more rusty, while not breaking the old conventions at this stage.

Just uploaded all of them to see what you think. You can cherry-pick or merge the ones you think is ok in the overall direction.

rust_src/src/lisp.rs
```diff
@@ -42,11 +43,17 @@ pub type EmacsInt = isize;
...
+#[cfg(dummy = "impossible")]
+pub const EMACS_LISP_FLOAT_SIZE: EmacsInt = 8;
 
 // This is dependent on CHECK_LISP_OBJECT_TYPE, a compile time flag,
 // but it's usually false.
-pub type LispObject = EmacsInt;
+#[repr(C)]
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+pub struct LispObject(EmacsInt);
 // TODO: set CHECK_LISP_OBJECT_TYPE and use a struct here, as it would
```
Wilfred: Now `LispObject` is a struct I think we can remove this todo.

rust_src/src/lisp.rs
```diff
@@ -130,13 +141,31 @@ pub enum LispType {
     Lisp_Float = 7,
 }
 
+impl LispObject {
+    #[allow(unused_unsafe)]
+    pub unsafe fn get_type_unchecked(self) -> LispType {
```
Wilfred: Should this be an unsafe function? I believe it's safe: you can't pass it a value that will break memory safety.

rust_src/src/lisp.rs
```diff
@@ -130,13 +141,31 @@ pub enum LispType {
...
+        let res = (self.to_raw() & !VALMASK) as u8;
+        unsafe { mem::transmute(res) }
+    }
+
+    #[inline]
+    pub unsafe fn get_untaggedptr_unchecked(self) -> * mut libc::c_void {
```
Wilfred: Likewise here: whilst dereferencing the pointer would be unsafe, I think is safe too, unless I'm missing something.

rust_src/src/lisp.rs
```diff
@@ -147,19 +176,206 @@ pub enum LispMiscType {
...
 #[allow(dead_code)]
-const MOST_NEGATIVE_FIXNUM: EmacsInt = (-1 - MOST_POSITIVE_FIXNUM);
+pub const MOST_NEGATIVE_FIXNUM: EmacsInt = (-1 - MOST_POSITIVE_FIXNUM);
 
-// This is also dependent on USE_LSB_TAG, which we're assuming to be 1.
-const VALMASK: EmacsInt = -(1 << GCTYPEBITS);
+
+// Lisp_Int0 (natnum)
+
+/// Natnums are inline integers that is great or equal to 0 and fit directly into
```
Wilfred: Minor nit: 'that are greater or equal'

Is this true? I'd thought Lisp_Int0 could be negative too.

rust_src/src/lisp.rs
```diff
@@ -246,118 +462,182 @@ macro_rules! defun {
 /// of arguments.
 pub const MANY: i16 = -2;
 
-/// Convert a LispObject to an EmacsInt.
-#[allow(non_snake_case)]
-fn XLI(o: LispObject) -> EmacsInt {
-    o as EmacsInt
-}
+mod deprecated {
```
Wilfred: We should add a comment here to clarify what we mean by deprecated. I think we mean functions that we've written to make porting easy, but we should be able to remove once the relevant functionality is Rust-only. Is that correct?

rust_src/src/lisp.rs
crlf0710: Yes, i agree.

rust_src/src/lisp.rs
crlf0710: Yes, i think so.

rust_src/src/lisp.rs
crlf0710: It is safe as long as every number in usize is a valid pointer. In casual cases it's always true. I'm not sure if there's corner cases. I'm remove the unsafe anyway.

rust_src/src/lisp.rs
crlf0710: That's what i mean. I'll add some comment here.

rust_src/src/lisp.rs
crlf0710: The sentence alone is correct (natnum = natural number) but it didn't explain the whole thing correct. And natnum doesn't have to be in Lisp_Int0 in LSB mode. I was wrong here :(

I'll add definition for fixnum and fix this comment. fixnum operations will be in TODO.
(The original code is limiting and confusing... maybe because of the LSB and MSB difference.
Might fix this later when more things are settled)

crlf0710:
I've made the changes. PTAL.

Wilfred:
Great :)

PR #51, opened 2017-01-14
[WIP] Feature/listp

mfeckie:
This follows on from the `feature/atom` branch, so some of the changes here will disappear once (if) that gets merged.

Had to bump libc because of compiler complaints around panic vs abort


src/lisp.h
```diff
@@ -1276,6 +1276,9 @@ CDR_SAFE (Lisp_Object c)
 Lisp_Object Fsetcar(Lisp_Object, Lisp_Object);
 Lisp_Object Fsetcdr(Lisp_Object, Lisp_Object);
 Lisp_Object Fcar(Lisp_Object);
+Lisp_Object Fatom(Lisp_Object, Lisp_Object);
+Lisp_Object Flistp(Lisp_Object, Lisp_Object);
+Lisp_Object Fnlistp(Lisp_Object, Lisp_Object);
```
Wilfred: These are functions of one argument, which is why the compile is failing on Travis.

rust_src/src/list.rs
```diff
@@ -0,0 +1,37 @@
+use std::os::raw::c_char;
```
Wilfred: Since lists are just collections of cons cells, I propose that we move these functions into cons.rs. We could move `Fatom` into cons.rs too.

rust_src/src/list.rs
```diff
@@ -0,0 +1,37 @@
...
+        Qnil
+    }
+}
+
+defun!("listp",
+       Flistp,
+       Slistp,
+       1, 1,
+       ptr::null(),
+       "Return t if OBJECT is a list, that is, a cons cell or nil. Otherwise, return nil.");
```
Wilfred: It's probably worth adding `(fn OBJECT)` at the end of the docstring, or `*Help*` will just show `(listp ARG)`.

0xAX:
Maybe also will be better to move `Fatom` and `Fnull` (https://github.com/Wilfred/remacs/blob/master/rust_src/src/strings.rs#L39) to the `list.rs` as both of these functions related to lists and described as `Predicates on Lists` in [documentation](https://www.gnu.org/software/emacs/manual/html_node/elisp/List_002drelated-Predicates.html)?

Wilfred:
@0xAX aha, I suggested something similar :). Since we're organising functions by type, I think we should either have a cons.rs or a list.rs, but not both. I don't feel strongly as to which.

0xAX:
> Since we're organising functions by type, I think we should either have a cons.rs or a list.rs, but not both

fully agree.

> I don't feel strongly as to which.

as for me - +1 for list.rs :)

mfeckie:
@Wilfred Finally got back to this, sorry for the delay.  Wanted to make sure everything was up to date with master.

Travis is now green

PR #54, opened 2017-01-14
Rust tutorial: exposing functions to the C code

Wilfred:
This has caused some confusion, so I thought it would be worth showing an example.


dpzmick:
This is really useful. I just went through the steps as given here and they worked, so probably good to merge.

PR #59, opened 2017-01-14
add implementation of XFASTINT()

0xAX:
@Wilfred now it is marked as `dead_code`, but I'm going to implement char fns, so it will be useful there


Wilfred:
Investigating this some more:

```c
#if DEFINE_KEY_OPS_AS_MACROS // false
# if USE_LSB_TAG // true
#  define XFASTINT(a) lisp_h_XFASTINT (a)
# endif
#endif

#if USE_LSB_TAG
# define lisp_h_XFASTINT(a) XINT (a)
#endif


#if USE_LSB_TAG // true


INLINE EMACS_INT
(XFASTINT) (Lisp_Object a)
{
  EMACS_INT n = lisp_h_XFASTINT (a);
  eassume (0 <= n);
  return n;
}

// ...

#else /* ! USE_LSB_TAG */


/* Like XINT (A), but may be faster.  A must be nonnegative.
   If ! USE_LSB_TAG, this takes advantage of the fact that Lisp
   integers have zero-bits in their tags.  */
INLINE EMACS_INT
XFASTINT (Lisp_Object a)
{
  EMACS_INT int0 = Lisp_Int0;
  EMACS_INT n = USE_LSB_TAG ? XINT (a) : XLI (a) - (int0 << VALBITS);
  eassume (0 <= n);
  return n;
}
```

There's a lot going on here, but I think we're doing the right thing.

0xAX:
Yes, that's why I've left tge comment about USE_LSB_TAG

rust_src/src/lisp.rs
```diff
@@ -506,3 +506,15 @@ pub fn XMARKER(a: LispObject) -> *const LispMarker {
...
     assert!(!MARKERP(Qnil))
 }
+
+// Implementation of the XFASTINT depends on the USE_LSB_TAG
+// in Emacs C. But we selected this implementation as in our
+// build that value is 1.
+// A must be nonnegative.
+#[allow(dead_code)]
+#[allow(non_snake_case)]
+fn XFASTINT(a: LispObject) -> EmacsInt {
```
Wilfred: Would you mind moving this into the `deprecated` module in this file?

0xAX: Yes, sure, updated. Moved it into `depreacted` mod in list.rs, did I understand you correctly?

0xAX:
>  Would you mind moving this into the deprecated module in this file?

Yes, sure, updated. Moved it into depreacted mod in list.rs, did I understand you correctly?

Wilfred:
I do think we should look at writing code that handles all values of `USE_LSB_TAG`, but this is good for now :)

0xAX:
> I do think we should look at writing code that handles all values of USE_LSB_TAG
exactly

Thanks for merge

PR #61, opened 2017-01-14
Add rust winapi dependencies.

### LINKED ISSUE #60: Linking libraries issue
Sorry i'm not very familiar with the auto{conf, ...} thing. But i hope someone who is more familiar with them can fix it. 

The problem itself is very clear:
Under windows the generated rust static library have 5 dependency libraries, three of them already linked but two not, namely  -lws2_32   -luserenv.  The cargo build can correctly displays this information on the screen. However the emacs linking doesn't have these flags. I think it needs to modified to either add these libraries on Windows platform or detect cargo output and do something.

crlf0710:
Resolves #60.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5126,6 +5126,10 @@ AC_SUBST(CARGO_FLAGS)
 
 LIB_REMACS="-lremacs -ldl"
 
+if test "${HAVE_W32}" = "yes"; then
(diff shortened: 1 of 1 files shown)


crlf0710:
No, it didn't work. There's order problem. Let me try again.

crlf0710:
This time the linking problem is gone.

Wilfred:
Thanks.

PR #62, opened 2017-01-14
Port "Fcdr()" and "cdr()" to Rust

TheKK:


rust_src/src/cons.rs
```diff
@@ -145,3 +162,16 @@ See Info node `(elisp)Cons Cells' for a discussion of related basic
 Lisp concepts such as car, cdr, cons cell and list.
 
 (fn LIST)");
+
+#[no_mangle]
+pub extern "C" fn Fcdr(list: LispObject) -> LispObject {
+    cdr(list)
+}
+
+defun!("cdr", Fcdr, Scdr, 1, 1, ptr::null(), " Return the cdr of LIST.  If arg is nil, return nil.
```
Wilfred: Minor nit: could you remove the space before `Return`?

TheKK: Fixed!

TheKK:
Sorry for bringing this up but I'd like to ask some questions about implementation details (hope they're not too trivial):
1. I saw how Lisp_Cons implemented in C and known there's an union inside it. My question is, how

``` rust
#[repr(C)]
#[allow(unused_variables)]
struct LispCons {
    car: LispObject,
    cdr: LispObject,
}

#[repr(C)]
#[allow(dead_code)]
pub struct LispConsChain {
    chain: *const LispCons,
}
```
achieves the same behaviour of C union? Does this mean we'll use `mem::transmute` to cast `LispCons.cdr` into `LispConsChain.chain`?

2. Is there any concern exposing `fn car(object: LispObject) -> LispObject` in cons.rs to C code?

3. In practical, what's the strategy while choosing between

```rust
unsafe fn Foo() { ... }
```
and

```rust
fn Foo() { unsafe { } }
```

Wilfred:
1: Yep, exactly. We'll use `mem::transmute` to cast from `LispCons.cdr` to `LispConsChain.cdr`. A struct is just a blob of bits, and since they're the same size this is fine.

I'm open to suggestions of cleaner approaches, but I think the current approach is reasonable.

2: `car` is currently not exposed to the C code, it's a function private to cons.rs. Note that it's not `extern "C"`. We could rename it to `CAR` and replace the `CAR` inline function lisp.h. I don't have any concerns with us doing so.

3: See #63, but do let me know if it's unclear.

PR #63, opened 2017-01-14
Document when functions should be marked unsafe

Wilfred:
@crlf0710 @dpzmick @jeandudey  (or anyone else who has an opinion!) does this seem like a good approach to you?


crlf0710:
Looks good.

dpzmick:
3 times the charm? Looks good to me as well.

PR #64, opened 2017-01-14
Porting 'base64-encode-string' to Rust

DavidDeSimone:
This PR aims to port 'base64-encode-string' into rust. While doing this, I added several string related functions to the rust layer, such as XSTRING and SBYTES. 

Some notes of this PR: 

- While implementing XSTRING, I gave a definition for the LispString struct. Part of a LispString in the c-layer is a pointer to an Interval struct. I did not write a rust equivalent for this struct, as base64-encode-string does not use that functionality. I have left a TODO for this. 
- I made some whitespace changes in build.rs and lisp.rs that were failing the precommit formatting hooks.
- I farmed out the actual function that does the base64-encoding to the c-layer function base64_encode_1. This function was a static c function used in fns.c across the different base64 related functions. I made it non-static, so that it can be used in rust. 
- The c-layer code uses SAFE_ALLOCA. As far as I know, rust does not have an alloca implementation currently. Instead, I used a Vec<libc::c_char> to store the memory used in base64 encoding.


rust_src/src/fns.rs
```diff
@@ -0,0 +1,43 @@
...
+
+static MIME_LINE_LENGTH: isize = 76;
+
+extern "C" {
+    fn make_unibyte_string(s: *const libc::c_char, length: libc::ptrdiff_t) -> LispObject;
+    fn base64_encode_1(from: *const libc::c_char, to: *mut libc::c_char, length: libc::ptrdiff_t,
+                       line_break: bool, multibyte: bool) -> libc::ptrdiff_t;
+}
+
+pub fn Base64EncodeString (string: LispObject, noLineBreak: LispObject) -> LispObject {
```
Wilfred: This doesn't need to be `pub` as far as I can see.

Also, could you use the standard Emacs naming here? `Fbase64_encode_string`, `Sbase64_encode_string`.

src/fns.c
```diff
@@ -3154,49 +3154,7 @@ into shorter lines.  */)
...
-  encoded_length = base64_encode_1 (SSDATA (string),
-				    encoded, length, NILP (no_line_break),
-				    STRING_MULTIBYTE (string));
-  if (encoded_length > allength)
-    emacs_abort ();
-
-  if (encoded_length < 0)
-    {
-      /* The encoding wasn't possible. */
-      error ("Multibyte character in data for base64 encoding");
```
Wilfred: You don't seem to be checking the encoded length, or handling the multibyte case in your Rust implementation. What happens if there's a multibyte character? I suspect we need the same error in the same situation.

rust_src/src/fns.rs
```diff
@@ -0,0 +1,43 @@
+extern crate libc;
```
Wilfred: Could we just keep all of this in strings.rs.?

Wilfred:
Regarding your notes on this PR, agreed with all of them. I suggest you remove the commit hooks (with `$ rm .git/hooks/*`) as they would have been automatically installed. That has been fixed in #55.

rust_src/src/fns.rs
DavidDeSimone: Good call out, will do.

DavidDeSimone:
Thanks for your feedback @Wilfred, I'll address your points and update the PR accordingly.

rust_src/src/fns.rs
DavidDeSimone: Sure. It put in fns.rs just to keep parity with fns.c, however I think it does make more sense to be in the 'strings' module.

DavidDeSimone:
@Wilfred PR updated, let me know if there is anything else you would like me to change.

rust_src/src/strings.rs
```diff
@@ -44,3 +51,39 @@ fn Fnull(object: LispObject) -> LispObject {
...
+        }
+        
+        make_unibyte_string(encoded, encodedLength)
+    }
+}
+
+defun!("base64-encode-string", Fbase64_encode_string, Sbase64_encode_string, 2, 2, ptr::null(),
+       "Base64-encode STRING and return the result.
+       Optional second argument NO-LINE-BREAK means do not break long lines
+       into shorter lines.");
```
Wilfred: You'll want to add a function signature to your docstring too, or `*Help*` will just show `(base64-encode-string ARG1 ARG2)`. You can do this by adding:

```rust
"... into shorter lines.

(fn STRING &optional NO-LINE-BREAK)");
```

DavidDeSimone:
Thank you @Wilfred for the feedback, and for maintaining such an awesome project. Let me know if there is anything else I can do.

Wilfred:
Merged, thank you for your contribution! :)

PR #65, opened 2017-01-15
Rollup

c-nixon:
This branch rolls up all of the changes made to the gnu emacs master branch since remacs was forked.

Various quite important pieces of functionality have landed (the concurrency branch and systemd support for starters)

It builds and passes tests on my machine (tm) but it could do with a test on non linux platforms.

I'm fairly sure it should work, but will fix any issues that crop up.


roobie:
Builds and runs on my machine too. I'm going to use it for daily editing stuff a while and see if I get any exceptional behavior.

c-nixon:
I've been using it all day and havn't noticed anything out of the ordinary. It seems to behave exactly the same as gnuemacs' master branch, so I' m fairly confident I fixed the (surprisingly few) conflicts correctly.

Wilfred:
Wowza, thank you for tackling this!

I initially forked remacs from the `emacs-25` branch, but I agree this is a good approach. This will help us track upstream much more easily in future.

PR #67, opened 2017-01-15
Implementation of CHECK_STRING()

0xAX:
Additionally `STRINGP` moved to `lisp::deprecated` to be consistent with `FLOATP`, `NUMBERP` and etc..


0xAX:
@Wilfred I've updated PR, rebased on master, so conflicts should be resolved now

PR #69, opened 2017-01-15
Use declared Qt in strings.rs instead of importing from C again

TheKK:


dpzmick:
Could also replace all uses of `Qt` with `LispObject::from_bool(true)` for less dependence on all these magical constants!

TheKK:
Should `Qnil` also be replaced by `LispObject::constant_nil()`? The only drawback I could figure out is it's more verbose, otherwise it's the same.

Wilfred:
@TheKK I'm happy to merge this as soon as the conflicts are fixed :)

TheKK:
Conflict solved!

crlf0710:
@Wilfred This is good! I actually thought about adding something like this last week. But i didn't just because i can't come up with a good name for those...

PR #71, opened 2017-01-15
add implementation of max-char function

0xAX:


rust_src/src/character.rs
```diff
@@ -0,0 +1,20 @@
...
+
+use lisp;
+
+extern crate libc;
+
+use lisp::{LispObject, LispSubr, EmacsInt};
+
+/// Maximum character code ((1 << CHARACTERBITS) - 1).
+pub const MaxChar: EmacsInt = 0x3FFFFF;
+
```
Wilfred: Why not do the calculation rather than leaving it in a comment?

We should also confirm that this is the right value on both 32 and 64 bit builds.

rust_src/src/character.rs
```diff
@@ -0,0 +1,20 @@
...
+fn Fmax_char() -> LispObject {
+    lisp::make_number(MaxChar)
+}
+
+defun!("max-char", Fmax_char, Smax_char, 0, 0, ptr::null(),
+"Return the character of the maximum code.
+
+(fn)");
```
Wilfred: Minor thing, but you don't need to declare the calling convention here. There are no arguments that we want to name.

rust_src/src/character.rs
0xAX: > Why not do the calculation rather than leaving it in a comment?

Will move this from comment to calculation

> We should also confirm that this is the right value on both 32 and 64 bit builds.

Yeah, I will check this

rust_src/src/character.rs
0xAX: Ok, will remove

0xAX:
@Wilfred updated

0xAX:
@Wilfred updated

0xAX:
conflicts are resolved

PR #76, opened 2017-01-16
More remacs renaming and packaging path fixups

c-nixon:
This commit does a few things:
* A lot of renaming of instances of emacs to remacs
* Renaming the emacsclient build artifact to remacs client
* Changing the manpage install to go in man2 rather than man1

This all makes it possible to package/install remacs side-by-side with
GNU Emacs without clobbering the GNU Emacs install.

Once caveat is that some info files still get put under /usr/share/info
I havn't worked out how you're meant to namespace things or how to
completely disable them during the build, but its easy to remove them
from the package tree


etc/remacs.appdata.xml
```diff
@@ -1,19 +1,19 @@
...
  <summary>An extensible text editor</summary>
  <description>
   <p>
-   GNU Emacs is an extensible, customizable text editor - and more.
+   remacs is an extensible, customizable text editor - and more.
    At its core is an interpreter for Emacs Lisp, a dialect of the Lisp
    programming language with extensions to support text editing.
   </p>
-  <p>The features of GNU Emacs include:</p>
+  <p>The features of rmacs include:</p>
```
Wilfred: Typo here.

etc/remacs.desktop
```diff
@@ -1,12 +1,12 @@
 [Desktop Entry]
-Name=Emacs
+Name=remacs
```
Wilfred: I'm inclined to capitalise this, what do you think?

etc/remacs.appdata.xml
c-nixon: This right here is why I love Rust and things like Haskell. The compiler keeps my stupid out of the world ;)

etc/remacs.desktop
c-nixon: Sure

PR #82, opened 2017-01-17
Feature/port number helpers

mfeckie:
Port a bunch of number helpers to Rust

- [x] integerp
- [x] natnump
- [x] numberp
- [x] integer-or-marker-p
- [x] number-or-marker-p

### DIFF
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -26,6 +26,7 @@ pub use math::Ftimes;
 pub use math::Fmax;
 pub use math::Fmin;
 pub use math::Fquo;
+pub use numbers::Fintegerp;
 pub use numbers::Ffloatp;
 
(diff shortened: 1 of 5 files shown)


rust_src/src/lib.rs
```diff
@@ -25,6 +26,12 @@ pub use math::Ftimes;
 pub use math::Fmax;
 pub use math::Fmin;
 pub use math::Fquo;
+pub use numbers::Fintegerp;
+pub use numbers::Finteger_or_marker_p;
+pub use numbers::Ffloatp;
+pub use numbers::Fnatnump;
+pub use numbers::Fnumberp;
+pub use numbers::Fnumber_or_marker_p;
```
Wilfred: It looks to me like these functions don't need to be exported. I can't see them being used elsewhere in the C.

PR #83, opened 2017-01-17
(Fix for os_x_build_on_travis branch) add --with-gnutls=no to travis.yml

c-nixon:
With this change travis successfully builds temacs

Now the failure is "emacs: Invalid function: cdr", which looks like an actual bug somewhere.

Question: Should mac and linux be given different configure args?


Wilfred:
Thanks. @c-nixon I've also given you the ability to push branches and review code on the repo itself :)

PR #84, opened 2017-01-17
Use cargo flags for features flag

c-nixon:
Enforce --features="strict" in travis via configure script rather than an explicit duplicate cargo build step

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,10 @@ rust:
   - nightly
 
 env:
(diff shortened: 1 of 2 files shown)


c-nixon:
```RUSTFLAGS= \

	cargo build --features 'strict' --release --manifest-path ./rust_src/Cargo.toml```

does indeed appear in the travis build log. Seems to be working.

Issue #80, opened 2017-01-17
Use a rust library for SHA* functions


Wilfred:
For example, rust-crypto.

Emacs has its own SHA256 implementation, which we don't need.

moosingin3space:
What about using [ring](https://github.com/briansmith/ring)? It's becoming the preferred Rust crypto library from what I can see.

Wilfred:
Thanks, I've updated the issue to be less explicit about which library we use.

moosingin3space:
I'd like to claim this, since I'm learning ring.

Wilfred:
@moosingin3space it's yours! :)

Have a look at `Fbase64_encode_string` for an example of creating strings in Rust. The elisp functions exposing this functionality are `md5` and `secure-hash`, which live in `src/fns.c`.

Alternatively, you don't need to redefined the elisp functions. You could look at the `secure_hash` C function in `src/fns.c`, which chooses a `hash_func` value like `sha256_buffer` (defined in `lib/sha256.c`). You'd then need to expose a function in lib.rs with the same calling convention:

```c
void *
sha256_buffer (const char *buffer, size_t len, void *resblock);
```

PR #110, opened 2017-01-23
Implement SHA1, SHA256, SHA384, and SHA512 in Rust using ring

moosingin3space:
Early implementation of #80


moosingin3space:
Looks like the test case failure is inconsistent. 
I'm new to the emacs codebase, how are the `sha*_stream` functions used? I'd like to be able to completely replace Emacs' own implementation of the hashes, but I didn't know what to do with the stream functions.

moosingin3space:
Not sure why it fails with "unable to parse TOML".

moosingin3space:
Also, what do you think I should do about SHA-224? It's not present in ring, but is needed in Emacs.

Wilfred:
We should ask nicely for SHA224 support :) https://github.com/briansmith/ring/issues/438

In the meantime, I'd be happy to merge this once the conflicts are fixed.

moosingin3space:
I'd like to try and get rid of the last usage of the Emacs SHA-1 implementation, which I left in for now.

Wilfred:
@moosingin3space friendly ping.

moosingin3space:
@Wilfred I'm sorry, I became pretty busy with school. I'll give it a shot at replacing the last SHA-1 usages (and removing all the C code I can) this week!

Wilfred:
This is all looking good, I think the way you've structured it is great.

I don't think Ring is going to be a good fit for us though. It will be awkward supporting SHA224 since upstream doesn't want to support it. Ring is also difficult to build on Windows.

Whilst [dagenix/rust-crypto](https://github.com/DaGenix/rust-crypto/) has been abandoned, there's a [rust-crypto](https://github.com/RustCrypto/hashes) project that is actively maintained. The code is pure Rust, so it should be easier to work with. They have a [sha2 crate](https://crates.io/crates/sha2) which I think would be a great fit.

Wilfred:
Continuing work in #162.

Issue #103, opened 2017-01-21
Port tempname to Rust


Wilfred:
We could straightforwardly replace `gen_tempname` and `try_tempname` with Rust implementations, probably from an existing crate.

This would let us remove tempname.c entirely.

DavidDeSimone:
I can grab this if no one else is interested.

Wilfred:
It's yours! :)

DavidDeSimone:
Update on this. Open invite to anyone who spots any errors to correct me. 

After investigation, it looks like __gen_tempname and __try_tempname are glibc functions, part of an internal API. If you attempt to compile emacs with a version of the c standard library that does not provide them, emacs will provide an implementation via tempname.c 

It looks like emacs only uses this functions for an implementation of mkostemp, which is used in a number of different places in the code base. w32.c provides a custom implementation of this function for windows that does not utilize __gen_tempname or __try_tempname.  __try_tempname only seems to be used by __gen_tempname. The ability to create a temporary directory, or to check if the temporary file exists via these functions seems to go unused in the emacs code base. 

From my perspective, it seems that we want a rust-based replacement for mkostemp, that can be used cross platform. It looks like there are a number of solid crates in this space that can give us what we need. After implementing this, I would then delete the implementation of mkostemp in w32.c, and the implementation of __gen_tempname and __try_tempname, and update the places in code that use mkostemp to use our new rust function.

DavidDeSimone:
This is complete as of https://github.com/Wilfred/remacs/issues/116, we should close this.

Wilfred:
Done :)

PR #116, opened 2017-01-27
[WIP] Removing gen_tempname and try_tempname from emacs.

DavidDeSimone:
More details of this approach can be found at https://github.com/Wilfred/remacs/issues/103

These functions were used as part of an alternative implementation of 'mkostemp', to be used if mkostemp was not available. Instead of using these functions, we provide an implementation of a mkostemp-like function from a rust crate used to generate temporary filenames. This library seems to provide functionality that will meet the current needs of the emacs usage of mkostemp. (Crate is located here: https://github.com/Stebalien/tempfile)

This is still WIP and should not be merged in yet. I still have some testing to do, and some possible tweaks to the build system/generated macros. I wanted to make this available for comment while I continue to work, in case anyone sees obvious errors, or if I'm making any serious architectural mistakes.


rust_src/src/files.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use std::os::unix::io::IntoRawFd;
+
+extern crate tempfile;
+extern crate libc;
+
+#[allow(unused_variables)]
+#[no_mangle]
+pub extern "C" fn mkrstemp(template: *mut libc::c_char,
+                          flags: libc::c_int) -> libc::c_int {
```
jeandudey: You can change `flags` to `_` and remove the unused variables attribute.

src/filelock.c
```diff
@@ -54,11 +54,12 @@ along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.  */
 #include "w32.h"	/* for dostounix_filename */
 #endif
 
-
 #ifdef HAVE_UTMP_H
 #include <utmp.h>
 #endif
 
+int mkrstemp(char *template, int flags);
```
jeandudey: Maybe you can change the function name to `rust_mkostemp` (to indicate that is from the rust codebase) or something more meaningful.

src/filelock.c
```diff
@@ -396,7 +397,7 @@ create_lock_file (char *lfname, char *lock_info_str, bool force)
       memcpy (nonce, lfname, lfdirlen);
       strcpy (nonce + lfdirlen, nonce_base);
 
-      fd = mkostemp (nonce, O_BINARY | O_CLOEXEC);
+      fd = mkrstemp (nonce, O_BINARY | O_CLOEXEC);
```
jeandudey: Also it would be great that you remove `nonce` and the flags parameters to remove unused code.

Wilfred:
👍 this is looking great!

I agree with @jeandudey -- we should remove the unused parameters, and make the function name clearer. We can then merge this! :)

DavidDeSimone:
There are some changes I'm going to have to make to this that I haven't pushed yet. I'll update this tonight with a larger write up of the work I had to do, and why I had to do it.

DavidDeSimone:
Updating as promised:

While the oneliner to tempfile::tempfile() was awesome, it does not fit the requirements of mkostemp for use in remacs. That function returns a temporary file without a path. The code in filelock.c manipulates the filename returned from mkostemp across several function calls.  tempfile has the concept of a "NamedTempFile", which also does not meet our needs. For these reasons, I dropped the tempfile dependency, and decided to implement the needed functionality in project. 

Work left to be done:

- mkostemp will set errorno in certain conditions. After googling, it seems that setting errorno from rust may be a pain, so I've changed the signature of rust_make_temp to take a mutable int pointer. 
- I need to implement a windows equivalent of files::open_temporary_file. I am planning on using the function 'sys_open' defined to w32 to accomplish this. 
- Unfortunately, the elisp tests don't have great coverage for anything using 'rust_make_temp'. Only one test module seems to hit this path at all, and it handles the failures quite gracefully. The only way I have been able to force that test to fail is to literally have 'rust_make_temp' panic on call. To have confidence in these changes, I written a number of rust layer tests. I have several more that I would like to implement. 

Related to point 2, has anyone tried building for windows yet? 

@jeandudey I would appreciate it if you would look over the diff again. Thanks.

rust_src/src/files.rs
```diff
@@ -0,0 +1,111 @@
+extern crate libc;
+extern crate rand;
```
jeandudey: I think this should be on the crate root, and in this file add the respective `use` declarations.

rust_src/src/files.rs
```diff
@@ -0,0 +1,111 @@
...
+
+fn generate_temporary_filename(name: &mut String) {
+    let len = name.len();
+    assert!(len >= 6);
+    unsafe {
+        let mut name_vec = &mut name.as_mut_vec();
+        let mut bytes = &mut name_vec[len - 6..len];
+        rand::thread_rng().fill_bytes(bytes);
+        for byte in bytes.iter_mut() {
+            *byte = match *byte % 62 {
```
jeandudey: I think a comment here is needed to explain what this algorithm does.

rust_src/src/files.rs
```diff
@@ -0,0 +1,111 @@
...
+        if file_handle != -1 {
+            unsafe { libc::strcpy(template, attempt_name.as_ptr()); }           
+            break;
+        }
+    };
+
+    file_handle
+}
+
+fn generate_temporary_filename(name: &mut String) {
```
jeandudey: Here instead of taking a borrowed `String`, you could just take a owned `String` and return the modified one.

src/callproc.c
```diff
@@ -52,6 +52,8 @@ along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.  */
 #include "nsterm.h"
 #endif
 
+int rust_make_temp(char *template, int flags, int *errorcode);
```
jeandudey: This should be on a header file (e. g.: `rust-temp.h`) to avoid code duplication.

rust_src/src/files.rs
```diff
@@ -0,0 +1,111 @@
...
+                v @ 10...35 => (v - 10 + b'a' as u8),
+                v @ 36...61 => (v - 36 + b'A' as u8),
+                _ => unreachable!(),
+            }
+        }
+    }
+}
+
+#[cfg(unix)]
+fn open_temporary_file(name: &CString, flags: libc::c_int) -> io::Result<libc::c_int> {
```
jeandudey: What about using `RawFd` for the return type?

rust_src/src/files.rs
DavidDeSimone: Good callout, I agree

rust_src/src/files.rs
DavidDeSimone: I chose to make this take a mutable reference, as https://doc.rust-lang.org/book/references-and-borrowing.html calls out taking ownership and then returning as 'unidiomatic'. I am opening to changing it if we prefer the style you described.

DavidDeSimone:
So this is _almost_ in a state that I am happy with. There are a few things in the current version I am still thinking about.

- This won't build/work on windows. I will try and play around with building remacs on windows this week. With the current architecture, it should be relatively easy to make this code work for windows.   
- https://github.com/Wilfred/remacs/issues/134 also brings up a legitimate concern. Since this is Rustifying code that used to live in lib, it may make more sense to include 'files.rs' in a "lib" crate. Travis is currently failing for nightly because lib-src/ has dependencies on mkostemp.  
- I need to add more comments for maintainability.

Wilfred:
@DavidDeSimone this is looking really good -- are you happy with it, or do you still plan to make further changes?

DavidDeSimone:
@Wilfred Apologies for the delay, I've been busy with work lately.

So I think this is almost ready to merge. Currently, this PR breaks OSX builds, so I think I just need to address that, and it will be ready to go.

rust_src/remacs-lib/Cargo.toml
```diff
@@ -8,4 +8,8 @@ libc = "0.2"
 
 [lib]
 path = "lib.rs"
-crate-type = ["staticlib", "rlib"]
\ No newline at end of file
+crate-type = ["staticlib", "rlib"]
+
+[features]
+# Treat warnings as a build error on Travis.
+strict = []
```
Wilfred: Well spotted! :)

Wilfred:
Looking good!

I guess that this will change the name of temporary files on remacs vs gnu emacs? If so, we should mention it in REMACS_COMPATIBILITY.md.

Other than that, the build is currently failing on building etags. Looks like we now need to link etags against libpthread due to depending on `remacs-lib`.

DavidDeSimone:
So rust_make_temp should have ALMOST identical to mkostemp in terms of named files. The only difference is that mkostemp will set errno, while rust_make_temp takes an int ptr that will serve as the error code. I might change this difference and introduce https://github.com/lfairy/rust-errno as a dep to remacs-lib.

DavidDeSimone:
So I'm pretty happy with the code, and I (finally) feel ready to mark this non-WIP once the build is passing. I think that the build box is failing to link libdl and librt. I am able to successfully build the products of lib-src/ locally on my own machine. I will look into this later in the week, and then hopefully we will be ready for final review and merge.

DavidDeSimone:
Hello @Wilfred. Sorry for the delay, I've had a lot going on in my personal life.

So I believe this PR should be in a solid state now. The only thing I'm confused about is why the build is failing. It looks like is complaining about improper formatting in base64.rs? I didn't touch that code as part of this PR. Should I update the formatting?

Wilfred:
@DavidDeSimone awesome!

It looks like rustfmt 0.8.1 has come out and it's changed the desired indentation on this file. If we push a commit to master, it will fail with the same error.

 I suggest you update the indentation on this branch and we merge 👍

DavidDeSimone:
Damn, it looks like were still getting linker errors. I'll have to look further into what could be causing this.

Wilfred:
@DavidDeSimone OK, I've pushed a fix for the formatting, but there's a linker error:

```
../rust_src/remacs-lib/target/release/libremacs_lib.a(std-13f36e2630c2d79b.0.o): In function `std::sys::imp::time::inner::now':

/buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/time.rs:345: undefined reference to `clock_gettime'

/buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/time.rs:345: undefined reference to `clock_gettime'

/buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/time.rs:345: undefined reference to `clock_gettime'

/buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/time.rs:345: undefined reference to `clock_gettime'

/buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/time.rs:345: undefined reference to `clock_gettime'

../rust_src/remacs-lib/target/release/libremacs_lib.a(std-13f36e2630c2d79b.0.o): In function `std::sys::imp::weak::fetch::hc8ee4ba6cea67d5c':

/buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/weak.rs:78: undefined reference to `dlsym'

../rust_src/remacs-lib/target/release/libremacs_lib.a(std-13f36e2630c2d79b.0.o): In function `std::sys::imp::condvar::Condvar::wait_timeout::h645b59ceb0926126':

/buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libstd/sys/unix/condvar.rs:77: undefined reference to `clock_gettime'

collect2: ld returned 1 exit status
```

Looks like this is `ctags` failing to link against a Rust library?

DavidDeSimone:
Aha, I've cracked the code. So this was building fine on my Debian box, but after setting up a new Ubuntu install I was able to reproduce the build issue. I'm going to switch my development on this project to Ubuntu to prevent issues like this in the future. This was just a matter of me not correctly including the -ldl and -lrt flags in the build command. 

So this PR is pretty big in terms of history at this point, we may want to squash it into a singular commit to keep the history clean. 

Also the Mac build failed, but with the following error: 
```
error: missing key: 'url'
rustup: command failed: /var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/tmp.tjV5EvPv/rustup-init --default-toolchain=nightly -y 1.62s
$ curl -sSf https://sh.rustup.rs | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION -y
info: downloading installer
info: syncing channel updates for 'nightly-x86_64-apple-darwin'
error: missing key: 'url'
The command "curl -sSf https://sh.rustup.rs | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION -y" failed and exited with 1 during .
```

Build https://travis-ci.org/Wilfred/remacs/builds/218647766 was able to succeed on Mac, so I think we should be fine if we can fix this travis script error. Thanks for everything involved in this PR @Wilfred.

Wilfred:
Great, looking good!

If you could squash the commits, that would be brilliant. Let me know if that's too fiddly and I'm happy to do the squashing.

I think the Mac failure is just https://github.com/rust-lang-nursery/rustup.rs/issues/1062 . I'll try restarting the build now, I'm hopeful that should work.

Wilfred:
OK, I've looked at rebasing these commits but it's messy. Let's merge this so we can take advantage of your efforts :)

Issue #95, opened 2017-01-20
Linux 32 bits build is broken


jeandudey:
Debug:
```gdb) run --batch --loadup bootstrap
Starting program: /home/jeandudey/rust-projects/remacs/src/temacs --batch --loadup bootstrap
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x0822d1b2 in remacs::lists::XCAR (object=LispObject = {...}) at /home/jeandudey/rust-projects/remacs/rust_src/src/lists.rs:151
151	    (*XCONS(object)).car
(gdb) bt
#0  0x0822d1b2 in remacs::lists::XCAR (object=LispObject = {...}) at /home/jeandudey/rust-projects/remacs/rust_src/src/lists.rs:151
#1  0x0822d291 in remacs::lists::car (object=LispObject = {...}) at /home/jeandudey/rust-projects/remacs/rust_src/src/lists.rs:166
#2  0x0822d454 in remacs::lists::Fcar (list=LispObject = {...}) at /home/jeandudey/rust-projects/remacs/rust_src/src/lists.rs:185
#3  0x081aa768 in print_error_message (data=<optimized out>, stream=<optimized out>, context=<optimized out>, caller=<optimized out>) at print.c:907
#4  0x081aaa2c in Ferror_message_string (obj=140488155) at print.c:868
#5  0x0818f827 in signal_or_quit (error_symbol=error_symbol@entry=26328, data=140488155, data@entry=140488139, keyboard_quit=keyboard_quit@entry=false) at eval.c:1603
#6  0x0818fb82 in Fsignal (error_symbol=26328, data=140488139) at eval.c:1483
#7  0x081901bc in xsignal (data=<optimized out>, error_symbol=26328) at lisp.h:3817
#8  0x081901bc in xsignal2 (error_symbol=26328, arg1=-1073749392, arg2=16056) at eval.c:1624
#9  0x0817b0e5 in wrong_type_argument (predicate=-1073749392, value=16056) at data.c:152
#10 0x0822d310 in remacs::lists::car (object=LispObject = {...}) at /home/jeandudey/rust-projects/remacs/rust_src/src/lists.rs:170
#11 0x0822d454 in remacs::lists::Fcar (list=LispObject = {...}) at /home/jeandudey/rust-projects/remacs/rust_src/src/lists.rs:185
[...]

PR #121, opened 2017-01-29
Revert "Remove --enable-check-lisp-object-type configure flag."

Wilfred:
Reverts Wilfred/remacs#111

32-bit linux requires that `Lisp_Object` is a struct. We need to pass this flag on 32-bit linux to avoid #95.

I will not merge this until we have CI confirming this is correct.


Wilfred:
I'm going to explore other approaches.

PR #136, opened 2017-02-06
WIP: Implement base64 encode/decode

ljos:
Hi -

I have tried to reimplement the base64 encode/decode functions in fn.c. I know they are not exactly how they should be, but I thought it would be nice to get some input before continuing.

The thing I'm struggling most with right now is the "multibyte"-string stuff. I am not entirely sure what that means and what is going on in the C code. To me it seems it seems it takes the input as a normal `*char` and if it is told it is multibyte it recasts to `*unsigned char`. However, if it finds a character that is larger than 128 it sends returns with `-1` instead.

The implementation seems to be working for me, but I don't have complete confidence in it yet. I am especially worried about all the casts to different types etc. I don't really know rust or c, but this was a fun experiment for me.


rust_src/src/base64.rs
```diff
@@ -0,0 +1,127 @@
+extern crate libc;
+extern crate base64;
+
+use base64::base64::{decode, encode};
+use std::ffi::CStr;
+use std::ffi::CString;
+use std::ptr;
+
+const MIME_LINE_LENGTH: usize = 76;
```
DavidDeSimone: This is also defined in strings.rs, we probably don't want to repeat the definition.

ljos: I am wondering; maybe the whole definition should be in strings... I don't know. I feel maybe it shouldn't be in its own file.

rust_src/src/base64.rs
```diff
@@ -0,0 +1,127 @@
...
+    to: *mut libc::c_char,
+    length: libc::ptrdiff_t,
+    multibyte: bool,
+    nchars_return: *mut libc::ptrdiff_t
+)
+    -> libc::ptrdiff_t
+{
+    match unsafe { CStr::from_ptr(from).to_str() } {
+        Ok(encoded) => {
+            let decoded = CString::new(decode(encoded).unwrap()).unwrap();
```
DavidDeSimone: I think it would be good to rework the code to avoid these raw unwraps. It seems like a good end-state would be that if either decode or CString::new, this function should return -1. 

I'm picturing a function which does the decoding, using a rust interface. That way it's easy for other rust code to use this functionality directly, without having to worry about unsafe mutable buffers. Something like

```
let foo = CStr::from_ptr(from)...
match decode_base64(foo) {
    Ok(encoded) => // copy bytes and return n
    Err(_) => -1
}

fn decode_base64(to_decode: &str) -> Result<CString> { // This can be a regular string one day, when we port
                                                       // the remaining base64-encode functions to rust.
       let decoded = try!(decode(to_decode)); // You may need to map_err here. 
       try!(CString::new(decoded)) // or map_err here
}
```

This way we have a clean separation between rust functionality using rust primitives, and c functionality using unsafe c-style code.

ljos: I get where you are going. I agree.

ljos:
I will refactor according to @DavidDeSimone tomorrow and fix the multibyte stuff. I think I get it  now. I think I should be able to implement the tests as suggested by @Wilfred. 

I also think maybe we should also put the functions inside strings or maybe somewhere else. I'm not convinced they need their own file. I will sleep on it.

ljos:
@Wilfred It seems to me that with the current code we don't have to fail on multibyte code. I am going to do some more tests, but for now it seems that if we receive `"日本"` it does the right thing and returns `"5pel5pys"`, and that is turned back to the right thing.

I can implement fail on multibyte so we provide the exact same API, but I would be very surprised if there is some code out there that depends on base64-encode to fail on multibyte input.

ljos:
After the recent changes I can now get this 
```
ELISP> (setq s "日本")
"日本"
ELISP> (multibyte-string-p s)
t
ELISP> (setq e (base64-encode-string s))
"5pel5pys"
ELISP>(setq s (base64-decode-string e))
"日本"
ELISP> (multibyte-string-p s)
t
ELISP> (setq s2 "abc")
"abc"
ELISP> (multibyte-string-p s2)
nil
ELISP> (base64-encode-string s2)
"YWJj"
ELISP> (setq s3 (string-to-multibyte "abc"))
"abc"
ELISP> (multibyte-string-p s3)
t
ELISP> (base64-encode-string s3)
"YWJj"
```

Wilfred:
Ooh, that's a really interesting idea.

Emacs currently returns raw bytes when decoding non-ASCII, but it is able to decode correctly:

```lisp
ELISP> (base64-decode-string "5pel5pys")
"\346\227\245\346\234\254"
ELISP> (string-as-multibyte (base64-decode-string "5pel5pys"))
"日本"
```

This suggests to me that making `base64-encode-string` work is the right thing to do, even though it's a slight behaviour change. We should probably document this change though.

This only leaves the question of what we should do with the `multibyte` argument to the function `base64_encode_1`, because this function is also used in `base64-encode-region`. What do you think?

ljos:
After a little testing it seems like `base64-decode-region` isn't ready for multibyte just yet; `encode` works fine. It doesn't particularly like when I use my multibyte respecting version. I don't think it should be difficult to get it to work though. I'll have look.

ljos:
I think we should support multibyte strings, but only if we can ensure that the region functions also work correctly. I have an idea for how to do it; I will look it over tomorrow.

ljos:
I think it should be solved. I hadn't understood completely how the nchar_return argument is used. We need the multibyte argument for decoding, as it is the only way we can now a region supports multibyte.

I think I should test a few more things before it is merged just to be confident that it works, but I don't think the code will change much at the time. I will also add a note in the compatibility section.

ljos:
So, the ` lisp/net/sasl-scram-rfc-tests` failing seems to be because of the changes in base64 library. I am starting to be convinced that there is something fishy about the test itself. It does pass, but I am not sure it should on standard emacs.

The thing that fails is that a variable called `client-proof` is base64 encoded. It is a multibyte string and looks like this `¿Eü¿psÙ=$fÉC!t_áÈá;`, it encoded by normal emacs results in `v0X8v3Bz2T0CJGbJQyF0X+HI4Ts=`, but running `base64-decode-string` on that returns `\277E\374\277ps\331=$f\311C!t_\341\310\341;`, and it doesn't get any better by calling `string-as-multibyte` which just returns the same. On the other hand, when run through the rust-base64 function it returns `wr9Fw7zCv3Bzw5k9AiRmw4lDIXRfw6HDiMOhOw==`, which when decoded returns to the same string that went in. In normal emacs, if I give it the base64 string made by our encoder it returns a string that becomes the right string after `string-as-multibyte`.

I believe the test is flawed. I have included a fix for the test.

ljos:
OK. There is something horribly wrong. I think I will need to implement the multibyte checks and stuff. Tramp is acting up and so are other parts.

ljos:
I think I solved it. I think all tests should pass now.

Wilfred:
Merged! Thank you :)

Issue #138, opened 2017-02-10
RFC: Use Rust codegen


jeandudey:
About the use of Rust codegen
-------------------------------------------

### Description
I've thinked that we could use nigthtly **Rust** features to make **Remacs**
development and porting easier. I was inspired how the *rocket.rs* crate
handled the *HTTP* request using Rust attributes, it looks something like this:

```rust
#[get("/")]
fn index() -> &'static str {
   "Hello, world!"
}
```

I think Remacs could do something similar and have a `defun` or a
`lisp_function` attribute to generate FFI code, also it would solve the problem
of having two lisp object types, because the generated code can hide all the
details of having a FFI `Lisp_Object` type.

An example of a native lisp function would look like this in Rust:

```rust
#[lisp_function(name = "symbolp", min = 1)]
#[lisp_doc("Return t if OBJECT is a symbol")]
#[lisp_doc("(fn OBJECT)")]
fn symbolp(object: LispObject) -> LispObject {
    if object.is_symbol() {
        LispObject::constant_t()
    } else 
        LispObject::constant_nil()
    }
}
```

And the generated code would look something like this:

```rust
lazy_static! {
    pub static ref Ssymbolp: remacs_sys::Lisp_Subr = remacs_sys::Lisp_Subr {
        header: VectorlikeHeader {
            size: ((lisp::PvecType::PVEC_SUBR as libc::c_int) << lisp::PSEUDOVECTOR_AREA_BITS) as libc::ptrdiff_t,
        },
        function: Fsymbolp as *const libc::c_void,
        min_args: 1,
        max_args: 1,
        symbol_name: b"symbolp\0".as_ptr(),
        intspec: std::ptr:null(),
        doc: b"Return t if OBJECT is a symbol\n(fn OBJECT)".as_ptr(),
    }
}

pub extern "C" Fsymbolp(p0: remacs_sys::Lisp_Object) -> remacs_sys::Lisp_Object {
    let p0 = LispObject::from_raw(p0);
    let ret = symbolp(p0);
    ret.to_raw()
}

fn symbolp(object: LispObject) -> LispObject {
    if object.is_symbol() {
        LispObject::constant_t()
    } else 
        LispObject::constant_nil()
    }
}
```

You will notice that there is an extra type in the `Fsymbolp` function, this is
used to keep the ABI compability with the **C** code.

### Implementation
[...]

PR #139, opened 2017-02-11
Fix issues on 32 bits platforms.

jeandudey:
This is a big change to the remacs Rust codebase, it basically changes
everything.

The `defun!` macro was modified to automatically create a wrapper
function that will be used outside of C, it means that you don't need to
actually use 'unsafe' code in your functions, however the existing
functions need to be updated to work without 'unsafe' blocks.

Also a new macro was added, it's called `defun_many!`, it is similar to
the normal 'defun 'macro however this only works for functions that
accept a dynamic list of parameters. It nicely creates a wrapper
function like the 'defun!' to pass a slice of arguments (`LispObject`s) to
the actual function.

I found that the code in` math.rs` is a bit complicated, specially
the arith_driver function, it needs to be refactored, i don't know if
that function actually works correctly.

The `src/Makefile.in` file was also modified to rebuild if the files
inside `remacs-sys` changed, including Cargo.toml.


rust_src/src/lisp.rs
```diff
@@ -34,50 +37,30 @@ use marker::{LispMarker, marker_position};
...
-pub const EMACS_FLOAT_SIZE: EmacsInt = 8;
-#[cfg(dummy = "impossible")]
-pub const GCTYPEBITS: EmacsInt = 3;
-#[cfg(dummy = "impossible")]
-pub const USE_LSB_TAG: bool = true;
-
 #[repr(C)]
 #[derive(PartialEq, Eq, Clone, Copy)]
-pub struct LispObject(EmacsInt);
+pub struct LispObject(Lisp_Object);
```
Wilfred: We should probably add a docstring comment here explaining the purpose of this struct (safety, ability to define methods).

I also think there's a risk that having both LispObject and Lisp_Object will be confusing. How do you feel about defining CLispObject, consistent with the CString naming convention?

rust_src/src/lisp.rs
```diff
@@ -538,21 +521,58 @@ unsafe impl Sync for LispSubr {}
...
-                symbol_name: ((concat!($lisp_name, "\0")).as_ptr()) as *const c_char,
+                symbol_name: ((concat!($lisp_name, "\0")).as_ptr()) as *const $crate::libc::c_char,
+                intspec: $intspec,
+                doc: (concat!($docstring, "\0").as_ptr()) as *const $crate::libc::c_char,
+            };
+        }
+    }
+}
+
+macro_rules! defun_many {
```
Wilfred: Needs a docstring, and probably a mention in the example in the readme.

rust_src/src/lisp.rs
jeandudey: Maybe we could leave the `Lisp_Object` name in `remacs-sys` and import it like `use remacs_sys::Lisp_Object as CLispObject`.

I think that declarations should match the exact name of their counterpart in C to avoid possible confusions. Also renaming `Lisp_Object` as `CLispObject` in `remacs-sys` can confuse future contributors because `CLispObject` looks like some kind of higher level abstraction.

jeandudey: Also i came up with another idea for #138, maybe we could have a lint that detects when a function has a `Lisp_Object` in the parameters and doesn't has a `extern "C"` marker.

Wilfred: > `use remacs_sys::Lisp_Object as CLispObject`

I really like this idea! This makes it explicit, keeps Rust conventions, but preserves the original type names from the C.

README.md
```diff
@@ -266,21 +266,20 @@ We can see we need to define a `Snumberp` and a `Fnumberp`. `Qt` and
...
-        Qnil
+        LispObject::constant_nil()
     }
 }
 
 // This defines a built-in function in elisp, which is a represented
 // with a static struct.
 defun!("numberp", // the name of our elisp function
-       Fnumberp, // the rust function we want to call
+       Fnumberp, // the function that will be called by C (this calls numberp).
```
Wilfred: Why isn't this `Fnumberp(object)`? It's not quite obvious to me when we need `Ffoo` vs `Ffoo(bar)` in the `defun!` declaration, would you mind clarifying in the docstring?

Wilfred:
The Linux build is passing on Travis, but the OS X build is failing on `info-xref-test-makeinfo`. I don't think it's related to these changes.

We should probably add a `(skip-unless (not (eq system-type 'darwin)))` to the test.

README.md
jeandudey: Wow, i didn't caught it. I'll fix it

jeandudey:
@Wilfred updated.

PR #140, opened 2017-02-13
Remove MS-DOS

Wilfred:
I've removed MS-DOS references from the code and from the manual. Each commit should be fairly self-contained and self-explanatory, but reviews welcome.

@c-nixon this increases the risk of conflicts on future roll-ups, FWIW. It should still be unlikely though -- I don't think the MS-DOS port sees much work.


ljos:
Just for completeness, should it be mentioned in REMACS_COMPATIBILITY.md that we don't support MS-DOS?

Wilfred:
Yep, that's a very sensible idea :). Done.

Issue #135, opened 2017-02-04
Run rustfmt on Travis CI


jeandudey:
We should be running `rusfmt` on the backend and fail the build if the code isn't formatted properly.

Wilfred:
Done in #151.

PR #144, opened 2017-02-18
Run rustfmt as part of CI process

mfeckie:
Implementation for #135


rust_src/src/globals.rs
```diff
@@ -1,5 +1,5 @@
-#![allow(dead_code)] // https://github.com/rust-lang/rust/issues/38780
-
+#![cfg_attr(rustfmt, rustfmt_skip)]
```
Wilfred: Why aren't we automatically formatting this file?

rust_src/src/globals.rs
```diff
@@ -1,5 +1,5 @@
-#![allow(dead_code)] // https://github.com/rust-lang/rust/issues/38780
```
Wilfred: I think this comment is worth keeping -- we only have this at the module level due to the bug linked. We can remove this in 1.16.

.travis.yml
```diff
@@ -19,8 +19,11 @@ env:
   # Ensure that we build without warnings.
   - CARGO_FLAGS="--features 'strict'"
 
+before_script: (cargo install rustfmt || true)
```
Wilfred: Why the `|| true` here?

jeandudey: To fail if `rustfmt` can't be installed i think

Wilfred: This will actually succeed if `rustfmt` can't be installed:

```
$ cargo install nosuchpackage || true
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: could not find `nosuchpackage` in `registry https://github.com/rust-lang/crates.io-index`
$ echo $?
0
```

However, we depend on it being installed when we call `cargo fmt` lower down. If I've understood correctly, at least.

jeandudey: From http://johannh.me/blog/rustfmt-ci.html:
> Since cargo returns an error if a package is installed already, I have to use (cargo install rustfmt || true).

However if `cargo install rustfmt` fails without installing it et all, the build will fail.

Wilfred: Aha, it's because `cache: cargo` saves the installed `rustfmt` so you don't need to install every time. Makes sense.

Wilfred:
Merged in #151, thanks for your PR! :)

PR #145, opened 2017-02-19
Move LispSubr to remacs-sys

jeandudey:
`LispSubr` is now `Lisp_Subr` in `remacs-sys`.

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -54,17 +54,38 @@ pub struct vectorlike_header {
     pub size: libc::ptrdiff_t,
 }
 
+/// Representation of an Emacs Lisp function symbol.
 #[derive(Debug)]
 #[repr(C)]
 pub struct Lisp_Subr {
     pub header: vectorlike_header,
-    // This field is actually an union
+
+    /// This is the function pointer that will be called when the user invokes
+    /// the Emacs Lisp function. Also, this field is actually an union in C.
     pub function: *const libc::c_void,
+
+    /// The minimum number of arguments that can be passed to the Emacs Lisp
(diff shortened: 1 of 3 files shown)


rust_src/src/lisp.rs
```diff
@@ -422,80 +421,6 @@ impl LispObject {
...
-    PVEC_SUB_CHAR_TABLE = 16,
-    PVEC_FONT = 17,
-}
-
-#[repr(C)]
-pub struct VectorLikeHeader {
-    pub size: libc::ptrdiff_t,
-}
-
-/// Represents an elisp function.
```
Wilfred: I think these docstrings are worth preserving, why have we removed them?

rust_src/remacs-sys/lib.rs
```diff
@@ -20,6 +20,62 @@ include!(concat!(env!("OUT_DIR"), "/definitions.rs"));
...
+#[repr(C)]
+pub struct Lisp_Subr {
+    pub header: vectorlike_header,
+    // This field is actually an union
+    pub function: *const libc::c_void,
+    pub min_args: libc::c_short,
+    pub max_args: libc::c_short,
+    pub symbol_name: *const libc::c_char,
+    pub intspec: *const libc::c_char,
+    // In, Emacs C, this is an EMACS_INT. Should this be changed?
```
Wilfred: Wow, well spotted! Apparently this used to be a `*char`: https://github.com/emacs-mirror/emacs/commit/c5461d03a411ff5c6f43885a0a9030e8a94bbc2e which is why it probably works.

Worth adding a `TODO` I think.

jeandudey:
@Wilfred updated.

PR #147, opened 2017-02-20
nlistp takes exactly one argument

Wilfred:


Wilfred:
Thanks @jeandudey.

I'll make you a reviewer too, so you can approve PRs :)

PR #152, opened 2017-02-21
[WIP] Implement #138 (codegen)

jeandudey:
At very early stages of implementation. If you find some minor nit, report it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: rust
 cache: cargo
 rust:
-  - stable
-  - beta
-  - nightly
+  - nightly-2017-02-18
 os:
   - linux
   - osx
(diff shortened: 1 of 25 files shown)


jeandudey:
@Wilfred
i've made some progress on this but i cannot keep contributing to the project (the [why][1]), it's really a crap that i cannot continue developing at the same rate i would like to, i have left a TODO list in the case anyone wants to continue the PR:

- [ ]  Handle `MANY` arguments, this should be calculated at compile time by checking if the type is `&[LispObject]`.
    - [ ] Write `compile-fail` tests for this.
- [ ] Check that the return type of the rust function is `LispObject`.
- [ ] Check that `min` is not larger than the number of arguments.
- [ ] Generate the FFI symbol (`Lisp_Subr`). Note: this can be done without the use of `lazy_static`.
- [ ] Add to the nightly the specific version needed to compile remacs.
- [ ] Make a list of files that need to be updated in case a new version of nightly Rust need to be updated.

Here a nice to have that i would like to see on remacs:
- Make `LispObject` type safe and memory safe, E. g.: using references and lifetimes on `to_*` methods.
    - Also a `to_mut_*` would be needed for each of them.
- Keep `lisp` free of FFI code, only the necessary.
- Add a type safe wrapper around `defsubr` (`register_function`?).
- Use cargo workspaces, this is a plus for productivity.

Please make the project good without me and keep doing a good work on this :+1: 

[1]: https://gist.github.com/jeandudey/a885d80cbedeaef568fbe389348aa076

Issue #142, opened 2017-02-14
Can't commit in magit


PythonNut:
(followup to #92)

When trying to commit staged changes in `magit`, the following error is produced:

> Nothing staged (or unstaged)

In regular Emacs, this works correctly.

Unfortunately, I don't currently have a guess as to why this happens.

vodik:
A hint with this one is the assert looking for staged files fails in remacs which succeeds in emacs. More specifically:

```elisp
(magit-git-exit-code "diff" "--quiet" "--cached")
```

returns 1 under emacs and 0 under remacs.

vodik:
And I think I know what it is. Try this in remacs:

```elisp
(process-file "false")
```

You incorrectly get back 0. Should be 1. Works like that in emacs.

vodik:
And looks like `call-process` is original C code, so don't have the faintest idea how a regression here was introduced.

vodik:
Copying callproc.c from upstream emacs and fixing it up so it would compile resulted in a correction of this problem. I'll investigate a bit further tomorrow, but at the very least I have a fix:

```diff
diff --git a/src/callproc.c b/src/callproc.c
index 113d3f41f6..873065027d 100644
--- a/src/callproc.c
+++ b/src/callproc.c
@@ -182,11 +182,11 @@ call_process_cleanup (Lisp_Object buffer)
     {
       kill (-synch_process_pid, SIGINT);
       message1 ("Waiting for process to die...(type C-g again to kill it instantly)");
-      immediate_quit = true;
-      maybe_quit ();
+
+      /* This will quit on C-g.  */
       wait_for_termination (synch_process_pid, 0, 1);
+
       synch_process_pid = 0;
-      immediate_quit = false;
       message1 ("Waiting for process to die...done");
     }
 }
@@ -293,6 +293,11 @@ call_process (ptrdiff_t nargs, Lisp_Object *args, int filefd,
 
   error_file = Qt;
 
+  /* Without asynchronous processes we cannot have BUFFER == 0.  */
+  if (nargs >= 3
+      && (INTEGERP (CONSP (args[2]) ? XCAR (args[2]) : args[2])))
+    error ("Operating system cannot handle asynchronous subprocesses");
+
   /* Decide the coding-system for giving arguments.  */
   {
     Lisp_Object val, *args2;
@@ -635,9 +640,6 @@ call_process (ptrdiff_t nargs, Lisp_Object *args, int filefd,
       process_coding.src_multibyte = 0;
[...]

PR #156, opened 2017-02-27
Port callproc.c from emacs 25

vodik:
I back-ported emacs 25 callproc.c back into remacs, while preserving existing cleanup efforts. With this patch in place, `call-process` correctly returns process exit codes, fixing #142.

I want to emphasis this is the first time I've ever looked at the emacs source code and don't have the slightest idea what part of this is the actual fix. But might not be a bad idea to merge since it seems like emacs has cleaned it up a bit. For example `reallocate_fd` is dead code.

### DIFF
--- a/src/callproc.c
+++ b/src/callproc.c
@@ -182,11 +182,11 @@ call_process_cleanup (Lisp_Object buffer)
     {
       kill (-synch_process_pid, SIGINT);
       message1 ("Waiting for process to die...(type C-g again to kill it instantly)");
-      immediate_quit = true;
-      maybe_quit ();
+
+      /* This will quit on C-g.  */
       wait_for_termination (synch_process_pid, 0, 1);
+
       synch_process_pid = 0;
-      immediate_quit = false;
       message1 ("Waiting for process to die...done");
     }
 }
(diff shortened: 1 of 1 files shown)


Wilfred:
Thanks for your contribution!

Remacs is a fork of the master branch of GNU Emacs, so it's effectively Emacs 26. Merging patches from Emacs 25 could cause pain when we next merge in GNU Emacs master.

Could you identify which commits this is? git-blame should be able to help you here. Looking at master on GNU Emacs, I suspect it's https://github.com/emacs-mirror/emacs/commit/b4c9f9120d8b0da0593f2fbde69b40374f56451d .

vodik:
Aaaand it isn't 25. Brain fart, my bad. I meant 26.

I pulled this straight from a emacs-git source tree I have kicking around. Teaches me to write PRs late in the night.

So, first thing I did was look at the log for callproc.c and I didn't see anything interesting there. In fact, that commit you suspect to be the problem isn't in the history.

I'll see if I can find anything more later tonight, at the very least I'll fix up the commit to correctly identify the version.

c-nixon:
#158 has the upstream commit containing the immediate_quit changes, along with everything else that went into emacs/master in the last month.

Could you test that PR? Being able to pull it all in at once makes the rollup merges much easier :)

vodik:
Sure, just a sec.

vodik:
The issue is fixed with #158, so I'll close this PR. Thanks for all the help!

PR #165, opened 2017-04-03
Port sig2str.c to Rust.

### LINKED ISSUE #122: Port str2sig to Rust
It's a simple C library, converting signal numbers to strings.

barrucadu:
- I wasn't sure what to do about sig2str.h, so I just stripped out the now-irrelevant bits and left it.
- `str2sig` is only used in `abbr_to_signal` in `signal-process` in process.c. When that is rustified it may be possible to adopt a safer interface.
- `sig2str` doesn't appear to be used anywhere, and isn't exposed as a lisp function. Should it just be removed?

Closes #122


Wilfred:
Looking good! I've opened #166 to fix the formatting issue which is breaking the build here.

Yep, feel free to remove unused functions. Emacs has quite a few of these: the C libraries are generic libraries that have been copy-pasted in, and not all the functionality is actually used.

Wilfred:
I think keeping sig2str.h is sensible, as it means the C code knows the calling convention of the functions we're exposing. As you say, we can refactor as the C codebase shrinks.

barrucadu:
I've rebased this against master and removed the `sig2str` function, I'm now happy with this PR if you don't have any further suggestions.

Wilfred:
Terrific, thank you :)

Issue #169, opened 2017-04-12
No docstrings on Rust primitive functions


Wilfred:
It looks like we don't have docstrings on Rust primitive functions if you build from clean.

We should either change how we initialise functions (see `rust_init_syms` in lib.rs) or we should integrate with snarf (see elisp primitive `Snarf-documentation` which writes etc/DOC).

birkenfeld:
Fixed in #203.

PR #170, opened 2017-04-16
Porting characterp

Titan-C:
This is my first experience on rust. I started porting a single function from character.c which also needed to port a macro from character.h which I did not delete after porting.

This compiles and the function works
But I noticed documentation is not working, neither in this function nor in other ported functions. like `max-char` or the one of the readme page `numberp`

Looking forward to listen from your reviews

### DIFF
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -19,12 +19,12 @@ defun!("max-char",
        ptr::null(),
        "Return the character of the maximum code.");
 
-/* Nonzero iff X is a character.  */
+// Nonzero iff X is a character.
 pub fn CHARACTERP(x: LispObject) -> bool {
     lisp::NATNUMP(x) && lisp::XFASTINT(x) <= MAX_CHAR
 }
 
-fn characterp(object: LispObject, ignore: LispObject) -> LispObject {
+fn characterp(object: LispObject, _ignore: LispObject) -> LispObject {
     if CHARACTERP(object) {
         LispObject::constant_t()
     } else {
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -18,3 +18,29 @@ defun!("max-char",
        0,
        ptr::null(),
        "Return the character of the maximum code.");
+
+/* Nonzero iff X is a character.  */
+pub fn CHARACTERP(x: LispObject) -> bool {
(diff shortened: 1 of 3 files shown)


rust_src/src/character.rs
```diff
@@ -18,3 +18,29 @@ defun!("max-char",
        0,
        ptr::null(),
        "Return the character of the maximum code.");
+
+/* Nonzero iff X is a character.  */
```
Wilfred: Could you make this a normal docstring, so `/// Return true iff ...`

rust_src/src/character.rs
```diff
@@ -18,3 +18,29 @@ defun!("max-char",
...
+       Scharacterp,
+       characterp,
+       1,
+       2,
+       ptr::null(),
+       "Return non-nil if OBJECT is a character.
+In Emacs Lisp, characters are represented by character codes, which
+are non-negative integers.  The function `max-char' returns the
+maximum character code.
+usage: (characterp OBJECT)");
```
Wilfred: This looks correct. The reason you don't see the docstring in Remacs is a known bug, see #169.

Titan-C:
Comments are addressed. I hope that travis builds now.
I think I know the answer(backwards compatibility), but I'll ask anyway. Why does this function `characterp` needs two arguments if the second one is going to be completely ignored anyway? It does not even show up in the docstring. Shouldn't we just drop it?

jeandudey:
>I hope that travis builds now

Nightly seems to be broken on Mac OS X, so it isn't related to your changes.

>Why does this function characterp needs two arguments if the second one is going to be completely ignored anyway?

Yes, you're right, we should remove the last argument, it doesn't breaks backwards compatibility. However we should document it at `REMACS_COMPATIBILITY.md` just "to be sure".

If you are going to remove it please let me know so i can merge this PR, if not i can do it after merging this.

jeandudey:
Also, can you squash these commits? we have a pretty big numbers of commits right now :smile: :+1:

Wilfred:
Yep, it's backward compatibility. This means that users can call `(characterp ?a nil)`. I think we should continue to support this so Remacs supports the same elisp that GNU Emacs supports.

(I don't know what the extra argument was for, nor am I aware of any elisp that depends on this. However, the extra argument is documented in the docstring, so users might expect it.)

PR #171, opened 2017-04-19
Cast to EmacsInt instead of i64 on strings.rs

jeandudey:
The 32 bit compilation was broken because of an invalid cast.


Wilfred:
Thanks. It's unfortunate that nightly is broken, but it's unrelated to these changes so I'm merging :)

PR #175, opened 2017-04-19
Remove rust_print_lisp_object function

jeandudey:
This function was used for debugging but now it isn't needed anymore.


Wilfred:
👍

PR #176, opened 2017-04-19
Change Emacs to Remacs in autogen.sh

jeandudey:
Just noted it when building the first time again :+1: 


Wilfred:
Nice. I'm not sure whether we should do a find and replace on the whole codebase: I see Remacs as an Emacs, just not *the* GNU Emacs. Certainly for build instructions I agree we should be specific.

PR #177, opened 2017-04-19
Add define_functions macro.

jeandudey:
I refactored the `rust_init_syms` function, i added the define_functions
macro. It handles the definition of the `rust_init_syms` function, also
it eventually could be used to export C functions and handle modules
definitions.

Also the `floatfns_init_syms` function was removed, the define_functions
allows us to initialize the symbols more cleanly without the need of another
function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,14 +293,16 @@ defun!("numberp", // the name of our primitive function inside elisp
 (fn OBJECT)");
 ```
 
-Finally, we need to delete the old C definition and call `defsubr`
-inside `rust_init_syms`:
+Finally, we need to delete the old C definition and put the
+symbol name in the `define_functions` macro inside `rust_init_syms`:
 
 ``` rust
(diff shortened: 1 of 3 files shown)


Wilfred:
I'm definitely in favour of removing the separate function for defining float functions.

I worry that the macro (whilst definitely pretty) is a little too magical though. What do you think about moving the function definition out of the macro?

jeandudey:
@Wilfred thanks for the feedback, i've thinked that too and tought it was magical, also it would allow us to define other `*_init_syms` functions. I'll change it

jeandudey:
@Wilfred i've updated it :+1:

jeandudey:
@Wilfred oops, i forgot to mention that i updated the README.md file

Wilfred:
I'm closing this due to inactivity. If you're still interested, do feel free to open a new PR.

PR #181, opened 2017-04-22
Implement `lisp_fn` procedural macro

jeandudey:
**Status:** Not ready for review


rust_src/remacs-macros/Cargo.lock
```diff
@@ -0,0 +1,42 @@
+[root]
```
jeandudey: Oops, this shouldn't be here, I'll delete it later.

jeandudey:
Check list:

- [ ] Better error handling, there are a lots of `unwrap`s around there, at least they should be formatted.
- [ ] Detect `MANY` arguments, the macro should check for `&[LispObject]`.
- [ ] Put the Rust function back in code.
- [ ] Detect if the arguments are `LispObject` or `&[LispObject]`
    - [ ] Detect if `&[LispObject]` and `LispObject` are mixed and throw an error.
    - [ ] Detect if `&mut [LispObject]` is used and throw and error.
- [ ] Remove that debug `panic` at the end.
- [ ] Add a README.md in `remacs-macros` explaining what `lisp_fn` does and how it behaves when different arguments are passed.
- [ ] Configure `.travis.yml` to use **nightly**
- [ ] Remove `defun` example in the main README.md, to use this?

jeandudey:
@Wilfred now ready to review, i don't have implemented support for `UNEVALLED` functions, it should be trivial to do but i left it as part to another PR.

rust_src/remacs-macros/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+In this example the attribute generates the `Fsame` function that is going to b called in **C**, and the `Ssame` structure that holds the function information. You still need to register the function with `defsubr` to make it visible in Emacs Lisp. To make a function visible to C you need to export it in the crate root (lib.rs) as follows:
+
+```rust
+use somemodule::Fsome;
+```
+
+### `MANY` functions
+
+This attributes handles the definitions of `MANY` functions, these functions can take an arbitrary number of arguments, but you *still can* specify a `min` number of arguments.
```
Wilfred: *This attribute

But I'm not sure what attribute you're referring too here. Do we need to use the term `MANY`? I think it's just a term the Emacs C uses, but readers of the README might not know it.

rust_src/remacs-macros/README.md
```diff
@@ -0,0 +1,44 @@
+remacs-macros
+=============
+
+# Attributes
+
+## `lisp_fn`
+This macros creates the necessary functions and symbols automatically.
+It handles `MANY` functions and normal functions too (support for `UNEVALLED`
```
Wilfred: I think the note about `UNEVALLED` belongs in a comment rather than in the docs that new users see first.

rust_src/remacs-macros/README.md
```diff
@@ -0,0 +1,44 @@
+remacs-macros
+=============
+
+# Attributes
+
+## `lisp_fn`
+This macros creates the necessary functions and symbols automatically.
```
Wilfred: *This macro

rust_src/remacs-macros/README.md
```diff
@@ -0,0 +1,44 @@
...
+It handles `MANY` functions and normal functions too (support for `UNEVALLED`
+needs to be done).
+
+It is used like this:
+
+```rust
+/// Return the same argument
+#[lisp_fn(name = "same", min = "1")]
```
Wilfred: Do we support setting a maximum number of arguments in this macro? We probably should, so we can use it in this example.

rust_src/remacs-macros/README.md
```diff
@@ -0,0 +1,44 @@
...
+fn same(obj: LispObject) -> LispObject {
+    obj
+}
+```
+
+Where the `name` argument specifies the **symbol name** that is going to be use in Emacs Lisp, `min` specifies the **minimum** number of arguments that can be passed to this function.
+
+In this example the attribute generates the `Fsame` function that is going to b called in **C**, and the `Ssame` structure that holds the function information. You still need to register the function with `defsubr` to make it visible in Emacs Lisp. To make a function visible to C you need to export it in the crate root (lib.rs) as follows:
```
Wilfred: *going to be

rust_src/remacs-macros/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+```rust
+/// Returns the first argument.
+#[lisp_fn(name = "first", min = "1")]
+fn first(args: &mut [LispObject]) -> LispObject {
+    args[0]
+}
+```
+
+## `lisp_doc`
```
Wilfred: I don't think we should have code that doesn't work. Let's just drop this and leave the docs as Rust docstrings.

.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: rust
 cache: cargo
 rust:
-  - stable
-  - beta
   - nightly
```
Wilfred: We should probably also test against a pinned nightly version too, so we have a known-good point in time to compare with when nightly changes.

rust_src/remacs-macros/function.rs
```diff
@@ -0,0 +1,244 @@
...
+            if is_unsafe(unsafety) {
+                return Err("lisp functions cannot be `unsafe`");
+            }
+            
+            if is_const(constness) {
+                return Err("lisp functions cannot be `const`");
+            }
+
+            if !is_rust_abi(abi) {
+                return Err("lisp functions can only use \"Rust\" ABI");
```
Wilfred: Nice defensive coding :)

rust_src/remacs-macros/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+            pub static ref #sname: ::remacs_sys::Lisp_Subr = ::remacs_sys::Lisp_Subr {
+                header: ::remacs_sys::vectorlike_header {
+                    size: (::remacs_sys::PVEC_SUBR << ::remacs_sys::PSEUDOVECTOR_AREA_BITS) as ::libc::ptrdiff_t,
+                },
+                function: self::#fname as *const ::libc::c_void,
+                min_args: #min_args,
+                max_args: #max_args,
+                symbol_name: (#symbol_name).as_ptr() as *const ::libc::c_char,
+                intspec: ::std::ptr::null(),
+                doc: ::std::ptr::null(),
```
Wilfred: It's annoying that Emacs doesn't look at this attribute in the struct. We should probably fix that in Remacs (not in this PR).

rust_src/src/character.rs
```diff
@@ -1,25 +1,18 @@
 use std::ptr;
-
 use lisp::{self, LispObject};
+use remacs_macros::{lisp_fn, lisp_doc};
 use remacs_sys::{EmacsInt, CHARACTERBITS};
 
 /// Maximum character code
 pub const MAX_CHAR: EmacsInt = (1 << CHARACTERBITS as EmacsInt) - 1;
 
+#[lisp_fn(name = "max-char")]
```
Wilfred: This is a really nice API, I like it.

rust_src/remacs-macros/README.md
jeandudey: The maximum number of arguments are calculated automatically counting the number of arguments in the function.

Wilfred: Aha, excellent. We should probably mention this.

.travis.yml
jeandudey: I think we should only use one nightly because the proc-macro doesn't change that much, also it would slow builds

jeandudey:
@Wilfred changes addressed

Wilfred:
Looks great! There's a compile warning about an unused macro `defun_many`, which seems to be why the build is failing. You should be able to merge once that's fixed :)

PR #191, opened 2017-05-25
Some functions moved to Rust

birkenfeld:
This is my first venture into remacs, please review carefully. Thanks :)

PR #192, opened 2017-05-28
Proc macro updates

birkenfeld:
This is based on #191, #181 and the `use-lisp-fn` branch.


Wilfred:
Looks like this is just failing due to rustfmt, though I imagine we'll get the OS X failure once rustfmt is happy.

rust_src/remacs-macros/function.rs
```diff
@@ -78,81 +87,29 @@ pub fn get_fn_arg_ident(fn_arg: &syn::FnArg) -> Result<syn::Ident> {
...
                     ArgType::Other => {
-                        return Err("lisp functions should contain only `LispObject`s");
+                        return Err("lisp functions should only have `LispObject` args");
                     },
                 }
             },
             _ => return Err("lisp functions cannot have `self` arguments"),
         }
     }
-
```
jeandudey: :100:, this looked like a mounstrosity. Thanks for doing this :smile:

rust_src/remacs-macros/lisp_attr.rs
```diff
@@ -1,63 +1,72 @@
...
-use std::collections::HashMap;
 use syn;
 use synom;
 
+use function::{Function, LispFnType};
+
+/// Arguments of the lisp_fn attribute.
 pub struct LispFnArgs {
+    /// Desired Lisp name of the function.
+    /// If not given, derived as the Rust name with "_" -> "-".
```
jeandudey: :+1: Very ingenious

birkenfeld:
Thanks for the reviews! I'll rebase this onto master, then any conflicts should be gone.

Wilfred:
@birkenfeld it's odd that the tests are still failing. I'll take a look. Would you mind pushing your old version of this branch somewhere? I'd like to bisect the various changes before they were squashed.

birkenfeld:
@Wilfred unfortunately I don't have that reflog with me currently. But it's still only Tramp, so your suggestion to disable the tests might be incomplete?

birkenfeld:
Merged via #203.

PR #194, opened 2017-05-29
Adding a README.md to better explain the purpose of the remacs-lib 

DavidDeSimone:
New contributors should be able to read this README to fully understand what code should live in remacs-lib.


Wilfred:
👍

PR #195, opened 2017-05-29
Converting 'count-trailing-zeros' and 'count_one_bits' to Rust

DavidDeSimone:
This PR will allow us to delete:
 ```
lib/count-one-bits.c
lib/count-one-bits.h
lib/count-trailing-zeros.c
lib/count-trailing-zeros.h
```
By replacing the functionality provided in these files by function provided by the Rust standard library. 

I will submit the deletion of those files as a separate PR, as it looks like people (including myself!) have been removing files from lib/ without running gnulib-tool to re-generate the make files. Running gnulib-tool now with the proper generation command leads to some slightly incorrect makefiles, and that will need to be corrected.

### DIFF
--- a/rust_src/remacs-lib/math.rs
+++ b/rust_src/remacs-lib/math.rs
@@ -1,4 +1,4 @@
-extern crate libc;
+use libc;
 
 // Macro used to generate a c function that wraps functionality provided by the Rust stdlib
 // for u32 and other unsigned types.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: rust_src/remacs-lib/math.rs [hunk 1]
```diff
@@ -0,0 +1,22 @@
+extern crate libc;
```
jeandudey: I think this should be on the crate root and import it in submodules with `use libc;`

DavidDeSimone:
@jeandudey Thanks for the feedback, I've updated the PR.

Wilfred:
Nice!

> I will submit the deletion of those files as a separate PR, as it looks like people (including myself!) have been removing files from lib/ without running gnulib-tool to re-generate the make files

I wonder if we can configure Travis to make sure we don't forget in future. I think I'm guilty of this too.

PR #205, opened 2017-06-11
Add docker development environment

Fuco1:
For the people like me who don't like tainting the host system with superfluous nonsense like dev libraries :)

It also makes it simpler to get up and running, virtually requiring only `docker-compose up -d` and you are set to hack.

The image is rather big (about 1 GB) but this is not a problem because it is a build environment and will not be used for distribution.

Instructions are in the README.


docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+
+services:
+  remacs:
+    build: .
+    image: remacs-build:latest
+    working_dir: ${PWD}
+    tty: true
+    # Unfortunately dumping doesn't work without this
```
Wilfred: Presumably this is due to: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=23529 (see also [this](https://emacs.stackexchange.com/a/22667/304)).

Fuco1: Probably, but that will never get fixed, not until the old timers run things :/

Maybe one day remacs will fix it itself, or just remove the misfeature completely.

Wilfred: Yep, I was mostly linking it for future reference :)

Eventually I want Remacs to have a sensible dumping mechanism, at which point this will just work.

Issue #193, opened 2017-05-29
Move base64 lisp functions from C completely


jeandudey:
Move base64 lisp functions from C completely

Currently in Rust two functions are exported to C, these functions are `base64_encode_1` and `base64_decode_1` and are defined in `rust_src/src/base64.rs`. These functions are reused too in `src/fns.c`in the lisp native functions `base64-encode-region` and `base64-decode-region`.

Ideally these functions could be moved to Rust in the `base64.rs` module removing the need of intermediary functions like `base64_encode_1` that are unsafe and deal with raw pointers. Also these unsafe functions are reused too in the `strings.rs` module.

Here is a list of what needs to be done (in order):

1. [ ] Move `base64-encode-region` and `base64-decode-region` to the Rust codebase. Probably these functions should belong to the `base64.rs` module.
2. [ ] Refactor `base64_encode_1` and `base64_decode_1` and make them more safe and clean.
3. [x] Move the `base64-encode-string` and `base64-encode-string` defined in `strings.rs` to the `base64.rs` module.
4. [x] **Optional:** modify the code to use the `base64` crate and drop the `rustc-serialize` dependency.

mjfeller:
I spent yesterday reading through the available docs and source. I think I'm going to give this one a go.

I'm still learning rust and Remacs, but this does look like a good first bug for me to try and tackle.

Wilfred:
I've updated the checklist as several of these have been done now.

@jeandudey `base64_encode_1` has had some refactoring, do you think it's sufficiently safe and clean now, or did you have further changes in mind?

Wilfred:
I'm going to close this as our base64 implementation is now pure Rust.

Feel free to open an issue if you have ideas for further cleanup :)

PR #209, opened 2017-06-18
Move from (deprecated) rustc-serialize to base64 crate

birkenfeld:
Part of #193.

### DIFF
--- a/REMACS_COMPATIBILITY.md
+++ b/REMACS_COMPATIBILITY.md
@@ -4,13 +4,6 @@ Remacs should generally behave identically to GNU Emacs. If you find a
 difference, please
 [file a bug](https://github.com/Wilfred/remacs/issues/new).
 
-## Differences
-
-   - `base64-encode-string` and `base64-decode-string` does not fail
-     on multibyte strings.
-   - `base64-encode-region` and `base64-decode-region` does not fail
-     on multibyte regions.
-
 ## Detecting Remacs
 
 You can detect if your current Emacs instance is Remacs by running the
--- a/rust_src/Cargo.lock
+++ b/rust_src/Cargo.lock
@@ -3,13 +3,13 @@ name = "remacs"
 version = "0.1.0"
 dependencies = [
  "alloc_unexecmacosx 0.1.0",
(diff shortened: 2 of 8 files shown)


ljos:
If I remember correctly, the error here in `lisp/net/sasl-scram-rfc-tests` is because the base64 crate is not able to decode the string "fyko+d2lbbFgONRv9qkxdawL3rfcNHYJY1ZVvWVs7j" as it is not actually validly encoded as base64. I do not know where this string comes for, but both rustc-serialize library and the emacs implementation decodes the string correctly. The string it decodes to is also very strange. I don't know how they got that string. It seems to me that there is something missing, or it is from a faulty base64 implementation. But as I said; rustc-serialize and emacs decode this string correctly as emacs expects.

ljos:
This was also the reason I chose the rustc-serialize library in the first place.

birkenfeld:
Interesting, thanks for the heads-up. I'll have to investigate that in detail.

birkenfeld:
Should work now. The problem was actually just that base64 did not allow embedded newlines in the default config.

REMACS_COMPATIBILITY.md
```diff
@@ -4,13 +4,6 @@ Remacs should generally behave identically to GNU Emacs. If you find a
 difference, please
 [file a bug](https://github.com/Wilfred/remacs/issues/new).
 
-## Differences
-
-   - `base64-encode-string` and `base64-decode-string` does not fail
```
Wilfred: Are we now erroring on multibyte strings? Whilst we're calling `error` in e.g. `base64_encode_string`, that was true before.

birkenfeld: Well, for *some* multibyte strings...

The difference is, Emacs only allows encoding ASCII + the Latin1 subset. Characters with codepoint 128-255, which are encoded by two bytes in the multibyte encoding (basically UTF-8), are reduced to one byte before the base64 transformation.

The current Rust implementation just encodes the individual multibyte encoding bytes. That means it can encode any string, but gets different results. That's why the vanilla test had to be changed.

I would argue that being compatible is the right thing to do here.

Wilfred: Makes sense, thanks.

PR #210, opened 2017-06-18
Converting 'string-to-unibyte' from C to Rust.

DavidDeSimone:
This PR converts 'string-to-unibyte' from C to Rust. I've also added const ptr getter methods to LispStringRef, since it seemed like we don't always want to take ownership of the string if we only want it as a const pointer.

### DIFF
--- a/rust_src/src/multibyte.rs
+++ b/rust_src/src/multibyte.rs
@@ -87,6 +87,14 @@ impl LispStringRef {
         self.data as *mut c_char
     }
 
+    pub fn const_data_ptr(&self) -> *const c_uchar {
+        self.data as *const c_uchar
+    }
+
+    pub fn const_sdata_ptr(&self) -> *const c_char {
+        self.data as *const c_char
+    }
+
     #[inline]
     pub fn as_slice(&self) -> &[u8] {
         unsafe { slice::from_raw_parts(self.data as *const u8, self.len_bytes() as usize) }
--- a/rust_src/src/strings.rs
+++ b/rust_src/src/strings.rs
@@ -7,7 +7,8 @@ use libc;
 use lisp::LispObject;
 use multibyte;
 use remacs_sys::{SYMBOL_NAME, EmacsInt, error, base64_encode_1, base64_decode_1, make_string,
-                 make_uninit_multibyte_string, string_to_multibyte as c_string_to_multibyte};
+                 make_uninit_multibyte_string, string_to_multibyte as c_string_to_multibyte,
+                 make_unibyte_string};
 use remacs_macros::lisp_fn;
 
 pub static MIME_LINE_LENGTH: isize = 76;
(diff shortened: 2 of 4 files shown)


src/fns.c
```diff
@@ -979,33 +979,6 @@ If STRING is multibyte and contains a character of charset
...
-  CHECK_STRING (string);
-
-  if (STRING_MULTIBYTE (string))
-    {
-      ptrdiff_t chars = SCHARS (string);
-      unsigned char *str = xmalloc (chars);
-      ptrdiff_t converted = str_to_unibyte (SDATA (string), str, chars);
-
-      if (converted < chars)
-	error ("Can't convert the %"pD"dth character to unibyte", converted);
```
DavidDeSimone: It took me a while to figure out that "pD" is a macro. Based on this comment in src/lisp.h
```
/* Use pD to format ptrdiff_t values, which suffice for indexes into
   buffers and strings.  Emacs never allocates objects larger than
   PTRDIFF_MAX bytes, as they cause problems with pointer subtraction.
   In C99, pD can always be "t"; configure it here for the sake of
   pre-C99 libraries such as glibc 2.0 and Solaris 8.  */
```
It looks like we can just make our rust code use %td for ptrdiff_t's?

rust_src/src/strings.rs
```diff
@@ -169,3 +170,35 @@ fn string_to_multibyte(string: LispObject) -> LispObject {
     let _ = string.as_string_or_error();
     unsafe { LispObject::from_raw(c_string_to_multibyte(string.to_raw())) }
 }
+
+/// Return a unibyte string with the same individual chars as STRING.
+/// -If STRING is unibyte, the result is STRING itself.
```
birkenfeld: Stray "-" left from the diff.

rust_src/src/multibyte.rs
```diff
@@ -87,6 +87,14 @@ impl LispStringRef {
         self.data as *mut c_char
     }
 
+    pub fn const_data_ptr(&self) -> *const c_uchar {
```
birkenfeld: These don't add much value, I'd just cast to `*const` where needed. Or can't you anyway pass a `*mut T` where a `*const T` is requested?

rust_src/src/strings.rs
```diff
@@ -169,3 +170,35 @@ fn string_to_multibyte(string: LispObject) -> LispObject {
...
+/// -Otherwise it is a newly created string, with no text properties,
+/// -where each `eight-bit' character is converted to the corresponding byte.
+/// -If STRING contains a non-ASCII, non-`eight-bit' character,
+/// -an error is signaled.
+#[lisp_fn]
+fn string_to_unibyte(string: LispObject) -> LispObject {
+    let mut string_mut = string;
```
birkenfeld: You can just put `mut string` in the function signature.
Or even better, write
```
if is_multibyte() {
    ...
    LispObject::from_raw(raw_ptr)
} else {
    string
}
```

rust_src/src/strings.rs
```diff
@@ -169,3 +170,35 @@ fn string_to_multibyte(string: LispObject) -> LispObject {
...
+    if lispstr.is_multibyte() {
+        let size = lispstr.len_bytes();
+        let mut buffer: Vec<libc::c_uchar> = Vec::with_capacity(size as usize);
+        let converted_size =
+            multibyte::str_to_unibyte(lispstr.const_data_ptr(), buffer.as_mut_ptr(), size);
+
+        unsafe {
+            if converted_size < size {
+                error(
+                    "Can't convert %dth character to unibyte\0".as_ptr(),
```
birkenfeld: `%td` as your comment suggests?

rust_src/src/multibyte.rs
DavidDeSimone: The value added by these functions in my mind isn't the return value, but in the argument. In order to get the data pointer currently, I **need** to do a mutable borrow. This would mean that in any case that I need to access data, I need mutability. This would translate into us making a lot of variables mutable that don't need to be, and could make structuring code awkward when the borrow checker enforces 'only one mutable borrow at a time'.

For example, if I use 'data_ptr', the variable lispstr must now be marked mutable, for no other reason except to satisfy data_ptr's need to do a mutable borrow.

rust_src/src/strings.rs
DavidDeSimone: So after further investigation, it looks like I may not fully understand what that comment is saying. On my system, that macro expands pD to "l", not "t". And in fact hard coding this template to %td instead of %d or %ld leads to the runtime error: 

`(error "Invalid format operation %t")`

I suppose this is something that we may want to define as a rust accessible variable?  I'm not sure the best way to handle this concept.

rust_src/src/strings.rs
DavidDeSimone: Done

rust_src/src/strings.rs
DavidDeSimone: Done

PR #213, opened 2017-06-20
moved vector functions into rust

mjfeller:
I've recreated this PR with a fresh branch from the master. Everything should be implemented and good to go.

Let me know if there is anything I've missed or can clean up. I'm still very new to Rust and Remacs

### DIFF
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -485,6 +485,12 @@ impl LispObject {
         })
     }
 
+    pub fn is_font(self) -> bool {
+        self.as_vectorlike().map_or(false, |v| {
+            v.is_psuedovector(PseudovecType::PVEC_FONT)
+        })
+    }
+
     pub fn is_font_spec(self) -> bool {
         // return FONTP (x) && (ASIZE (x) & PSEUDOVECTOR_SIZE_MASK) == FONT_SPEC_MAX;
         unimplemented!()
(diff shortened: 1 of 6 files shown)


rust_src/src/lisp.rs
```diff
@@ -394,6 +394,101 @@ impl LispObject {
...
+
+    pub fn is_array(self) -> bool {
+        self.is_vector() || self.is_string() || self.is_char_table() || self.is_bool_vector()
+    }
+
+    pub fn is_sequence(self) -> bool {
+        self.is_cons() || self.is_nil() || self.is_array()
+    }
+
+    pub fn is_configuration(self) -> bool {
```
birkenfeld: I'd prefer `is_window_configuration`, even if it's long, for clarity.

rust_src/src/lisp.rs
```diff
@@ -394,6 +394,101 @@ impl LispObject {
...
+
+    pub fn is_bool_vector(self) -> bool {
+        self.as_vectorlike().map_or(
+            false,
+            |v| v.is_pseudovector(PseudovecType::PVEC_BOOL_VECTOR),
+        )
+    }
+
+    pub fn is_array(self) -> bool {
+        self.is_vector() || self.is_string() || self.is_char_table() || self.is_bool_vector()
```
birkenfeld: This *could* be made more efficient by only matching once for `is_vectorlike`, and then ORing the subtype checks.

But it's probably not worth the slightly smaller readability, or even optimized away in release.

mjfeller: I'll leave it as is for now unless someone says it should be changed.

PR #214, opened 2017-06-20
travis formatting for remacs-macros

mjfeller:
I updated the travis script to format remacs-macros.

Cargo fmt would hang when run as
```bash
cargo fmt -- --write-mode=diff
```
but by specifying the base file to format seems to prevent it from hanging.
```bash
cargo fmt -- --write-mode=diff lib.rs
```

example:
```bash
remacs/rust_src/remacs-macros
▶ cargo fmt -- -v --write-mode=diff lib.rs
Formatting /Users/mark/prog/remacs/rust_src/remacs-macros/function.rs
Formatting /Users/mark/prog/remacs/rust_src/remacs-macros/lib.rs
Formatting /Users/mark/prog/remacs/rust_src/remacs-macros/lisp_attr.rs
```

So the question is, is this an acceptable fix?

I also ran cargo fmt over the remacs-macros source and updated the readme to reflect the additions of the format checker.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -24,8 +24,7 @@ cd "$DIR/rust_src/remacs-lib"
 cargo fmt -- --write-mode=diff
 
 cd "$DIR/rust_src/remacs-macros"
-# TODO: not yet - cargo fmt seems to enter an infinite loop here.
-# cargo fmt -- --write-mode=diff
+cargo fmt -- --write-mode=diff lib.rs
 
 cd "$DIR/rust_src/alloc_unexecmacosx"
 cargo fmt -- --write-mode=diff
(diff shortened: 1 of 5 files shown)


mjfeller:
I've looked through the upstream bug reports and didn't see anything yet. I haven't done an exhaustive search by any means. I'll do a little more looking and maybe submit an issue asking about this.

birkenfeld:
Did some investigating and found the issue: `cargo fmt` did not recognize "proc-macro" crates.

https://github.com/rust-lang-nursery/rustfmt/pull/1738 should fix it.

Wilfred:
Merged, thanks.

PR #215, opened 2017-06-22
Port overlayp and buffer-live-p to rust

shanavas786:


birkenfeld:
Looks good, just needs a `cargo fmt`

birkenfeld:
Thanks! The remaining failure is because some test is looking for the buffer-live-p defun in C. We can just point it to another defun for now.

shanavas786:
Exporting `Fbuffer_live_p` is not enough ?

How can we point it to another defun ?

Wilfred:
Looks great, thank you :)

PR #217, opened 2017-06-25
Port 'string-lessp' to Rust

DavidDeSimone:
This is a WIP PR for porting string-lessp to Rust. As part of this work, it seemed useful to add an iterator for a LispStringRef, in order to replace the macro FETCH_STRING_CHAR_ADVANCE

Work left to do:

- [x] Port basic string-lessp functionality to Rust
- [x] Remove string-lessp implmentation in C.
- [x] Add code point iterator for LispStringRef
- [x] Add safe API for working with LispSymbol's, similar to LispStringRef/as_string/as_string_or_error
   - [x] Fix issue where the value coming out of the LispSymbolRef's mem::transmute seem to be nonsense
- [x] Verify that the alignment/size/padding of Rust's Lisp_Symbol matches the alignment/size/padding of C's Lisp_Symbol
- [x] Basic profiling to make sure that these new concepts do not add significant cost over their C equivalents. (Based on my findings, performance for this function is equivalent to it's C counterpart)

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -665,9 +665,28 @@ pub struct Lisp_String {
     pub data: *mut libc::c_char,
 }
 
-// @TODO
+#[repr(C)]
+pub union SymbolUnion {
+    pub value: Lisp_Object,
+    pub alias: *mut Lisp_Symbol,
+    pub blv: *mut libc::c_void, // @TODO implement Lisp_Buffer_Local_Value
+    pub fwd: *mut libc::c_void, // @TODO implement Lisp_Fwd
+}
+
+/// This struct has 4 bytes of padding, representing the bitfield that 
+/// lives at the top of a Lisp_Symbol. The first 10 bits of this field are
+/// used
+
+// @TODO check the value of name post and pre transmutation, it seems that name is surviving but
+// may not be the correct value
 #[repr(C)]
 pub struct Lisp_Symbol {
+    pub symbol_bitfield: u32,
+    pub name: Lisp_Object,
+    pub val: SymbolUnion,
+    pub function: Lisp_Object,
+    pub plist: Lisp_Object,
+    pub next: *mut Lisp_Symbol,
 }
 
 extern "C" {
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -676,9 +676,6 @@ pub union SymbolUnion {
 /// This struct has 4 bytes of padding, representing the bitfield that 
 /// lives at the top of a Lisp_Symbol. The first 10 bits of this field are
 /// used
-
-// @TODO check the value of name post and pre transmutation, it seems that name is surviving but
-// may not be the correct value
 #[repr(C)]
 pub struct Lisp_Symbol {
     pub symbol_bitfield: u32,
@@ -732,7 +729,8 @@ extern "C" {
     pub static Qfont_spec: Lisp_Object;
     pub static Qfont_entity: Lisp_Object;
     pub static Qfont_object: Lisp_Object;
-
+    pub static lispsym: Lisp_Symbol;
+    
     pub fn Fcons(car: Lisp_Object, cdr: Lisp_Object) -> Lisp_Object;
     pub fn Fcurrent_buffer() -> Lisp_Object;
     pub fn Fget_buffer(buffer_or_name: Lisp_Object) -> Lisp_Object;
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -665,6 +665,11 @@ pub struct Lisp_String {
     pub data: *mut libc::c_char,
 }
 
+// @TODO
+#[repr(C)]
+pub struct Lisp_Symbol {
+}
+
 extern "C" {
     pub static mut globals: emacs_globals;
     pub static Qt: Lisp_Object;
@@ -675,6 +680,7 @@ extern "C" {
     pub static Qintegerp: Lisp_Object;
     pub static Qfloatp: Lisp_Object;
     pub static Qstringp: Lisp_Object;
+    pub static Qsymbolp: Lisp_Object;
     pub static Qlistp: Lisp_Object;
     pub static Qmarkerp: Lisp_Object;
     pub static Qwholenump: Lisp_Object;
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -669,11 +669,11 @@ pub struct Lisp_String {
 pub union SymbolUnion {
     pub value: Lisp_Object,
     pub alias: *mut Lisp_Symbol,
-    pub blv: *mut libc::c_void, // @TODO implement Lisp_Buffer_Local_Value
-    pub fwd: *mut libc::c_void, // @TODO implement Lisp_Fwd
+pub blv: *mut libc::c_void, // @TODO implement Lisp_Buffer_Local_Value
+pub fwd: *mut libc::c_void, // @TODO implement Lisp_Fwd
 }
 
-/// This struct has 4 bytes of padding, representing the bitfield that 
+/// This struct has 4 bytes of padding, representing the bitfield that
 /// lives at the top of a Lisp_Symbol. The first 10 bits of this field are
 /// used
 #[repr(C)]
@@ -730,7 +730,7 @@ extern "C" {
     pub static Qfont_entity: Lisp_Object;
     pub static Qfont_object: Lisp_Object;
     pub static lispsym: Lisp_Symbol;
-    
+
     pub fn Fcons(car: Lisp_Object, cdr: Lisp_Object) -> Lisp_Object;
     pub fn Fcurrent_buffer() -> Lisp_Object;
     pub fn Fget_buffer(buffer_or_name: Lisp_Object) -> Lisp_Object;
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -174,6 +174,14 @@ impl LispObject {
             unsafe { wrong_type_argument(Qsymbolp, self.to_raw()) }
         }
     }
+
+    #[inline]
+    pub fn as_string_or_symbol(string: LispObject) -> LispStringRef {
+        match string.as_symbol() {
+            Some(sym) => sym.symbol_name().as_string().expect("Expected a symbol name?"),
+            None => string.as_string_or_error()
+        }
+    }
 }
 
 // Misc support (LispType == Lisp_Misc == 1)
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -19,10 +19,10 @@ use vectors::LispVectorlikeRef;
 use buffers::LispBufferRef;
 
 use remacs_sys::{EmacsInt, EmacsUint, EmacsDouble, EMACS_INT_MAX, EMACS_INT_SIZE,
-                 EMACS_FLOAT_SIZE, USE_LSB_TAG, GCTYPEBITS, wrong_type_argument, Qstringp, Qsymbolp,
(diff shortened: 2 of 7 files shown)


rust_src/src/lisp.rs
```diff
@@ -155,6 +155,10 @@ impl LispObject {
     pub fn is_symbol(self) -> bool {
         self.get_type() == LispType::Lisp_Symbol
     }
+
+    pub fn symbol_name(&self) -> LispObject {
```
birkenfeld: This is not safe - it'll treat the `LispObject` as a symbol regardless of its type.

You'll have to introduce a `LispSymbolRef` newtype and respective `as_symbol` and `as_symbol_or_error` methods on `LispObject` to convert safely.

rust_src/src/multibyte.rs
```diff
@@ -106,6 +106,22 @@ impl LispStringRef {
     }
 }
 
+// Substitue for FETCH_STRING_CHAR_ADVANCE
+impl Iterator for LispStringRef {
```
birkenfeld: Great idea to use an iterator here!

rust_src/src/strings.rs
```diff
@@ -202,6 +202,43 @@ fn string_to_unibyte(string: LispObject) -> LispObject {
...
+    let mut i1 = 0;
+    while i1 < end {
+        // Unwraps should be fine here, due to our manual tracking of
+        // valid length
+        let (codept1, _) = lispstr1.next().unwrap();
+        let (codept2, _) = lispstr2.next().unwrap();
+        i1 += 1;
+        
+        if codept1 != codept2 {
+            if codept1 < codept2 {
```
birkenfeld: This is just `return LispObject::from_bool(codept1 < codept2)`

rust_src/src/strings.rs
```diff
@@ -202,6 +202,43 @@ fn string_to_unibyte(string: LispObject) -> LispObject {
...
+                return LispObject::constant_t();
+            } else {
+                return LispObject::constant_nil();
+            }
+        }
+    }
+
+    if i1 < lispstr2.len_bytes() {
```
birkenfeld: same here (without `return`)

rust_src/src/strings.rs
```diff
@@ -202,6 +202,43 @@ fn string_to_unibyte(string: LispObject) -> LispObject {
...
+}
+
+fn string_lessp(string1: LispObject, string2: LispObject) -> LispObject {
+    let mut lispstr1 = get_string_or_symbol(string1);
+    let mut lispstr2 = get_string_or_symbol(string2);
+
+    let end = cmp::min(lispstr1.len_bytes(), lispstr2.len_bytes());
+    let mut i1 = 0;
+    while i1 < end {
+        // Unwraps should be fine here, due to our manual tracking of
```
birkenfeld: hmm, could this be done with a `zip`? I don't think we need `i1` afterwards since if we finish the loop it will always be `= end`.

rust_src/src/lisp.rs
DavidDeSimone: Good callout, I will do this next.

rust_src/src/strings.rs
DavidDeSimone: Also another good call out, I've changed the implementation to use zip

rust_src/src/multibyte.rs
```diff
@@ -106,6 +106,44 @@ impl LispStringRef {
...
+    fn next(&mut self) -> Option<(Codepoint, usize)> {
+        if self.cur < self.string_ref.len_bytes() as usize {
+            let codepoint: Codepoint;
+            let ref_slice = self.string_ref.as_slice();
+            if self.string_ref.is_multibyte() {
+                let (cp, advance) = multibyte_char_at(&ref_slice[self.cur..]);
+                codepoint = cp;
+                self.cur += advance;
+            } else {
+                codepoint = ref_slice[self.cur] as u32;
```
birkenfeld: .. as Codepoint?

rust_src/src/multibyte.rs
```diff
@@ -106,6 +106,44 @@ impl LispStringRef {
...
+                self.cur += 1;
+            }
+            
+            Some((codepoint, self.cur))
```
birkenfeld: Hm, this returns the index of the *next* codepoint. Is this intended?

rust_src/src/multibyte.rs
```diff
@@ -106,6 +106,44 @@ impl LispStringRef {
...
+        } else {
+            None
+        }
+    }
+}
+
+impl LispStringRef {
+    pub fn iter(&self) -> LispStringRefIterator {
```
birkenfeld: Depending on need by other APIs, I think two iterators a la `chars()` and `char_indices()` would make sense.

rust_src/src/strings.rs
```diff
@@ -202,6 +202,32 @@ fn string_to_unibyte(string: LispObject) -> LispObject {
...
+    let mut count = 0;
+    let zip = lispstr1.iter().zip(lispstr2.iter());
+    for ((codept1, _), (codept2, _)) in zip {
+        count += 1;
+        if codept1 != codept2 {
+            return LispObject::from_bool(codept1 < codept2);
+        }
+    }
+
+    LispObject::from_bool(count < lispstr2.len_chars())
```
birkenfeld: I think this is just `lispstr1.len_chars() < lispstr2.len_chars()`, and keeping `count` is unnecessary.

(If lispstr2 is shorter or equally long, this will test `len2 < len2` which is false.)

rust_src/src/strings.rs
```diff
@@ -1,6 +1,6 @@
 //! Functions operating on strings.
 
-use std::ptr;
+use std::{ptr, cmp};
```
birkenfeld: unused import?

rust_src/src/lisp.rs
birkenfeld: Thanks!

rust_src/src/strings.rs
```diff
@@ -202,10 +202,9 @@ fn string_to_unibyte(string: LispObject) -> LispObject {
     }
 }
 
-// @TODO need to rework this function to use new as_symbol_or_error API
 fn get_string_or_symbol(mut string: LispObject) -> multibyte::LispStringRef {
     if string.is_symbol() {
-        string = string.symbol_name()
+        string = string.as_symbol_or_error().symbol_name()
```
birkenfeld: This still duplicates the symbol check. I'd make it
```
match string.as_symbol() {
    Some(sym) => sym.symbol_name().as_string().expect("symbol name not a string?")
    None => string.as_string_or_error()
}
```
If you like, you can also make `string_equal` use this function, it currently uses SYMBOL_NAME directly.

rust_src/src/lisp.rs
```diff
@@ -150,11 +151,37 @@ impl LispObject {
...
 impl LispObject {
+    #[inline]
     pub fn is_symbol(self) -> bool {
         self.get_type() == LispType::Lisp_Symbol
     }
+
+    #[inline]
+    pub fn as_symbol(&self) -> Option<LispSymbolRef> {
+        if self.is_symbol() {
+            Some(LispSymbolRef::new(unsafe { mem::transmute(self.get_untaggedptr()) }))
```
DavidDeSimone: This looks to be incorrect. The code in lisp.h for XSYMBOL looks like
```
  intptr_t i = (intptr_t) XUNTAG (a, Lisp_Symbol);
  void *p = (char *) lispsym + i;
  return p;
```

and it looks like we will have to emulate this logic for getting the address to mem::transmute.

rust_src/remacs-sys/lib.rs
```diff
@@ -665,6 +665,30 @@ pub struct Lisp_String {
...
+}
+
+/// This struct has 4 bytes of padding, representing the bitfield that 
+/// lives at the top of a Lisp_Symbol. The first 10 bits of this field are
+/// used
+
+// @TODO check the value of name post and pre transmutation, it seems that name is surviving but
+// may not be the correct value
+#[repr(C)]
+pub struct Lisp_Symbol {
```
shanavas786: gcmarkbit, redirect ...etc are missing ?

rust_src/src/multibyte.rs
DavidDeSimone: This sounds reasonable to me, I like the idea of a chars() and char_indicies().

rust_src/remacs-sys/lib.rs
DavidDeSimone: So this is my understanding of the situation re: the mentioned variables. If anyone spots any incorrect information please do not hesitate to correct me:

Lisp_Symbol has the following definition in lisp.h (I've stripped out the comments for clarity):
```
struct Lisp_Symbol
{
  bool_bf gcmarkbit : 1;
  ENUM_BF (symbol_redirect) redirect : 3;
  ENUM_BF (symbol_trapped_write) trapped_write : 2;
  unsigned interned : 2;
  bool_bf declared_special : 1;
  bool_bf pinned : 1;
  Lisp_Object name;
  union {
    Lisp_Object value;
    struct Lisp_Symbol *alias;
    struct Lisp_Buffer_Local_Value *blv;
    union Lisp_Fwd *fwd;
  } val;
  Lisp_Object function;
  Lisp_Object plist;
  struct Lisp_Symbol *next;
```
This struct is using the C notation for bit fields. According to my research, Rust does not (https://github.com/rust-lang/rfcs/issues/314, https://users.rust-lang.org/t/c-structs-with-bit-fields-and-ffi/1429)  support setting up an equivalent for C bitfieldsin it's `#[repr(C)]` directive.

On my system (64-bit Ubuntu 16.04), no matter what the typedef of bf_bool or ENUM_BF, the bit field section of the struct takes 4 bytes. Due to alignment, the struct is padded, and offsetof(struct Lisp_Symbol, name) reports 8. My initial solution to representing this struct in Rust was to represent the bit field block as a u32, taking up the 4 bytes I mentioned earlier. I have not fully convinced myself that this is 100% the correct and portable thing to do. 

Even if we can safely represent this block with a u32 on every system we support, it seems that there are 'gotcha's with accessing these bit fields via the bit wise operators (due to endianness, and compiler differences w.r.t bit field implementation.) 

Overall I am not sure the best way to handle the interop for C structs that use bit fields that we need to access in Rust. It seems that one option is to simply pad the Rust struct as best we can, and if we need to access these fields, we will need to maintain C bindings that access them for us.

rust_src/src/multibyte.rs
```diff
@@ -136,13 +138,37 @@ impl<'a> Iterator for LispStringRefIterator<'a> {
...
             string_ref: self,
             cur: 0,
         }
     }
+
+    pub fn chars(&self) -> LispStringRefCharIterator {
+        LispStringRefCharIterator(self.iter())
+    }
+
+    pub fn char_indices(&self) -> LispStringRefIndexIterator {
```
birkenfeld: Sorry to nitpick again, but actually `char_indices` returns the `(index, char)` tuple - you just need to rename `iter`.

DavidDeSimone: No need to apologize, this should be fixed now.

DavidDeSimone:
Overall, I feel pretty confident with the PR now. If any maintainers have additional feedback, I will be happy to address their concerns, otherwise I think this is ready to merge.

rust_src/src/multibyte.rs
```diff
@@ -106,6 +106,59 @@ impl LispStringRef {
...
 }
 
+pub struct LispStringRefIterator<'a> {
+    string_ref: &'a LispStringRef,
+    cur: usize,
+}
+
+pub struct LispStringRefCharIterator<'a>(LispStringRefIterator<'a>);
+
+// Substitue for FETCH_STRING_CHAR_ADVANCE
```
Wilfred: *Substitute

rust_src/src/strings.rs
```diff
@@ -202,6 +202,21 @@ fn string_to_unibyte(string: LispObject) -> LispObject {
     }
 }
 
+#[lisp_fn]
```
Wilfred: Please put the docstring here.

Wilfred:
Marvellous :)

PR #227, opened 2017-07-14
Mid-July rollup

birkenfeld:


birkenfeld:
tramp tests on OSX are failing again... I don't remember what we did about that last time...

birkenfeld:
Thanks for merging. To be honest, I'm not very happy about the complete DOS support removal, because it makes the upstream<->remacs diff very noisy. I'd just have removed support from anything ported to Rust, and error out once while configuring.

Similarly, the removal of the ChangeLog files is a bit unfortunate since they generate a merge conflict on every rollup (which might be prudent to do more often).

PR #234, opened 2017-07-16
Port the rest of floatfns, remove bit-counting gnulib modules

birkenfeld:


DavidDeSimone:
Thanks for cleaning up the gnulib modules! I had a half finished patch to do that, but it looks like you already have it all taken care of.

rust_src/src/eval.rs
```diff
@@ -14,11 +10,21 @@ fn xsignal(error_symbol: LispObject, data: LispObject) -> ! {
...
-    unreachable!();
 }
 
 /// Convenience function for calling `xsignal` with an empty list.
 pub fn xsignal0(error_symbol: LispObject) -> ! {
     xsignal(error_symbol, Qnil);
 }
+
+/// Convenience function for calling `xsignal` with a one-element list.
+// pub fn xsignal1(error_symbol: LispObject, arg: LispObject) -> ! {
```
Wilfred: Why is this commented out?

rust_src/src/floatfns.rs
```diff
@@ -147,15 +176,220 @@ fn ftruncate(arg: LispObject) -> LispObject {
...
+    if libm::signbit(f1) != libm::signbit(f2) {
+        LispObject::from_float(-f1)
+    } else {
+        x1
+    }
+}
+
+/// Get significand and exponent of a floating point number.
+/// Breaks the floating point number X into its binary significand SGNFCAND
+/// \(a floating point value between 0.5 (included) and 1.0 (excluded))
```
Wilfred: I believe you don't actually need to escape the `(` here.

The Emacs convention for escaping parentheses is because something gets confused if you start a line with a `(` (I think it's either highlighting or indentation). The C code suffers from this:

```c
DEFUN ("frexp", Ffrexp, Sfrexp, 1, 1, 0,
       doc: /* Get significand and exponent of a floating point number.
Breaks the floating point number X into its binary significand SGNFCAND
\(a floating point value between 0.5 (included) and 1.0 (excluded))
and an integral exponent EXP for 2, such that:
```

but the Rust code doesn't, because the line starts with `//`.

rust_src/src/eval.rs
birkenfeld: It's currently not used - I could remove it or add a `allow(dead_code)` instead.

rust_src/src/floatfns.rs
birkenfeld: Right, I wasn't paying attention while copy-pasting :)

PR #235, opened 2017-07-16
A few more fns from fns.c

birkenfeld:


rust_src/src/numbers.rs
```diff
@@ -39,3 +47,40 @@ fn numberp(object: LispObject) -> LispObject {
...
+/// With positive integer LIMIT, return random number in interval [0,LIMIT).
+/// With argument t, set the random number seed from the system's entropy
+/// pool if available, otherwise from less-random volatile data such as the time.
+/// With a string argument, set the seed based on the string's contents.
+/// Other values of LIMIT are ignored.
+///
+/// See Info node `(elisp)Random Numbers' for more details.
+#[lisp_fn(min = "0")]
+fn random(limit: LispObject) -> LispObject {
+    let mut rng = RNG.lock().unwrap();
```
Wilfred: I presume that this will change the output of the random number generator in Remacs, even when seeded with the same value as GNU Emacs? If so, we should mention it in REMACS_COMPATIBILITY.md

birkenfeld: Yes, although as far as I can see Emacs uses the C library's generator, which I assume does not guarantee any specific sequence either.

Wilfred: Aha, OK.

PR #236, opened 2017-07-16
Port multibyte-char-to-unibyte

shanavas786:


rust_src/src/multibyte.rs
```diff
@@ -191,6 +191,18 @@ fn raw_byte_from_codepoint(cp: Codepoint) -> c_uchar {
     (cp - 0x3F_FF00) as c_uchar
 }
 
+/// Same as the CHAR_TO_BYTE_SAFE macro.
+#[inline]
+pub fn raw_byte_from_codepoint_safe (cp: Codepoint) -> c_uchar {
```
birkenfeld: The space before the argument list here is causing failure due to rustfmt.

rust_src/src/multibyte.rs
```diff
@@ -191,6 +191,18 @@ fn raw_byte_from_codepoint(cp: Codepoint) -> c_uchar {
...
+    if cp < 0x80 {
```
birkenfeld: Should be 0x100? (And maybe include the comment from the C version.)

rust_src/src/multibyte.rs
```diff
@@ -191,6 +191,18 @@ fn raw_byte_from_codepoint(cp: Codepoint) -> c_uchar {
...
+        cp as c_uchar
+    } else if cp > MAX_5_BYTE_CHAR {
+        (cp - 0x3F_FF00) as c_uchar
```
birkenfeld: this should use `multibyte::raw_byte_from_codepoint`

rust_src/src/multibyte.rs
shanavas786: why 0x100? ASCII_P compares with 0x80

birkenfeld: But there's an "if c < 256" in the C source of `multibyte-char-to-unibyte`.

shanavas786: Yeah, updated

PR #237, opened 2017-07-16
Port windowp to Rust

brotzeit:
I followed the tutorial and it seems to work. But I have no idea how the function is_window actually works.

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -4,7 +4,7 @@ use lisp::LispObject;
 use remacs_macros::lisp_fn;
 
 
-/// Return t if OBJECT is a window.
+/// Return t if OBJECT is a window and nil otherwise.
 #[lisp_fn]
 fn windowp(object: LispObject) -> LispObject {
     LispObject::from_bool(object.is_window())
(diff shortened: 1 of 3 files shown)


rust_src/src/windows.rs
```diff
@@ -0,0 +1,11 @@
+//! Functions operating on windows.
+
+use lisp::LispObject;
+use remacs_macros::lisp_fn;
+
+
+/// Return t if OBJECT is a window.
```
birkenfeld: The docstring should be copied exactly from the C function.

Wilfred:
@brotzeit did you mean to close this?

If you're wondering, `is_window` is defined in lisp.rs and looks like this:

```rust
    pub fn is_window(self) -> bool {
        self.as_vectorlike().map_or(false, |v| {
            v.is_pseudovector(PseudovecType::PVEC_WINDOW)
        })
    }
```

This looks at the value, and returns false if it's not a psuedovector. If it is a psuedovector, we see if it has a pseudovector type of `PVEC_WINDOW`, which is how Emacs represents windows.

brotzeit:
I see. Thanks for the explanation. I'm a noob ;)

PR #239, opened 2017-07-16
Refactor math functions a bit with more Rusty matching and less checking

birkenfeld:
Also port float_arith_driver to avoid awkward recasting as raw lisp objects.

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -700,6 +700,7 @@ extern "C" {
     pub static Qt: Lisp_Object;
     pub static Qarith_error: Lisp_Object;
     pub static Qnumber_or_marker_p: Lisp_Object;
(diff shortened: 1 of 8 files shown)


rust_src/remacs-sys/lib.rs
```diff
@@ -700,6 +700,7 @@ extern "C" {
     pub static Qt: Lisp_Object;
     pub static Qarith_error: Lisp_Object;
     pub static Qnumber_or_marker_p: Lisp_Object;
+    pub static Qinteger_or_marker_p: Lisp_Object;
```
Wilfred: Sooner or later we should look at defining built-in symbols in Rust. It's the main thing stopping us writing useful unit tests.

birkenfeld: Yep, and also gets rid of the `unsafe` wherever you use them.

PR #246, opened 2017-07-17
Adding nightly validation to autogen.sh. 

DavidDeSimone:
We have had a number of new contributors not realize that remacs requires nightly rust, and this has caused problems. This should make the build fail fast if nightly is not installed.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -30,13 +30,16 @@
 
 ### Code:
 
+## Used when remacs is dependent on nightly
+## Will print an error message telling people they need nightly rust
+require_nightly=true
+
 ## Tools we need:
 ## Note that we respect the values of AUTOCONF etc, like autoreconf does.
 progs="autoconf"
 
 ## Minimum versions we need:
 autoconf_min=`sed -n 's/^ *AC_PREREQ(\([0-9\.]*\)).*/\1/p' configure.ac`
-
 
 ## $1 = program, eg "autoconf".
 ## Echo the version string, eg "2.59".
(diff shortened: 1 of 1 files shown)


autogen.sh
```diff
@@ -130,6 +133,15 @@ case $do_autoconf,$do_git in
     do_autoconf=true;;
 esac
 
+if $require_nightly; then
+    echo "Validation Rust install ..."
```
birkenfeld: "Validating"?

autogen.sh
```diff
@@ -130,6 +133,15 @@ case $do_autoconf,$do_git in
...
+    command -v rustc >/dev/null 2>&1 || { echo >&2 "Remacs requires rust to be installed in order to build. Please install it via rustup: https://www.rustup.rs/; Aborting."; exit 1; }
+    command -v cargo >/dev/null 2>&1 || { echo >&2 "I require foo but it's not installed.  Aborting.Remacs requires cargo to be installed in order to build. Please install it via rustup: https://www.rustup.rs/; Aborting."; exit 1; }
```
birkenfeld: message seems jumbled?

autogen.sh
```diff
@@ -130,6 +133,15 @@ case $do_autoconf,$do_git in
...
+
+    rustc --version | grep nightly || { echo >&2 "Remacs currently requires nightly Rust. If you do not have nightly Rust, you should install rustup at https://www.rustup.rs/, and run 'rustup install nightly && rustup default nightly"; exit 1; }
```
birkenfeld: I'd suggest `rustup override ...` instead, to make the nightly choice only for Remacs.

autogen.sh
DavidDeSimone: Whoops, bade copy/paste, let me fix that.

autogen.sh
```diff
@@ -130,6 +134,15 @@ case $do_autoconf,$do_git in
     do_autoconf=true;;
 esac
 
+if $require_nightly; then
+    echo "Validating Rust install ..."
+    command -v rustc >/dev/null 2>&1 || { echo >&2 "Remacs requires rust to be installed in order to build. Please install it via rustup: https://www.rustup.rs/; Aborting."; exit 1; }
```
birkenfeld: hmm, actually these checks are not depending on nightly/stable, should they be moved outside the `if`?

autogen.sh
```diff
@@ -130,6 +134,15 @@ case $do_autoconf,$do_git in
...
+    command -v cargo >/dev/null 2>&1 || { echo >&2 "Remacs requires cargo to be installed in order to build. Please install it via rustup: https://www.rustup.rs/; Aborting."; exit 1; }
+
+    rustc --version | grep nightly || { echo >&2 "Remacs currently requires nightly Rust. If you do not have nightly Rust, you should install rustup at https://www.rustup.rs/, and run 'rustup install nightly && rustup override nightly' in the remacs directory."; exit 1; }
```
birkenfeld: it's `rustup override set nightly` I think.

DavidDeSimone: you are correct, updating

autogen.sh
DavidDeSimone: good callout

Wilfred:
Excellent, thanks :)

PR #248, opened 2017-07-17
Port fontp to Rust

### LINKED ISSUE #233: Port fontp to Rust
Relevant C code:

```c
DEFUN ("fontp", Ffontp, Sfontp, 1, 2, 0,
       doc: /* Return t if OBJECT is a font-spec, font-entity, or font-object.
Return nil otherwise.
Optional 2nd argument EXTRA-TYPE, if non-nil, specifies to check
which kind of font it is.  It must be one of `font-spec', `font-entity',
`font-object'.  */)
  (Lisp_Object object, Lisp_Object extra_type)
{
  if (NILP (extra_type))
    return (FONTP (object) ? Qt : Qnil);
  if (EQ (extra_type, Qfont_spec))
    return (FONT_SPEC_P (object) ? Qt : Qnil);
  if (EQ (extra_type, Qfont_entity))
    return (FONT_ENTITY_P (object) ? Qt : Qnil);
  if (EQ (extra_type, Qfont_object))
    return (FONT_OBJECT_P (object) ? Qt : Qnil);
  wrong_type_argument (intern ("font-extra-type"), extra_type);
}
```

The EQ calls are symbol comparisons, so `EQ (extra_type, Qfont_spec)` is equivalent to the elisp `(eq extra_type 'font-spec)`.

The macros being used are:

```c
/* True iff X is one of font-spec, font-entity, and font-object.  */
INLINE bool
FONTP (Lisp_Object x)
{
  return PSEUDOVECTORP (x, PVEC_FONT);
}

/* True iff X is font-spec.  */
INLINE bool
FONT_SPEC_P (Lisp_Object x)
{
  return FONTP (x) && PVSIZE (x) == FONT_SPEC_MAX;
}

/* True iff X is font-entity.  */
INLINE bool
FONT_ENTITY_P (Lisp_Object x)
{
  return FONTP (x) && PVSIZE (x) == FONT_ENTITY_MAX;
}
```

atheriel:
This is my first pass at porting `fontp` to Rust, and should be considered a work in progress. I have a few questions:

1) ~~Should I change the naming convention for the `font_property_index` enum?~~
2) ~~Is there a convention for calling `intern` from Rust code?~~ See #255.
3) ~~Can or should I add `is_font_spec()` (etc) methods to `LispObject` in `lisp.rs`, so that the checking can be moved out of `fontp` and into those methods?~~

This should close #233.

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -1202,3 +1202,47 @@ extern "C" {
         nchars_return: *mut ptrdiff_t,
     ) -> ptrdiff_t;
 }
+
+/// Contains C definitions from the font.h header.
+pub mod font {
+    use libc::c_int;
+
+    /// Represents the indices of font properties in the contents of a font
+    /// vector.
+    ///
+    /// # C Porting Notes
+    ///
+    /// The equivalent C enum is `font_property_index`. Since it is meant to
+    /// represent indices for three different length vectors, the C definition
+    /// contains duplicate variants, e.g `FONT_OBJLIST_INDEX = FONT_SPEC_MAX`,
+    /// to represent sizes. These have been moved out of this enum and are
+    /// available as constant `c_int` values on this module.
+    #[allow(non_camel_case_types, dead_code)]
+    #[repr(C)]
+    pub enum FontPropertyIndex {
+        FONT_TYPE_INDEX,
+        FONT_FOUNDRY_INDEX,
+        FONT_FAMILY_INDEX,
+        FONT_ADSTYLE_INDEX,
+        FONT_REGISTRY_INDEX,
+        FONT_WEIGHT_INDEX,
+        FONT_SLANT_INDEX,
+        FONT_WIDTH_INDEX,
+        FONT_SIZE_INDEX,
+        FONT_DPI_INDEX,
+        FONT_SPACING_INDEX,
+        FONT_AVGWIDTH_INDEX,
+        FONT_EXTRA_INDEX,
+        // In C, we have FONT_SPEC_MAX, FONT_OBJLIST_INDEX = FONT_SPEC_MAX here.
+        FONT_OBJLIST_INDEX,
+        // In C, we have FONT_ENTITY_MAX, FONT_NAME_INDEX = FONT_ENTITY_MAX here.
+        FONT_NAME_INDEX,
+        FONT_FULLNAME_INDEX,
+        FONT_FILE_INDEX,
+        // In C, we have FONT_OBJECT_MAX here.
+    }
+
+    pub const FONT_SPEC_MAX: c_int = FontPropertyIndex::FONT_OBJLIST_INDEX as c_int;
+    pub const FONT_ENTITY_MAX: c_int = FontPropertyIndex::FONT_NAME_INDEX as c_int;
+    pub const FONT_OBJECT_MAX: c_int = (FontPropertyIndex::FONT_FILE_INDEX as c_int) + 1;
+}
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -1,7 +1,6 @@
 use remacs_macros::lisp_fn;
-use remacs_sys::{PseudovecType, Qfont_spec, Qfont_entity, Qfont_object, Qsymbolp,
-                 wrong_type_argument};
-use lisp::{LispObject, Qnil};
+use remacs_sys::{Qfont_spec, Qfont_entity, Qfont_object, Qsymbolp, wrong_type_argument};
+use lisp::LispObject;
 
 // See font_property_index in font.h for details.
 #[allow(non_camel_case_types, dead_code)]
@@ -29,9 +28,9 @@ enum font_property_index {
     // In C, we have FONT_OBJECT_MAX here.
 }
 
-const FONT_SPEC_MAX: i32 = font_property_index::FONT_OBJLIST_INDEX as i32;
-const FONT_ENTITY_MAX: i32 = font_property_index::FONT_NAME_INDEX as i32;
-const FONT_OBJECT_MAX: i32 = (font_property_index::FONT_FILE_INDEX as i32) + 1;
+pub const FONT_SPEC_MAX: i32 = font_property_index::FONT_OBJLIST_INDEX as i32;
+pub const FONT_ENTITY_MAX: i32 = font_property_index::FONT_NAME_INDEX as i32;
+pub const FONT_OBJECT_MAX: i32 = (font_property_index::FONT_FILE_INDEX as i32) + 1;
 
 /// Return t if OBJECT is a font-spec, font-entity, or font-object.
 /// Return nil otherwise.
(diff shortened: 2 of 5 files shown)


rust_src/src/fonts.rs
```diff
@@ -0,0 +1,64 @@
+use remacs_macros::lisp_fn;
+use remacs_sys::{PseudovecType, Qfont_spec, Qfont_entity, Qfont_object, Qsymbolp,
+                 wrong_type_argument};
+use lisp::{LispObject, Qnil};
+
+// See font_property_index in font.h for details.
+#[allow(non_camel_case_types, dead_code)]
+#[repr(C)]
+enum font_property_index {
```
DavidDeSimone: This enum will probably want to be defined in the remacs-sys crate instead of here. That crate is where we house the code related to FFI and representing structures that are defined in the C layer.

rust_src/src/fonts.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Optional 2nd argument EXTRA-TYPE, if non-nil, specifies to check
+/// which kind of font it is.  It must be one of `font-spec', `font-entity',
+/// `font-object'.
+#[lisp_fn(min = "1")]
+pub fn fontp(object: LispObject, extra_type: LispObject) -> LispObject {
+    object.as_vectorlike().map_or(Qnil, |v| {
+        if v.is_pseudovector(PseudovecType::PVEC_FONT) {
+            if extra_type.eq(Qnil) {
+                LispObject::constant_t()
+            } else if extra_type.eq(LispObject::from_raw(unsafe { Qfont_spec })) {
```
DavidDeSimone: Yeah I don't like the raw unsafe Q... usage here. Personally I like the idea of having safe methods implemented for LispObject like "is_font_spec", etc.

rust_src/src/fonts.rs
```diff
@@ -0,0 +1,64 @@
...
+                LispObject::from_bool(v.pseudovector_size() == FONT_OBJECT_MAX as i64)
+            } else {
+                // TODO: This should actually be equivalent to
+                // intern("font-extra-type"), not Qsymbolp.
+                unsafe { wrong_type_argument(Qsymbolp, extra_type.to_raw()) }
+            }
+        } else {
+            // As with the C version, checking that object is a font takes priority
+            // over checking that extra_type is well-formed.
+            Qnil
```
DavidDeSimone: We generally prefer calls to LispObject::constant_nil() over raw Qnil usage.

DavidDeSimone:
Thanks for the contribution! This is looking good to me so far, I've left a few inline comments on a couple things I saw.

rust_src/src/fonts.rs
atheriel: Ok, I will move it there.

rust_src/src/fonts.rs
atheriel: I thoought it would be "safe enough" to get at these symbols this way -- is there another option?

DavidDeSimone: Let me clarify:

Accessing these symbols the way you are doing is fine, but we have to use unsafe blocks in Lisp binding code. If we have another lisp binding that needs to do similar logic, we will need to add more unsafe blocks. In general, I think a Rust code base is at it's best when it minimizes unsafe code blocks. 

We can improve this situation by adding a safe wrapper around unsafe code (since the code here is "unsafe" due to its interaction with C). For example something like:

```
impl LispObject {
    pub fn is_font_spec(&self) -> bool {
        self.eq(unsafe { /* ... */ });  
    }
}
```

atheriel:
Ok, I've added some `is_font_*` methods to `LispObject`.

rust_src/src/fonts.rs
atheriel: That's fair. I think that `fontp` is the only place that uses of `Qfont_spec` and friends in this manner in the codebase, though. Is it worth writing a safe interface (for example, creating a `FontEntity::from_object()`-style enum method)?

DavidDeSimone: That's a fair point, I'm personally fine either way.

atheriel:
I will probably wait on #244 before moving the struct to `remacs-sys`.

rust_src/src/fonts.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// Return t if OBJECT is a font-spec, font-entity, or font-object.
+/// Return nil otherwise.
+/// Optional 2nd argument EXTRA-TYPE, if non-nil, specifies to check
+/// which kind of font it is.  It must be one of `font-spec', `font-entity',
+/// `font-object'.
+#[lisp_fn(min = "1")]
+pub fn fontp(object: LispObject, extra_type: LispObject) -> LispObject {
+    if object.is_font() {
+        if extra_type.eq(LispObject::constant_nil()) {
```
birkenfeld: shorter: `extra_type.is_nil()`

rust_src/src/lisp.rs
```diff
@@ -559,6 +560,27 @@ impl LispObject {
             v.is_pseudovector(PseudovecType::PVEC_FONT)
         })
     }
+
+    pub fn is_font_spec(self) -> bool {
```
birkenfeld: Instead of adding these to `LispObject`, I would create an `as_font` method on `is_vectorlike` that returns a "LispFontRef" or similar.

`is_spec`, `is_entity` etc can the be methods on that struct.

rust_src/src/fonts.rs
shanavas786: use camelcase names.

atheriel: Moved, with a new camelcase name.

rust_src/src/fonts.rs
atheriel: I wrote a little enum to abstract this pattern out.

rust_src/src/fonts.rs
atheriel: Changed.

rust_src/src/fonts.rs
atheriel: Changed.

rust_src/src/lisp.rs
atheriel: Ok, I've taken this approach.

rust_src/src/fonts.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub enum FontExtraType {
+    Spec,
+    Entity,
+    Object,
+}
+
+impl FontExtraType {
+    pub fn from_object_or_error(extra_type: LispObject) -> FontExtraType {
```
birkenfeld: I'd prefer `from_symbol_or_error`.

rust_src/src/fonts.rs
```diff
@@ -0,0 +1,72 @@
...
+        if extra_type.eq(LispObject::from_raw(unsafe { Qfont_spec })) {
+            FontExtraType::Spec
+        } else if extra_type.eq(LispObject::from_raw(unsafe { Qfont_entity })) {
+            FontExtraType::Entity
+        } else if extra_type.eq(LispObject::from_raw(unsafe { Qfont_object })) {
+            FontExtraType::Object
+        } else {
+            // TODO: This should actually be equivalent to
+            // intern("font-extra-type"), not Qsymbolp.
```
birkenfeld: In order to call `intern`, you can add it to the imported functions in `remacs-sys`. Signature should be straightforward.

rust_src/remacs-sys/lib.rs
```diff
@@ -1096,3 +1096,47 @@ extern "C" {
         nchars_return: *mut ptrdiff_t,
     ) -> ptrdiff_t;
 }
+
+/// Contains C definitions from the font.h header.
+pub mod font {
```
birkenfeld: Good idea to add some modules to structure this crate!

rust_src/src/fonts.rs
atheriel: Done.

rust_src/src/fonts.rs
atheriel: This turns out to be more difficult than it first appears. See #255.

birkenfeld:
Can be updated to use `intern()` now.

rust_src/src/fonts.rs
atheriel: Done.

atheriel:
Updated to use `intern()`. Is there anything else extant on this PR?

birkenfeld:
Nope, looks, good!

rust_src/src/fonts.rs
```diff
@@ -0,0 +1,71 @@
+use remacs_macros::lisp_fn;
+use remacs_sys::{font, Qfont_spec, Qfont_entity, Qfont_object, wrong_type_argument};
+use lisp::LispObject;
+use symbols::intern;
+use vectors::LispVectorlikeRef;
+
+// A font is not a type in and of itself, it's just a group of three kinds of
+// pseudovector. This newtype allows us to define methods that yield the actual
```
thisch: This could be improved. 
* plural of pseudovectors 
* missing space between new and type.
* First you mention that a font is not a type and then you continue with "This new type"

atheriel: I'm glad someone is reading the docstrings :) Not quite sure I agree about the plural -- a font is "one of three kinds of pseudovector", so fonts are "a group of three kinds of pseudovector". Might be wrong here, though.

The term "newtype" [refers to](https://aturon.github.io/features/types/newtype.html) the Rust pattern of `struct X(Y)`. It's not a typo. A "font" does not have its own structure in the C code. It refers to three possible structs. Hence a font is not "a type in and of itself". It's just a label for a group of structures. So I'm not actually defining any structure here, I just want to add some methods to `LispVectorlikeRef` -- hence the newtype.

PR #250, opened 2017-07-18
[WIP] Port mapcar/mapc/mapcan/mapconcat.

birkenfeld:
Also creates a bit more API around vectors and bool vectors.

Doesn't run cleanly through the test suite yet.

### DIFF
--- a/rust_src/src/lists.rs
+++ b/rust_src/src/lists.rs
@@ -557,7 +557,7 @@ fn mapcar1<T: iter::FromIterator<LispObject>>(seq: LispObject, func: LispObject)
 pub fn mapconcat(function: LispObject, sequence: LispObject, separator: LispObject) -> LispObject {
     let mapped: Vec<_> = mapcar1(sequence, function);
     if mapped.is_empty() {
-        return LispObject::from(unsafe { empty_unibyte_string });
+        return LispObject::from_raw(unsafe { empty_unibyte_string });
     }
 
     let mut with_sep: Vec<_> = mapped
@@ -565,7 +565,7 @@ pub fn mapconcat(function: LispObject, sequence: LispObject, separator: LispObje
         .map(LispObject::to_raw)
         .intersperse(separator.to_raw())
         .collect();
-    unsafe { LispObject::from(Fconcat(with_sep.len() as ptrdiff_t, with_sep.as_mut_ptr())) }
+    unsafe { LispObject::from_raw(Fconcat(with_sep.len() as ptrdiff_t, with_sep.as_mut_ptr())) }
 }
 
 /// Apply FUNCTION to each element of SEQUENCE, and make a list of the results.
@@ -594,7 +594,7 @@ pub fn mapcan(function: LispObject, sequence: LispObject) -> LispObject {
     let mut mapped: Vec<_> = mapcar1(sequence, function);
     unsafe {
         let raw_slice = mem::transmute::<&mut [LispObject], &mut [Lisp_Object]>(&mut mapped);
-        LispObject::from(Fnconc(raw_slice.len() as ptrdiff_t, raw_slice.as_mut_ptr()))
+        LispObject::from_raw(Fnconc(raw_slice.len() as ptrdiff_t, raw_slice.as_mut_ptr()))
     }
 }
 
--- a/rust_src/src/lists.rs
+++ b/rust_src/src/lists.rs
@@ -9,10 +9,10 @@ use libc::ptrdiff_t;
 use remacs_macros::lisp_fn;
 use remacs_sys::{EmacsInt, Lisp_Object};
 use remacs_sys::{Fconcat, Fnconc};
-use remacs_sys::{Qcircular_list, Qplistp, Qsequencep};
+use remacs_sys::{Qcircular_list, Qlistp, Qplistp, Qsequencep};
 use remacs_sys::globals;
 
-use lisp::{LispCons, LispObject};
+use lisp::LispObject;
 use lisp::defsubr;
 use symbols::LispSymbolRef;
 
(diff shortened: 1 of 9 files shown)


Wilfred:
Code looks sensible to me 👍

birkenfeld:
Rebased, but still haven't figured out the reason for the crash...

rust_src/src/lists.rs
```diff
@@ -486,3 +491,87 @@ pub fn merge(mut l1: LispObject, mut l2: LispObject, pred: LispObject) -> LispOb
...
+        iter::empty().collect()
+    } else if let Some(vec) = seq.as_vectorlike().and_then(|v| v.as_vector()) {
+        vec.iter().map(|item| call!(func, item)).collect()
+    } else if let Some(bvec) = seq.as_vectorlike().and_then(|v| v.as_bool_vector()) {
+        bvec.iter().map(|item| call!(func, item)).collect()
+    } else if let Some(string) = seq.as_string() {
+        string
+            .chars()
+            .map(|ch| call!(func, LispObject::from_fixnum(ch as EmacsInt)))
+            .collect()
```
brotzeit: Is it possible that this is a multibyte char issue ? Can char() handle those ?

birkenfeld: It's not Rust `chars`, but an iterator for `LispStringRef` which handles multibyte and unibyte. But it's still possible that it is buggy, I'll have a look.

shaleh:
I have rebased your work on a branch in my fork. Look at update-maps-branch. What I did was keep the original C version and your version is now `rust-mapcar`, etc. I only exposed `mapconcat` from Rust. This is enough to see the failure.

make lisp/char-fold-tests.log demonstrates the crash.

shaleh:
This resolves one of the segfaults.

     #[lisp_fn]
     pub fn mapconcat(function: LispObject, sequence: LispObject, separator: LispObject) -> LispObject {
         let mapped: Vec<_> = rust_mapcar1(sequence, function);
    +    if mapped.is_empty() {
    +        return LispObject::from(unsafe { empty_unibyte_string });
    +    }
    +

rust_src/src/lists.rs
shaleh: The code is actually failing in the `is_cons` case. I added debug prints to catch it.

I tweaked mapconcat:

    +    println!("Begin {:?}", function);
    +    for i in mapped.iter() {
    +        match i.as_string() {
    +            Some(s) => println!("mapconcat(r): {} == {:?}", i.to_raw(), *s),
    +            None => println!("mapconcat(r): {}", i.to_raw()),
    +        };
    +    }
    +    println!("End");

Along with a derive(Debug) for Lisp_String to avoid sending bogus pointers to `display_string`. Here is the output for $ gdb --args ../src/remacs -Q --no-site-file --no-site-lisp -batch -L :. -l ert -l lisp/sort-tests.el

    Begin '('closure '('('n . 8 't ) '('_ ) '('string '('let '('('c '('random 52 ) ) ) '('+ '('if '('> 'c 25 ) 71 65 ) 'c ) ) ) )
    mapconcat(r): 16041668 == Lisp_String { size: 1, size_byte: -1, intervals: 0x0, data: 0xedba98 }
    mapconcat(r): 16041700 == Lisp_String { size: 16041600, size_byte: -1, intervals: 0x0, data: 0x0 }
    mapconcat(r): 16041732 == Lisp_String { size: 16041696, size_byte: -1, intervals: 0x0, data: 0x0 }
    mapconcat(r): 16041764 == Lisp_String { size: 16041728, size_byte: -1, intervals: 0x0, data: 0x0 }
    mapconcat(r): 16041796 == Lisp_String { size: 16041760, size_byte: -1, intervals: 0x0, data: 0x0 }
    mapconcat(r): 16041828 == Lisp_String { size: 1, size_byte: -1, intervals: 0x0, data: 0xedbaa8 }
    mapconcat(r): 15757748 == Lisp_String { size: 1, size_byte: -1, intervals: 0x0, data: 0xedbab8 }
    mapconcat(r): 15757716 == Lisp_String { size: 1, size_byte: -1, intervals: 0x0, data: 0xedbac8 }
    End

The data is bad straight out of the invocation by `call!` near as I can tell.

shaleh:
I have updated my branch with debugging help. @birkenfeld I won't have time for this for a bit. Hopefully this gets you closer.

birkenfeld:
Great, thanks! I'll take a look when I can.

rust_src/src/lists.rs
shaleh: I debugged this a little further today.

    seq.iter_tails_safe()
            .map(|cons| {
                let result = call!(func, cons.car());
                println!("Call!: {} -> {:?}", result.to_raw(), result);
                result
            })
            .collect();

The data is printed correctly in the `map`. If I switch to a loop and manually populate the vector I can see that the data goes south as the vector is being filled:

     for i in seq.iter_tails_safe() {
            let result = call!(func, i.car());
            println!("Call!: {} -> {:?}", result.to_raw(), result);
            output.push(result);
            for g in output.iter() {
                match g.as_string() {
                    Some(_) => {
                        println!("mapcar1(r): {}", g.to_raw());
                        println!("== {:?}", g);
                    }
                    None => println!("mapcar1(r): {}", g.to_raw()),
                }
            }

Now you can see that after a few iterations the pointers are off and becoming garbage. It has nothing to do with the call to `random` in the elisp test. I dummied out the function using `/dev/urandom` and it still fails.

shaleh:
Ok, I rebased this against master. The segfaults are still there.

shaleh:
I have a suspicion we are hitting either a GC issue or something similar. Emacs uses a constant called `ARRAY_MARK_FLAG` to indicate an object is in the GC. We might be missing this flag. Or maybe a similar flag that is not GC somewhere else. That would explain the odd values.

shaleh:
We have partially implemented these functions. This PR is too out of sync with the current tree to attempt to revive it.

Thanks @birkenfeld

cireu:
> We have partially implemented these functions. This PR is too out of sync with the current tree to attempt to revive it.
> 
> Thanks @birkenfeld

Have `mapconcat` be implemented? I' m interested in it.

shaleh:
No. It triggers deep failures in the garbage collection code today. Give it a look but know this will lead you very deep into the system.

PR #251, opened 2017-07-18
Bringing over some hash-table related functions to remacs.

DavidDeSimone:
This includes creating several different new APIs, including pseudovector allocation. This code has yet to be thoroughly tested, so I am expecting some bugs along the line. I am hoping for feedback on the API for creating and working with hash tables, and potential other pseudovector types in the future.


rust_src/remacs-sys/lib.rs
```diff
@@ -94,7 +94,7 @@ pub enum Lisp_Type {
     Lisp_Float = 7,
 }
 
-#[repr(isize)]
```
DavidDeSimone: I still need to look into why we were repr(isize) here, and if we need to switch over to repr(C) now that we are invoking a C function with this enum.

rust_src/src/hashtable.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl LispHashTableRef {
+    pub fn allocate() -> LispHashTableRef {
+        let vec_ptr =
+            allocate_pseudovector!(Lisp_Hash_Table, count, PseudovecType::PVEC_HASH_TABLE);
+        LispHashTableRef::new(unsafe { mem::transmute(vec_ptr) })
+    }
+
+    pub fn copy(&mut self, other: &LispHashTableRef) {
+        *self = *other
```
DavidDeSimone: Looking at this again, I don't believe this is what I want to be doing. I don't believe this is copying the structs into each other.

rust_src/src/lisp.rs
```diff
@@ -480,6 +485,31 @@ impl LispObject {
...
+        if self.is_hash_table() {
+            Some(LispHashTableRef::new(
+                unsafe { mem::transmute(self.get_untaggedptr()) },
+            ))
+        } else {
+            None
+        }
+    }
+
+    /// Equivalent to XSET_HASH_TABLE in C
```
birkenfeld: Do we need this?

rust_src/remacs-sys/lib.rs
```diff
@@ -1044,8 +1069,13 @@ extern "C" {
     pub static Qfont_spec: Lisp_Object;
     pub static Qfont_entity: Lisp_Object;
     pub static Qfont_object: Lisp_Object;
+    pub static Qhash_table_p: Lisp_Object;
     pub static lispsym: Lisp_Symbol;
 
+    pub static header_size: size_t;
```
birkenfeld: Even if these are cased like in Emacs, can we publicly alias them in uppercase?

DavidDeSimone:
@birkenfeld I agree 100%, I am very interested in seeing if we can switch over to using Rust's hashmap. I figure starting out with a 'copy' function will be a good first babystep, and will help lay the ground work for interacting with hash tables at the Rust layer.

rust_src/remacs-sys/lib.rs
DavidDeSimone: Based on discussions in other threads, I believe that changing this to repr(C) is the right call.

rust_src/src/lisp.rs
```diff
@@ -96,6 +97,22 @@ impl LispObject {
         unsafe { mem::transmute(res) }
     }
 
+    pub fn tag_ptr<T>(external: ExternalPtr<T>, ty: Lisp_Type) -> LispObject {
```
Wilfred: Handy!

rust_src/src/lisp.rs
```diff
@@ -485,6 +506,36 @@ impl LispObject {
     }
 }
 
+impl LispObject {
+    pub fn as_hash_table_or_error(&self) -> LispHashTableRef {
+        if self.is_hash_table() {
+            LispHashTableRef::new(unsafe { mem::transmute(self.get_untaggedptr()) })
+        } else {
+            unsafe { wrong_type_argument(Qhash_table_p, self.to_raw()) }
```
Wilfred: Once #266 lands we should use `wrong_type!` here.

Wilfred:
Merged, thank you :)

PR #253, opened 2017-07-19
Define a thread_state type and the current-buffer function

Wilfred:
Defining `current-buffer` in Rust requires us to have the `thread_state` type defined.

This definitely isn't finished yet, but any feedback welcome :)

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -1,11 +1,12 @@
 //! Functions operating on buffers.
 
-use libc::{c_uchar, ptrdiff_t};
+use libc::{c_void, c_uchar, ptrdiff_t};
 
 use lisp::{LispObject, ExternalPtr};
-use remacs_sys::{Lisp_Buffer, Lisp_Type, Vbuffer_alist, current_thread, make_lisp_ptr};
+use remacs_sys::{Lisp_Buffer, Lisp_Type, Vbuffer_alist, make_lisp_ptr};
 use strings::string_equal;
 use lists::{car, cdr};
+use threads::ThreadState;
 
 use remacs_macros::lisp_fn;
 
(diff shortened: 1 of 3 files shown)


db48x:
aside from some TODOs, it works for me: https://github.com/db48x/remacs/blob/secure-hash/rust_src/src/crypto/mod.rs

Wilfred:
OK, this is now ready to go.

@db48x thanks for giving it a try. :) Glad to hear it's working. I'm afraid I've amended my commits, so you'll need to rebase.

rust_src/src/buffers.rs
```diff
@@ -103,3 +103,12 @@ pub fn get_buffer(buffer_or_name: LispObject) -> LispObject {
         ))
     }
 }
+
+/// Return the current buffer as a Lisp object.
+#[lisp_fn]
+pub fn current_buffer() -> LispObject {
```
birkenfeld: Do we prefer a free function or something like  `ThreadState::current_buffer`?

Wilfred: Good question!

The C code definitely uses Fcurrent_buffer in quite a few places, but if we define a `ThreadState::current_buffer` then we can offer an interface with a better Rusty type.

There's also the question of whether or not `ThreadState::current_buffer` should be a static function, or a method on a static value `current_thread`.

I've pushed an additional commit: what do you think? I'm sure it could be prettier somehow, but I'm not sure how.

birkenfeld: LGTM. The only way I see to make it prettier would be to have `unsafe fn LispObject::from_ptr()`, but it wouldn't save much.

DavidDeSimone: @birkenfeld I've added a "from_ptr" equivalent in https://github.com/Wilfred/remacs/pull/251, and was planning on changing this call site to use that if this PR gets merged first.

Wilfred:
Rebased and formatted.

Wilfred:
If we merge #260 before this, then this PR is redundant because #260 includes these commits.

Wilfred:
#260 has been merged, but it didn't include `ThreadState::current_buffer`, so I'm rebasing this.

PR #255, opened 2017-07-20
Adds intern and an intern_static! macro

atheriel:
It's a common pattern in the C codebase to call `intern("some static c string")`. I thought that it would be nice to have a similar API available in Rust. Unfortunately, there's no way of doing this with the `std::cstring` interface that avoids heap-allocating and unnecessary copies. I've written a macro that avoids this by concatenating a null byte to the end of a string at compile time, and casting to a `*const c_char`. This allows us to write 

```rust
let symbol = unsafe { intern_static!("my-symbol-1") };
```

This macro is, and always will be, unsafe. If you pass it a string with internal NULL bytes, e.g. `intern("my-\0dangerous-\0string")`, then C will not correctly deduce the length of the string, and we'll leak memory. However, for symbols defined at compile time and clearly audited by the project, I don't think this is a problem.

One could also add a safe, non-static version of `intern` using the `std::cstring` machinery.

### DIFF
--- a/rust_src/src/symbols.rs
+++ b/rust_src/src/symbols.rs
@@ -60,12 +60,14 @@ fn symbol_plist(object: LispObject) -> LispObject {
     object.as_symbol_or_error().get_plist()
 }
 
-/// Intern (e.g. create a symbol from) a string slice.
-pub fn intern(s: &str) -> LispObject {
-    unsafe {
-        LispObject::from_raw(intern_1(
+/// Intern (e.g. create a symbol from) a string.
+#[allow(dead_code)]
+pub fn intern<T: AsRef<str>>(string: T) -> LispObject {
+    let s = string.as_ref();
+    LispObject::from_raw(unsafe {
+        intern_1(
             s.as_ptr() as *const libc::c_char,
             s.len() as libc::ptrdiff_t,
-        ))
-    }
+        )
+    })
 }
--- a/rust_src/src/symbols.rs
+++ b/rust_src/src/symbols.rs
@@ -1,6 +1,7 @@
+use libc;
 use remacs_macros::lisp_fn;
 use lisp::{LispObject, ExternalPtr};
-use remacs_sys::Lisp_Symbol;
+use remacs_sys::{Lisp_Symbol, intern_1};
 
 pub type LispSymbolRef = ExternalPtr<Lisp_Symbol>;
 
@@ -57,4 +58,14 @@ fn symbol_function(object: LispObject) -> LispObject {
 #[lisp_fn]
 fn symbol_plist(object: LispObject) -> LispObject {
     object.as_symbol_or_error().get_plist()
+}
+
+/// Intern (e.g. create a symbol from) a string slice.
+pub fn intern(s: &str) -> LispObject {
+    unsafe {
+        LispObject::from_raw(intern_1(
+            s.as_ptr() as *const libc::c_char,
+            s.len() as libc::ptrdiff_t,
+        ))
+    }
 }
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -1060,6 +1060,8 @@ extern "C" {
     pub fn make_uninit_multibyte_string(nchars: EmacsInt, nbytes: EmacsInt) -> Lisp_Object;
     pub fn string_to_multibyte(string: Lisp_Object) -> Lisp_Object;
 
+    pub fn intern_1(s: *const c_char, length: ptrdiff_t) -> Lisp_Object;
+
     pub fn SYMBOL_NAME(s: Lisp_Object) -> Lisp_Object;
     pub fn CHECK_IMPURE(obj: Lisp_Object, ptr: *const c_void);
     pub fn internal_equal(


rust_src/src/symbols.rs
```diff
@@ -58,3 +59,21 @@ fn symbol_function(object: LispObject) -> LispObject {
...
+/// already).
+///
+/// This is an implementation equivalent of the C version.
+pub unsafe fn intern(s: *const libc::c_char) -> LispObject {
+    LispObject::from_raw(intern_1(s, libc::strlen(s) as libc::ptrdiff_t))
+}
+
+/// Intern (e.g. create a symbol for) a `&'static str`, avoiding
+/// allocation. This macro is inherently unsafe: if the string contains internal
+/// NULL bytes (`\0`), we will violate memory safety.
```
birkenfeld: Why would it violate memory safety? It would just use the shorter string since you're using `strlen` to determine the length.

rust_src/src/symbols.rs
```diff
@@ -58,3 +59,21 @@ fn symbol_function(object: LispObject) -> LispObject {
...
+macro_rules! intern_static {
+    ($s:expr) => ({
+        let __intern_s = concat!($s, "\0") as *const str as *const [::libc::c_char] as *const ::libc::c_char;
```
birkenfeld: This should be `concat!(...).as_ptr() as *const libc::c_char`

birkenfeld: BTW, nice idea to use a macro here. We could use the same approach to call `error` (where we currently added the \0 manually everywhere).

s-kostyaev:
> This macro is, and always will be, unsafe. If you pass it a string with internal NULL bytes, e.g. intern("my-\0dangerous-\0string"), then C will not correctly deduce the length of the string, and we'll leak memory. However, for symbols defined at compile time and clearly audited by the project, I don't think this is a problem.

I'm newbe in rust, but can it be checked at compile time inside macro?

birkenfeld:
Actually, since `intern_1` gets the length as well, I don't think we even need the trailing NULL byte.

rust_src/src/symbols.rs
DavidDeSimone: By the way, Rust has hygienic macros. No longer do we have to name our variables in macros horrible underscore salads (unless you like the convention and did it for a reason besides to avoid shadowing at macro call sites).

rust_src/src/symbols.rs
atheriel: My comment here comes from my understanding is that handling of strings with internal NULL bytes is undefined behaviour in C, and from the "unsafe" denotation on [the equivalent std method](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_vec_unchecked) . You might be right, though.

rust_src/src/symbols.rs
atheriel: I did it for convention, although it has been a while since I've looked at the source code for the standard library macros. Happy to change it, though.

atheriel:
> Actually, since `intern_1` gets the length as well, I don't think we even need the trailing NULL byte.

@birkenfeld Well, that seems obvious in retrospect. Don't know why I was overthinking it. I'll change it.

atheriel:
@s-kostyaev Using a procedural macro, yes. Using the basic `macro_rules!`, no.

rust_src/src/symbols.rs
atheriel: Right.

s-kostyaev:
@atheriel Thanks for reply. Maybe would be better use procedural macro in this case and add this compile time check? It can prevent memory leak.

rust_src/src/symbols.rs
birkenfeld: Let's assume intern_1 didn't take the length; then as far as the C API is concerned the string would end at the first null byte. That's not undefined behavior, but of course not what you intended either.

The method on CString is unsafe because it maintains the invariant that its string has the same length for Rust and C consumers, so not checking must be an unsafe function.

birkenfeld:
> Maybe would be better to use procedural macro in this case and add this compile time check? It can prevent memory leak.

No leaks possible in this case.

s-kostyaev:
>  If you pass it a string with internal NULL bytes, e.g. intern("my-\0dangerous-\0string"), then C will not correctly deduce the length of the string, and we'll leak memory.

birkenfeld:
There's no "passing a string". There's only passing a pointer to a string constant in static memory.

s-kostyaev:
Static memory can't leak. Thanks for explanation.

atheriel:
I should be able to rewrite this and rebase tonight. Thanks for all the comments!

atheriel:
@birkenfeld How does this look?

birkenfeld:
Can't `intern` directly take a `&str`, and the macro can be removed?

atheriel:
@birkenfeld Yep.

atheriel:
@birkenfeld I've fixed the dead code issue. I also realized that this function can clearly be generic, to accomodate other string-like types.

birkenfeld:
Nice!

Issue #232, opened 2017-07-15
Port window-live-p to Rust


Wilfred:
The function:

```c
DEFUN ("window-valid-p", Fwindow_valid_p, Swindow_valid_p, 1, 1, 0,
       doc: /* Return t if OBJECT is a valid window and nil otherwise.
A valid window is either a window that displays a buffer or an internal
window.  Windows that have been deleted are not valid.  */)
  (Lisp_Object object)
{
  return WINDOW_VALID_P (object) ? Qt : Qnil;
}
```

The macro:

```c
/* True if WINDOW is a live window.  */
#define WINDOW_LIVE_P(WINDOW)					\
  (WINDOWP (WINDOW) && BUFFERP (XWINDOW (WINDOW)->contents))
```

gregkatz:
Hi,

I'm new to the project (and emacs lisp though I have decent rust proficiency) and I was thinking about taking this one. Quick question: In this bug, it has ```WINDOW_LIVE_P``` listed for the macro but the function is for ```window-valid-p```. Is that intentional?

Thanks,

Greg

Wilfred:
Woops! Sorry, that was my mistake. The function is actually this one:

```c
DEFUN ("window-live-p", Fwindow_live_p, Swindow_live_p, 1, 1, 0,
       doc: /* Return t if OBJECT is a live window and nil otherwise.
A live window is a window that displays a buffer.
Internal windows and deleted windows are not live.  */)
  (Lisp_Object object)
{
  return WINDOW_LIVE_P (object) ? Qt : Qnil;
```

it's in `src/window.c`.

gregkatz:
Ok, no problem and thanks. I should have a PR opened tomorrow. 

I'm planning on mimicking the implementation of ```buffer-live-p```. Sound like a good plan to you? It will involve adding a a window struct to remacs-sys.

Wilfred:
Merged! :)

PR #258, opened 2017-07-20
Port window-live-p to Rust

gregkatz:
This PR ports window-live-p to Rust. Because I followed the approach used by buffer-live-p, it includes a struct Lisp_Window in remacs-sys/lib.rs. If this isn't the right approach let me know, and I'll try again!

It seems to work as expected in ielm.

Closes Issue #232.

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -124,6 +124,17 @@ pub enum PseudovecType {
     PVEC_FONT, /* Should be last because it's used for range checking.  */
 }
 
+#[repr(C)]
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum TextCursorKinds {
+    DEFAULT_CURSOR = -2,
+    NO_CURSOR = -1,
+    FILLED_BOX_CURSOR,
+    HOLLOW_BOX_CURSOR,
+    BAR_CURSOR,
+    HBAR_CURSOR,
+}
+
 // XXX: this can also be char on some archs
 pub type bits_word = size_t;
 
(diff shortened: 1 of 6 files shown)


rust_src/remacs-sys/lib.rs
```diff
@@ -286,6 +297,97 @@ pub struct Lisp_Marker {
...
+    pub base_line_number: ptrdiff_t,
+    pub base_line_pos: ptrdiff_t,
+    pub column_number_displayed: ptrdiff_t,
+    pub nrows_scale_factor: c_int,
+    pub ncols_scale_factor: c_int,
+    pub cursor: CursorPos,
+    pub phys_cursor: CursorPos,
+    pub output_cursor: CursorPos,
+    pub last_cursor_vpos: c_int,
+    pub pys_cursor_type: TextCursorKinds,
```
Wilfred: phys_cursor_type

rust_src/remacs-sys/lib.rs
```diff
@@ -286,6 +297,97 @@ pub struct Lisp_Marker {
...
+    pub temslot: Lisp_Object,
+    pub vertical_scroll_bar: Lisp_Object,
+    pub vertical_scroll_bar_type: Lisp_Object,
+    pub horizontal_scroll_bar: Lisp_Object,
+    pub horizontal_scroll_bar_type: Lisp_Object,
+    pub display_table: Lisp_Object,
+    pub dedicated: Lisp_Object,
+    pub redisplay_end_trigger: Lisp_Object,
+    pub combination_limit: Lisp_Object,
+    pub window_paremeters: Lisp_Object,
```
Wilfred: window_parameters

gregkatz:
Thanks for the review! I'll make the changes.

rust_src/remacs-sys/lib.rs
```diff
@@ -124,6 +124,17 @@ pub enum PseudovecType {
     PVEC_FONT, /* Should be last because it's used for range checking.  */
 }
 
+#[repr(isize)]
```
birkenfeld: Is this really correct for a C enum? I thought the default type there was a c_int.

Better use `#[repr(C)]`?

gregkatz: Thanks a ton for the review! I'll be glad to make this change. I used repr(isize) to match ```PseudovecType```. Maybe I should also change that to ```repr(c)```?

birkenfeld: Good idea IMO.

birkenfeld: But let's wait for @Wilfred 's opinion.

rust_src/remacs-sys/lib.rs
mrluanma: The typo is still there: par[e]meters -> parameters.

gregkatz: Sorry about this. I think I accidentally reverted my first patch with my second patch. I'm fairly new to Github, so I appreciate everyone's patience.

rust_src/remacs-sys/lib.rs
Wilfred: Based on my current understanding of Rust FFI, I think we should use `repr(C)` for all our enums.

> It has fairly simple intent: do what C does. The order, size, and alignment of fields is exactly what you would expect from C or C++. Any type you expect to pass through an FFI boundary should have repr(C), as C is the lingua-franca of the programming world. 
>
> -- https://doc.rust-lang.org/nomicon/other-reprs.html

The C enum in question is:

```c
enum text_cursor_kinds
{
  DEFAULT_CURSOR = -2,
  NO_CURSOR = -1,
  FILLED_BOX_CURSOR,
  HOLLOW_BOX_CURSOR,
  BAR_CURSOR,
  HBAR_CURSOR
};
```

What size does this have?

> An enum is only guaranteed to be large enough to hold int values. The compiler is free to choose the actual type used based on the enumeration constants defined so it can choose a smaller type if it can represent the values you define.
> 
> -- https://stackoverflow.com/a/366026/509706

So I think `repr(C)` ensures that our enums match the size the C codebase expects.

PR #260, opened 2017-07-21
Implement md5 and secure-hash functions in Rust

db48x:
This was a bit of an adventure; it calls out to many parts of Emacs which are not yet converted into Rust. Most of those things don't even have safe wrappers either, so I guess it's possible that you won't even want this patch. On the other hand, this passes all the tests. (On the gripping hand, there are no tests for these functions. Oops.)


birkenfeld:
Wow, that's a big one :) But looks very nice at a glance, I'll review it in detail this weekend.

db48x:
Why did the build fail? I tried to load the build log but it's just sitting there.

DavidDeSimone:
@db48x So the build "failed" due to our `cargo fmt` setup. If you're code doesn't match what `cargo fmt` would output, travis will mark the build as failure. 

You'll need to run `cargo fmt`, and check in the changes it outputs. There should be more info about this process in the README.

db48x:
Oh, right, I forgot about that. I rather dislike what cargo fmt does to the code, but...

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+use remacs_sys::{current_thread, record_unwind_current_buffer, set_buffer_internal,
+                 make_buffer_string, call4};
+use remacs_sys::{globals, Fbuffer_file_name, Ffind_operation_coding_system, Flocal_variable_p};
+use remacs_sys::{Qmd5, Qsha1, Qsha224, Qsha256, Qsha384, Qsha512, Qstringp, Qraw_text,
+                 Qcoding_system_error, Qwrite_region, Qbuffer_file_coding_system};
 use remacs_macros::lisp_fn;
+use symbols::{symbol_name, fboundp};
+
+enum HashAlg {
+    HashMD5,
```
birkenfeld: Duplicate prefix (`Hash`) for the variants should be omitted.

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+    };
+    if start_byte > end_byte {
+        std::mem::swap(start_byte, end_byte);
+    }
+    //if !(buffer.begv <= start_byte && end_byte <= buffer.zv) {
+    //    args_out_of_range(start, end);
+    //}
+    let string = LispObject::from_raw(unsafe { make_buffer_string(*start_byte, *end_byte, false) });
+    unsafe { set_buffer_internal(prev_buffer) };
+    unsafe { (*current_thread).m_specpdl_ptr.offset(-40) }; // TODO: this needs to be std::mem::size_of<specbinding>()
```
birkenfeld: Also, this doesn't modify `m_specpdl_ptr`, it just returns a new pointer.

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+    start_byte: &mut ptrdiff_t,
+    end_byte: &mut ptrdiff_t,
+) -> LispObject {
+    let prev_buffer = unsafe { (*current_thread).m_current_buffer };
+    unsafe { record_unwind_current_buffer() };
+    unsafe { set_buffer_internal(buffer.as_ptr() as *const _ as *const libc::c_void) };
+    *start_byte = if start.is_nil() {
+        buffer.begv
+    } else {
+        match start.as_number_coerce_marker_or_error() {
```
birkenfeld: Hm, actually `as_number_coerce_marker_or_error` should not allow floats. I'll have to fix that.

That's a bit of inconsistent naming on the C side ("number" used instead of "fixnum" or "integer").

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+                    )
+                }).as_string_or_error(),
+            )
+        }
+        get_input_from_string(object, string.unwrap(), start, end).as_string_or_error()
+    } else if object.is_buffer() {
+        let mut start_byte: ptrdiff_t = 0;
+        let mut end_byte: ptrdiff_t = 0;
+        let s = get_input_from_buffer(buffer.unwrap(), start, end, &mut start_byte, &mut end_byte);
+        if s.as_string_or_error().is_multibyte() {
```
birkenfeld: Can you call `s.as_string_or_error` only once?

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+    _secure_hash(hash_alg(algorithm), input.as_slice(), binary.is_nil())
+}
 
 #[no_mangle]
-pub unsafe extern "C" fn sha1_buffer(
-    buffer: *const c_char,
-    len: size_t,
-    dest_buf: *mut c_void,
-) -> *mut c_void {
+fn _secure_hash(algorithm: HashAlg, input: &[u8], hex: bool) -> LispObject {
```
birkenfeld: If it's `no_mangle`, it should also be `extern "C"`. But I guess the function is not called from C anymore?

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+        // digest's buffer long enough
+        hash_func(input, digest_str.as_mut_slice());
+    }
+    if hex {
+        make_digest_string(digest_str.as_mut_slice(), digest_size);
+    }
+    digest
+}
+
+fn make_digest_string(buffer: &mut [u8], len: usize) {
```
birkenfeld: The slice should know its length already.

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+    static hexdigit: [u8; 16] = *b"0123456789abcdef";
+    for i in (0..len).rev() {
+        let v = buffer[i];
+        buffer[2 * i] = hexdigit[(v >> 4) as usize];
+        buffer[2 * i + 1] = hexdigit[(v & 0xf) as usize];
+    }
+}
+
+unsafe fn md5_buffer(buffer: &[u8], dest_buf: &mut [u8]) {
```
birkenfeld: These shouldn't be unsafe anymore.

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+    let output = md5::compute(buffer);
+    ptr::copy_nonoverlapping(output.as_ptr(), dest_buf.as_ptr() as *mut u8, output.len());
```
birkenfeld: Now that we're using slices, you can use `dest_buf.copy_from_slice` which is not unsafe.

src/fns.c
```diff
@@ -3995,7 +3720,7 @@ this variable.  */);
   defsubr (&Swidget_apply);
   defsubr (&Sbase64_encode_region);
   defsubr (&Sbase64_decode_region);
-  defsubr (&Smd5);
-  defsubr (&Ssecure_hash);
+//  defsubr (&Smd5);
```
birkenfeld: Can be just removed.

rust_src/Cargo.toml
```diff
@@ -14,6 +14,7 @@ remacs-lib = { version = "0.1.0", path = "remacs-lib/" }
 remacs-macros = { version = "0.1.0", path = "remacs-macros" }
 libc = "0.2"
 rand = "0.3.15"
+md5 = "0.3.5"
```
birkenfeld: We should consider using the `md-5` and `sha-1` crates which have the same interface (and maintainer) as sha2 and a host of other digest crates. (Yes, the naming is unfortunate.)

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,470 @@
...
+        HashAlg::HashSHA512
+    } else {
+        let name = symbol_name(algorithm).as_string_or_error();
+        unsafe {
+            error(b"Invalid algorithm arg: %s\0".as_ptr(), name.as_slice());
+        }
+    }
+}
+
+fn validate_coding_system(coding_system: LispObject, noerror: LispObject) -> LispObject {
```
db48x: Heh, I'll rename it to avoid confusion.

rust_src/src/crypto/mod.rs
db48x: This one is curious; len is the length of the digest (16 bytes for md5, for example), and the buffer is twice that so that each character of the digest can be expanded to 2 hex digits. Perhaps a better name is in order, or a comment.

rust_src/src/crypto/mod.rs
db48x: Yea, no_mangle can go.

rust_src/src/crypto/mod.rs
db48x: Almost; dest_buf might have been made long enough to hold the eventual hex string, so copy_from_slice will panic.

birkenfeld: That's not unsafe :)

db48x:
I didn't get this when I built it locally:

    binding stack not balanced (serious byte compiler bug)

/me grumbles

rust_src/src/crypto/mod.rs
```diff
@@ -414,20 +415,30 @@ fn _secure_hash(algorithm: HashAlg, input: &[u8], hex: bool) -> LispObject {
...
         buffer[2 * i] = hexdigit[(v >> 4) as usize];
         buffer[2 * i + 1] = hexdigit[(v & 0xf) as usize];
     }
 }
 
+// For the following hash functions, the caller must ensure that the
+// destination buffer is at least long enough to hold the
+// digest. Additionall, the caller may have been asked to return a hex
+// string, in which case dest_buf will be twice as long as the digest.
+// Thus, these functions are unsafe.
```
birkenfeld: Ah, sorry, I misremembered that `copy_from_slice` requires same length. But that's easy: make it `dest_buf[..output.len()].copy_from_slice(output)`.

rust_src/src/crypto/mod.rs
```diff
@@ -414,20 +415,30 @@ fn _secure_hash(algorithm: HashAlg, input: &[u8], hex: bool) -> LispObject {
...
     digest
 }
 
-fn make_digest_string(buffer: &mut [u8], len: usize) {
+/// To avoid a copy, buffer is both the source and the destination of
+/// this transformation. Buffer must contain len bytes of data and
+/// 2*len bytes of space for the final hex string.
+fn hexify_digest_string(buffer: &mut [u8], len: usize) {
     static hexdigit: [u8; 16] = *b"0123456789abcdef";
+    debug_assert!(buffer.len() == 2 * len, "buffer must be long enough to hold 2*len hex digits");
```
birkenfeld: +1

db48x:
Can we merge this one in to master today?

rust_src/src/crypto/mod.rs
```diff
@@ -1,78 +1,467 @@
...
+            LispNumber::Fixnum(n) => n as ptrdiff_t,
+            LispNumber::Float(n) => n as ptrdiff_t,
+        }
+    };
+    if start_byte > end_byte {
+        std::mem::swap(start_byte, end_byte);
+    }
+    //if !(buffer.begv <= start_byte && end_byte <= buffer.zv) {
+    //    args_out_of_range(start, end);
+    //}
```
Wilfred: Why is this commented out?

Wilfred:
👍 from me, looks good :)

rust_src/src/crypto/mod.rs
db48x: Oops, I simply forgot about it. We don't yet have it declared in remacs_sys, and it's not yet included amongst the new macros either.

db48x:
I went ahead and wrote a macro

Wilfred:
:tada: :sparkles: :tada:

db48x:
Thanks!

Wilfred:
@db48x I've given you the power to approve PRs too if you wish :)

Issue #178, opened 2017-04-19
Use rust-bindgen


jeandudey:
Using rust-bindgen would solve us a lot of problems and we wouldn't need to maintain the C bindings manually.

db48x:
FWIW, that's a somewhat experimental patch; there are a few compile errors

brotzeit:
#781

PR #264, opened 2017-07-22
use bindgen rather than hand-writing all the FFI stuff

db48x:
This was a bit of work. bindgen is pretty close to automatic on this code-base, because it's pure C, but it does miss a few things (MOST_POSITIVE_FIXNUM, a few functions, etc; I don't know the entire extent of this) and it also gets a few things wrong (it thinks emacs_abort and similar functions return () rather than never returning). Clean-ups for these appear in rust_src/remacs-sys/lib.rs and rust_src/remacs-sys/build.rs.

Once that was done it was just a matter of tweaking the code already written in rust_src/src/ and rust_src/remacs-macros/ so that it works with the generated bindings. This is mostly due to differences in naming and typing, but there were some very tricky changes to the rust_fn decorator macro that are quite important.

This passes all of the tests.


rust_src/remacs-sys/lib.rs
```diff
@@ -18,159 +18,8 @@ extern crate libc;
...
-    pub intspec: *const c_char,
-
-    // TODO: Change this to EMACS_INT
-    //
-    // If you wan't to give it a try and solve this you should see this commit:
-    // https://github.com/Wilfred/remacs/commit/c5461d03a411ff5c6f43885a0a9030e8a94bbc2e
-    /// The docstring of the Emacs Lisp function.
-    pub doc: *const c_char,
-}
+include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
```
birkenfeld: Do we need the generated `definitions.rs` anymore after that?

rust_src/remacs-sys/lib.rs
```diff
@@ -180,1022 +29,51 @@ pub struct Lisp_Subr {
...
-    pub fn string_to_multibyte(string: Lisp_Object) -> Lisp_Object;
-
-    pub fn SYMBOL_NAME(s: Lisp_Object) -> Lisp_Object;
-    pub fn CHECK_IMPURE(obj: Lisp_Object, ptr: *const c_void);
     pub fn internal_equal(
         o1: Lisp_Object,
         o2: Lisp_Object,
         kind: EqualKind,
-        depth: c_int,
+        depth: std::os::raw::c_int,
```
birkenfeld: Can bindgen use `libc` types instead? (At least on the command line there is an option for that.)

rust_src/src/lisp.rs
```diff
@@ -88,17 +87,17 @@ impl LispObject {
...
-                   } else {
-                       raw >> VALBITS
-                   }) as u8;
-        unsafe { mem::transmute(res) }
+        let res = if USE_LSB_TAG {
+            raw & (!unsafe { VALMASK } as EmacsUint)
+        } else {
+            raw >> (Lisp_Bits::VALBITS as u32)
+        };
+        unsafe { mem::transmute(res as u32) }
```
birkenfeld: If `Lisp_Type` is `#[repr(C)]`, that transmute probably won't work everywhere.

rust_src/src/lisp.rs
```diff
@@ -88,17 +87,17 @@ impl LispObject {
...
     pub fn get_type(self) -> Lisp_Type {
         let raw = self.to_raw() as EmacsUint;
-        let res = (if USE_LSB_TAG {
-                       raw & (!VALMASK as EmacsUint)
-                   } else {
-                       raw >> VALBITS
-                   }) as u8;
-        unsafe { mem::transmute(res) }
+        let res = if USE_LSB_TAG {
+            raw & (!unsafe { VALMASK } as EmacsUint)
```
birkenfeld: Why is VALMASK unsafe?

rust_src/src/multibyte.rs
```diff
@@ -37,16 +37,15 @@ use std::slice;
...
+use remacs_sys::{emacs_abort, char_bits, EmacsInt, Lisp_String, error};
 
 pub type LispStringRef = ExternalPtr<Lisp_String>;
 
 // cannot use `char`, it takes values out of its range
 pub type Codepoint = u32;
 
 /// Maximum character code
-pub const MAX_CHAR: Codepoint = (1 << CHARACTERBITS) - 1;
+pub const MAX_CHAR: Codepoint = (1 << (char_bits::CHARACTERBITS as u32)) - 1;
```
birkenfeld: looks like we have to alias these `char_bits` somewhere, readability quickly degrades here.

db48x:
This will fix #178

rust_src/remacs-sys/lib.rs
db48x: Yes, it has specific values for a few things that wouldn't otherwise be known. Could probably trim it slightly, we do end up with both EmacsInt and EMACS_INT which are the same thing.

rust_src/src/lisp.rs
db48x: It's a static global, so there's no way for the compiler to verify that it's been initialized.

rust_src/src/multibyte.rs
db48x: Agreed. Goes along with creating safe wrappers for everything.

db48x:
../../src/widget.h:24:10: fatal error: 'X11/IntrinsicP.h' file not found
    thread 'main' panicked at 'Unable to generate bindings: ()', src/libcore/result.rs:860:4

Who can install things on the CI server?

birkenfeld:
> Who can install things on the CI server?

At least for Linux, you can put Debian package names into `.travis.yml` under `addons`. Not that does anything on OSX though. On the other hand, doesn't Emacs have native UI for OSX, and doesn't need X11 there?

db48x:
No, the stock Emacs doesn't have a native UI on OSX. There's a port that everybody uses.

Wilfred:
I'm really excited about this! It saves us duplicating type definitions, prevents a variety of mistakes, and saves us work going forward. 👍 💯

DavidDeSimone:
@db48x is there any easy way to inspect the code that bindgen is generating? I'd be curious to see how it handled things bit fields and flexible array members.

db48x:
@DavidDeSimone, sure, just open up the generated bindings.rs file.

db48x:
For an example of a bitfield, check out Lisp_Symbol. It's got a _bitfield_1: [u8; 2usize] member, then an impl that lets you get/set each bitfield.

db48x:
Something went wrong while updating this; it now sometimes corrupts arrays when building them.

db48x:
Oh, nice. I have to do another merge. I guess it will wait until after lunch.

db48x:
Well, that was a fun merge. Tripped over a bindgen bug: https://github.com/servo/rust-bindgen/issues/744

db48x:
oops, I forgot something

Issue #229, opened 2017-07-15
Port processp to Rust


Wilfred:

jsoo1:
I'm looking into this using the walkthrough and I have some beginner questions about some errors I'm seeing during the c macro expansion:

```
In file included from src/systime.h:22:0,
                 from src/thread.h:29,
                 from src/lisp.h:1830,
                 from src/dummy.c:1:
lib/timespec.h:25:3: error: #error "Please include config.h first."
  #error "Please include config.h first."
   ^
In file included from src/dummy.c:1:0:
src/lisp.h:4155:17: error: missing binary operator before token "("
 #if GNUC_PREREQ (3, 0, 0) && ! GNUC_PREREQ (4, 6, 0)
                 ^
src/lisp.h:4537:65: error: missing binary operator before token "("
      && defined __GNUC__ && !defined __clang__ && ! GNUC_PREREQ (4, 3, 2)
```

`src/dummy.c` looks like:
```
#include "lisp.h"


DEFUN ("processp", Fprocessp, Sprocessp, 1, 1, 0,
       doc: /* Return t if OBJECT is a process.  */)
     (Lisp_Object object)
{
  return PROCESSP (object) ? Qt : Qnil;
}
```
What is going on here?

jsoo1:
Ah, so I commented out the GCC version checks in `lisp.h` on 4537 and 4158. And I commented out
``` c
/* #ifndef _GL_INLINE_HEADER_BEGIN */
/*  #error "Please include config.h first." */
/* #endif */
```
in `timespec.h`.  Will my expansion be legitimate? I have gcc 5+ in this container.

atheriel:
Can't help you with the expansion issue, but you might want to look at [the definition of `PROCESSP` itself](https://github.com/Wilfred/remacs/blob/aa5b1277ebad491a361a7bc8d71650310227764c/src/process.h#L206), and the Rust implementation of similar functions (e.g. [this one](https://github.com/Wilfred/remacs/blob/master/rust_src/src/lisp.rs#L545)). Good luck!

DavidDeSimone:
@jsoo1 I believe what you have done is good enough to see what the DEFUN macro is doing. You won't need to do this expansion every time, this exercise is to give you an idea of the magic behind defun, and #[lisp_fn] 

Once you've done that, I would follow @atheriel 's suggestion and begin moving the definition of PROCESSP over, following the linked example.

jsoo1:
[...]

PR #265, opened 2017-07-23
Port processp

jsoo1:
Porting `processp` to rust per [#229](https://github.com/Wilfred/remacs/issues/229).

Not much more to it.  Went smooth once I found `is_process` :smile:. Great community support, too, thanks!

### DIFF
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -151,6 +151,7 @@ pub extern "C" fn rust_init_syms() {
         defsubr(&*windows::Swindowp);
         defsubr(&*windows::Swindow_live_p);
         defsubr(&*process::Sget_process);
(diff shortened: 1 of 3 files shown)


rust_src/src/lib.rs
```diff
@@ -151,6 +151,7 @@ pub extern "C" fn rust_init_syms() {
         defsubr(&*windows::Swindowp);
         defsubr(&*windows::Swindow_live_p);
         defsubr(&*process::Sget_process);
+        defsubr(&*process::Sprocessp);
```
jsoo1: Do I need to export this?

birkenfeld: Do you mean a `pub use` in `lib.rs`? Not if the build succeeds, which means it's not used from C.

PR #266, opened 2017-07-23
Add some macros for convenience and safety

birkenfeld:
Vararg constructs in C should be replaced by macros, such as `error(fmt, ...)` which now uses Rust's formatting machinery.

Multitudes of functions with the same purpose like `xsignalN` and `callN` can also be replaced by macros on the Rust side.

### DIFF
--- a/rust_src/src/floatfns.rs
+++ b/rust_src/src/floatfns.rs
@@ -3,11 +3,10 @@
 use std::mem;
 use libc;
 
-use eval::{xsignal0, xsignal2};
 use math::ArithOp;
 use lisp::{LispObject, LispNumber};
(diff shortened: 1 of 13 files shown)


DavidDeSimone:
ha, I started this exact PR last night. LGTM.

rust_src/src/floatfns.rs
```diff
@@ -86,14 +85,14 @@ pub fn float_arith_driver(
                     accum = next;
                 } else {
                     if next == 0. {
-                        xsignal0(LispObject::from_raw(unsafe { Qarith_error }));
+                        xsignal!(Qarith_error);
```
Wilfred: Lovely! :)

PR #267, opened 2017-07-23
Add LispObarrayRef and port intern functions to Rust

atheriel:
This PR ports a chunk of the obarray-related code from `lread.c` to Rust, including the lisp functions `intern` and `intern-soft`. I had to add some extern functions to the C code in order to get access to some global state.

### DIFF
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -48,6 +48,7 @@ mod process;
 mod fonts;
 mod threads;
 mod chartable;
+mod category;
 
 #[cfg(all(not(test), target_os = "macos"))]
 use alloc_unexecmacosx::OsxUnexecAlloc;
@@ -125,6 +126,9 @@ pub use buffers::Fcurrent_buffer;
 
 // used in chartab.c
 pub use chartable::Fset_char_table_parent;
+
+// used in category.c
+pub use category::Fcategory_table_p;
 
 // Used in process.c
 pub use str2sig::str2sig;
@@ -264,6 +268,8 @@ pub extern "C" fn rust_init_syms() {
         defsubr(&*chartable::Schar_table_subtype);
         defsubr(&*chartable::Schar_table_parent);
         defsubr(&*chartable::Sset_char_table_parent);
+        defsubr(&*category::Scategory_table_p);
+        defsubr(&*category::Scategory_table);
 
         defsubr(&*floatfns::Sisnan);
         defsubr(&*floatfns::Sacos);
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -49,6 +49,7 @@ mod fonts;
 mod threads;
 mod chartable;
 mod category;
+mod obarray;
 
 #[cfg(all(not(test), target_os = "macos"))]
 use alloc_unexecmacosx::OsxUnexecAlloc;
@@ -123,6 +124,9 @@ pub use vectors::Fsort;
 pub use lists::merge;
 pub use buffers::Fget_buffer;
 pub use buffers::Fcurrent_buffer;
+pub use obarray::intern_1;
+pub use obarray::Fintern;
+pub use obarray::Fintern_soft;
 
 // used in chartab.c
 pub use chartable::Fset_char_table_parent;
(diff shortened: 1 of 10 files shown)


rust_src/src/lisp.rs
```diff
@@ -942,3 +943,14 @@ impl Debug for LispObject {
         Ok(())
     }
 }
+
+/// Intern (e.g. create a symbol from) a string.
+#[allow(dead_code)]
```
birkenfeld: dead_code still needed?

rust_src/src/obarray.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// Return the symbol that matches NAME (either a symbol or string). If
+    /// there is no such symbol, return the integer bucket number of where the
+    /// symbol would be if it were present.
+    ///
+    /// # C Porting Notes
+    ///
+    /// This is a wrapper around the C function `oblookup`. It is capable of
+    /// handling multibyte strings, unlike intern_1 and friends.
+    pub fn lookup(&self, name: LispObject) -> LispObject {
+        let string = if name.is_symbol() {
```
birkenfeld: It's still better to use `as_symbol` which does the conversion already.

Anyway, this is basically `LispObject::symbol_or_string_as_string`.

src/lread.c
```diff
@@ -4009,6 +4009,20 @@ check_obarray (Lisp_Object obarray)
   return obarray;
 }
 
+/* Like check_obarray, but for the global Vobarray. */
```
birkenfeld: Is it not possible to access the `V...` from Rust?

atheriel: No, it's not exposed in any .h files. I got a linker error when I tried. Is there another way?

rust_src/src/lisp.rs
atheriel: Nope. Fixed.

rust_src/src/obarray.rs
atheriel: Fixed.

src/lread.c
DavidDeSimone: it looks like Vobarray is defined as `#define Vobarray globals.f_Vobarray` in globals.h. We set up the emacs globals as remacs_sys::globals, so it seems like if we need access to the `V...` variables in Rust, we can just use the global object (unless I am misunderstanding the situation).

atheriel: So... at first I thought this would work. But if I replace

```rust
let obarray = unsafe { check_vobarray() };
```

with

```rust
let obarray = unsafe { check_obarray(globals.f_Vobarray) };
```

The bootstrapping process fails on the first call to `intern_1` in the C code, and Emacs will not build past the `temacs` stage, because `globals.f_Vobarray` is inexplicably `nil` at that point. This is a totally bizarre bug I haven't been able to track down. My best guess at the moment is that the `globals` object might have different contents for `obarray` during the bootstrap process and the regular Emacs incarnation.

A similar issue occurs with `globals.V_fpurify_flag` -- purify is not detected properly during bootstrapping, and symbols are not interned into pure storage when they should be (although the build does complete).

Any thoughts on this would be appreciated.

Wilfred: Ouch, that's odd. Let's open an issue and fix that, but I don't think it should hold up this PR.

atheriel:
@Wilfred I've actually been working on other major symbol table work that builds on this PR, and I fixed one of the `globals` issues in the process. Let me upstream that fix and rebase tonight, and then this first part can be merged.

atheriel:
This should be good to go now.

rust_src/src/lib.rs
```diff
@@ -122,6 +123,9 @@ pub use vectors::Fsort;
 pub use lists::merge;
 pub use buffers::Fget_buffer;
 pub use buffers::Fcurrent_buffer;
+pub use obarray::intern_1;
```
birkenfeld: You're declaring `intern_1` in remacs-sys but also exporting it here...

rust_src/src/obarray.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Return the symbol that matches NAME (either a symbol or string). If
+    /// there is no such symbol, return the integer bucket number of where the
+    /// symbol would be if it were present.
+    ///
+    /// # C Porting Notes
+    ///
+    /// This is a wrapper around the C function `oblookup`. It is capable of
+    /// handling multibyte strings, unlike intern_1 and friends.
+    pub fn lookup(&self, name: LispObject) -> LispObject {
+        let string = LispObject::symbol_or_string_as_string(name);
```
birkenfeld: Should use normal method call syntax here.

birkenfeld:
Resolved conflicts and my two suggestions.

atheriel:
@birkenfeld Thanks for doing that, I wasn't able to get back to it this past weekend.

PR #273, opened 2017-07-25
port point

brotzeit:
I've tried another one, though I'm not really sure what I'm doing ;)


rust_src/src/editfns.rs
```diff
@@ -0,0 +1,14 @@
...
+use remacs_macros::lisp_fn;
+use lisp::{LispObject};
+use remacs_sys::{Fcurrent_buffer, EmacsInt};
+
+
+/// Return value of point, as an integer.
+#[lisp_fn]
+pub fn point() -> LispObject {
+    let buffer = LispObject::from_raw(unsafe { Fcurrent_buffer() });
+    let b = buffer.as_vectorlike().unwrap().as_buffer().unwrap();
```
DavidDeSimone: I think this can be just `let b = buffer.as_buffer().unwrap();`, since as_buffer() will do the as_vectorlike call for you.

DavidDeSimone:
The only thing I see here is that the old point function used PT, which is just
`#define PT (current_buffer->pt + 0)`, which bypasses the type checking that `buffer.as_vectorlike().unwrap().as_buffer().unwrap()` gives us. We might want a 
``` rust
unsafe to_buffer_unchecked(&self) -> LispBufferRef {
    // ... 
}
``` 

added to lisp.rs for this case. I'd be curious what @birkenfeld thinks of that idea.

birkenfeld:
We probably want an API that gives us the current buffer as a buffer ref, for example `LispBufferRef::current()`.

brotzeit:
You mean like in #253 ?

birkenfeld:
Yep.

brotzeit:
Ok, then I wait for the other commit.

Wilfred:
Build is failing for a trivial formatting issue -- run `cargo fmt` in the `rust_src` directory and that should fix it.

brotzeit:
Now it works without the unwrapping stuff. I guess that's because the method current_buffer returns a LispBufferRef instead of LispObject ?

brotzeit:
@Wilfred Are you ok with that ?

brotzeit:
I didn't see #279. I'll rebase tomorrow.

brotzeit:
Why failed the check ?

birkenfeld:
`lisp/net/puny-tests.el` failed. And consistently between OSX and Linux, so it looks genuine.

You can run it locally with `make test/lisp/net/puny-tests.log`.

birkenfeld:
One possible reason is that you're using `z_byte`, while the original talks about characters. You probably have to port Z to a method on `LispBufferRef`.

brotzeit:
My commit from yesterday didn't fail and I just removed the code that was already pushed in #279 and now it fails.

birkenfeld:
It looks like the `buffer_size` commits were never checked though (no green check mark next to them).

birkenfeld:
Can you reproduce the failure locally?

brotzeit:
I could. But I made my commits again locally and now all checks pass. I'll create a new PR and hope it still works.

PR #279, opened 2017-07-30
Define buffer-file-name in Rust

Wilfred:


rust_src/src/lisp.rs
```diff
@@ -434,6 +434,13 @@ impl LispObject {
         self.as_vectorlike().map_or(None, |v| v.as_buffer())
     }
 
+    pub fn as_buffer_or_error(self) -> LispBufferRef {
+        if !self.is_buffer() {
```
birkenfeld: Better to match on the result of `as_buffer` - it's more idiomatic and avoids a second check.

Wilfred: Updated, thanks.

PR #280, opened 2017-08-02
port point and buffer-size

brotzeit:


brotzeit:
Hmmm. I can reproduce it locally again. I probably was in the wrong directory.

2 files contained unexpected results:
  lisp/net/network-stream-tests.log
  lisp/net/puny-tests.log

brotzeit:
Now I know what you meant @birkenfeld. And it seems to work. I'll fix it.

PR #283, opened 2017-08-06
WIP: Translate validate_region

ljos:
I have been trying to implement validate_region in rust, and it looks to me that I have done it correctly. It compiles, remacs runs, but when I run the tests it will seqfault at what seems like the end of every test. I tried to do some debug printing, and `validate_region` does run and completes the way it should at least some of the time, but every test file fails with a segfault. It seems to work when I revert to the old code.

Could anyone guide me on how to debug this?

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -10,14 +10,25 @@ use threads::ThreadState;
 
 use remacs_macros::lisp_fn;
 
+pub const BEG: ptrdiff_t = 1;
 pub const BEG_BYTE: ptrdiff_t = 1;
 
 pub type LispBufferRef = ExternalPtr<Lisp_Buffer>;
 
 impl LispBufferRef {
     #[inline]
+    pub fn zv(&self) -> ptrdiff_t {
+        self.zv
+    }
+
+    #[inline]
     pub fn beg_addr(&self) -> *mut c_uchar {
         unsafe { (*self.text).beg }
+    }
+
+    #[inline]
+    pub fn beg(&self) -> ptrdiff_t {
+        BEG
     }
 
     #[inline]
@@ -53,6 +64,11 @@ impl LispBufferRef {
     #[inline]
     pub fn z_byte(&self) -> ptrdiff_t {
         unsafe { (*self.text).z_byte }
+    }
+
+    #[inline]
+    pub fn z(&self) -> ptrdiff_t {
+        unsafe { (*self.text).z }
     }
 
     // Check if buffer is live
--- a/rust_src/src/editfns.rs
+++ b/rust_src/src/editfns.rs
@@ -1,0 +1,43 @@
+//! Lisp functions pertaining to editing.
+
+use remacs_macros::lisp_fn;
+use lisp::LispObject;
+use remacs_sys::EmacsInt;
+use threads::ThreadState;
+use buffers::get_buffer;
+
+
+/// Return value of point, as an integer.
+/// Beginning of buffer is position (point-min).
+#[lisp_fn]
+pub fn point() -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
+    LispObject::from_natnum(buffer_ref.pt as EmacsInt)
+}
+
+/// Return the number of characters in the current buffer.
+/// If BUFFER is not nil, return the number of characters in that buffer
+/// instead.
+///
+/// This does not take narrowing into account; to count the number of
+/// characters in the accessible portion of the current buffer, use
+/// `(- (point-max) (point-min))', and to count the number of characters
+/// in some other BUFFER, use
+/// `(with-current-buffer BUFFER (- (point-max) (point-min)))'.
+#[lisp_fn(min = "0")]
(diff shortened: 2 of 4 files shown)


birkenfeld:
One thing I see with only a quick review is that you're not swapping the actual pointers passed into the function, only the locals. That might be enough to fix the crashes.

birkenfeld:
Related, the marker coercion changes the arguments in the C version but not the Rust version.

ljos:
I think you are right; the biggest problem is the marker coercion. I didn't read the code there correctly. I thought they just needed to check things, but it actually needs to change, not just check.

I will fix it and see what happens.

rust_src/src/lisp.rs
```diff
@@ -339,6 +339,17 @@ impl LispObject {
         }
     }
 
+    #[inline]
+    pub fn as_fixnum_coerce_marker_or_error(self) -> EmacsInt {
```
shanavas786: This is duplication, use `as_number_coerce_marker_or_error` instead

ljos: No, because as_number also allows floats. This only allows fixnums.

Wilfred: I think it's slightly unfortunate that the other `.as_foo` methods return Rust wrappers, whereas `.as_fixnum_coerce_marker_or_error` just returns an `EmacsInt`. I don't have any better ideas though.

Wilfred:
Build is failing due to rustfmt. Is the PR ready to go otherwise? Code looks equivalent to the C to me.

ljos:
I fixed the rustfmt issue, but I don't think it is ready. Like I said eariler, I get errors on test and I haven't been able to figure out how to fix it yet.

rust_src/src/buffers.rs
```diff
@@ -129,3 +131,26 @@ pub fn buffer_file_name(buffer: LispObject) -> LispObject {
...
     LispObject::from_raw(buf.filename)
 }
+
+
+#[no_mangle]
+pub extern "C" fn validate_region(b: *mut Lisp_Object, e: *mut Lisp_Object) {
+    let start = LispObject::from_raw(unsafe { *b });
+    let stop = LispObject::from_raw(unsafe { *e });
+
+    let mut beg = start.as_fixnum_coerce_marker_or_error();
```
birkenfeld: You're still not coercing the original pointers, or swapping the original pointers.

ljos: Yes - after some testing this is definitely the problem, but I don't really understand how to fix it.  I was thinking if I used `mem::replace` on the marker with the fixnum at `m.position()` in `as_fixnum_coerce_marker_or_error`, so:
```rust
mem::replace(&mut self, Self::from_fixnum(m.position() as EmacsInt))
``

I get the right return from `as_fixnum_coerce...`, but it doesn't seem to change the value at the original pointer. I thought that since I got the LispObject by converting it from the raw pointer from the input of `validate_region` that should be the right thing to do.

Like I said, I don't seem to be understanding what is going on here.

**Comment by ljos** [comment]
Seems I was overcomplicating what was going on. I didn't realise completely how the lispobjects worked, but now it seems to be right. There are still some errors, but I think that is because I am not swapping the values at the pointer when b > e yet. Looking to fix that later.

File: rust_src/src/buffers.rs [hunk 3]
```diff
@@ -129,3 +131,31 @@ pub fn buffer_file_name(buffer: LispObject) -> LispObject {
...
+    let mut end = stop.as_fixnum_coerce_marker_or_error();
+
+    if beg > end {
+        mem::swap(&mut beg, &mut end);
+    }
+
+    unsafe {
+        *b = start.to_raw();
```
birkenfeld: Just use `beg.to_raw()` and `end.to_raw()` and that should be it.

ljos: `beg` and `end` are `EmacsInt` and not `LispObject`.

birkenfeld: Right then, to_raw -> from_fixnum

ljos: I fixed it in the latest commit; I swap start and end as well. It seems to work.

ljos:
I don't think this is the most elegant code. I would prefer if we would have a fixnum type which implemented PartialOrd that I could return from `as_fixnum_coerce_marker_or_error`. I think that would make life easier. Maybeit is possible to use LispNumber for that, but I don't know yet.

birkenfeld:
I don't think this is correct; the C code coerces a marker and writes the fixnum back to `*b`.

ljos:
Isn't that what I am doing?

I take the value at '*b' convert it to 'LispObject'. That object, if it is a 'Marker', is coerced to a 'Fixnum' and then the '*b' is set to the raw value of the (coerced) LispObject. It might be less efficient since I always set '*b', but I am not entirely sure how I would do it otherwise as `as_fixnum_coerce_marker_or_error` doesn't have access to `*b`.

The error is because of some unneeded `mut` that will be fixed with #288. `make check` works on my machine.

ljos:
Actually, I think you are right. I am testing a fix now.

Issue #276, opened 2017-07-26
[RFC] Using Rust's HashMap for a LispHashTable


DavidDeSimone:
[Rendered](https://gist.github.com/DavidDeSimone/400e0f6041d2f4a1eb7938f113a6036f)

birkenfeld:
Very nice writeup! It would be cool to try (but that's easy to say when you're not the one doing it).

As for rooting: not that I oppose an API to do explicit rooting, but doesn't the GC scan the stack and mark any LispObjects found there anyway? (Remembering my brief past forays into Emacs C code, there used to be these `GCPRO` macros that C functions use to make sure their variables weren't collected, but they're completely gone now.)

DavidDeSimone:
Thanks for the feedback @birkenfeld . It looks like you are correct that the lisp GC does scan the stack and mark what it believes are Lisp_Objects. The GC calls mark_threads defined in thread.c which comes back to mark_stack in alloc.c. I am going to play around with this to make sure that it is recognizing Rust LispObjects, but assuming that all works, I think we can remove the rooting API from the spec.

Wilfred:
This looks great, thanks for taking the time to write this up.

## Performance

I would be interested in benchmarks, to see if Rust's HashMap is faster regardless of how a user tunes `rehash-size` and `rehash-threshold`. Naively I think HashMap may be faster, because it's had more people use it in more situations.

We should probably consider a hash function like [fnv](https://doc.servo.org/fnv/). I don't believe there's any DoS protection in the current hashmap implementation, so we should take the faster hash function.

## Compatibility

`rehash-size` and `rehash-threshold` are exposed to user elisp, so we need to preserve backwards compatibility.

We should probably just return placeholder values, so something like:

```
;; ignore :rehash-size and :rehash-threshold
ELISP> (setq h (make-hash-table :rehash-size 1.9 :rehash-threshold 0.9))
;; don't bother showing rehash-size or rehash-threshold in the
;; printed representation
#s(hash-table size 65 test eql data ())

ELISP> (hash-table-rehash-size h)
nil ;; or 1.5?
ELISP> (hash-table-rehash-threshold h)
nil ;; or 0.8125?
```

Note that the default value of `rehash-threshold` has changed from 0.8 to 0.8125 in Emacs 26 (see commits 7207b63c8e290ddec33908ce8d38be5793388318 and 83c9c6fc1cc943f239a021b42a8ca9b0e162198c).

## Usage of rehash-size and rehash-threshold

[...]

PR #286, opened 2017-08-07
[WIP] Port Remacs Lisp_Hash_Table to use FnvHashMap. 

DavidDeSimone:
This is the  WIP commit for https://github.com/Wilfred/remacs/issues/276. 

TODO List:
- [x]  Manually derive PartialEq and Eq for HashableLispObject
  - [x] Implement user defined hash function and cmp function. 
- [x] Add "purecopy" functionality. 
- [x] Remove def. of Lisp_Hash_Table from C/Rust
- [x] Rename all "LispHashMap" functions to "LispHashTable" equiv. 
  - [x] Make "LispHashTable" be properly tagged as a HashTable pseudovector, with proper sizing. 
  - [x] Convert "get_hash_table_or_error" to return a LispHashTableRef that references a LispHashTable, instead of a Lisp_Hash_Table. 
- [x] Have "LispHashTable" be allocated by the Lisp GC, and managed by the Lisp GC
  - [x] Add a "hashtable_finalize" method for the cleanup at the rust layer.
- [x] Add weak table support. 
- [ ] Handle profiler.c's custom hash/cmp function for it's hash_table_test
- [x] Fix temacs build stage segfaults
- [ ] Final test/integration pass
- [ ] Final performance/benchmarking pass


DavidDeSimone:
So right now I am dealing with some problems in the _unexec_ codepath. temacs builds and runs fine, however attempting to access a particular allocation is failing when running bootstrap-emacs. 

The Vcoding_system_hash_table Hash Table seems to be valid when running temacs, but is not valid when running bootstrap-emacs. Via GDB, I can verify the state of this hash table. Internally, it's rust HashMap looks like: 

table = {capacity_mask = 1023,  size = 899, hashes = 0x7fffee429010}

When running temacs, the 'hashes' memory address is a valid address that I can deference. However, running bootstrap-emacs, that address is no longer valid, and attempting to deference it will lead to a seg fault. This address is not always the same, however it seems to always be in the 0x7fffe.. range. My first thought was there was some trouble mixing Rust's default allocator, with the unexec flow, so I switched remacs over to globally using malloc, similar to how we use unexec_malloc on OSX. However, this has not helped. Interestingly enough, the vector that stores the keys and values of the hash map always seems to be valid. It is ONLY this hashmap's internal hash data that is invalid. It's also interesting to note that this is not happening for ALL hashmaps. Most of them seem to work fine in the dumped bootstrap-emacs.  

I am going to keep investigating, however I am open to community ideas on what paths I should go down on this.

DavidDeSimone:
After further investigation, it turns out that sometimes our HashMaps will have indivdual allocations above M_MMAP_THRESHOLD, which is set to 64 * 1024 in alloc.c. Since the allocation is above that threshold, it will use mmap over brk/sbrk, and thus will not be copied into the new bss segment when emacs unexec's. 

Looking back, it seems obvious, but this was a bit of a doozy to figure out haha.

DavidDeSimone:
So this is making good progress, and all but one testing is passing! This ended up being slightly more difficult then I anticipated, due to the fact that most of the C-layer makes a lot of assumptions about the internals of HashTables and how they work. 

I need to spend some time benchmarking and optimizing this guy to make sure that we are still getting the advertised gains, but I hope to have this PR ready for wide-review by the end of the month.

DavidDeSimone:
After a long journey, and a lot of code, I am going to advise that we do *not* use Rust's HashMap at this time, and that we make a straight port of existing HashTable code. 

Reasons:

- While the lisp layer has a very clean API for interacting with a HashTable, the C layer makes a lot of assumptions on the structure of the HashTable. 
- There are certain operations that update the key of the hash table in place. While I have a possible way of doing this, I am very leery of it, because we are making Rust's HashMap do something it explicitly doesn't support. 
- In order to support APIs that the C code bases uses to interact with hash tables (including an API to update a key/value in constant time by directly writing into it's private data vector), we have to maintain a lot of boilerplate. 
- After testing and playing around with the current code writing, I believe there is a strong chance that this refactor will introduce divergence and incompatibility between upstream in the future. 

If no one objects, I am going to close this PR, and will open on a PR that focuses on converting the existing HashTable from C to Rust piecemeal.

brotzeit:
If you think that's the best solution do it. Maybe we should avoid to make things too complicated for now.

Wilfred:
Whilst this hasn't resulted in us landing code that uses FnvHashMap, it's a really useful conclusion. Thanks for exploring this so thoroughly!

Could you comment on where the code assumes that it can mutate keys in hash tables? I would have thought we will eventually want to refactor that out.

PR #290, opened 2017-08-14
Translate validate_region

ljos:
I think the code as it is now should be functionally equal to the C code. It might be a little less efficient as we are setting the `*b` and `*e` on every validation regardless if something changed or not.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -3,10 +3,12 @@
 use libc::{c_void, c_uchar, ptrdiff_t};
 
 use lisp::{LispObject, ExternalPtr};
-use remacs_sys::{Lisp_Buffer, Lisp_Type, Vbuffer_alist, EmacsInt, make_lisp_ptr};
+use remacs_sys::{Lisp_Object, EmacsInt, Lisp_Buffer, Lisp_Type, Vbuffer_alist, make_lisp_ptr};
 use strings::string_equal;
 use lists::{car, cdr};
 use threads::ThreadState;
+
+use std::mem;
 
 use remacs_macros::lisp_fn;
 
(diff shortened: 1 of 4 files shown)


rust_src/src/buffers.rs
```diff
@@ -211,3 +213,30 @@ fn buffer_modified_tick(buffer: LispObject) -> LispObject {
...
+        mem::swap(&mut beg, &mut end);
+    }
+
+    unsafe {
+        *b = LispObject::from_fixnum(beg).to_raw();
+        *e = LispObject::from_fixnum(end).to_raw();
+    }
+
+    let buffer = current_buffer();
+    let buf = buffer.as_buffer_or_error();
```
birkenfeld: I think this is just `let buf = ThreadState::current_buffer();`

rust_src/src/buffers.rs
```diff
@@ -211,3 +213,30 @@ fn buffer_modified_tick(buffer: LispObject) -> LispObject {
...
+    let mut beg = start.as_fixnum_coerce_marker_or_error();
+    let mut end = stop.as_fixnum_coerce_marker_or_error();
+
+    if beg > end {
+        mem::swap(&mut beg, &mut end);
+    }
+
+    unsafe {
+        *b = LispObject::from_fixnum(beg).to_raw();
+        *e = LispObject::from_fixnum(end).to_raw();
```
birkenfeld: This looks good now!

rust_src/src/buffers.rs
ljos: I need the current buffer as a LispObject for the `args_out_of_range!` call, but I can change it if you want and just call `current_buffer` there.

birkenfeld: Sounds good!

PR #292, opened 2017-08-16
port marker-position

brotzeit:
I would do it this way and change the old method to charpos_or_error ? As `Fmarker_position` is still used, I couldn't remove marker-position.

### DIFF
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -28,7 +28,7 @@ use remacs_sys::{EmacsInt, EmacsUint, EmacsDouble, VALMASK, VALBITS, INTTYPEBITS
                  make_float, circular_list, internal_equal, Fcons, CHECK_IMPURE, Qnil, Qt,
                  Qnumberp, Qfloatp, Qstringp, Qsymbolp, Qnumber_or_marker_p, Qwholenump, Qvectorp,
                  Qcharacterp, Qlistp, Qintegerp, Qhash_table_p, Qchar_table_p, Qconsp, Qbufferp,
-                 SYMBOL_NAME, PseudovecType, EqualKind};
+                 Qmarkerp, SYMBOL_NAME, PseudovecType, EqualKind};
 
 // TODO: tweak Makefile to rebuild C files if this changes.
 
@@ -847,7 +847,7 @@ impl LispObject {
         } else if let Some(f) = self.as_float() {
             LispNumber::Float(f)
         } else if let Some(m) = self.as_marker() {
-            LispNumber::Fixnum(m.position() as EmacsInt)
+            LispNumber::Fixnum(m.charpos_or_error() as EmacsInt)
         } else {
             wrong_type!(Qnumber_or_marker_p, self)
         }
@@ -879,6 +879,12 @@ impl LispObject {
             } else {
                 None
             },
+        )
+    }
+
+    pub fn as_marker_or_error(self) -> LispMarkerRef {
+        self.as_marker().unwrap_or_else(
+            || wrong_type!(Qmarkerp, self),
         )
     }
 
--- a/rust_src/src/marker.rs
+++ b/rust_src/src/marker.rs
@@ -12,8 +12,7 @@ pub type LispMarkerRef = ExternalPtr<Lisp_Marker>;
 
 impl LispMarkerRef {
     pub fn charpos(self) -> Option<ptrdiff_t> {
-        let buf = self.buffer;
-        if buf.is_null() {
+        if self.buffer.is_null() {
             None
         } else {
             Some(self.charpos)
@@ -21,8 +20,7 @@ impl LispMarkerRef {
     }
 
     pub fn charpos_or_error(self) -> ptrdiff_t {
-        let buf = self.buffer;
-        if buf.is_null() {
+        if self.buffer.is_null() {
             error!("Marker does not point anywhere");
         }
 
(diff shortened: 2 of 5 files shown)


rust_src/src/marker.rs
```diff
@@ -17,10 +17,33 @@ impl LispMarkerRef {
...
 #[lisp_fn]
 fn markerp(object: LispObject) -> LispObject {
     LispObject::from_bool(object.is_marker())
 }
+
+/// Return the position of MARKER, or nil if it points nowhere.
+#[lisp_fn]
+fn marker_position(object: LispObject) -> LispObject {
+    if object.is_marker() {
+        let pos = object.as_marker().unwrap().charpos();
```
shanavas786: I would prefer a `as_marker_or_error` method.

src/marker.c
```diff
@@ -761,7 +761,6 @@ verify_bytepos (ptrdiff_t charpos)
 void
 syms_of_marker (void)
 {
-  defsubr (&Smarker_position);
```
shanavas786: remove `marker-position` definition too.

rust_src/src/marker.rs
```diff
@@ -17,10 +17,33 @@ impl LispMarkerRef {
         // TODO: add assertions that marker_position in marker.c has.
         self.charpos
     }
+
+    pub fn charpos(self) -> Option<ptrdiff_t> {
```
shanavas786: `charpos()` and `position()` methods are ambiguous. following the conventions, `charpos()` and `charpos_or_error()` would be better.

src/marker.c
shanavas786: Add `pub use markers::Fmarker_position;` to `src/lib.rs`, then you can remove `C` defun

brotzeit: Could have figured that out by myself. Thanks ;)

rust_src/src/lisp.rs
```diff
@@ -882,6 +882,12 @@ impl LispObject {
         )
     }
 
+    pub fn as_marker_or_error(self) -> LispMarkerRef {
+        self.as_marker().unwrap_or_else(
+            || wrong_type!(Qmarkerp, self),
```
shanavas786: closure is not necessary. you can use `self.as_marker().unwrap_or(wrong_type!(Qmarkerp, self))`

rust_src/src/marker.rs
```diff
@@ -24,3 +32,13 @@ impl LispMarkerRef {
...
     LispObject::from_bool(object.is_marker())
 }
+
+/// Return the position of MARKER, or nil if it points nowhere.
+#[lisp_fn]
+fn marker_position(object: LispObject) -> LispObject {
+    let pos = object.as_marker_or_error().charpos();
+    match pos {
+        None => LispObject::constant_nil(),
+        _ => LispObject::from_natnum(pos.unwrap() as EmacsInt),
```
shanavas786: use destructuring:  `Some(var)` in place of `_`

brotzeit: Would be `Some(ptrdiff_t) => LispObject::from_natnum(ptrdiff_t as EmacsInt),` ok to indicate the type ?

rust_src/src/lisp.rs
brotzeit: I get an error `#[warn(unreachable_code)] on by default`. Is it possible that the macro needs a new scope to work ?

rust_src/src/marker.rs
shanavas786: variable name, not type `Some(t) => LispObject::from_natnum(t as EmacsInt)`

brotzeit: Yeah I know. I was just wondering if this might be ok in this case. But you're right.

rust_src/src/lisp.rs
shanavas786: well, I got this from @levans:safaradeg.net on #rust:matrix.org,

> it's because the arguments of a function are evaluated before the call is made

brotzeit: Interesting. Thanks for the info.

birkenfeld: @shanavas786 just to be clear, that is the argument for *having* the closure.

rust_src/src/marker.rs
```diff
@@ -1,14 +1,26 @@
...
 
+use buffers::LispBufferRef;
+
 pub type LispMarkerRef = ExternalPtr<Lisp_Marker>;
 
 impl LispMarkerRef {
-    /// Return the char position of marker MARKER, as a C integer.
-    pub fn position(self) -> ptrdiff_t {
+    pub fn charpos(self) -> Option<ptrdiff_t> {
+        let buf = self.buffer;
```
birkenfeld: more style than anything, but the separate `let` seems unnecessary here.

rust_src/src/marker.rs
```diff
@@ -17,10 +29,46 @@ impl LispMarkerRef {
         // TODO: add assertions that marker_position in marker.c has.
         self.charpos
     }
+
+    pub fn buffer(self) -> Option<LispBufferRef> {
+        let buf = self.buffer;
+        if buf.is_null() {
+            None
+        } else {
+            Some(unsafe { mem::transmute(self.buffer) })
```
birkenfeld: here, if you keep `buf`, use it :)

rust_src/src/lisp.rs
Wilfred: Yep. An example might be helpful here. The following snippets are not equivalent:

```rust
fn bar() -> u32 {
    println("bar called");
    42
}

foo.unwrap_or_else(|| bar()); // will print 'bar called' only when foo is None
foo.unwrap_or(bar()); // will always print 'bar called'
```

Wilfred:
Merged in #296. Thank you :)

PR #293, opened 2017-08-16
Port bolp and eolp to rust

kidd:
With the PR as it is, I get `^^^^^^^^^^ expected u8, found char`, which makes sense because I'm comparing the address offset address with the char '\n' itself.   

I couldn't find the function that would get the char at a given addr (something with a signature like: buffer.char_at_addr(addr)).    Do we have such function? 

Also, is that use of unsafe, safe? :) or at least, sensible?

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -99,13 +99,13 @@ impl LispBufferRef {
 
     #[inline]
     pub fn fetch_byte(&self, n: ptrdiff_t) -> u8 {
-        let base_addr = if n >= self.gpt_byte() {
+        let offset = if n >= self.gpt_byte() {
             self.gap_size()
         } else {
             0
         };
 
-        unsafe { self.beg_addr().offset(base_addr + n - self.beg_byte()) as u8 }
+        unsafe { self.beg_addr().offset(offset + n - self.beg_byte()) as u8 }
     }
 }
 
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -96,6 +96,19 @@ impl LispBufferRef {
     pub fn is_live(self) -> bool {
         LispObject::from_raw(self.name).is_not_nil()
     }
+
+    #[inline]
+    pub fn fetch_byte(&self, n: ptrdiff_t) -> u8 {
+        let base_addr = unsafe {
+            if n  >= self.gpt_byte() {
+                self.gap_size()
+            } else {
+                0
+            } };
+        let byte_addr = (base_addr + n +
+                         self.beg_addr() as ptrdiff_t - self.beg_byte()) as *const u8;
+        unsafe {(*byte_addr) as u8 }
+    }
 }
 
 impl LispObject {
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -42,6 +42,11 @@ impl LispBufferRef {
     }
 
     #[inline]
+    pub fn gap_size(&self) -> ptrdiff_t {
+        unsafe { (*self.text).gap_size }
+    }
+
+    #[inline]
     pub fn gap_end_addr(&self) -> *mut c_uchar {
         unsafe {
             (*self.text).beg.offset(
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -107,7 +107,7 @@ impl LispBufferRef {
             0
         };
 
-        unsafe { self.beg_addr().offset(offset + n - self.beg_byte()) as u8 }
+        unsafe { *(self.beg_addr().offset(offset + n - self.beg_byte())) as u8 }
     }
 }
 
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -99,12 +99,12 @@ impl LispBufferRef {
 
     #[inline]
     pub fn fetch_byte(&self, n: ptrdiff_t) -> u8 {
-        let base_addr = unsafe {
+        let base_addr =
             if n  >= self.gpt_byte() {
                 self.gap_size()
             } else {
                 0
-            } };
+            };
         let byte_addr = (base_addr + n +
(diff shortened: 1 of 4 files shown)


Wilfred:
I think the use of `unsafe` here is reasonable. I believe `.text` is never NULL (surely a buffer always has text?).

We don't have an API for accessing bytes in buffers yet. Looking at the relevant C macros:

```c
/* Return the byte at byte position N.  */

#define FETCH_BYTE(n) *(BYTE_POS_ADDR ((n)))

#define BYTE_POS_ADDR(n) \
  (((n) >= GPT_BYTE ? GAP_SIZE : 0) + (n) + BEG_ADDR - BEG_BYTE)
```

This looks like it's doing pointer arithmetic of bytes before or after the gap. We should probably do the same thing.

kidd:
I think now it's in a mergeable state.  There was some trial and error in the castings (casting everything to u32, then to isize, and ended up using pdiff_t (I'd appreciate any feedback)).

Also ported eolp.

There's clearly some space for refactoring in both (they are basically (bobp() || .....)  and (eobp() || ....)  and we're not reusing them, but I'm not confident enough we want to create new internal functions with different signatures (bool instead of LispObject).

kidd:
mmm... It looks like it's a formatting issue.

The thing is that if I run `[~/remacs/rust_src] $ cargo fmt -- --write-mode=diff` I get lots of other diffs. 

Any hints?

Thanks!

birkenfeld:
You might have an old version of rustfmt. The default style changed quite a lot with 0.9.0.

rust_src/src/buffers.rs
```diff
@@ -91,6 +96,19 @@ impl LispBufferRef {
...
+    #[inline]
+    pub fn fetch_byte(&self, n: ptrdiff_t) -> u8 {
+        let base_addr =
+            if n  >= self.gpt_byte() {
+                self.gap_size()
+            } else {
+                0
+            };
+        let byte_addr = (base_addr + n +
+                         self.beg_addr() as ptrdiff_t - self.beg_byte()) as *const u8;
```
birkenfeld: looks like that should be `self.beg_addr().offset(base_addr + n - self.beg_byte())`

rust_src/src/editfns.rs
```diff
@@ -49,3 +49,22 @@ pub fn bobp() -> LispObject {
     let buffer_ref = ThreadState::current_buffer();
     LispObject::from_bool(buffer_ref.pt == buffer_ref.begv)
 }
+
+/// Return t if point is at the beginning of a line.
+#[lisp_fn]
+pub fn bolp() -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
+    let pt_byte = buffer_ref.pt_byte - 1;
```
birkenfeld: hmm, calling `pt_byte - 1` also `pt_byte` is a bit confusing.

rust_src/src/editfns.rs
```diff
@@ -49,3 +49,22 @@ pub fn bobp() -> LispObject {
...
+    LispObject::from_bool(buffer_ref.pt == buffer_ref.begv ||
+                          buffer_ref.fetch_byte(pt_byte) as char == '\n')
```
birkenfeld: instead of `as char`, compare with `b'\n'`

kidd:
Thanks for the comments! I'll fix those asap.



_Sent from my Lge LG-K580 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

rust_src/src/buffers.rs
```diff
@@ -91,6 +96,17 @@ impl LispBufferRef {
     pub fn is_live(self) -> bool {
         LispObject::from_raw(self.name).is_not_nil()
     }
+
+    #[inline]
+    pub fn fetch_byte(&self, n: ptrdiff_t) -> u8 {
+        let base_addr = if n >= self.gpt_byte() {
```
Wilfred: I'd prefer we called this `offset` -- it's not actually an address (0 wouldn't make sense as an address here).

rust_src/src/editfns.rs
```diff
@@ -49,3 +49,22 @@ pub fn bobp() -> LispObject {
...
+    )
+}
+
+/// Return t if point is at the end of a line.
+///`End of a line' includes point being at the end of the buffer.
+#[lisp_fn]
+pub fn eolp() -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
+    LispObject::from_bool(
+        buffer_ref.pt == buffer_ref.zv() || buffer_ref.fetch_byte(buffer_ref.pt_byte) == b'\n',
```
kidd: cargo fmt told me about that final comma, but.... does it make sense? I find it a bit strange, no?

kidd:
still timeouts :(. Next week I'm going to have some time to bisect over the commits, or try to investigate a bit more. It's a pity it doesn't happen in my local machine, just in travisCI :(

rust_src/src/editfns.rs
Wilfred: Agreed, it's a little odd. Perhaps it's a rustfmt bug.

I like that rustfmt is a tool that you just run and it fixes formatting without human input. We could file a bug upstream, but I'd leave the comma for now if rustfmt wants it.

Wilfred:
@kidd have you tried running the test suite yourself? Does it terminate? I've been thinking that `(while (not (bolp)) ...)` is quite common, so it's possible a bug has made a test go into an infinite loop with this PR.

rust_src/src/editfns.rs
birkenfeld: I don't think it's a big, it's just consistent in always having a final comma on multi-line lists of expressions. The argument for having the trailing comma in the first place, less churn when an element is added, applies to single arguments as well.

kidd:
@Wilfred you're right, the infinite loop was caused by not returning eolp => t in certain situations. 

The bug was introduced during the latest refactors and went under the radar. Hopefully now it's green.

Wilfred:
Awesome :)

PR #298, opened 2017-08-27
port some functions

brotzeit:


rust_src/src/minibuf.rs
```diff
@@ -0,0 +1,33 @@
...
+    } else if object.is_string() {
+        get_buffer(object)
+    } else {
+        unsafe {
+            LispObject::from_raw(make_lisp_ptr(
+                object.as_buffer_or_error().as_ptr() as *mut c_void,
+                Lisp_Type::Lisp_Vectorlike,
+            ))
+        }
+    };
```
brotzeit: This doesn't look good to me. Is it better to check if object is a buffer and throw the error here instead of calling `as_buffer_or_error` ?

src/window.c
```diff
@@ -1526,27 +1526,6 @@ column 0.  */)
...
-Note that, when WINDOW is selected, the value returned is the same as
-that returned by `point' for WINDOW's buffer.  It would be more strictly
-correct to return the top-level value of `point', outside of any
-`save-excursion' forms.  But that is hard to define.  */)
-  (Lisp_Object window)
-{
-  register struct window *w = decode_live_window (window);
-
-  if (w == XWINDOW (selected_window))
-    return make_number (BUF_PT (XBUFFER (w->contents)));
```
Wilfred: You don't seem to be covering this case in the Rust implementation -- it's always getting the marker position. Am I missing something?

brotzeit: I forgot it, but I still don't understand why this has to be handled separately. But I fxied it. Thanks.

Wilfred:
Awesome, this all looks good. In my (totally biased opinion), the code you've written is significantly more readable than the C it replaces :)

Sorry for the delay in reviewing this, I have less time to review PRs in the next week or so.

Wilfred:
@brotzeit I've also given you collaborator powers, so you also have the power to approve PRs if you fancy ✨

PR #300, opened 2017-08-31
Fix new warning on latest Rust nightly

Wilfred:
This changed in https://github.com/rust-lang/rust/pull/42588.


kidd:
I guess this will be the solution to https://github.com/Wilfred/remacs/pull/293#issuecomment-326349978 , right?

Wilfred:
@kidd yep!

PR #302, opened 2017-09-04
Port point-min and point-max to Rust

kidd:


rust_src/src/editfns.rs
```diff
@@ -68,3 +68,19 @@ pub fn eolp() -> LispObject {
         buffer_ref.pt == buffer_ref.zv() || buffer_ref.fetch_byte(buffer_ref.pt_byte) == b'\n',
     )
 }
+
+/// Return the minimum permissible value of point in the current
+/// buffer.  This is 1, unless narrowing (a buffer restriction) is in
+/// effect.
+#[lisp_fn]
+pub fn point_min() -> LispObject {
+    LispObject::from_natnum(ThreadState::current_buffer().begv as EmacsInt)
```
kidd: begv (the same as zv) is both an "attribute" of buffer and an accessor. Better use the accessor, right?

PR #303, opened 2017-09-07
WIP: Port markers to rust

kidd:
This might be a bit too much for me, but I'm giving it a try.

In https://github.com/Wilfred/remacs/blob/master/rust_src/remacs-sys/lib.rs#L287 , the fields 'insertion_type' and 'need_adjustment', are not migrated. is it because we won't need that in Rust, or it is something we have to do and what's there is a hack?

Any tip/advice/alert for porting markers?

### DIFF
--- a/rust_src/src/marker.rs
+++ b/rust_src/src/marker.rs
@@ -3,7 +3,7 @@ use std::mem;
 
 use remacs_sys::make_lisp_ptr;
 use lisp::{LispObject, ExternalPtr};
-use remacs_sys::{Lisp_Marker, EmacsInt, Lisp_Type};
+use remacs_sys::{Lisp_Misc_Type, Lisp_Marker, EmacsInt, Lisp_Type};
 use remacs_macros::lisp_fn;
 
 use buffers::LispBufferRef;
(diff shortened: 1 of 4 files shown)


rust_src/src/marker.rs
```diff
@@ -39,6 +39,19 @@ impl LispMarkerRef {
     }
 }
 
+fn build_marker(buf: LispObject, charpos: ptrdiff_t, byte_pos: ptrdiff_t) -> LispObject {
```
Wilfred: I suggest you add the original docstring here too.

rust_src/src/marker.rs
```diff
@@ -39,6 +39,19 @@ impl LispMarkerRef {
...
+    let m = Lisp_Marker{
```
Wilfred: I think we'd benefit from using the same assertions that are in the C. `eassert` in C is equivlaent to `debug_assert` in Rust.

Wilfred: You'll need to call `allocate_misc` rather than `Lisp_Marker{...}`. As written, this is creating a struct on the stack and returning it, which won't work.

kidd:
Thanks Wilfred!

The "easy" things I'm already doing them. It'll take more time to do the more complex tasks:

- [ ] Write and use allocate_misc.
- [ ] Create functionality for those bit fields. If Rust doesn't support that out of the box, we can use them as one field and have accessors that use bitmasks to extract the values we want.
- [ ] Where to use LispObject vs Lisp_Marker

jaccarmac:
Will the work being done on bitfields make this stuff easier to port?

shaleh:
I am inclined to drop this PR for now. We are not ready to replace functions in alloc.c.

kidd:
@shaleh , yes, feel free to delete this one. I went totally MIA, and moreso if you think it's not in the stage to be moved forward.

shaleh:
@kidd the project will be here when you have time.

Issue #299, opened 2017-08-31
Writing tests with remacs-sys dependencies


rlupton20:
Hi, I jut discovered this project yesterday, and first would like to say this is a cool idea, even if just a chance to dig into the emacs internals. So I decided to get stuck in, first by writing a few tests in `lisp.rs`, just to get oriented. This issue is mostly to track my progress, gather input, and help future wayfarers.

I wrote this little test

```
#[test]
fn test_float_promotion_on_exceeding_max_int() {
    let x = LispObject::int_or_float_from_fixnum(MOST_POSITIVE_FIXNUM+1);
    assert!(x.as_fixnum() == None);
}
```

but on running `cargo test` I get a linker error
```
  = note: /home/richard/remacs/rust_src/target/debug/deps/remacs-c65f4237c5aa0389.0.o: In function `remacs::lisp::{{impl}}::from_float':
          /home/richard/remacs/rust_src/src/lisp.rs:78: undefined reference to `make_float'
          collect2: error: ld returned 1 exit status
```

(`make check` also fails with this test in place, `cargo build` works - to be expected since presumably on build we don't need to do any linking).

Tracing this through, I found `make_float` is declared as an C function in `remacs-sys`, and defined in `src/alloc.c`. I could work all the way down, and port all of the functions across to rust, but perhaps in general there is some merit to being able to pin these functions with tests before doing the port. This would also make it more viable for people to contribute tests to the project as a way to get started. 

Searching around, I couldn't find an obvious way to make `cargo test` link against the C code - is it known if there is a way to do this? Presumably we want `cargo build` to ignore linking.

Wilfred:
Great question! Yes, we don't have a solution yet: the few tests we have don't link against `remacs-sys` because of exactly this problem.

We've mostly been porting elisp functions, so you can test them by starting an Remacs instance and just calling the functions in ielm. This works, but of course it's not well-suited for other bits of code.

We would really benefit from a better solution, but we don't have one yet.

DavidDeSimone:
(Warning Shameless Plug): 

Once I finish up my work on, https://github.com/Wilfred/remacs/pull/286, I plan on finishing the next milestone for my mocking library, [mock_derive](https://github.com/DavidDeSimone/mock_derive). The next milestone will allow us to mock the results of extern "C" calls for tests. While not a perfect solution in my mind, it will allow us to write tests and piece meal bring over functions without having to figure out linking our C code to the Rust tests.

[...]

PR #304, opened 2017-09-13
Adding ability to mock extern "C" functions 

DavidDeSimone:
Adding ability to mock extern "C" functions  defined by the emacs C layer. This is done via a mocking library using proc macros. Adding a basic test as an example.

This is the solution that was discussed in https://github.com/Wilfred/remacs/issues/299 It is using a library that I maintain, which can be found here: https://github.com/DavidDeSimone/mock_derive


brotzeit:
I don't feel confident enough to review any PRs. Hopefully somebody with more experience has the time to take a look at it soon.

DavidDeSimone:
No problem @brotzeit . As a call out, this crate should only affect the app when running in test configuration. It should have zero affect on symbols/variables/etc when running cargo build or cargo build --release

PR #306, opened 2017-09-17
minor style & naming fixes

birkenfeld:


birkenfeld:
Thanks for the review!

rust_src/src/buffers.rs
```diff
@@ -321,12 +324,10 @@ fn set_buffer(buffer_or_name: LispObject) -> LispObject {
     if buffer.is_nil() {
         unsafe { nsberror(buffer_or_name.to_raw()) }
     };
-    if !buffer.as_buffer().unwrap().is_live() {
+    let buf = buffer.as_buffer_or_error();
```
Wilfred: This isn't quite equivalent. The previous version would panic if `buffer` had any other type (shouldn't ever happen), whereas this will signal an error in elisp.

Is this intentional? AFAICS this isn't actually changing user visible behaviour, but it's doing more work (slower) and the code reflects the intent less.

birkenfeld: If you look below, `as_buffer_or_error` was called anyway in the unsafe block. If you think `unwrap()` is semantically better here, it can be used instead, sure.

Wilfred: Aha, I totally missed that. Thanks :)

PR #309, opened 2017-09-18
Add cargo bin directory to path before checking rustfmt

shanavas786:
`which rustfmt` always fails as `$HOME/.cargo/bin` is not in `PATH`. As a result `rustfmt` is always reinstalled.


shanavas786:
The failure is due to check `[ "$(rustfmt --version)" = 0.9.0-* ]`

PR #310, opened 2017-09-18
Fix rustfmt version check

shanavas786:


.travis.yml
```diff
@@ -20,7 +20,7 @@ env:
 before_script:
   # Install rustfmt 0.9.0 if it isn't already installed on
   # Travis. This can be slow to install, so raise the Travis timeout.
-  - (which rustfmt && rustfmt --version && [ "$(rustfmt --version)" = 0.9.0-* ]) || travis_wait cargo install --force rustfmt --vers 0.9.0
+  - (which rustfmt && rustfmt --version && [[ "$(rustfmt --version)" =~ "0.9.0" ]]) || travis_wait cargo install --force rustfmt --vers 0.9.0
```
birkenfeld: I would have thought this is bash-only, so non-portable, but Travis' OSX build seems to be fine with it...

PR #311, opened 2017-09-18
Port some math functions

shanavas786:


rust_src/src/math.rs
```diff
@@ -354,3 +354,51 @@ fn geq(args: &mut [LispObject]) -> LispObject {
...
+        LispNumber::Float(num) => {
+            LispObject::from_float(num - 1.0)
+        },
+    }
+}
+
+/// Return the bitwise complement of NUMBER.  NUMBER must be an integer.
+#[lisp_fn]
+fn lognot(number: LispObject) -> LispObject {
+    LispObject::from_fixnum(!number.as_fixnum_or_error())
```
birkenfeld: I think this needs to be `from_fixnum_truncated`.

shanavas786: travis failure was due to format errors,
why do you think `from_fixnum_truncated` is needed here ?

birkenfeld: never mind, I was thinking about unsigned integers.

PR #316, opened 2017-09-19
port char-after

brotzeit:


src/editfns.c
```diff
@@ -1098,38 +1098,6 @@ At the beginning of the buffer or accessible region, return 0.  */)
...
-      pos_byte = PT_BYTE;
-      XSETFASTINT (pos, PT);
-    }
-
-  if (MARKERP (pos))
-    {
-      pos_byte = marker_byte_position (pos);
-      if (pos_byte < BEGV_BYTE || pos_byte >= ZV_BYTE)
-	return Qnil;
-    }
```
brotzeit: This looks to me like a bug. `BEGV_BYTE` and `ZV_BYTE` belong to the current buffer instead of the buffer the marker points to.

src/editfns.c
```diff
@@ -1098,38 +1098,6 @@ At the beginning of the buffer or accessible region, return 0.  */)
...
-    {
-      CHECK_NUMBER_COERCE_MARKER (pos);
-      if (XINT (pos) < BEGV || XINT (pos) >= ZV)
-	return Qnil;
-
-      pos_byte = CHAR_TO_BYTE (XINT (pos));
-    }
-
-  return make_number (FETCH_CHAR (pos_byte));
-}
```
brotzeit: `FETCH_CHAR` also uses the current buffer.

src/editfns.c
birkenfeld: Well, the docstring says that it uses the current buffer. So if the marker points to a different buffer the offset is still used, for the current buffer.

Not very clear, I agree, but appears nevertheless intended.

brotzeit: I can't imagine anybody is expecting this behaviour. Why use a marker if you don't use its location ? I will ask upstream what they think about it.

brotzeit:
> This is according to the documentation, which explicitly says the
> position is looked up in the current buffer.  So this sounds like an
> intentional and documented behavior.

I will make the according changes, though I don't see any reason why I would want the position of a marker, that actually points to another buffer, in the current buffer. No matter if the documentation mentions it explicitly, this is an unnecessary source of potential bugs. There won't be many cases where this function is called with a marker, especially not a marker that points to another buffer. But I think sometimes it's worth to at least consider to change something that makes no sense, even if it was intended. No offense against the emacs devs, they are doing a great job, but I hope such changes will be possible at some point.

src/editfns.c
brotzeit: How can I get this to work with multibyte chars ?

rust_src/src/editfns.rs
```diff
@@ -139,25 +139,24 @@ pub fn point_max() -> LispObject {
 /// If POS is out of range, the value is nil.
 #[lisp_fn(min = "0")]
 pub fn char_after(pos: LispObject) -> LispObject {
-    let cur_buffer = ThreadState::current_buffer();
+    let buffer_ref = ThreadState::current_buffer();
     if pos.is_nil() {
-        LispObject::from_natnum(cur_buffer.fetch_byte(cur_buffer.pt_byte) as EmacsInt)
+        LispObject::from_natnum(buffer_ref.fetch_byte(buffer_ref.pt_byte) as EmacsInt)
     } else if pos.is_marker() {
```
birkenfeld: `... else if let Some(marker) = pos.as_marker()` ?

rust_src/src/editfns.rs
```diff
@@ -139,25 +139,24 @@ pub fn point_max() -> LispObject {
...
-        let m = pos.as_marker().unwrap();
-        let m_buffer = m.buffer().unwrap();
-        let m_pos_byte = m.bytepos_or_error();
-        if m_pos_byte < m_buffer.begv_byte() || m_pos_byte >= m_buffer.zv_byte() {
+        let pos_byte = pos.as_marker().unwrap().bytepos_or_error();
+        // Only works correctly if marker points to current buffer
```
birkenfeld: Expand this a bit (for compatibility with C impl ...)

brotzeit:
I need help with this. I don't know how to handle multibyte chars. There is a function called `string_char` in multibyte.rs and I guess it has a similiar purpose like `STRING_CHAR` that is used in `FETCH_MULTIBYTE_CHAR`.

Wilfred:
@brotzeit did you ever send a message upstream? If not, I will do so :)

brotzeit:
Yes, here the answer again:

> This is according to the documentation, which explicitly says the
> position is looked up in the current buffer. So this sounds like an
> intentional and documented behavior.
> 

There are several functions that treat markers this way, so I think there maybe was a good reason for doing it.

Wilfred:
Test failures looks legitimate:

```
Test electric-pair-electric-quote-context-sensitive-after-space-double-at-point-3-in-text-mode condition:
    (ert-test-failed
     ((should
       (equal
	(buffer-substring-no-properties ... ...)
	expected-string))
      :form
      (equal " ‘’" " “")
      :value nil :explanation
      (arrays-of-different-length 3 2 " ‘’" " “" first-mismatch-at 1)))
   FAILED  129/454  electric-pair-electric-quote-context-sensitive-after-space-d
```

I suspect fixing the marker behaviour to use the current buffer rather than the marker's buffer will fix the tests.

brotzeit:
The function already takes the current buffer instead of the marker's buffer.

Take a look at the result of test-json-read-string.

```
Test test-json-read-string condition:
    (ert-test-failed
     ((should
       (equal
	(json-read-string)
	"abcαβγ"))
      :form
      (equal "abcÎÎÎ" "abcαβγ")
      :value nil :explanation
      (array-elt 3
		 (different-atoms
		  (206 "#xce" "?Î")
		  (945 "#x3b1" "?α")))))
```

I got the idea, that this is a multibyte related problem from this output. The c implementation makes use of `FETCH_CHAR` while I'm using fetch_byte as we don't have a fetch_char and I'm not sure how to implement it correctly.

Issue #318, opened 2017-09-20
Port following-char to Rust


brotzeit:
```c
DEFUN ("following-char", Ffollowing_char, Sfollowing_char, 0, 0, 0,
       doc: /* Return the character following point, as a number.
At the end of the buffer or accessible region, return 0.  */)
  (void)
{
  Lisp_Object temp;
  if (PT >= ZV)
    XSETFASTINT (temp, 0);
  else
    XSETFASTINT (temp, FETCH_CHAR (PT_BYTE));
  return temp;
}
```

peter-b:
This issue appears to have been fixed by #318 and should be closed!

Wilfred:
There was #320, but that was closed by the original author. This is still up for grabs :)

brotzeit:
He had problems implementing FETCH_CHAR, just like me #316 :)

enebo:
Ah so the issue was that we need multi-byte character read.  I got really confused when I looked at the already ported fetch_char as it appears to be the multi-byte character logic and not the single-bytes version with a single-byte return value.  Am I wrong about that?

brotzeit:
@enebo We have a `fetch_char` method now, if you want to complete your PR.

brotzeit:
#436

PR #320, opened 2017-09-22
Ports following-char #318.

enebo:
Trying my hand at my first Rust code and first time looking at Emacs C source :)  I was largely able to pattern match other similar commits and it appears to be working so let's do this!


enebo:
Ok this is broken (did not know about make check) because I did a fetch_byte but I need to fetch_char which does not exist yet...hmm not quite ready

PR #329, opened 2017-09-25
Port identity function to rust

kidd:


rust_src/src/lib.rs
```diff
@@ -64,6 +64,7 @@ mod editfns;
 mod util;
 mod minibuf;
 mod cmds;
+mod fns;
```
birkenfeld: I'd put it in `objects.rs` instead.

kidd: I wanted to keep the file structure the same as with original emacs sources, but I have no problem moving it to objects.rs . will do it right away. 

Thanks!

birkenfeld: We've sort of picked apart fns.c already since it is such a grab-bag. Thanks for the update!

kidd:
ready to be merged :)

rust_src/src/objects.rs
```diff
@@ -52,3 +52,9 @@ fn equal_including_properties(o1: LispObject, o2: LispObject) -> LispObject {
     };
     LispObject::from_bool(res)
 }
+
+/// Return the argument unchanged.
+#[lisp_fn]
+fn identity(x: LispObject) -> LispObject {
```
birkenfeld: argument should be called `arg` like in the C version.

Wilfred:
@kidd I've added you as a collaborator! This enables you to review other people's PRs too, if you wish :)

Issue #330, opened 2017-09-27
libremacs_lib.a: No such file or directory


royalaid:
I am trying to build remacs for Windows under msys2. I keep running into 

```
make -C nt all
make[1]: Entering directory '/c/emacs/remacs/nt'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/c/emacs/remacs/nt'
make -C lib all
make[1]: Entering directory '/c/emacs/remacs/lib'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/c/emacs/remacs/lib'
make -C lib-src all
make[1]: Entering directory '/c/emacs/remacs/lib-src'
RUSTFLAGS= \
cargo build  --manifest-path=../rust_src/remacs-lib/Cargo.toml
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
  CCLD     etags.exe
x86_64-w64-mingw32-gcc.exe: error: ../rust_src/remacs-lib/target/debug/libremacs_lib.a: No such file or directory
make[1]: *** [Makefile:400: etags.exe] Error 1
make[1]: Leaving directory '/c/emacs/remacs/lib-src'
make: *** [Makefile:413: lib-src] Error 2
```
The file does in fact not exist. Any ideas?

DavidDeSimone:
Thanks for the report @royalaid! I'll have time to look at this during weekend if no one gets to the bottom of it before then. We don't have many contributors that use windows or iterate on msys2, so we appreciate the report.

royalaid:
```
$ find rust_src/remacs-lib/target
rust_src/remacs-lib/target
rust_src/remacs-lib/target/debug
rust_src/remacs-lib/target/debug/.cargo-lock
rust_src/remacs-lib/target/debug/.fingerprint
rust_src/remacs-lib/target/debug/.fingerprint/errno-2c25c2596184936f
rust_src/remacs-lib/target/debug/.fingerprint/errno-2c25c2596184936f/dep-lib-errno-2c25c2596184936f
rust_src/remacs-lib/target/debug/.fingerprint/errno-2c25c2596184936f/lib-errno-2c25c2596184936f
rust_src/remacs-lib/target/debug/.fingerprint/errno-2c25c2596184936f/lib-errno-2c25c2596184936f.json
rust_src/remacs-lib/target/debug/.fingerprint/kernel32-sys-7ec75af9d036e79c
[...]

PR #336, opened 2017-10-02
Fixing remacs build on mingw.

DavidDeSimone:
Fixes https://github.com/Wilfred/remacs/issues/330. The problem here is that cargo will create static libs named "x_lib.lib" instead of the expected "libx_lib.a" on gnu/linux or osx.

NOTE: While temacs will be able to build with this patch, we will not finish building our final emacs executable. This is further discussed here: https://github.com/Wilfred/remacs/issues/337


brotzeit:
I can't reproduce the new error on my local machine with latest nightly.

PR #347, opened 2017-10-07
Port current-column to Rust

### LINKED ISSUE #327: Port current-column to Rust
```c
// Defined in ~/src/emacs/src/indent.c
DEFUN ("current-column", Fcurrent_column, Scurrent_column, 0, 0, 0,
       doc: /* Return the horizontal position of point.  Beginning of line is column 0.
This is calculated by adding together the widths of all the displayed
representations of the character between the start of the previous line
and point (e.g., control characters will have a width of 2 or 4, tabs
will have a variable width).
Ignores finite width of frame, which means that this function may return
values greater than (frame-width).
Whether the line is visible (if `selective-display' is t) has no effect;
however, ^M is treated as end of line when `selective-display' is t.
Text that has an invisible property is considered as having width 0, unless
`buffer-invisibility-spec' specifies that it is replaced by an ellipsis.  */)
  (void)
{
  Lisp_Object temp;
  XSETFASTINT (temp, current_column ());
  return temp;
}
```

This will require adding a declaration in remacs-sys for the `current_column` C function.

peter-b:
Adds a binding to the C API to `remacs-sys`, and adds a new `indent`
module.

Closes #327.

### DIFF
--- a/rust_src/src/indent.rs
+++ b/rust_src/src/indent.rs
@@ -1,0 +1,25 @@
+use remacs_macros::lisp_fn;
+use lisp::LispObject;
+use remacs_sys;
+use remacs_sys::MOST_POSITIVE_FIXNUM;
+use remacs_sys::EmacsInt;
+
+/// Return the horizontal position of point.  Beginning of line is
+/// column 0.  This is calculated by adding together the widths of all
+/// the displayed representations of the character between the start
+/// of the previous line and point (e.g., control characters will have
+/// a width of 2 or 4, tabs will have a variable width).
+/// Ignores finite width of frame, which means that this function may
+/// return values greater than (frame-width).
+/// Whether the line is visible (if `selective-display' is t) has no
+/// effect; however, ^M is treated as end of line when
+/// `selective-display' is t.
+/// Text that has an invisible property is considered as having width
+/// 0, unless `buffer-invisibility-spec' specifies that it is replaced
+/// by an ellipsis.
+#[lisp_fn]
+fn current_column() -> LispObject {
+    let col = unsafe { remacs_sys::current_column() };
+    assert!(col <= MOST_POSITIVE_FIXNUM as isize);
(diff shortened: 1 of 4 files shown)


rust_src/src/indent.rs
```diff
@@ -0,0 +1,25 @@
...
+/// `selective-display' is t.
+/// Text that has an invisible property is considered as having width
+/// 0, unless `buffer-invisibility-spec' specifies that it is replaced
+/// by an ellipsis.
+#[lisp_fn]
+fn current_column() -> LispObject {
+    let col = unsafe { remacs_sys::current_column() };
+    assert!(col <= MOST_POSITIVE_FIXNUM as isize);
+    LispObject::from_fixnum(col as EmacsInt)
+}
```
brotzeit: The c code makes use of make_natnum. In rust you can use from_natnum, which also takes care of the assertion. I think you can remove the namespace of current_column and do it in one line.

rust_src/src/indent.rs
```diff
@@ -0,0 +1,25 @@
+use remacs_macros::lisp_fn;
+use lisp::LispObject;
+use remacs_sys;
+use remacs_sys::MOST_POSITIVE_FIXNUM;
+use remacs_sys::EmacsInt;
+
```
brotzeit: It would be better if it would look like this `use remacs_sys::{EmacsInt, current_column};`

rust_src/src/indent.rs
```diff
@@ -0,0 +1,25 @@
+use remacs_macros::lisp_fn;
```
brotzeit: You forgot to add `//! Indentation functions.` in the first line.

Titan-C:
This in now in competition with PR #392

brotzeit:
The function was ported in #392, sorry. Feel free to take another issue.

PR #350, opened 2017-10-10
Ported minibuffer-selected-window

### LINKED ISSUE #326: Port minibuffer-selected-window to Rust
```c
DEFUN ("minibuffer-selected-window", Fminibuffer_selected_window, Sminibuffer_selected_window, 0, 0, 0,
       doc: /* Return the window which was selected when entering the minibuffer.
Returns nil, if selected window is not a minibuffer window.  */)
  (void)
{
  if (minibuf_level > 0
      && MINI_WINDOW_P (XWINDOW (selected_window))
      && WINDOW_LIVE_P (minibuf_selected_window))
    return minibuf_selected_window;

  return Qnil;
}
```

Yarikx:
Closes #326

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -144,7 +144,7 @@ pub fn minibuffer_selected_window() -> LispObject {
     let level = unsafe { minibuf_level };
     let current_minibuf = unsafe { LispObject::from_raw(current_minibuf_window) };
     if level > 0 && selected_window().as_window_or_error().is_minibuffer() &&
-        current_minibuf.as_window_or_error().is_live()
+        current_minibuf.as_window().unwrap().is_live()
     {
         current_minibuf
     } else {
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -2,7 +2,8 @@
 
 use lisp::{LispObject, ExternalPtr};
 use remacs_macros::lisp_fn;
-use remacs_sys::{Lisp_Window, selected_window as current_window};
+use remacs_sys::{Lisp_Window, selected_window as current_window,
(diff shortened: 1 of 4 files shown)


rust_src/src/windows.rs
```diff
@@ -135,3 +136,18 @@ pub fn window_minibuffer_p(window: LispObject) -> LispObject {
...
+
+/// Return the window which was selected when entering the minibuffer.
+/// Returns nil, if selected window is not a minibuffer window.
+#[lisp_fn]
+pub fn minibuffer_selected_window() -> LispObject {
+    let level = unsafe { minibuf_level };
+    let current_minibuf = unsafe { LispObject::from_raw(current_minibuf_window) };
+    if level > 0 && selected_window().as_window_or_error().is_minibuffer() &&
+        current_minibuf.as_window_or_error().is_live()
+    {
```
brotzeit: I think the c implementation isn't throwing errors here.

brotzeit: I think the c implementation isn't throwing errors here.

brotzeit: I think the c implementation does not throw errors here.

brotzeit: I think you shouldn't raise errors here.

birkenfeld: The alternative would be `unwrap()`... not sure that is better.

brotzeit: I don't think it's better, but it looks like the c code does not throw errors here.

Yarikx: Makes sense. As only windows object can be in `current_minibuf` static var, an error should not occur there. I think we can safely unwrap().

PR #356, opened 2017-10-11
WIP Port set-buffer-redisplay

rski:
the caveat with this change is that it uses `current_buffer().as_ptr() as *const _ as  *const c_void` to pass the buffer ptr, which I picked up from elsewhere in remacs. My rust-fu isn't up to par yet to figure out in a reasonable amount of time how to properly do this.

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -35,16 +35,16 @@ pub const CHAR_HYPER: char_bits = 0x1000000;
 pub const CHAR_SHIFT: char_bits = 0x2000000;
 pub const CHAR_CTL: char_bits = 0x4000000;
 pub const CHAR_META: char_bits = 0x8000000;
-pub const CHAR_MODIFIER_MASK: char_bits =
-    CHAR_ALT | CHAR_SUPER | CHAR_HYPER | CHAR_SHIFT | CHAR_CTL | CHAR_META;
+pub const CHAR_MODIFIER_MASK: char_bits = CHAR_ALT | CHAR_SUPER | CHAR_HYPER | CHAR_SHIFT |
+    CHAR_CTL | CHAR_META;
 pub const CHARACTERBITS: char_bits = 22;
 
 pub const PSEUDOVECTOR_FLAG: ptrdiff_t = std::isize::MAX - std::isize::MAX / 2;
 pub const PSEUDOVECTOR_SIZE_BITS: ptrdiff_t = 12;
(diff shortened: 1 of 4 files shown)


rust_src/remacs-sys/lib.rs
```diff
@@ -522,6 +522,8 @@ pub struct Lisp_Buffer {
     pub overlay_center: ptrdiff_t,
 
     pub undo_list: Lisp_Object,
+
+    pub prevent_redisplay_optimizations_p: bool,
```
birkenfeld: This is not the right location for this member - it is part of the `flags` bitfield declared above.

Since Rust doesn't support bitfields in structs, you'll have to extract it from there.

rust_src/remacs-sys/lib.rs
```diff
@@ -1040,6 +1042,8 @@ extern "C" {
     ) -> ptrdiff_t;
 
     pub fn hash_remove_from_table(h: *mut Lisp_Hash_Table, key: Lisp_Object);
+
+    pub fn bset_update_mode_line(buffer: *const libc::c_void) -> Lisp_Object; // TODO: buffer*
```
birkenfeld: You can use `*mut Lisp_Buffer`.

rust_src/src/xdisp.rs
```diff
@@ -0,0 +1,22 @@
...
+/// This function may be passed to `add-variable-watcher`.
+#[lisp_fn]
+fn set_buffer_redisplay(
+    _symbol: LispObject,
+    _newval: LispObject,
+    _op: LispObject,
+    _where: LispObject,
+) -> LispObject {
+    unsafe {
+        bset_update_mode_line(ThreadState::current_buffer().as_ptr() as *const _ as
```
birkenfeld: With the change above, this can use `as_mut()` and not need a cast.

rust_src/remacs-sys/lib.rs
rski: ok i'll look into fixing this

rust_src/remacs-sys/lib.rs
rski: would it make sense to change this for set_buffer_internal as well?

rust_src/src/xdisp.rs
rski: consider it done. I had used as_mut(), but I didn't get the signature right so of course it wouldn't compile. Thanks.

rust_src/remacs-sys/lib.rs
birkenfeld: Yes, I guesss that function predates our definition of the struct.

rski:
I haven't forgotten about this, but I don't think I have either the skill or time to see it through. Maybe someone better will make it.

PR #357, opened 2017-10-12
Port indirect-variable functions

shanavas786:


rust_src/remacs-sys/lib.rs
```diff
@@ -208,6 +208,14 @@ pub enum Symbol_Interned {
     InternedInInitialObarray = 2,
 }
 
+#[repr(C)]
+pub enum Symbol_Redirect {
+    Plainval = 4,
+    Varalias = 1,
```
Wilfred: Rust enums are generally camelcase, so I'd be tempted to go with `PlainVal` and `VarAlias`.

shanavas786: Good catch, updated.

Wilfred:
Perfect :)

PR #360, opened 2017-10-13
make the REmacs icon distinct from the default Emacs icon

chenl:
Please excuse the crudity of this model. I didn't have time to build
it to scale or paint it...

Err, actually painting it, is the only thing I did.  Also it took
Emacs a long time to move over from using ugly design, and I am afraid
this might be considered a regression.

If there is someone knowing what he/she is doing, please feel free to
make this nicer.

Also I will not be offended if this PR would be rejected.


Wilfred:
Looks great! Having some sort of distinctive icon is definitely a good thing. We might want to explore a different design (perhaps a stylised R?) but this a really good start.

remacs

ZelphirKaltstahl:
How about an "R" as beautifully written as the "E" of Emacs?

chenl:
As I see it, there are several design elements that can be combined:

From the Emacs side we have the stylised E [that was intended to resemble the Gnu's horns](https://www.ee.ryerson.ca/~elf/emacs/logo/), we also have the purple sphere, that although rather new, is wildly excepted, and became part of the design element of the Emacs branding.

From the Rust side there is the industrial R, the gear, a rusty color pallet (that is unofficial but wildly used), and [Ferris](http://www.rustacean.net/).

Any of those can be combined, including as suggested just the styling of the E with the R letter (or the other way around).

The main point, IMHO, is that for a long time the Emacs visuals where so ugly as to detract from its adoption, that stemmed for the fact the people involved where infatuated with the idea of combining the word Emacs with a image of a Gnu, with no regard to actual aesthetics and with no professional to help them with this regard. Please don't go that way with REmacs.

chenl:
And one more observation: I was using the modified icon for about a day, and:

1. It isn't as ugly as I was fearing. It looks good both in a large and small scale. I also became to like the concept of a rusty Emacs.
2. It is *really* nice to have a distinct icon for each version of REmacs and to easily know if I am in a normal or experimental editor.

shanavas786:
Finally managed to create some :)

1. remacs

2. remacs-4

3. remacs-3

source: [remacs.zip](https://github.com/Wilfred/remacs/files/1385263/remacs.zip)

ZelphirKaltstahl:
Ha, using the cog wheel - that looks good too. It can also be interpreted as "something that works with Emacs on the inside".

chenl:
@shanavas786 suggestions look grate, just make sure that they also work well at small size.

shanavas786:
Svg files are attached. Scaling should not be an issue.

On October 15, 2017 12:12:40 PM GMT+03:00, Chen Rotem Levy <notifications@github.com> wrote:
>@shanavas786 suggestions look grate, just make sure that they also work
>well at small size.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/Wilfred/remacs/pull/360#issuecomment-336697000

-- 
ഷാനവാസ്
Sent from my Android device with K-9 Mail.

Wilfred:
Those look brilliant! 👑 

I think the more limited colour palette in 2 and 3 will probably work better at small sizes.

chenl:
@shanavas786 perhaps I wasn't expressing myself correctly. When I say  *also work well at small size*, I meant visually (i.e. when shrunk it should still look good).

shanavas786:
I see. I will try, though my design skills are not good enough.
Feel free to modify those. Any help is appreciated.

chenl:
Although I thing the E + Cog design is very good, and can be a grate logo, at a small scale the gear teeth tend to get washed out.

Here we have:

[Emacs original] [@chenl's suggestion]
[@shanavas786's no.2 (simplified)] [@shanavas786's no.3]

## just the icons
just the icons
## within a screenshot
screenshot

WolfeCub:
In my opinion the orange colored Emacs logo is the best. The cogs end up being too small to look good in smaller icons whereas the orange really differentiates itself from the regular Emacs logo.

nasciiboy:
3 > orange > 2 > 1

ZelphirKaltstahl:
An idea is to give the cog wheel less teeth and bigger ones. Then it might be more recognizable in small sizes.

shaleh:
Should we merge the current version and then work on improving it?

Wilfred:
I agree with @shaleh: let's merge this PR, and we can follow up with improvements.

PR #364, opened 2017-10-16
Porting puthash/maphash/hash_table_count

DavidDeSimone:
One common pattern in C to loop over key/value pairs in a Lisp_Hash_Table is:

``` c
for (i = 0; i < HASH_TABLE_SIZE (h); ++i) 
  {
    if (NILP (HASH_HASH (h, i))
      continue;

    key = HASH_KEY (h, i);
    value = HASH_VALUE (h, i);
    // ...
  }
```

I believe it would be more idomatic Rust to use an iterator, so I wrote one. There is a base iterator, that iterates over the indices of `key_and_value`. This iterator is used to implement a key and value iterator. Usage looks like

``` rust
for (key, value) in hash_table.iter() {
// ...
}

```

### DIFF
--- a/rust_src/src/hashtable.rs
+++ b/rust_src/src/hashtable.rs
@@ -89,6 +89,74 @@ impl LispHashTableRef {
     pub fn remove(mut self, key: LispObject) {
         unsafe { hash_remove_from_table(self.as_mut(), key.to_raw()) };
     }
+
+    pub fn get_hash_hash(self, idx: isize) -> LispObject {
+        let index = LispObject::from_natnum(idx as EmacsInt);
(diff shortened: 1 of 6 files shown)


rust_src/src/hashtable.rs
```diff
@@ -62,23 +62,101 @@ impl LispHashTableRef {
...
+                return Some((key, value));
+            }
+        }
+
+        None
+    }
+}
+
+impl LispHashTableRef {
+    pub fn indicies(&self) -> HashTableIter {
```
Wilfred: *indices

PR #371, opened 2017-10-19
Replace LispObject::from_raw with std::convert::From implementation

shanavas786:


brotzeit:
I would prefer to only change `from_raw` to `from`. IMO the verbosity of the other method names is helpful.

shaleh:
Agreed. Rust is about explicit types. Why hide them?

birkenfeld:
I have to agree with @brotzeit - to me this seems to stretch the purpose of `From` too far.

birkenfeld:
Next question: should we now replace `LispObject::from(x)` with `x.into()` which is less wordy?

shaleh:
That it is a little more interesting:

    #[inline]
    pub fn point_marker(self) -> LispObject {
        self.pointm.into()
    }

Versus current style

    #[inline]
    pub fn point_marker(self) -> LispObject {
        LispObject::from(self.pointm)
    }

shaleh:
And:

    #[lisp_fn]
    pub fn selected_window() -> LispObject {
        unsafe { current_window }.into()
    }

shaleh:
Tweaking this:

    /// Internal function to get a displayable string out of a Lisp string.
    fn display_string<LO: Into<LispObject>>(obj: LO) -> String {
        let mut s = obj.into().as_string().unwrap();
        let slice = unsafe { slice::from_raw_parts(s.data_ptr(), s.len_bytes() as usize) };
        String::from_utf8_lossy(slice).into_owned()
    }

Also makes this possible:

display_string(unsafe { SYMBOL_NAME(self.to_raw()) })

brotzeit:
@Wilfred what do you think ?

shaleh:
Style guideline time?

Use `into` for the examples I showed where the value is being returned and it is clear that the Lisp_Object is being upcast into a Rust LispObject. Otherwise, use ::from() within functions?

PR #372, opened 2017-10-19
port FETCH_CHAR and char-after

brotzeit:


rust_src/src/buffers.rs
```diff
@@ -135,6 +136,31 @@ impl LispBufferRef {
 
         unsafe { *(self.beg_addr().offset(offset + n - self.beg_byte())) as u8 }
     }
+
+    #[inline]
+    pub fn fetch_multibyte_char(&self, n: ptrdiff_t) -> c_int {
```
Wilfred: It's a shame we don't have a type for characters, but that would require changes to multibyte.rs.

rust_src/src/editfns.rs
```diff
@@ -152,3 +153,31 @@ pub fn goto_char(position: LispObject) -> LispObject {
...
+/// If POS is out of range, the value is nil.
+#[lisp_fn(min = "0")]
+pub fn char_after(mut pos: LispObject) -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
+    if pos.is_nil() {
+        pos = point();
+    }
+    if pos.is_marker() {
+        let pos_byte = pos.as_marker().unwrap().bytepos_or_error();
+        // Only works correctly if marker points to current buffer
```
Wilfred: I think this comment is misleading: the code works correctly, it's just an unfortunate/surprising behaviour that's implemented.

Could you rephrase this to something like: "Note that this considers the position in the current buffer, even if the marker is from another buffer"?

brotzeit: Yeah, I was also thinking about changing this.

Wilfred:
:sparkles:

PR #373, opened 2017-10-19
mention elisp tests in readme

brotzeit:


README.md
```diff
@@ -330,6 +330,11 @@ Then you can run this in the checkout root to reformat all Rust code:
 $ make rustfmt
 ```
 
+### Running elisp tests
+
+You can run the elisp test suite from the directory "test".
+Use `make` to run all tests. It's also possible to run single tests, e.g. `make src/data-tests`.
```
birkenfeld: Why not mention toplevel `make check` and `make check-maybe`? (Together with touching `.el` files to rerun with `check-maybe`?)

brotzeit: I didn't know about these commands. Thanks.

brotzeit:
While we're at it, does anybody know how to run elisp tests in a compilation buffer ?

README.md
shaleh: Can we also get a pointer to writing new tests? I tried to add a new file this morning but the make system claimed to not know about it.

Testing from Rust is a bunch of boiler plate to make LispObjects so I figured I'd write elisp tests.

brotzeit: I don't think this is suitable for the walkthrough. Elisp tests need to be merged upstream and too much text can be confusing. But maybe we can put it in the README in the test directory ?

brotzeit:
I just noticed, that we already have a few rust tests. The tests don't run when I use make in "test", so I remove this part ok ?

brotzeit:
Is it possible that you overlooked my latest comment and commit ? `make check` and `make check maybe` are the commands we want, right ?

shaleh:
Your commands are accurate. However, unless run _IN_ the tests directory they do not work as documented. This is missing from the README text you added.

brotzeit:
I also didn't know this works until @birkenfeld mentioned it.

> Why not mention toplevel make check and make check-maybe?

Problem with running the tests in the directory test is that our rust tests are irgnored.

shaleh:
Hmm. I must not have been in the top level when I ran `make check-maybe` because it failed for me with unknown make command and that is why I added that comment. Never mind. Both do work from the toplevel as you say.

README.md
```diff
@@ -330,6 +330,17 @@ Then you can run this in the checkout root to reformat all Rust code:
...
 ```
 
+### Running tests
+
+* make check
+  Run all tests as defined in the directory.  Expensive tests are
+  suppressed.  The result of the tests for <filename>.el is stored in
+  <filename>.log.
+
+* make check-maybe
```
Wilfred: Could you put the commands in backticks (e.g. `` `make check` ``), so they're clearly distinguishable form the text?

README.md
```diff
@@ -330,6 +330,17 @@ Then you can run this in the checkout root to reformat all Rust code:
 $ make rustfmt
 ```
 
+### Running tests
+
+* make check
+  Run all tests as defined in the directory.  Expensive tests are
```
Wilfred: Which directory is the directory? I'd be tempted to say "Run elisp and Rust tests. If run in a subdirectory, only run the tests in that directory." (assuming I've understood correctly)

README.md
```diff
@@ -330,6 +330,17 @@ Then you can run this in the checkout root to reformat all Rust code:
...
+  suppressed.  The result of the tests for <filename>.el is stored in
+  <filename>.log.
+
+* make check-maybe
+  Like "make check", but run only the tests for files which have
+  unresolved prerequisites.
```
Wilfred: Perhaps worth saying what we mean by unresolved prerequisites. I presume this means files that have been modified since the last build (i.e. usual Makefile logic)?

README.md
brotzeit: You're right, I already thought about adding such a note.

README.md
brotzeit: I actually don't know what it means, I just took it from the README in the test directory ;)

shaleh: @Wilfred you are reading that correctly. Make check looks for <file>.log which depends on <file>.elc which in turn depends on <file>.el.

PR #374, opened 2017-10-20
port sleep-for

brotzeit:


rust_src/src/dispnew.rs
```diff
@@ -0,0 +1,42 @@
...
+use lisp::LispObject;
+use std::{ptr, cmp};
+use floatfns::extract_float;
+
+
+/// Pause, without updating display, for SECONDS seconds.
+/// -SECONDS may be a floating-point value, meaning that you can wait for a
+/// -fraction of a second.  Optional second arg MILLISECONDS specifies an
+/// -additional wait period, in milliseconds; this is for backwards compatibility.
+/// -\(Not all operating systems support waiting for a fraction of a second.)
```
Wilfred: These hyphens look like they've been copy-pasted by accident?

Also, you don't need to escape parens in Rust docstrings (because the paren is not the beginning of the line so Emacs doesn't get confused).

PR #376, opened 2017-10-21
Port position-bytes to Rust

### LINKED ISSUE #344: Port position-bytes to Rust
```c
DEFUN ("position-bytes", Fposition_bytes, Sposition_bytes, 1, 1, 0,
       doc: /* Return the byte position for character position POSITION.
If POSITION is out of range, the value is nil.  */)
  (Lisp_Object position)
{
  CHECK_NUMBER_COERCE_MARKER (position);
  if (XINT (position) < BEG || XINT (position) > Z)
    return Qnil;
  return make_number (CHAR_TO_BYTE (XINT (position)));
}
```

shaleh:
Closes #344

### DIFF
--- a/rust_src/src/editfns.rs
+++ b/rust_src/src/editfns.rs
@@ -239,7 +239,7 @@ pub fn char_after(mut pos: LispObject) -> LispObject {
 /// First argument is the string to copy.
 /// Remaining arguments form a sequence of PROPERTY VALUE pairs for text
 /// properties to add to the result.
-/// usage: (propertize STRING &rest PROPERTIES)  */
+/// usage: (propertize STRING &rest PROPERTIES)
 #[lisp_fn(min = "1")]
 pub fn propertize(args: &mut [LispObject]) -> LispObject {
     /* Number of args must be odd. */
--- a/rust_src/src/editfns.rs
+++ b/rust_src/src/editfns.rs
@@ -161,12 +161,10 @@ pub fn position_bytes(position: LispObject) -> LispObject {
     let pos = position.as_fixnum_coerce_marker_or_error() as ptrdiff_t;
     let cur_buf = ThreadState::current_buffer();
 
-    if pos < cur_buf.begv || pos > cur_buf.zv {
-        return LispObject::constant_nil();
-    }
-
-    unsafe {
-        let bytepos = buf_charpos_to_bytepos(cur_buf.as_ptr(), pos);
+    if pos >= cur_buf.begv && pos <= cur_buf.zv {
+        let bytepos = unsafe { buf_charpos_to_bytepos(cur_buf.as_ptr(), pos) };
         LispObject::from_natnum(bytepos as EmacsInt)
+    } else {
+        LispObject::constant_nil()
     }
 }
--- a/rust_src/src/editfns.rs
+++ b/rust_src/src/editfns.rs
@@ -1,7 +1,7 @@
 //! Lisp functions pertaining to editing.
 
 use remacs_macros::lisp_fn;
-use lisp::LispObject;
+use lisp::{LispObject, LispNumber};
 use util::clip_to_bounds;
 use remacs_sys::{buf_charpos_to_bytepos, globals, set_point_both, EmacsInt, Qinteger_or_marker_p,
                  Qmark_inactive};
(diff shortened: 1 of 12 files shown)


shaleh:
I used `return` explicitly to avoid small one line else clauses. If that style is not OK, let me know.

rust_src/src/editfns.rs
```diff
@@ -152,3 +152,27 @@ pub fn goto_char(position: LispObject) -> LispObject {
     };
     position
 }
+
+/// Return the byte position for character position POSITION.
+/// If POSITION is out of range, the value is nil.
+#[lisp_fn]
+pub fn position_bytes(position: LispObject) -> LispObject {
+    if let LispNumber::Fixnum(pos) = position.as_number_coerce_marker_or_error() {
+        let cur_buf = ThreadState::current_buffer();
```
brotzeit: It seems like we use `as_fixnum_coerce_marker_or_error` for `CHECK_NUMBER_COERCE_MARKER` and `as_number_coerce_marker_or_error` for `CHECK_NUMBER_OR_FLOAT_COERCE_MARKER`. I think this is a little irritating, what do you think @Wilfred.

rust_src/src/editfns.rs
```diff
@@ -152,3 +152,27 @@ pub fn goto_char(position: LispObject) -> LispObject {
...
+        }
+
+        unsafe {
+            let bytepos = buf_charpos_to_bytepos(cur_buf.as_ptr(), pos as isize);
+            return LispObject::from_natnum(bytepos as EmacsInt);
+        }
+    };
+
+    wrong_type!(Qinteger_or_marker_p, position)
+}
```
brotzeit: `as_number_coerce_marker_or_error` takes care of the error for you.

rust_src/src/editfns.rs
```diff
@@ -152,3 +152,27 @@ pub fn goto_char(position: LispObject) -> LispObject {
...
+        let cur_buf = ThreadState::current_buffer();
+
+        // cast up to 64bit before comparing with pos
+        let begv = cur_buf.begv as EmacsInt;
+        let zv = cur_buf.zv as EmacsInt;
+
+        if pos < begv || pos > zv {
+            return LispObject::constant_nil();
+        }
+
```
brotzeit: I think you don't need the cast here. This can be done in one if else block with implicit returning. Take a look how the *or_error methods are used in other functions. That's what I do when I'm not sure how to use a function :)

shaleh:
Thank you for the comments. I will give this another crack.

I needed the casts because of a disagreement between taking `isize` and `i64`.

brotzeit:
Yeah sorry that wasn't clear enough. I would do it like this `let pos = position.as_fixnum_coerce_marker_or_error() as ptrdiff_t;`. This way you only have to make one cast. I don't know the difference between `ptrdiff_t` and `isize`, but the first one is the type of `begv` and `zv` ;).

You can also just use `if pos < cur_buf.begv || pos > cur_buf.zv`. I don't think we need variables for this.

shaleh:
Thanks. I will update the PR this evening.

rust_src/src/editfns.rs
shaleh: I totally missed `as_fixnum`. And yes, having it return a 'real' number and `as_number` returning a `LispNumber` is confusing.

shaleh:
Could 'Porting Elisp Primitive Functions: Walkthrough' add the suggestion to leave off `lisp_fn` until the code compiles? I was super confused by the type complaints until I found the old ticket about that and commented it out.

In my case it claimed at the same line of code I had a failure 'expected i64, have isize' AND 'expect isize, have i64'. AARRGGG :-)

brotzeit:
I know, that's really annoying. Hopefully they will change the macro implementation soon.

Two more things. Can you use an if else instead of the return and you should also remove the from_natnum call from the unsafe block.

brotzeit:
LGTM

Wilfred:
Looks good to me too :)

shaleh:
@Wilfred should I resolved the conflicts or is that handled on the committer's end?

brotzeit:
My bad. Maybe we should make a rollup with all those PRs waiting to be merged.

Wilfred:
@shaleh if you handle conflicts, that always helps us merge PRs. If you allow collaborators to push commits, as I see you've done here, a collaborator might help you fix the conflicts too.

shaleh:
Grumble grumble.

I have been bit by the merge failure fairy......

shaleh:
Just gonna start over...

PR #382, opened 2017-10-22
ported beginning-of-line

galczo5:


brotzeit:
I just realized, that this is an interactive function. I can't find any ported function with the lisp_fn option intspec and I can't get it to work. I guess you have to wait until this is fixed, sorry.

galczo5:
Ok, no problem :) I'll take another issue. Please let me know when fix will be merged in master.

shaleh:
This tweak gets things working.

    diff --git a/rust_src/src/cmds.rs b/rust_src/src/cmds.rs
    index 22c768c..91b6905 100644
    --- a/rust_src/src/cmds.rs
    +++ b/rust_src/src/cmds.rs
    @@ -10,7 +10,7 @@ pub fn forward_point(n: LispObject) -> LispObject {
         LispObject::from_fixnum(n.as_fixnum_or_error() + pt as EmacsInt)
     }
 
    -#[lisp_fn(min = "0")]
    +#[lisp_fn(min = "0", intspec = "^p")]
      pub fn beginning_of_line(n: LispObject) -> LispObject {
 
          let x = if n.is_nil() {

rust_src/src/cmds.rs
```diff
@@ -9,3 +9,19 @@ pub fn forward_point(n: LispObject) -> LispObject {
     let pt = ThreadState::current_buffer().pt();
     LispObject::from_fixnum(n.as_fixnum_or_error() + pt as EmacsInt)
 }
+
+#[lisp_fn(min = "0")]
+pub fn beginning_of_line(n: LispObject) -> LispObject {
+
```
brotzeit: I think it would look better if you make n mutable and avoid an additional variable and the else block.

shaleh: The C code is modifying a local copy. That is what this code is explicitly doing. Adding the `mut` keyword makes reading the function signature more complex because now I have to look, does it take a reference? Why are they just mutating a copy? Is LispObject now a type wrapper around a reference? All of that is not work saving the creation of a local variable in my opinion. Or is code golf the intention here?

brotzeit: I have noticed some other spots in our codebase, where it was done this way. It's not important to me how we handle it, but maybe we should discuss a guide line for this in order to create some consistency.

shaleh: Sounds good.

galczo5:
Still failing, even after rustfmt, any idea?

shaleh:
-use remacs_sys::{EmacsInt, Fline_beginning_position, set_point};⏎
    +use remacs_sys::{set_point, EmacsInt, Fline_beginning_position};⏎

It is not passing the rustfmt test. This is probably because you are running rustfmt 0.9.0 and not rustfmt 0.2.9 aka 'rustfmt-nightly'. cargo install rustfmt-nightly will help. Not it conflicts with rustfmt. What I did was rename all of the rustfmt related binaries in ~/.cargo/bin/ to *-stable and then installed rustfmt-nightly.

galczo5:
Thanks.

rust_src/src/cmds.rs
```diff
@@ -9,3 +9,21 @@ pub fn forward_point(n: LispObject) -> LispObject {
     let pt = ThreadState::current_buffer().pt();
     LispObject::from_fixnum(n.as_fixnum_or_error() + pt as EmacsInt)
 }
+
+//Move point to beginning of current line (in the logical order).
+//With argument N not nil or 1, move forward N - 1 lines first.
+//If point reaches the beginning or end of buffer, it stops there.
```
Wilfred: Please use a triple-slash docstring, and use the full docstring that was in the C code.

galczo5:
Will end it today. I had busy weekend.

galczo5:
Done

src/cmds.c
```diff
@@ -136,30 +136,6 @@ go to its beginning.  */)
   return make_number (count <= 0 ? - shortage : shortage);
 }
 
-DEFUN ("beginning-of-line", Fbeginning_of_line, Sbeginning_of_line, 0, 1, "^p",
-       doc: /* Move point to beginning of current line (in the logical order).
-With argument N not nil or 1, move forward N - 1 lines first.
-If point reaches the beginning or end of buffer, it stops there.
-
-This function constrains point to the current field unless this moves
```
Wilfred: Just need to add this second paragraph to the docstring, and we're good to go :)

galczo5: Done. Are we ready to merge?

brotzeit: We have to wait for #456.

PR #385, opened 2017-10-23
Updating the 'lisp_fn' macro to support interactive functions.

DavidDeSimone:
An example would be 
``` rust
#[lisp_fn(intspec="r")]
fn blah....
```

To define the function as interactive, acting on the current region.

### DIFF
--- a/rust_src/remacs-macros/lib.rs
+++ b/rust_src/remacs-macros/lib.rs
@@ -24,6 +24,12 @@ pub fn lisp_fn(attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
     let mut rargs = quote::Tokens::new();
     let mut body = quote::Tokens::new();
     let max_args = function.args.len() as i16;
+    let intspec = if let Some(intspec) = lisp_fn_args.intspec {
+        quote!{ (#intspec).as_ptr() as *const ::libc::c_char }
+    } else {
+        quote!{ ::std::ptr::null() }
+    };
+    
     match function.fntype {
         function::LispFnType::Normal(_) => {
             for ident in function.args {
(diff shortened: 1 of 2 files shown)


rust_src/remacs-macros/lisp_attr.rs
```diff
@@ -39,13 +44,15 @@ fn parse_kv(kv_list: Vec<(syn::Ident, syn::StrLit)>, function: &Function) -> Lis
...
             "min" => min = i16::from_str(&string.value).unwrap(),
+            "intspec" => intspec = Some(string.value),
             _ => (), // TODO: throw a warning?
         }
     }
     LispFnArgs {
         name: name.unwrap_or_else(|| function.name.to_string().replace("_", "-")),
         c_name: c_name.unwrap_or_else(|| function.name.to_string()),
         min: min,
+        intspec: intspec,
```
birkenfeld: You can just say `intspec,` here (new feature from a few versions back)

rust_src/remacs-macros/lib.rs
```diff
@@ -24,6 +24,12 @@ pub fn lisp_fn(attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
     let mut rargs = quote::Tokens::new();
     let mut body = quote::Tokens::new();
     let max_args = function.args.len() as i16;
+    let intspec = if let Some(intspec) = lisp_fn_args.intspec {
+        quote!{ (#intspec).as_ptr() as *const ::libc::c_char }
```
birkenfeld: Don't we need a null byte here?

DavidDeSimone: Good catch. I incorrectly thought that was being handled by the lisp_attr parsing, but it is not.

rust_src/remacs-macros/lisp_attr.rs
DavidDeSimone: Updated.

PR #390, opened 2017-10-23
port provide

brotzeit:


rust_src/src/fns.rs
```diff
@@ -32,3 +32,50 @@ fn featurep(feature: LispObject, subfeature: LispObject) -> LispObject {
...
+    subfeature.list_or_error();
+    unsafe {
+        if LispObject::from_raw(Vautoload_queue).is_not_nil() {
+            Vautoload_queue = Fcons(
+                Fcons(LispObject::from_fixnum(0).to_raw(), globals.f_Vfeatures),
+                Vautoload_queue,
+            )
+        }
+    }
+    let mut tem = memq(
```
Wilfred: I'd be tempted to just inline this: `memq(...).is_nil()`. I know the C used `tem`, but I'm not sure it makes sense for us (it's not that helpful in the C either really).

rust_src/src/lisp.rs
```diff
@@ -745,6 +745,16 @@ impl LispObject {
         }
     }
 
+    // Same as the `CHECK_LIST` function.
+    #[inline]
+    pub fn list_or_error(self) -> LispObject {
```
Wilfred: This function is only used in one place, and we're not using the return value :)

Generally we have `.as_foo_or_error` when we are unwrapping a value. How about defining an `.is_list` and explicitly calling `wrong_type!` in `provide`?

brotzeit: I count 19, but I also thought about doing it that way. I also didn't implement `LOADHIST_ATTACH`, but that function is only used 8 times. I don't mind changing this.

rust_src/src/fns.rs
brotzeit: I think it makes sense as it's also used here

```
  tem = Fassq (feature, Vafter_load_alist);
  if (CONSP (tem))
    Fmapc (Qfuncall, XCDR (tem));
```

brotzeit:
Should be ok now.

rust_src/src/fns.rs
```diff
@@ -30,4 +30,44 @@ fn featurep(feature: LispObject, subfeature: LispObject) -> LispObject {
...
+    feature.as_symbol_or_error();
+    if !subfeature.is_list() {
+        wrong_type!(Qlistp, subfeature)
+    }
+    unsafe {
+        if LispObject::from(Vautoload_queue).is_not_nil() {
+            Vautoload_queue = Fcons(
+                Fcons(LispObject::from_fixnum(0).to_raw(), globals.f_Vfeatures),
+                Vautoload_queue,
+            )
```
shaleh: Semicolon?

rust_src/src/fns.rs
```diff
@@ -30,4 +30,44 @@ fn featurep(feature: LispObject, subfeature: LispObject) -> LispObject {
...
+    if memq(feature, LispObject::from(unsafe { globals.f_Vfeatures })).is_nil() {
+        unsafe {
+            globals.f_Vfeatures = Fcons(feature.to_raw(), globals.f_Vfeatures);
+        }
+    }
+    if subfeature.is_not_nil() {
+        put(feature, LispObject::from(Qsubfeatures), subfeature);
+    }
+    unsafe {
+        globals.f_Vcurrent_load_list = Fcons(
```
shaleh: This loses the name `LOADHIST_ATTACH`. Should we add a Rust function to keep the name alive?

brotzeit: I've thought a while about that one. It's only used in a few places and I think it's easy to read with the formatting.

brotzeit:
Anything else ?

PR #393, opened 2017-10-23
Close #375 Port process-buffer to rust

### LINKED ISSUE #375: Port process-buffer to Rust
```c
DEFUN ("process-buffer", Fprocess_buffer, Sprocess_buffer,
       1, 1, 0,
       doc: /* Return the buffer PROCESS is associated with.
The default process filter inserts output from PROCESS into this buffer.  */)
{
  CHECK_PROCESS (process);
  return XPROCESS (process)->buffer;
}
```
Take a look at `process_name` in process.rs.

Titan-C:
I'm renewing my interest in practicing rust, I'm an early beginner in it.
I followed the porting of process-name, this compiled and seems to work.
Is there a way to write tests for this? I very naively used `(run-python)` to have a process launched and then `(process-buffer (get-buffer-process "*Python*"))` to use this function. 

A side question. Why are Pull Requests at this beginner level not squashed when merged? It makes it quite complicated when reviewing the history see various commits that where the trials of the same patch. Having some cleanup of the commit history seems to help in the long run. Squashing on merge is very simple on github.

### DIFF
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -259,6 +259,7 @@ pub extern "C" fn rust_init_syms() {
         defsubr(&*process::Sget_process);
         defsubr(&*process::Sprocessp);
         defsubr(&*process::Sprocess_name);
+        defsubr(&*process::Sprocess_buffer);
         defsubr(&*process::Sget_buffer_process);
         defsubr(&*lists::Satom);
         defsubr(&*lists::Slistp);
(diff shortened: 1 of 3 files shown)


brotzeit:
Thanks for your PR! 

> Is there a way to write tests for this?

Yes, but others here mentioned, that it's complicated due to the FFI stuff, but I don't really know. Maybe somebody else can answer this question.

> Why are Pull Requests at this beginner level not squashed when merged?

I agree. We should take better care of it.

Wilfred:
Good suggestion regarding squash commits. It is enabled, we should probably use it more (particularly for PRs with a small number of commits).

PR #402, opened 2017-10-24
Port set-minibuffer-window

### LINKED ISSUE #389: Port set-minibuffer-window to Rust
```c

DEFUN ("set-minibuffer-window", Fset_minibuffer_window,
       Sset_minibuffer_window, 1, 1, 0,
       doc: /* Specify which minibuffer window to use for the minibuffer.
This affects where the minibuffer is displayed if you put text in it
without invoking the usual minibuffer commands.  */)
  (Lisp_Object window)
{
  CHECK_WINDOW (window);
  if (! MINI_WINDOW_P (XWINDOW (window)))
    error ("Window is not a minibuffer window");

  minibuf_window = window;

  return window;
}
```

shaleh:
Make `minibuf_window` mutable.
Add `as_minibuffer_or_error()`

Closes #389

### DIFF
--- a/rust_src/src/minibuf.rs
+++ b/rust_src/src/minibuf.rs
@@ -34,3 +34,17 @@ fn active_minibuffer_window() -> LispObject {
         LispObject::from_raw(unsafe { minibuf_window })
     }
 }
+
+/// Specify which minibuffer window to use for the minibuffer.
+/// This affects where the minibuffer is displayed if you put text in it
+/// without invoking the usual minibuffer commands.
+#[lisp_fn]
+pub fn set_minibuffer_window(window: LispObject) -> LispObject {
+    window.as_minibuffer_or_error(); // just for the checks
+
+    unsafe {
+        minibuf_window = window.to_raw();
+    }
+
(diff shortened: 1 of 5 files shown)


rust_src/src/minibuf.rs
```diff
@@ -34,3 +34,17 @@ fn active_minibuffer_window() -> LispObject {
...
     }
 }
+
+/// Specify which minibuffer window to use for the minibuffer.
+/// This affects where the minibuffer is displayed if you put text in it
+/// without invoking the usual minibuffer commands.
+#[lisp_fn]
+pub fn set_minibuffer_window(window: LispObject) -> LispObject {
+    window.as_minibuffer_or_error(); // just for the checks
+
```
shanavas786: I would prefer `w.is_minibuffer()` check rather than new `as_minibuffer_or_error` method.

shaleh: `is_minibuffer` is insufficient based on the C code. It does not check if it is a valid window, etc.

PR #409, opened 2017-10-25
Port window-frame

shaleh:
Add a `frame` method to `LispWindowRef` to get the window's frame.

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -32,7 +32,7 @@ impl LispWindowRef {
 
     #[inline]
     pub fn frame(&self) -> LispObject {
-        LispObject::from_raw(self.frame)
+        LispObject::from(self.frame)
     }
 
     #[inline]
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -31,6 +31,11 @@ impl LispWindowRef {
     }
 
     #[inline]
+    pub fn frame(&self) -> LispObject {
+        LispObject::from_raw(self.frame)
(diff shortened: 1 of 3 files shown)


shaleh:
The `frame` method will reduce redundancy as port the rest of the frame based functions in window.c.

rust_src/src/windows.rs
```diff
@@ -179,3 +184,17 @@ pub fn minibuffer_selected_window() -> LispObject {
...
+#[lisp_fn(min = "0")]
+pub fn window_frame(window: LispObject) -> LispObject {
+    let win = if window.is_nil() {
+        selected_window()
+    } else {
+        window
+    };
+
+    win.as_window_or_error().frame()
+}
```
brotzeit: It looks like the c code makes use of `CHECK_VALID_WINDOW` which throws a `Qwindow_valid_p` error. We don't have this implemented yet, but I wonder if we already have ported functions with `decode_valid_window` incorrectly, hmmm.

shaleh: I have a PR almost ready :-)

shaleh: The window-buffer routines also should have used `CHECK_VALID`

shaleh: I am adding `window_valid_or_selected` and `window_or_selected` to windows.rs.

shaleh: Bike shed while I am away. I will commit this afternoon.

shaleh:
Waiting on PR #411

PR #411, opened 2017-10-25
Window valid or selection

shaleh:


shaleh:
This is a refactor of the Rust code in Window.rs. I realized this morning that I was not properly porting items. When I looked closer at the code I saw there was a fair amount of duplication and that was only going to continue.

shaleh:
If the style of `as_valid_window_or_error` is accepted I think that `as_live_window_or_error` should follow suit. The current code calls `as_window` twice which is more work than it should be and is inefficient as the `Lisp_Object` is run through the vectorization code twice.

It sure would be nice if the [request](1) for `filter` on `Option` would pass. Then the code would be:

    object.as_window()
          .filter(|w| w.is_valid())
          .unwrap_or_else(|| wrong_type!())

[1]: https://github.com/rust-lang/rfcs/issues/1485

shaleh:
Once this lands I can update my PRs. If there are no objections I can focus on clearing out most of window.c over the next week.

rust_src/src/windows.rs
```diff
@@ -153,11 +168,7 @@ pub fn window_margins(window: LispObject) -> LispObject {
             LispObject::constant_nil()
         }
     }
-    let win = if window.is_nil() {
-        selected_window()
-    } else {
-        window
-    }.as_live_window_or_error();
```
shaleh: Note this code was not following what C did. This enforced live window on `selected_window` which the C code simply assumes is true. Some goes for valid checks too.

brotzeit: I have also noticed it, but I think something would be seriously broken if this not true.

shaleh:
Triggering a recheck

Issue #408, opened 2017-10-25
New compiler warning from recent hashtable PR


shaleh:
fns.c:2851:1: warning: no previous prototype for ‘hash_clear’ [-Wmissing-prototypes]
     hash_clear (struct Lisp_Hash_Table *h)

DavidDeSimone:
Good catch, I missed this. I'll submit a PR that fixes it. It should be a harmless for the time being.

shaleh:
It is harmless. No worries.

PR #414, opened 2017-10-26
Fixing minor warning in fns.c introduced in recent PR.

DavidDeSimone:
Fixes https://github.com/Wilfred/remacs/issues/408

### DIFF
--- a/src/fns.c
+++ b/src/fns.c
@@ -43,6 +43,7 @@ along with GNU Emacs.  If not, see <https://www.gnu.org/licenses/>.  */
 
 enum equal_kind { EQUAL_NO_QUIT, EQUAL_PLAIN, EQUAL_INCLUDING_PROPERTIES };
 bool internal_equal (Lisp_Object, Lisp_Object, enum equal_kind, int, Lisp_Object);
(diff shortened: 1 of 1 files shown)


src/fns.c
```diff
@@ -43,6 +43,7 @@ along with GNU Emacs.  If not, see <https://www.gnu.org/licenses/>.  */
 
 enum equal_kind { EQUAL_NO_QUIT, EQUAL_PLAIN, EQUAL_INCLUDING_PROPERTIES };
 bool internal_equal (Lisp_Object, Lisp_Object, enum equal_kind, int, Lisp_Object);
+void hash_clear (struct Lisp_Hash_Table *h);
```
shaleh: maybe a comment here saying this has moved to Rust?

DavidDeSimone: It hasn't yet moved to Rust. We need to add this declaration because this function was previously static, but since I am now calling it from Rust, it cannot be marked static.

shaleh: Ah. Carry :-)

Wilfred:
Merging to keep our build beautiful :)

Issue #345, opened 2017-10-05
Add lisp functions automatically


brotzeit:
I wonder if there would be a downside to adding lisp functions inside of the lisp_fn macro instead of appending them manually to rust_init_syms in lib.rs ? It would be really convenient if this would be handled automatically.

Wilfred:
Since functions are added in different places in the C codebase, I assumed there was a reason. I'm not aware of any reason though.

I do like that the code is relatively explicit, and I'm not sure how we'd add a function call inside a macro expansion. There's a verbosity-magic tradeoff, we may not be at the sweet spot yet though.

brotzeit:
If we decide to keep it this way, we should consider to move the functions to the according files like in the c codebase. I don't think there is any value in keeping all lisp functions in one big block in lib.rs or how would you want it to look like after all lisp functions were ported ?

Wilfred:
Agreed, it might be tidier to split them up. I don't feel too strongly either way though.

shaleh:
One argument in favor of moving them into the Rust files and out of lib.rs is porters are just adding to the end of the list. This is losing the connection of editfns v. indent v. .....

With each of us chewing at this little by little moving the parts to the file that owns it should help cut down on merge conflicts too.

brotzeit:
Done in #456 :thumbsup:

PR #415, opened 2017-10-26
Removing the need to manually declare symbols, and wrap them in unsafe.

DavidDeSimone:
This works similar to how our setup for emacs_globals works.

It would be nice if we also used this to generate LispObject equivalents for these symbols, i.e. `LispObject(Qnil)` so we don't have to litter the codebase with LispObject::from_raw(Qhash_table_p) etc, but we can tackle that in the future.


DavidDeSimone:
Actually I have a different idea for this. I will close this and update later.

rust_src/remacs-sys/build.rs
```diff
@@ -103,33 +110,60 @@ fn generate_globals() {
...
     let out_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("globals.rs");
     let mut out_file = File::create(out_path).expect("Failed to create definition file");
-
+    let mut parse_state = ParseState::ReadingGlobals;
+    
     write!(out_file, "#[allow(unused)]\n").expect("Write error!");
     write!(out_file, "#[repr(C)]\n").expect("Write error!");
     write!(out_file, "pub struct emacs_globals {{\n").expect("Write error!");
 
+
```
shaleh: dead line

shaleh:
Looking good

brotzeit:
This looks really nice. I would like to bring up [this topic](https://github.com/Wilfred/remacs/issues/345) again, while we're at it. Do we really need a huge list of lisp functions in lib.rs ? I can live with keeping it, but I think we should consider this. We can add them to the corresponding files after all lisp functions are ported, but I think this is useless at this point. And the merge conflicts are really annoying...

shaleh:
Agreed. What are the steps to resolve it? Want to open an issue to track? Or can you just submit a PR and be done? Because I am +10 on this

On October 26, 2017 12:12:50 PM PDT, brotzeit <notifications@github.com> wrote:
>This looks really nice. I would like to bring up [this
>topic](https://github.com/Wilfred/remacs/issues/345) again, while we're
>at it. Do we really need a huge list of lisp functions in lib.rs ? I
>can live with keeping it, but I think we should consider this. We can
>add them to the corresponding files after all lisp functions are
>ported, but I think this is useless at this point. And the merge
>conflicts are really annoying...
>
>-- 
>You are receiving this because you commented.
>https://github.com/Wilfred/remacs/pull/415#issuecomment-339769786

-- 

brotzeit:
I don't know how to do it, I just assumed it's possible ;)
But sure, I'll give it a try.

shaleh:
PR #429 shows how to move `defsubr`.

shaleh:
Trigger build

PR #423, opened 2017-10-26
Add rustfmt helper example.

shaleh:
This works around the issue on Linux where rustfmt-nightly cannot find the
Rust compiler's shared libraries.

### DIFF
--- a/rust_src/rustfmt.helper.sh
+++ b/rust_src/rustfmt.helper.sh
@@ -4,6 +4,6 @@
 #
 #     (setq rust-format-on-save t)
 #
-# Then make sure this script is in your path before ~/.cargo/bin
+# Then make sure this script is named `rustfmt` and in your path before ~/.cargo/bin.
 
 LD_LIBRARY_PATH=$(rustc --print sysroot)/lib:$LD_LIBRARY_PATH ~/.cargo/bin/rustfmt "$@"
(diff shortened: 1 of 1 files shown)


Wilfred:
Does this offer any advantages over `cargo fmt`?

shaleh:
Look at the comment in the script. Emacs runs it whenever I save

On October 28, 2017 3:57:48 AM PDT, Wilfred Hughes <notifications@github.com> wrote:
>Does this offer any advantages over `cargo fmt`?
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/Wilfred/remacs/pull/423#issuecomment-340182772

-- 

Wilfred:
I'm probably being really dumb, but I can't see any comments in the script.

shaleh:
No @Wilfred I committed the wrong version. New push on its way.

rust_src/rustfmt.helper.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Add this to your Emacs config and it will format the file whenever you save.
+#
+#     (setq rust-format-on-save t)
+#
+# Then make sure this script is named `rustfmt` and in your path before ~/.cargo/bin.
```
shanavas786: setting `LD_LIBRARY_PATH=$(rustc --print sysroot)/lib:$LD_LIBRARY_PATH` seems easier than these tweaks. I would rather suggest to add instructions to set `LD_LIBRARY_PATH` in readme.

shaleh: Why pollute that variable because of one busted command?

shaleh: No one is telling anyone to use it. This is offered as an example solution to the problem. Hopefully the compiler gets fixed and this can go away.

PR #424, opened 2017-10-26
Add 32bit Dockerfile.

shaleh:


shaleh:
I added to enable people to find bugs in 32bit.

https://docs.travis-ci.com/user/docker/#Building-a-Docker-Image-from-a-Dockerfile this seems to say we can run docker builds and then run the docker image.

Issue #412, opened 2017-10-25
Style guideline time?


shaleh:
Should we order imports alphabetically?

Should we group sections in rust_src/src/lib.rs?

As discussed elsewhere, if the From PR lands, should we allow `into` for returns but use `::from` otherwise?

Is `return` verboten or does @brotzeit  just not like it :)

jaccarmac:
Your point about imports is addressed by #394 :)

As a newcomer, section grouping would be very nice.

shaleh:
You want to start a PR?

On October 25, 2017 6:13:29 PM PDT, Jacob MacDonald <notifications@github.com> wrote:
>Your point about imports is addressed by #394 :)
>
>As a newcomer, section grouping would be very nice.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/Wilfred/remacs/issues/412#issuecomment-339520203

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jaccarmac:
Plan to as soon as I finish navigating the contribution process and solidify my knowledge of it!

shaleh:
No,the imports are not sorted or ordered:

    use lisp::{ExternalPtr, LispObject};
    use remacs_macros::lisp_fn;
    use remacs_sys::{minibuf_level, minibuf_selected_window as current_minibuf_window,
                 selected_window as current_window, EmacsInt, Lisp_Window};
    use marker::marker_position;
    use editfns::point;
    use libc::c_int;

This grab bag makes it a little confusing for the new contributor. Where should I add code? imports? The current plan appears to be just add everything to the end. Which is not satisfying :-)

jaccarmac:
Which file is that from? I had seen that the imports from each module got sorted, and extrapolated that to the modules themselves; Clearly I was wrong.

shaleh:
That is windows.rs

jaccarmac:
rust-lang-nursery/rustfmt#298 is a discussion of this exact issue and has some links to maybe-useful configuration options. I haven't finished reading the thread but it might be useful to other interested parties in this one.

shaleh:
reorder_imports and reorder_import_names are supported now. We can tweak the config and keep rolling.

On October 26, 2017 7:33:00 PM PDT, Jacob MacDonald <notifications@github.com> wrote:
>rust-lang-nursery/rustfmt#298 is a discussion of this exact issue and
[...]

PR #426, opened 2017-10-27
Re-order imports to minimal merge conflicts. Include a rustfmt config to keep it this way.

shaleh:
The config was generated with --dump-default-config. Then the values
for `reorder_imports` and `reorder_imports_group` were set to true.

Part of #412


shaleh:
This config will sort all imports. The `reorder_import_groups` option means if we put imports in groups they will be honored. For instance:

    use std::thing;
    use libc::foo;
    use std::bar;

    use libc;
    use core::blah;

Will reorder to:

    use libc::foo;
    use std::bar;
    use std::thing;

    use core::blah;
    use libc;

What this means is we will still have some control over the order of imports even if they are being sorted. So we could have system crates, then external crates, then local crates and each sub group will be sorted.

shaleh:
I did not commit the Rust changes because I wanted commentary before this is committed. I can update the PR when the time comes.

jaccarmac:
The config need only contain values we explicitly want to set. Others will be set to their default values. In my opinion, this is preferable as it reduces the number of rules to keep in mind when reading the file or deciding on style changes.

Since the version of `rustfmt-nightly` is managed explicitly upgrades will not cause formatting to change in pull requests which do not update the Travis script, which is one objection I can see to that idea.

shaleh:
Yeah, I was not sure either. This is why I dumped the default config and tweaked the values. Explicit versus implicit. Even if we moved to a new version of rustfmt the coding standard applied to remacs would remain (mostly) the same until we also updated the config.

jaccarmac:
It's a tradeoff between being explicit and having to duplicate the effort of the `rustfmt` maintainers each time they update their list of default rules.

In any case, the two tweaks look good, regardless of how they are serialized :+1:

shaleh:
Toggling to kick build

shaleh:
I have run the new fmt. I also ordered the groups in order from system, remacs libs, local modules.

shaleh:
Yes, we do. Each call to 'cargo fmt' needs its own config. We could store the config as a symlink but what happens on Windows?

shaleh:
Actually, I figured out how to get `cargo fmt` to do the right thing. It should would be nice to have a simple `--config` option. But bah, usability is worthless :-)

Issue #419, opened 2017-10-26
Port process-query-on-exit-flag to Rust


brotzeit:
```c
DEFUN ("process-query-on-exit-flag",
       Fprocess_query_on_exit_flag, Sprocess_query_on_exit_flag,
       1, 1, 0,
       doc: /* Return the current value of query-on-exit flag for PROCESS.  */)
  (register Lisp_Object process)
{
  CHECK_PROCESS (process);
  return (XPROCESS (process)->kill_without_query ? Qnil : Qt);
}
```

tonyday567:
I'll give this a go.

PR #430, opened 2017-10-28
port process_query_on_exit_flag and set_process_query_on_exit_flag

tonyday567:
#419 

The type of kill_without_query is actually a bool_bf rather than a bool.  See conf_post.h

```
/* The type of bool bitfields.  Needed to compile Objective-C with
   standard GCC.  It was also needed to port to pre-C99 compilers,
   although we don't care about that any more.  */
#if NS_IMPL_GNUSTEP
typedef unsigned int bool_bf;
#else
typedef bool bool_bf;
#endif
```

Both my c-fu and rust-fu are insufficient to work out what to do about that.  Everything compiles but I get a segfault - I think whenever remacs attempts to access kill_without_query.

The switch from 'query-on-exit' at the lisp level to the boolean opposite 'kill-without-query' at the c level was annoying, but I guessed this is not the time or place to try and fix that.

I originally went looking for an `as_bool_or_error`, just pattern matching on other PRs, and it took me a while to work out that the emacs-lisp convention that nil means false and non-nil (not just t!) means true is the dominant boolean.  So `is_not_nil` is the closest thing there is to an `as_bool`, and that it will never error.  Could we switch `is_not_nil` to `as_bool`?

### DIFF
--- a/rust_src/remacs-sys/build.rs
+++ b/rust_src/remacs-sys/build.rs
@@ -13,6 +13,12 @@ const WIDE_EMACS_INT: bool = true;
 #[cfg(not(feature = "wide-emacs-int"))]
 const WIDE_EMACS_INT: bool = false;
 
+#[cfg(feature = "ns-impl-gnustep")]
+const NS_IMPL_GNUSTEP: bool = true;
+
+#[cfg(not(feature = "ns-impl-gnustep"))]
+const NS_IMPL_GNUSTEP: bool = false;
+
 fn integer_max_constant(len: usize) -> &'static str {
     match len {
         1 => "0x7F_i8",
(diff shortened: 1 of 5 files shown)


shaleh:
I add the help wanted label so others will see this hopefully. I do not have the time at the moment -- yay Halloween.

Wilfred:
Thanks for your PR! :)

I'm surprised this compiles at all. `Lisp_Process` in `remacs-sys/lib.rs` doesn't have `kill_without_query` defined. The fields in `Lisp_Process` in `process.h` depend on `HAVE_GNUTLS`, and `HAVE_GETADDRINFO_A`, so we'll need to do some work to ensure our struct is always of a matching size.

Regarding `bool_bf`, take a look at how we've defined `EmacsInt` in `rust_src/remacs-sys/build.rs`. That enables us to define types of different sizes according to compile time flags. I imagine we'll want to do the same thing for `bool_bf`.

As for `.as_bool`, I think `.is_nil` and `.is_not_nil` are explicit and avoid confusion. 0 is truthy in elisp (as are empty strings, vectors and hashmaps) and I worry `.as_bool` would cause confusion. I'm open to persuasion, of course :)

tonyday567:
@Wilfred thanks!

Will start chasing down the leads and try and tighten up this PR.  This will involve a fair amount of learning about C and Rust, so might take a bit.

tonyday567:
Just saw #442 which seems to be a different strategy than being followed here.

shaleh:
#442 is following other PRs that recently landed. It is the easy and working way. Once we move the core off of C we can switch to a different solution.

tonyday567:
refactored towards using accessors for Lisp_Process. I wasn't exactly sure whether a C setter was also required for set_process_query_on_exit_flag, but it seemed symmetrical.

I just pattern matched with the build.rs changes, so any subtlety may have passed me by.

tonyday567:
Awesome - this is a great learning experience!

Issue #386, opened 2017-10-23
Verify ported functions are interactive, if interactive on upstream.


DavidDeSimone:
Our `lisp_fn` macro is only now in the process of gaining support for interactive functions. We have ported a number of lisp functions up to this point, so I think it would be a good idea to make sure that all ported functions that are supposed to be interactive actually are interactive. 

Related to https://github.com/Wilfred/remacs/issues/383

Wilfred:
Sounds like a good idea. Were you thinking of checking by hand, or something automated?

galczo5:
Hi,
Here is python script that returns (if you run it in `/rust_src/src`) all ported functions.
```
import subprocess
import re

grep = 'grep -A 1 -h -r "#\[lisp_fn"'
result = subprocess.Popen(grep, shell=True, stdout=subprocess.PIPE).stdout.read().decode("utf-8")

matched = re.findall(r'fn .*\(', result)

for line in matched:
    resultLine = line.replace("fn ", "").replace("(", "").replace("_", "-")
    print(resultLine)

```

I have no idea how to check if functions are interactive (non-manual way) in emacs. Any idea? I can modify script to generate elisp code which will return list of interactive functions.

brotzeit:
Try `(commandp 'function)`

Wouldn't it be easier to do this completely in elisp ?

galczo5:
Maybe :)

DavidDeSimone:
I think we are good. Using @galczo5 's python script to get all the ported functions, I made a quick and dirty elisp file to run on mainline emacs to identify if any of our ported functions are interactive. It looks like **none of our ported functions are interactive**, so we should be good on that front. I am going to close this issue as complete. 

elisp for anyone interested:

``` elisp
(setq ifns '(
            'base64-encode-string
            'base64-decode-string
            'overlayp
            'buffer-live-p
            'get-buffer
            'current-buffer
            'buffer-file-name
            'buffer-modified-p
            'buffer-name
            'buffer-modified-tick
            'buffer-chars-modified-tick
            'overlay-start
            'overlay-end
            'overlay-buffer
            'set-buffer
            'category-table-p
            'category-table
            'max-char
            'characterp
            'char-or-string-p
            'unibyte-char-to-multibyte
            'multibyte-char-to-unibyte
            'char-table-subtype
            'char-table-parent
            'set-char-table-parent
            'forward-point
            'md5
            'secure-hash
            'buffer-hash
            'indirect-function-lisp
[...]

PR #434, opened 2017-10-28
add intspec to goto-char

brotzeit:
@DavidDeSimone your code from #386 doesn't work correctly. It seems `goto-char` is the only interactive function we have ported so far.


shaleh:
Stuck on #438

DavidDeSimone:
You're right, there was a flaw in my elisp. Thanks for catching that @brotzeit

brotzeit:
@DavidDeSimone I only checked it again as I knew `goto-char` was interactive ;)

PR #435, opened 2017-10-28
Port the preceding-char function

robbert-vdh:
As referenced in issue #407. I've never had to do pointer arithmetic in Rust before, so there might be a cleaner way to implement the `DEC_POS` macro.

### DIFF
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -4,6 +4,7 @@ use libc::{c_uchar, ptrdiff_t};
 
 use remacs_macros::lisp_fn;
 use remacs_sys::EmacsInt;
+
 use threads::ThreadState;
 
 /// True iff byte starts a character in a multibyte form.
@@ -18,13 +19,11 @@ pub fn char_read_p(byte: c_uchar) -> bool {
 /// to the previous character boundary. No range checking of POS.
 ///
 /// Can be used instead of the `DEC_POS` macro.
-// XXX: Pointer arithmetic is wild. Please modify this if you know a
-//      more idiomatic way to move a pointer.
 #[inline]
-pub unsafe fn dec_pos(pos_byte: &ptrdiff_t) -> ptrdiff_t {
+pub unsafe fn dec_pos(pos_byte: ptrdiff_t) -> ptrdiff_t {
     let buffer_ref = ThreadState::current_buffer();
-    let mut new_pos = pos_byte - 1;
 
+    let mut new_pos = pos_byte - 1;
     let mut offset = new_pos - buffer_ref.beg_byte();
     if new_pos >= buffer_ref.gpt_byte() {
         offset += buffer_ref.gap_size();
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -1,7 +1,43 @@
 //! Operations on characters.
 
+use libc::{c_uchar, ptrdiff_t};
+
 use remacs_macros::lisp_fn;
 use remacs_sys::EmacsInt;
+use threads::ThreadState;
+
+/// True iff byte starts a character in a multibyte form.
+///
+/// Same as the `CHAR_READ_P` macro.
+#[inline]
+pub fn char_read_p(byte: c_uchar) -> bool {
+    (byte & 0xC0) != 0x80
+}
+
+/// Decrement the buffer byte position POS_BYTE of the current buffer
+/// to the previous character boundary. No range checking of POS.
+///
+/// Can be used instead of the `DEC_POS` macro.
+// XXX: Pointer arithmetic is wild. Please modify this if you know a
+//      more idiomatic way to move a pointer.
+#[inline]
+pub unsafe fn dec_pos(pos_byte: &ptrdiff_t) -> ptrdiff_t {
+    let buffer_ref = ThreadState::current_buffer();
+    let mut new_pos = pos_byte - 1;
+
+    let mut offset = new_pos - buffer_ref.beg_byte();
+    if new_pos >= buffer_ref.gpt_byte() {
+        offset += buffer_ref.gap_size();
+    }
+    let mut chp = buffer_ref.beg_addr().offset(offset);
+
+    while !char_read_p(*chp) {
+        chp = chp.offset(-1);
+        new_pos -= 1;
+    }
+
+    new_pos
+}
 
 use lisp::LispObject;
 use lisp::defsubr;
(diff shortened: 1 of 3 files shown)


shaleh:
Stuck on #438

rust_src/src/character.rs
```diff
@@ -1,9 +1,42 @@
...
+pub fn char_read_p(byte: c_uchar) -> bool {
+    (byte & 0xC0) != 0x80
+}
+
+/// Decrement the buffer byte position POS_BYTE of the current buffer
+/// to the previous character boundary. No range checking of POS.
+///
+/// Can be used instead of the `DEC_POS` macro.
+#[inline]
+pub unsafe fn dec_pos(pos_byte: &ptrdiff_t) -> ptrdiff_t {
```
shaleh: The C code from src/char.h

    #define DEC_POS(pos_byte)			\
     do {						\
        unsigned char *chp;				\
    	    					\
        pos_byte--;					\
        if (pos_byte < GPT_BYTE)			\
          chp = BEG_ADDR + pos_byte - BEG_BYTE;	\
        else					\
          chp = BEG_ADDR + GAP_SIZE + pos_byte - BEG_BYTE; \
        while (!CHAR_HEAD_P (*chp))			\
          {						\
	      chp--;					\
	      pos_byte--;				\
          }						\
      } while (false)


Notice that `pos_byte` is decremented. This is not what the Rust code does. the value of `pos_byte` is copied. This might lead to subtle bugs.

robbert-vdh: I originally implemented it that way, but as far as I could tell the macro was never used for side effects (i.e. for pointer arithmetic). That's why I'd decided to implement it in a more idiomatic way instead of mutating a reference. The rust compiler should be able to optimize the copy away, so it's purely a matter of personal preference ([example](https://godbolt.org/g/GTyMu6)).

If you'd prefer to stay true to the original C code I could of course change the function to mutate the reference!

shaleh: I just looked at the code and you are right `DEC_POS`. They just modify the pos in place because they cannot return.

shaleh: Next question. Does the whole function have to be unsafe? The rest of remacs tries to limit the scope of that.

robbert-vdh: Yeah that's a weird one. The way it's used in the C codebase it should never panic, but it doesn't do any bounds checking so an incorrect `pos_byte` argument could lead to segfaults.

shaleh:
I helped with the merge conflicts. Now for reviews.

rust_src/src/character.rs
```diff
@@ -1,12 +1,46 @@
...
 use remacs_sys::EmacsInt;
 
 use lisp::LispObject;
 use lisp::defsubr;
 use multibyte::{make_char_multibyte, raw_byte_from_codepoint_safe};
 use multibyte::MAX_CHAR;
+use threads::ThreadState;
+
+/// True iff byte starts a character in a multibyte form.
+///
```
brotzeit: Copied typo "iff" from upstream.

rust_src/src/character.rs
```diff
@@ -1,12 +1,46 @@
...
+/// Same as the `CHAR_READ_P` macro.
```
brotzeit: Same as `CHAR_HEAD_P` macro ?

brotzeit: Same as `CHAR_HEAD_P` ?

rust_src/src/character.rs
shaleh: 'iff'' is not a typo. It means "if and only if". There is no need for that extra strong wording here though. 'If' is sufficient.

brotzeit: Something else that I've learned :)

rust_src/src/character.rs
shaleh: yes. `read` is a typo in the comment and the function name.

rust_src/src/character.rs
shaleh: If `pos_byte` is not being modified there is no reason to pass a reference to an integer. Just pass `pos_byte` by value.

shaleh:
Ping. @robbert-vdh can you update this PR?

robbert-vdh:
I've updated `dec_pos` to take a value instead of a reference and rebased onto the current master branch!

rust_src/src/character.rs
```diff
@@ -1,7 +1,41 @@
 //! Operations on characters.
 
+use libc::{c_uchar, ptrdiff_t};
+
 use remacs_macros::lisp_fn;
 use remacs_sys::EmacsInt;
+use threads::ThreadState;
```
shaleh: You want a space here. It should be:

    use std; // if any

    use external crates; // if any. Such as libc, regex, etc.

    use remacs crates;

    use modules in the sys directory

Note the empty line between each.

rust_src/src/editfns.rs
```diff
@@ -297,6 +298,24 @@ pub fn following_char() -> EmacsInt {
     }
 }
 
+/// Return the character preceding point, as a number. At the
+/// beginning of the buffer or accessible region, return 0.
+#[lisp_fn(c_name = "previous_char")]
```
shaleh: Why the special `c_name`?

rust_src/src/editfns.rs
```diff
@@ -297,6 +298,24 @@ pub fn following_char() -> EmacsInt {
...
+pub fn preceding_char() -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
+
+    if buffer_ref.pt <= buffer_ref.begv {
+        return LispObject::from_natnum(0);
```
shaleh: Since you are returning `natnum` this routine can return `EmacsInt` instead of `LispObject`.

rust_src/src/editfns.rs
```diff
@@ -297,6 +298,24 @@ pub fn following_char() -> EmacsInt {
...
+    }
+
+    let pos = if LispObject::from_raw(buffer_ref.enable_multibyte_characters).is_not_nil() {
```
shaleh: instead of making a LispObject just for the comparison you can say `if buffer_ref.enable_multibyte_characters != Qnil`

robbert-vdh:
Thanks for the feedback! I've (finally) updated the commit.

shaleh:
Thanks for keeping at it @robbert-vdh

PR #436, opened 2017-10-28
Port the following-char function

robbert-vdh:


shaleh:
Stuck on #438

shaleh:
Kicking the build

Issue #438, opened 2017-10-28
Travis CI is broken. rustfmt-nightly is no longer installing


shaleh:
It looks like a dependency shifted. I do not have time this weekend.

shaleh:
It is due to the new nightly 1.23 from 10.27 fails. 1.22 from 10.22 works. So we can pin to that version of nightly or look into resolving the failure in rustfmt 0.2.9.

jaccarmac:
rustfmt 0.2.12 does appear to build on the latest nightly. We can bump that version instead.

shaleh:
As long as 0.2.12 is drop in clean go for it.

We probably want to freeze the compiler too. Just to avoid this in the future.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

shaleh:
@jaccarmac can you make a PR? Lock Travis to 1.23 and rustfmt 0.2.12. Maybe at a bit to the README too so in a few weeks we don't catch people. Maybe it is time to start a CONTRIBUTING.remacs.md?

shaleh:
remacs looks like it builds ok with 1.23. I have a local compile almost finished.

jaccarmac:
`CONTRIBUTING` is a good idea. However, I don't think locking the Rust version is a good idea. I see most nightly Rust projects in the wild just follow the nightly.

How does one even lock the nightly version? I haven't had occasion to do it before.

shaleh:
Specify a version to rustup.

jaccarmac:
rustup isn't used directly in Travis. Not sure if the Travis Rust version specification supports nightly versions. Trying a pull request with one part, then I'll roll the other change in separately and see if it works.

Wilfred:
Fixed by #439 I think, but feel free to reopen if you think there's more we should do.

shaleh:
Its baaaacccckkkk.

shaleh:
I locked down the Rust version used by the rustfmt check. #441

PR #439, opened 2017-10-28
Bump rustfmt version.

jaccarmac:
See #438.


jaccarmac:
Changing the reference to the rustfmt version in multiple places is somewhat annoying, perhaps time for a shell script to do that maintenance for us?

shaleh:
I will bless this once Travis completes

shaleh:
Thanks for the quick turn around

PR #442, opened 2017-10-29
Port process-id to Rust Closes #431

### LINKED ISSUE #431: Port process-id to Rust
```c
DEFUN ("process-id", Fprocess_id, Sprocess_id, 1, 1, 0,
       doc: /* Return the process id of PROCESS.
This is the pid of the external process which PROCESS uses or talks to.
For a network, serial, and pipe connections, this value is nil.  */)
{
  pid_t pid;

  CHECK_PROCESS (process);
  pid = XPROCESS (process)->pid;
  return (pid ? make_fixnum_or_float (pid) : Qnil);
}
```

Titan-C:
I gave a try to this, but I got stuck.
I added the `pid` attribute to the `Lisp_Process` structure, but where do I assign the value for it when the struct is instantiated?

I was also wondering about the `make_fixnum_or_float`, does it make sense to use it here in remacs? When would the process_id overflow to be transformed as an int. Maybe is there for some legacy reasons but the type `libc::pid_t=i32` that should be enough for all process_id?

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -821,7 +821,14 @@ pub struct Lisp_Process {
     /// Queue for storing waiting writes.
     pub write_queue: Lisp_Object,
 
-    // TODO: this struct is incomplete.
+    // This struct is incomplete.
+    // To access remaining fields use access functions written in
+    // src/process.c and export them here for use in Rust.
+}
+
+/// Functions to access members of `struct Lisp_Process`.
+extern "C" {
+    pub fn pget_pid(p: *const Lisp_Process) -> libc::pid_t;
 }
 
 #[repr(C)]
--- a/rust_src/src/process.rs
+++ b/rust_src/src/process.rs
@@ -1,12 +1,12 @@
 //! Functions operating on process.
 
-use remacs_macros::lisp_fn;
-use remacs_sys::{Fmapcar, Lisp_Process, Qcdr, Qlistp, Vprocess_alist};
-
 use buffers::get_buffer;
 use lisp::{ExternalPtr, LispObject};
 use lisp::defsubr;
 use lists::{assoc, cdr};
+use remacs_macros::lisp_fn;
+use remacs_sys::{EmacsInt, Fmapcar, Lisp_Process, Qcdr, Qlistp, Vprocess_alist};
+use remacs_sys::pget_pid;
 
 pub type LispProcessRef = ExternalPtr<Lisp_Process>;
 
@@ -63,6 +63,19 @@ fn process_buffer(process: LispObject) -> LispObject {
     process.as_process_or_error().buffer()
 }
 
+/// Return the process id of PROCESS.
+/// This is the pid of the external process which PROCESS uses or talks to.
+/// For a network, serial, and pipe connections, this value is nil.
+#[lisp_fn]
+fn process_id(process: LispObject) -> LispObject {
+    let pid = unsafe { pget_pid(process.as_process_or_error().as_ptr()) };
+    if pid != 0 {
+        LispObject::from_fixnum(pid as EmacsInt)
+    } else {
+        LispObject::constant_nil()
+    }
+}
+
 /// Return the (or a) live process associated with BUFFER.
 /// BUFFER may be a buffer or the name of one.
 /// Return nil if all processes associated with BUFFER have been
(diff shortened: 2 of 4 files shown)


rust_src/remacs-sys/lib.rs
```diff
@@ -823,6 +823,13 @@ pub struct Lisp_Process {
     /// Queue for storing waiting writes.
     pub write_queue: Lisp_Object,
 
+    /// Process ID.  A positive value is a child process ID.
+    /// Zero is for pseudo-processes such as network or serial connections,
+    /// or for processes that have not been fully created yet.
+    /// -1 is for a process that was not created successfully.
+    /// -2 is for a pty with no process, e.g., for GDB.  */
+    pub pid: libc::pid_t,
```
shaleh: This will not work because there are more objects in the struct between `write_queue` and `pid_t`. Look at #427 for a guide to how you need to get this working.

rust_src/src/process.rs
```diff
@@ -18,6 +18,11 @@ impl LispProcessRef {
     fn buffer(&self) -> LispObject {
         LispObject::from(self.buffer)
     }
+    #[inline]
```
shaleh: Mind your whitespace

rust_src/src/process.rs
```diff
@@ -18,6 +18,11 @@ impl LispProcessRef {
...
+    fn pid(&self) -> LispObject {
+        LispObject::from(self.pid as EmacsInt)
```
shaleh: What you want is `from_fixnum`.

rust_src/src/process.rs
```diff
@@ -61,6 +66,14 @@ fn process_buffer(process: LispObject) -> LispObject {
     process.as_process_or_error().buffer()
 }
 
+/// Return the process id of PROCESS.
+/// This is the pid of the external process which PROCESS uses or talks to.
+/// For a network, serial, and pipe connections, this value is nil.
+#[lisp_fn]
+fn process_id(process: LispObject) -> LispObject {
+    process.as_process_or_error().pid()
```
shaleh: You do not handle the case where pid is invalid and return nil.

rust_src/remacs-sys/lib.rs
brotzeit: Maybe we should add this to the walkthrough as it will be a frequent problem.

shaleh: Agreed.

Titan-C:
This seems to be working now. I'm not entirely sure it was the correct path to the solution, and that I correctly represent the previous code. 
To test I very naively used (run-python) to have a process launched and then (process-id (get-buffer-process "*Python*")) to use this function. It works, but I have no idea how to check for the other cases.

`lib.rs` seems to have changes quite a lot since my last commit. I'll have a look to rebase this PR.

rust_src/src/process.rs
```diff
@@ -1,7 +1,7 @@
 //! Functions operating on process.
 
 use remacs_macros::lisp_fn;
-use remacs_sys::{Fmapcar, Lisp_Process, Qcdr, Qlistp, Vprocess_alist};
+use remacs_sys::{fget_pid, EmacsInt, Fmapcar, Lisp_Process, Qcdr, Qlistp, Vprocess_alist};
```
shaleh: Please put `fget_pid` on its own line. We are breaking up the big imports to reduce conflicts.

src/process.h
```diff
@@ -201,6 +201,8 @@ struct Lisp_Process
     bool_bf gnutls_complete_negotiation_p : 1;
 #endif
 };
+/* Accessors for Rust */
+pid_t fget_pid(const struct Lisp_Process *p);
```
shaleh: Follow the C style. The return type is on a line then the rest of the definition is on the next.

Titan-C: I did not noticed that style from PR  #427. But sure I'll follow the C style.

src/process.c
```diff
@@ -392,6 +392,13 @@ pset_stderrproc (struct Lisp_Process *p, Lisp_Object val)
   p->stderrproc = val;
 }
 
+/* Accessors to enable Rust code to get data from the Lisp_Process struct */
+pid_t fget_pid(const struct Lisp_Process *p)
```
shaleh: This should be `pget_pid`. The 'f' in mine is for 'frame'. Look at process.h you can see others.

Titan-C: I did not figure it out why the f was there. Thank you! Is fixed now.

Titan-C:
I rebased and squashed the commit history to solve the conflict with master and have a clean history of this PR

shaleh:
We can squash on commit, so it is appreciated but not necessary.

PR #445, opened 2017-10-30
Fix Windows builds

DavidDeSimone:
The main problems with windows build's I found: 

- The size and implementation of bitfields differ for gcc on msys2 vs gcc on my debian machine. The size of these structs is much larger on msys2, so I have created a "BitfieldPadding" struct that is used to pad out Lisp_Symbols, etc. We should not access these fields directly, and instead should use C functions to get/set those values.
- In C, the Lisp_Subr's are static. This means there is no work to make them survive an unexec. However, in Rust we make the Lisp_Sub'rs `lazy_static`. This means that the Lisp_Subr itself lives on the heap, allocated by the default system allocator (which I believe is je_malloc). Allocations made by je_malloc are not guaranteed to survive unexec, which is happening on Windows. The easiest solution that I could think of was to explicitly allocate our Lisp_Subr's with a call to xmalloc. This changes what our calls to "defsubr" look like. 
- On msys2, ld.exe does not seem to always link to chkstk.obj when it needs to. This means that ld.exe can produce binaries that fail to link with __chkstk.  Reading online, a lot of people seem to pass the blame on this "chkstk.obj" issue, but I believe it lies with ld.exe. It looks like to get around this, we were having __chkstk call ___chkstk, but this was causing a SEGFAULT on my machine whenever we would allocate a large enough struct on the stack. In order to satisfy the linker, I've just stubbed __chkstk, making it a no-op. 

TODOs:
- [x] Set up appveyor so we can have CI for windows. I will most likely need help from others to get this set up (I don't know if any collaborator can do this, or if it needs to be @Wilfred ). In the meantime, I can create the appveyor.yml iterating on my own fork of remacs. 
- [x] Address the failing tests for windows.


shaleh:
It looks like I can add Wilfred/remacs as a project on Appveyor

DavidDeSimone:
So appveyor is setup for my fork, the results of which can be found here: 
https://ci.appveyor.com/project/DavidDeSimone/remacs

It should just be a matter of a maintainer setting up remacs on appveyor now.

From appveyor, it looks like some of our tests are failing. I will investigate. I'm first going to look upstream to see if these tests are also failing on the mainline, or if we have introduced these errors. 
```
src/process-tests.log
lisp/vc/vc-tests.log
lisp/net/network-stream-tests.log
lisp/ibuffer-tests.log
lisp/files-tests.log
lisp/calendar/icalendar-tests.log
```

Wilfred:
OK, I've created remacs as a project on Appveyor: https://ci.appveyor.com/project/Wilfred/remacs

Let me know if there are any settings that you can't modify and need me.

DavidDeSimone:
I'm in the process of fixing these merge commits, and evaluating if these tests are failing on upstream. I should have an update very shortly.

DavidDeSimone:
My results from testing upstream, from our last rollup:
```
UPSTREAM (a7ed4ec0958662265136820333da1a8fbd3a864e):
src/fileio-tests.log
lisp/vc/vc-tests.log
lisp/net/network-stream-tests.log
lisp/ibuffer-tests.log

REMACS:
lisp/vc/vc-tests.log
lisp/net/network-stream-tests.log
lisp/ibuffer-tests.log
lisp/files-tests.log
```

The other two tests seem to have been fixed on my local machine after the merge and some code cleanup. I think we should ignore these tests on Windows for the time being.

DavidDeSimone:
`cargo test` also fails due to linker errors. It looks like it's trying to link against all on our function's defined in remacs-sys, even if they aren't used by the tests. For the time being I've punted on that, and am just not running cargo test on windows. I figure we can either fix this by fiddling with linker settings in our Cargo.toml, or use `mock_derive` to stub out those functions to get the program to link.

DavidDeSimone:
So appveyor seems to have the problem that sometimes, their builds will just stop in their tracks, and not recover. They will timeout after an hour, and not recover. I have contacted the appveyor team with this issue here: https://help.appveyor.com/discussions/problems/9476-build-randomly-stalling

DavidDeSimone:
yay appveyor CI caught a windows specific issue!
```
error[E0412]: cannot find type `pid_t` in module `libc`
   --> C:\projects\remacs\rust_src\remacs-sys\lib.rs:816:54
    |
816 |     pub fn pget_pid(p: *const Lisp_Process) -> libc::pid_t;
    |                                                      ^^^^^ not found in `libc`
error: Could not compile `remacs-sys`.
```

However, configure reports ```checking for pid_t... yes```
I would assume that this is a bug/unsupported feature of libc?

DavidDeSimone:
Also appveyor got back to me (very quickly) about the build stalling issue, and gave me some advice about our setup. They suggest  adding a `non-blocking RDP` statement to the beginning of the build. I'll have time to do that later today if no one else gets around to it.

shaleh:
pid_t is indeed not defined. I opened https://github.com/rust-lang/libc/issues/832

shaleh:
https://github.com/rust-lang/libc/pull/833

shaleh:
@DavidDeSimone I made you a collaborator on my fork of the rust-lang/libc project. I won't have time to touch this again until at least Sunday night. Feel free to update the PR before then.

shaleh:
BAH!!!

libc upstream just refused the PR because it is a synthetic addition. So we have to fix this internally. They actually believe this is some other crate's problem. Yay people who have not solved problems across platforms.

rust_src/remacs-sys/lib.rs
```diff
@@ -23,6 +23,10 @@ pub mod libm;
 use libc::{c_char, c_double, c_float, c_int, c_short, c_uchar, c_void, intmax_t, off_t, ptrdiff_t,
            size_t, time_t, timespec};
 
+// libc refuses to merge pid_t as an alias for c_int in Windows, so we will not use libc::pid_t
```
Wilfred: Perhaps 'prefers not to merge'? 'Refuses' sounds like we think they're being unreasonable. It's a shame, but I think their position is defensible, and being nice makes it easier to file bug reports in future :)

DavidDeSimone: Good suggestion, will update.

DavidDeSimone:
Build is looking green. I personally feel satisfied with this PR, I think it's ready to merge unless anyone else has additional concerns.

PR #447, opened 2017-10-31
port framep, frame-live-p, set-frame-selected-window

brotzeit:


rust_src/src/frames.rs
```diff
@@ -1,21 +1,110 @@
...
+use objects::eq;
+use std::ptr;
 use lisp::defsubr;
 
 pub type LispFrameRef = ExternalPtr<Lisp_Frame>;
 
+impl LispFrameRef {
+    #[inline]
+    pub fn is_live(&self) -> bool {
+        if unsafe { fget_terminal(self.as_ptr()) == ptr::null() } {
```
shaleh: Why not write this as `fget_terminal != null()` and skip the if/else?

rust_src/src/frames.rs
```diff
@@ -1,21 +1,110 @@
...
+/// argument NORECORD non-nil means to neither change the order of recently
+/// selected windows nor the buffer list.  WINDOW must denote a live window.
+/// Return WINDOW.
+#[lisp_fn(min = "2")]
+pub fn set_frame_selected_window(
+    frame: LispObject,
+    window: LispObject,
+    norecord: LispObject,
+) -> LispObject {
+    let mut f = frame;
```
shaleh: This pattern is repeated in a few functions in window.c. I was working in the same place this morning and wrote this:

    pub fn frame_live_or_selected(object: LispObject) -> LispFrameRef {
        let frame = if object.is_nil() {
            selected_frame()
        } else {
            object
        }.as_frame_or_error();
    
        if !frame.is_live() {
            wrong_type!(Qframe_live_p, object);
        }
    
        frame
    }
    
    pub fn window_frame_live_or_selected(window: LispObject) -> LispFrameRef {
        if window.is_nil() {
            selected_frame().as_frame_or_error()
        } else {
            window
                .as_valid_window()
                .map_or(frame_live_or_selected(window), |w| {
                    w.frame().as_frame_or_error()
                })
        }
    }

rust_src/remacs-sys/lib.rs
```diff
@@ -924,10 +924,25 @@ pub struct Lisp_Frame {
...
+pub enum OutputMethod {
+    output_initial,
+    output_termcap,
+    output_x_window,
+    output_w32,
+    output_ns,
+}
+
+#[repr(C)]
+pub struct terminal {}
```
shaleh: Does it make sense to include at least the first element which is the header?

rust_src/src/frames.rs
brotzeit: Oops :)

rust_src/src/frames.rs
brotzeit: LGTM

rust_src/src/frames.rs
birkenfeld: Or rather, `!fget_terminal.is_null()`

rust_src/remacs-sys/lib.rs
```diff
@@ -924,10 +924,25 @@ pub struct Lisp_Frame {
...
+
 /// Functions to access members of `struct frame`.
+#[allow(improper_ctypes)]
 extern "C" {
     pub fn fget_column_width(f: *const Lisp_Frame) -> c_int;
     pub fn fget_line_height(f: *const Lisp_Frame) -> c_int;
+    pub fn fget_terminal(f: *const Lisp_Frame) -> *const terminal;
+    pub fn fget_output_method(f: *const Lisp_Frame) -> OutputMethod;
```
birkenfeld: This is, unfortunately, not sound. See the announcement of bindgen 0.31, which changed from representing C enums as Rust enums: https://www.reddit.com/r/rust/comments/795n0k/announcing_bindgen_0310/

> C/C++ enums are now translated into constants by default, rather than Rust enums. The old behavior was a big footgun because rustc assumes that the only values of an enum are its variants, whereas a lot of C/C++ code uses random values as enums. Put these two things and it leads to undefined behavior.

Changing this will also fix the need to include an "unnecessary" match pattern below.

rust_src/src/frames.rs
```diff
@@ -1,21 +1,110 @@
...
+        if unsafe { fget_terminal(self.as_ptr()) == ptr::null() } {
+            false
+        } else {
+            true
+        }
+    }
+
+    #[inline]
+    pub fn set_selected_window(&mut self, window: LispObject) {
+        self.selected_window = window.to_raw();
```
birkenfeld: should use `fset_selected_window`?

rust_src/src/frames.rs
```diff
@@ -1,21 +1,110 @@
...
+    norecord: LispObject,
+) -> LispObject {
+    let mut f = frame;
+    if f.is_nil() {
+        f = selected_frame();
+    }
+    let mut frame_ref = f.as_live_frame_or_error();
+    let w = window.as_live_window_or_error();
+
+    if eq(frame, w.frame()).is_nil() {
```
birkenfeld: `eq(x, y).is_nil()` is just `x.ne(y)`. (There's also `x.eq(y)` of course.)

rust_src/src/objects.rs
```diff
@@ -14,7 +14,7 @@ fn null(object: LispObject) -> LispObject {
 
 /// Return t if the two args are the same Lisp object.
 #[lisp_fn]
-fn eq(obj1: LispObject, obj2: LispObject) -> LispObject {
+pub fn eq(obj1: LispObject, obj2: LispObject) -> LispObject {
```
birkenfeld: (unneeded)

rust_src/src/frames.rs
```diff
@@ -56,4 +62,71 @@ pub fn selected_frame() -> LispObject {
...
+    norecord: LispObject,
+) -> LispObject {
+    let f = if frame.is_nil() {
+        selected_frame()
+    } else {
+        frame
+    };
+    let mut frame_ref = f.as_live_frame_or_error();
+    let w = window.as_live_window_or_error();
+
```
brotzeit: I have to do it this way as I need the updated frame `LispObject` below.

rust_src/src/frames.rs
```diff
@@ -56,4 +62,71 @@ pub fn selected_frame() -> LispObject {
...
+    if !f.eq(w.frame()) {
+        error!("In `set-frame-selected-window', WINDOW is not on FRAME")
+    }
```
brotzeit: Would it be better to define a `ne` method for `LispObject` ?

shaleh: `PartialEq` between a Frame and Window would invoke `window.frame()`?

brotzeit: I mean for LispObject. But this was just an idea. Maybe it's better to use eq and ne to make it obvious what is actually compared.

birkenfeld: There is already a `ne` method if I'm not mistaken.

I chose not to implement `PartialEq` because there are so many Lisp equality predicates (eq, eql, equal, possibly more), so this is more explicit.

rust_src/remacs-sys/lib.rs
```diff
@@ -941,13 +941,22 @@ pub struct terminal {
     pub header: Lisp_Vectorlike_Header,
 }
 
+pub enum OutputMethod {
```
birkenfeld: Again, I'm not comfortable with C functions being declared as directly returning Rust enums...

shaleh: @birkenfeld would you put a note in the contributing file about that?

brotzeit: What's the right way to do it ?

birkenfeld: Named constants, I think.

shaleh: But that means that `match` does not support these names very well, right?

birkenfeld: `match` should work fine with constants. You just have to include a catch-all case, which is as should be.

shaleh: Right. But the compiler does not know enough to warn that you have captured all cases without the catch all. Why not manually set the value of the enum? Foo(0), Bar(1), etc.

birkenfeld: That is not the point; please read the text I posted at https://github.com/Wilfred/remacs/pull/447#discussion_r148715654

shaleh: Which is why I asked you to document it in our CONTRIBUTING. This is a change in expectation. Documenting why, with a link to more discussions, experiences, etc. and maybe a snippet showing what to do.

brotzeit: If I can get this right, I can also do the documentation, I guess.

birkenfeld: Thanks!

rust_src/src/frames.rs
```diff
@@ -118,11 +118,13 @@ pub fn set_frame_selected_window(
...
-            OutputMethod::output_w32 => LispObject::from(Qw32),
-            OutputMethod::output_ns => LispObject::from(Qns),
+            0 => LispObject::constant_t(),
+            1 => LispObject::constant_t(),
+            2 => LispObject::from(Qx),
+            3 => LispObject::from(Qw32),
+            4 => LispObject::from(Qpc),
+            5 => LispObject::from(Qns),
+            _ => panic!("Invalid output method"),
         }
```
brotzeit: I don't know how to avoid the _ pattern ?

birkenfeld: You don't - that's the consequence of having unreliable enums in C.

But you should introduce some named constants for the existing cases.

shaleh: What about using `unreachable!` or implemented our own similar macro?

brotzeit: @birkenfeld I wasn't sure if you really mean that. I will try to fix it.

brotzeit:
Is this ok or do I need to return constants for every case ? And is it really that important for now ? I mean hopefully the enum issue will be fixed soon and this seems to be at least a safe way ?

brotzeit:
I have to make the enum matching on the c side or it won't work without the _ pattern. I don't know why I need additonal constants.

birkenfeld:
I'd say the problem is using `u8` - an enum is usually `u32` IIRC. But we already have a module that calculates C type sizes, so you can do the same for enums there.

birkenfeld:
Even better would be to give the C function (`fget_output_method`) a normal integer return type (C has no problems with that).

brotzeit:
But this still doesn't solve the _ pattern thing.

birkenfeld:
What exactly do you mean by "that _ pattern thing"? Please elaborate.

Also, why does the Rust function not handle "5" now?

brotzeit:
You wrote 2 weeks ago 

> Changing this will also fix the need to include an "unnecessary" match pattern below.

So I have tried to remove it in my last commits.

> Also, why does the Rust function not handle "5" now?

Disappeared somehow :)

birkenfeld:
> So I have tried to remove it in my last commits.

Ah. Sorry, that was ambiguous - what I meant is that the `_` case is needed, but now rustc doesn't call it "unncessary" anymore, and you don't have to suppress a warning.

brotzeit:
Anything else ?

rust_src/src/frames.rs
shaleh: `frame_live_or_selected` landed while this was in the works. That simplifies the logic here.

rust_src/remacs-sys/lib.rs
```diff
@@ -948,6 +948,7 @@ extern "C" {
     pub fn fget_minibuffer_window(f: *const Lisp_Frame) -> Lisp_Object;
     pub fn fget_root_window(f: *const Lisp_Frame) -> Lisp_Object;
     pub fn fget_terminal(f: *const Lisp_Frame) -> *const terminal;
+    pub fn fget_output_method(f: *const Lisp_Frame) -> c_int;
```
shaleh: Add as well:

    pub type output_method = u32;
    pub const output_initial: output_method = 0;
    pub const output_termcap: output_method = 1;
    pub const output_x_window: output_method = 2;
    pub const output_msdos_raw: output_method = 3;
    pub const output_w32: output_method = 4;
    pub const output_ns: output_method = 5;

This way you do not have magic numbers in `framep` and you avoid the C enum issue @birkenfeld raised.

shaleh: yes it is ugly. and annoying to hand type out those numbers. I'd love a macro for it.

birkenfeld: One more suggestion would be to define the enum nevertheless, and use one of the crates that generate a `TryFrom<int>` impl for it automatically.

rust_src/src/frames.rs
```diff
@@ -77,4 +91,73 @@ pub fn selected_frame() -> LispObject {
...
+/// argument NORECORD non-nil means to neither change the order of recently
+/// selected windows nor the buffer list.  WINDOW must denote a live window.
+/// Return WINDOW.
+#[lisp_fn(min = "2")]
+pub fn set_frame_selected_window(
+    frame: LispObject,
+    window: LispObject,
+    norecord: LispObject,
+) -> LispObject {
+    let f = if frame.is_nil() {
```
shaleh: `frame_live_or_selected` handles this.

rust_src/src/frames.rs
```diff
@@ -77,4 +91,73 @@ pub fn selected_frame() -> LispObject {
...
+    let mut frame_ref = frame_live_or_selected(f);
+    let w = window.as_live_window_or_error();
+
+    if f.ne(w.frame()) {
+        error!("In `set-frame-selected-window', WINDOW is not on FRAME")
+    }
+    if f.eq(selected_frame()) {
+        unsafe { LispObject::from(Fselect_window(window.to_raw(), norecord.to_raw())) }
+    } else {
+        frame_ref.fset_selected_window(window);
```
shaleh: Why the odd name? Shouldn't this be `frame_ref.set_selected_window(window)`?

rust_src/remacs-sys/lib.rs
brotzeit: Sounds more elegant, but is this worth the effort considering that we want to remove the c structs eventually ? It seems to me, that we will have many "ugly solutions" while there's c code. 

And here the constant solution looks simple and explicit.

birkenfeld: No real preference, just an idea I had :)

rust_src/src/frames.rs
```diff
@@ -77,4 +91,68 @@ pub fn selected_frame() -> LispObject {
...
+) -> LispObject {
+    let mut frame_ref = frame_live_or_selected(frame);
+    let w = window.as_live_window_or_error();
+
+    if frame_ref != w.frame().as_frame().unwrap() {
+        error!("In `set-frame-selected-window', WINDOW is not on FRAME")
+    }
+    if frame_ref == selected_frame().as_frame().unwrap() {
+        unsafe { LispObject::from(Fselect_window(window.to_raw(), norecord.to_raw())) }
+    } else {
```
shaleh: I have been pondering updating the API because the `as_foo` methods are almost always returning a foo so the unwrap is just noise. However, we clearly also need the Option return version as well.

rust_src/src/frames.rs
```diff
@@ -77,4 +91,68 @@ pub fn selected_frame() -> LispObject {
     unsafe { LispObject::from(current_frame) }
 }
 
+/// Return non-nil if OBJECT is a frame which has not been deleted.
+/// Value is nil if OBJECT is not a live frame.  If object is a live
+/// frame, the return value indicates what sort of terminal device it is
+/// displayed on.  See the documentation of `framep' for possible
+/// return values.
+#[lisp_fn]
+fn frame_live_p(object: LispObject) -> LispObject {
```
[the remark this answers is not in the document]
brotzeit: Hmmm, I think this should be clear. But I just looked at our first bug issues and wonder how we can make them more appealing.

Where would I document a list that contains frequent c macros/functions and the rust counterparts. I think it's too much for the walkthrough ? 

It would make things a lot easier for newcomers if they could lookup how to use `XSETFASTINT` for example.

rust_src/src/frames.rs
brotzeit: Yeah I also think that's annoying, but I couldn't think of a better solution.

rust_src/src/frames.rs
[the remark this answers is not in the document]
shaleh: I added a porting cookbook to the wiki based on the README. I need to link to it in the README.

rust_src/src/frames.rs
shaleh: One possibility is moving to `Result` returns for lisp_fn. Then we could use `?` to simplify.

`foo.as_string()?` this way we can handle it locally or let the error bubble out. This would also reduce the need for `as_foo_or_error`

birkenfeld: Result returns might be coming in place of `error!()` at some point but are a long way off.

In any case, wherever we're currently using `unwrap()` would stay that way even with Results, since it marks places where internal invariants dictate presence of `Some`. A controlled abort is the correct way to handle a violation.

shaleh:
toggling a new build

shaleh:
Blocked waiting on #511

shaleh:
Toggling the build again.

PR #448, opened 2017-10-31
port minibuffer-window and frame-root-window

shaleh:


shaleh:
@brotzeit this fails make check and I have not had time to debug it. But this shows my refactors.

shaleh:
My failure was `map_or` ran `frame_live_or_selected` for the 'or' case even when it was not needed. I should have used `map_or_else`. But I decided the if/else was easier to read and debug for the long term.

The code now passes make check locally.

rust_src/src/frames.rs
```diff
@@ -1,14 +1,57 @@
...
 pub type LispFrameRef = ExternalPtr<Lisp_Frame>;
 
+impl LispFrameRef {
+    #[inline]
+    pub fn is_live(&self) -> bool {
+        unsafe { fget_terminal(self.as_ptr()) != ptr::null() }
+    }
+}
+
+pub fn frame_or_selected(object: LispObject) -> LispFrameRef {
```
birkenfeld: Add some docstrings?

rust_src/src/frames.rs
```diff
@@ -1,14 +1,57 @@
...
 
 use lisp::{ExternalPtr, LispObject};
 use lisp::defsubr;
 
 pub type LispFrameRef = ExternalPtr<Lisp_Frame>;
 
+impl LispFrameRef {
+    #[inline]
+    pub fn is_live(&self) -> bool {
+        unsafe { fget_terminal(self.as_ptr()) != ptr::null() }
```
birkenfeld: should be using `ptr.is_null()` method instead?

shaleh: Good question. I will have to try it and see what happens.

You are suggesting:

    ! fget_terminal(self.as_ptr()).is_null()

correct?

rust_src/src/frames.rs
shaleh: What could I say that is not obvious from the code? if nil return selected frame else frame

rust_src/src/frames.rs
birkenfeld: yep

PR #450, opened 2017-11-01
Port window-parent

shaleh:


src/window.c
```diff
@@ -218,6 +218,12 @@ wset_update_mode_line (struct window *w)
     w->update_mode_line = true;
 }
 
+Lisp_Object
+wget_parent(struct window *w)
+{
+  return w->parent;
```
birkenfeld: We have `Lisp_Window` defined in remacs-sys -- why this alternate way of accessing the struct member?

shaleh: ifdefs and bitfields. Rust simply does not know enough about the C side. Is libxpm, Xft, whatever in use? Is this a Mac? because of this the struct varies in size so we cannot simply copy all of the field/member names over to Rust. Depending on how emacs is built the field is no longer where you think iti is.

shaleh: basically, the structs we have copied need to be emptied to just the header and these hacky C access methods need to be written. It has to be this way until enough of the code has migrated to Rust. Then we can work out proper solutions.

birkenfeld: Ok, makes sense. Can I then just ask for better names? `wget_parent` doesn't tell me that it's about a window, it makes me wonder if we're trying to fetch an URL :)

shaleh: We did not choose these names. This is the emacs naming standard we are following.

'w' for window, then 'get' then the field name.
'f' is for frame. 'p' is for process.

shaleh: Think of it as incentive to move to Rust quicker :-)

shaleh:
repeat build

PR #451, opened 2017-11-01
Add a no window build and test

shaleh:
It is currently set to ignore failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+---
 language: rust
 cache:
   cargo: true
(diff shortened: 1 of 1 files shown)


shaleh:
Ok then. This version appears to be the right form.

.travis.yml
```diff
@@ -43,7 +27,46 @@ jobs:
...
+        - TRAVIS_ALLOW_FAILURE=1
+      os: linux
+      addons:
+        apt:
+          packages:
+            - texinfo
+      before_script:
+        - if [ $TRAVIS_OS_NAME = linux ]; then ./autogen.sh && ./configure --without-makeinfo --with-x=no --without-gconf --without-gsettings; fi
+        # TODO: add the necessary dependencies so we can use the same configure flags on OS X.
+        - if [ $TRAVIS_OS_NAME = osx ]; then ./autogen.sh && ./configure --without-makeinfo --with-gnutls=no --with-x=no --without-gconf --without-gsettings; fi
```
birkenfeld: gnutls could be installed in this run, too? then the configure options would be the same.

.travis.yml
```diff
@@ -43,7 +27,46 @@ jobs:
...
+      script:
+        - make -j 3 && echo '==> Running nowindow tests' && make check
+    -
+      <<: *NO_WINDOW_SYSTEM_TEST
+      os: osx
+  allow_failures:
+    - env: TRAVIS_ALLOW_FAILURE=1
```
birkenfeld: Will that only refer to the no-window-system tests?

shaleh: Any env that has that variable set. At the moment, that is the no window system. Unless I messed up.

.travis.yml
shaleh: I need to validate that brew actually worked. I have been busy on my other PR :-)

shaleh: /usr/local/Homebrew/Library/Homebrew/brew.rb:12:in `<main>': Homebrew must be run under Ruby 2.3! You're running 2.0.0. (RuntimeError)

And it did not succeed. But it should be possible.

.travis.yml
birkenfeld: Ah, it's checked in that direction. Makes sense.

birkenfeld:
From the `brew install texinfo` output:
```
This formula is keg-only, which means it was not symlinked into /usr/local,
because software that uses TeX, such as lilypond and octave, require a newer
version of these files.
If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/texinfo/bin:$PATH"' >> ~/.bash_profile
For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/texinfo/lib
```

shaleh:
Thank you. Good info.

shaleh:
Grumble grumble.

I have the no windows version working. But I am having issues getting the OSX build to find makeinfo. I have updated the PATH. Anyone?

birkenfeld:
Your `PATH` is missing another double quote?

PR #452, opened 2017-11-01
Add init_modules macro 

shanavas786:


shaleh:
I can take over leave this one. However, I think this would be a great idea for the `defsubr` stanzas at the end of each module.

shaleh:
See #454

shaleh:
@shanavas786 I absorbed this in #456. The `rust_init_syms` function in lib.rs is now generated so I did not need the macro.

shanavas786:
Closing in favor of #456

PR #456, opened 2017-11-02
DRY is always better.

shaleh:


shaleh:
Thank you @shanavas786 your PR gave me the insight I needed.

shaleh:
I tried to also generate the `mod` lines in lib.rs but alas it was not to be.

rust_src/src/buffers.rs
```diff
@@ -360,20 +360,18 @@ fn set_buffer(buffer_or_name: LispObject) -> LispObject {
...
+    buffer_modified_p,
+    buffer_modified_tick,
+    buffer_name,
+    current_buffer,
+    get_buffer,
+    overlay_buffer,
+    overlay_end,
+    overlay_start,
+    overlayp,
+    set_buffer
```
birkenfeld: please put a trailing comma in those lists.

shaleh: It does not compile when I do.

shaleh:
I am trying to get that working now @birkenfeld. This is a proof of concept.

shaleh:
I cannot catch all of them. For instance the floatfns are generated by a macro.

shaleh:
What I almost have working is look for lisp_fn macro, add function name to a vector, write out the vector for `export_lisp_fns`. I like having the macro handle the actual setup of `defsubr` because it means that can be changed by editing the code. The build scripts are not an obvious place to go looking to make changes.

birkenfeld:
> I cannot catch all of them. For instance the floatfns are generated by a macro.

Right, that makes it awkward. Let's keep *some* explicitness :)

shaleh:
@birkenfeld here is something close to the final version. Now all the developer has to do is mark a function with the `lisp_fn` macro and the right things happen.

If someone adds a new file with new lisp_fn calls in it the build will panic until they add the include magic.

shanavas786:
Hope #452 can handle that magic too. :)

brotzeit:
Nice. That's what I actually wanted in #345 ;) @shaleh

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+        let line = line?;
+
+        match parse_state {
+            ParseState::Looking => {
+                if line.starts_with("#[lisp_fn") {
+                    if let Some(begin) = line.find(C_NAME) {
+                        let input = line.to_string();
+                        let start = begin + C_NAME.len();
+                        let end = &input[start..].find('"').unwrap() + start;
+                        let name = (&line[start..end]).to_string();
```
birkenfeld: see above.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+    for line in in_file.lines() {
+        let line = line?;
+
+        match parse_state {
+            ParseState::Looking => {
+                if line.starts_with("#[lisp_fn") {
+                    if let Some(begin) = line.find(C_NAME) {
+                        let input = line.to_string();
+                        let start = begin + C_NAME.len();
+                        let end = &input[start..].find('"').unwrap() + start;
```
birkenfeld: unnecessary `&`.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+    let mut has_include = false;
+
+    for line in in_file.lines() {
+        let line = line?;
+
+        match parse_state {
+            ParseState::Looking => {
+                if line.starts_with("#[lisp_fn") {
+                    if let Some(begin) = line.find(C_NAME) {
+                        let input = line.to_string();
```
birkenfeld: This string copy seems unnecessary.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+            "export_lisp_fns! {{ {} }}",
+            exported.join(", ")
+        )?;
+
+        if !has_include {
+            panic!(
+                [
+                    modname,
+                    ".rs is missing the required include for lisp_fn exports"
+                ].concat()
```
birkenfeld: seeing the concat here is kind of strange. `format!()` is more idiomatic I think.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+            continue;
+        }
+
+        let mut name: Option<String> = None;
+
+        if mod_path.is_dir() {
+            name = Some(
+                mod_path
+                    .file_name()
+                    .and_then(|s| s.to_str())
```
birkenfeld: You've already `unwrap()`ped the `mod_path.file_name()` once above, so handling failure differently here seems unnecessary.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+                );
+            }
+        }
+
+        if let Some(modname) = name {
+            let path = mod_path
+                .to_str()
+                .map_or_else(|| panic!("Cannot understand string"), |s| s.to_string());
+
+            let fp = match File::open(Path::new(&path)) {
```
birkenfeld: Constructing a new `Path` (and a `String` before) is unnecessary, `open` should get the `mod_path` directly.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+                Ok(f) => f,
+
+                Err(e) => {
+                    eprintln!("Failed to open {}", path);
```
birkenfeld: Then use `mod_path.display()` here.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+    Looking,
+    IgnoreComments,
+    NoMangleFn,
+    LispFn(Option<String>),
+}
+
+fn get_function_name(line: &str) -> Option<String> {
+    if let Some(pos) = line.find('(') {
+        if let Some(fnpos) = line.find("fn ") {
+            return Some((&line[(fnpos + 3)..pos]).to_string());
```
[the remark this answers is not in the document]
shaleh: Would not compile without the parens last I looked. I will give it another go.

Good call on the trim.

rust_src/build.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    None
+}
+
+static C_NAME: &'static str = "c_name = \"";
```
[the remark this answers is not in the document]
shaleh: Thanks

rust_src/build.rs
shaleh: It was complaining about temporaries and borrowing if I recall.

rust_src/build.rs
shaleh: I think this was the first code. I will give it a once over.

rust_src/build.rs
shaleh: Hmm, again with borrow complaints. I will see.

shaleh:
@birkenfeld thanks for the sanity checking.

shaleh:
I think we do. defsubr may need tweaks. Ideally that happens in tree and not in the build logic.

I know there is a Windows side fix where the `&*` becomes `as_ptr`.
-- 

shaleh:
@brotzeit @Wilfred @shanavas786 any further comments before I merge this?

birkenfeld:
LGTM.

birkenfeld:
Argh, sorry for creating more conflicts. I'll hold off merging anything else until this is in.

shaleh:
The conflicts will end up somewhere :-) no worries.

shaleh:
This is a big change. I would feel better if I had some more +1s.

brotzeit:
I think you should merge this, so we can start rebasing the other PRs. If there should be problems we can still fix them. But this is definitely a big step, good job !

rust_src/src/lisp.rs
```diff
@@ -429,6 +431,7 @@ impl LispObject {
             wrong_type!(Qvectorp, self)
         }
     }
+    */
```
Wilfred: A belated +1 from me: it's nice to see the code get less verbose.

Why did you comment out code though? Can we delete it?

shaleh: I expected they would get used. It can go away if that turns out to be a false idea

PR #460, opened 2017-11-04
Add `rust-toolchain` file.

shaleh:
This documents the version of the toolchain that is know to compile
the tools we use as well as remacs.


shaleh:
Probably. I can look at that.

shaleh:
As committed this should do what we want:

    $ cargo new foo --bin
    $ cd foo
    $ echo 'nightly-2017-09-01' > rust-toolchain
    $ cargo build
    error: override toolchain 'nightly-2017-09-01' is not installed
    info: caused by: the toolchain file at 'foo/rust-toolchain' specifies an uninstalled toolchain

shaleh:
@Wilfred I updated the Travis config to use the version from the toolchain file. So now the only version in the config is the rustfmt version.

In case you did not notice, the OSX and Linux builds are now using the same configure options. I only expect build stages to improve.

shaleh:
Caching is enabled so the tests should setup fast. We shall see.

.travis.yml
```diff
@@ -56,20 +58,25 @@ jobs:
...
+        - travis_wait rustup install $(cat rust_src/rust-toolchain)
+        - rustup default $(cat rust_src/rust-toolchain)
         # Travis docs say to avoid brew update, but it works around the incorrect Ruby version
         - brew update
         - brew install gnutls
         - brew install texinfo
-  allow_failures:
-    - env: TRAVIS_ALLOW_FAILURE=1
+  #allow_failures:
+  #  - env: TRAVIS_ALLOW_FAILURE=1
```
Wilfred: Can we remove this?

shaleh: I left it in to show people how to enable it is we need it in the future. With the stages system it took me a bit to figure out how to allow for expected failure. The current Travis docs are not written for stages.

I can remove it or leave it. Do you want more comments? Move it out to a comment block? Put it in a document like say contributing?

Wilfred: I'd prefer that we remove it from here -- there's a risk that commented code bitrots. Putting in the contributing guidelines sounds like a great idea.

shaleh: Doc updated.

PR #461, opened 2017-11-05
port nsberror

brotzeit:


rust_src/src/buffers.rs
```diff
@@ -360,4 +361,17 @@ fn set_buffer(buffer_or_name: LispObject) -> LispObject {
     buffer
 }
 
+/// No such buffer error.
+#[no_mangle]
+pub fn nsberror(spec: LispObject) {
+    if let Some(s) = spec.as_string() {
+        error!(
+            "No buffer named {}",
+            std::str::from_utf8(s.as_slice()).unwrap()
```
shaleh: Should that be made a function or method?

brotzeit: I don't think this is a good fit for a method. We have our *_or_error methods and that would maybe confusing.

shaleh: This code says "s is a string, then slice it, turn it into utf8. Trust me" Which is a whole lot of 's is kinda but not really a string'.

birkenfeld: Especially since as an Emacs string it might *not* be utf8.

brotzeit: So I use `from_utf8_lossy` ?

birkenfeld: That could work, yes. But really, in the end you'd want to insert the bytes literally, so that Emacs on the other end can display them as intended.

brotzeit: It should be forbidden to use multibyte chars for buffer names ;)
I'll try to fix this.

brotzeit: You mean multibyte chars in buffer names ?

shaleh: If birkenfeld is write that non UTF8 can slip through there should be a test for it.

brotzeit: You're right, but wouldn't it be better to simply not allow creating those buffers ? And I already see the answer to my mail "The docstring does not say it's not allowed to use non UTF8 chars, so it was probably intended that way" ;)

brotzeit: Maybe it would also be good to deny the creation of non UTF8 buffer names.

birkenfeld: That is probably impossible since buffer names may be allowed to be unibyte strings (haven't checked this).

brotzeit:
I don't know a better way to do this.

rust_src/src/buffers.rs
```diff
@@ -360,4 +360,25 @@ fn set_buffer(buffer_or_name: LispObject) -> LispObject {
...
+        ];
+        xsignal!(
+            ::remacs_sys::Qerror,
+            LispObject::from(Fconcat(
+                argsarray.len() as ptrdiff_t,
+                argsarray.as_mut_ptr(),
+            ))
+        );
+    } else {
+        error!("Invalid buffer argument");
```
shaleh: My preference would be to see this if/else inverted. That way the code exits early and the rest of the function is not indented in an if.

rust_src/src/buffers.rs
```diff
@@ -360,4 +360,25 @@ fn set_buffer(buffer_or_name: LispObject) -> LispObject {
...
+/// No such buffer error.
+#[no_mangle]
+pub fn nsberror(spec: LispObject) {
+    if spec.is_string() {
+        let msg = "No buffer named ";
+        let mut argsarray = [
+            unsafe { make_string(msg.as_ptr() as *const c_char, msg.len() as ptrdiff_t) },
+            spec.to_raw(),
+        ];
+        xsignal!(
```
shaleh: Why `xsignal` here when the original called `error` for both cases? You have re-created the `error!` macro.

brotzeit: Yeah, I asked in our gitter room a few days ago. And @db48x suggested "well, in that case, error! probably should never call format!". 

You said "@brotzeit what @db48x suggests is also basically what display_string in lisp.rs does. It makes the slice using byte length.".
The problem with `display_string` is that it also makes use of `from_utf8_lossy`, which doesn't work with multi and unibyte chars.

rust_src/src/buffers.rs
brotzeit: Ok.

rust_src/src/buffers.rs
brotzeit: I think `error!` isn't the problem. It has to be possible to cast a multi/unibyte string to rust str, but I just don't know how...

shaleh: The C code has not cared. I say port nsberror as written. Then open an issue to add more tests. If we can make the code break then we fix it.

In the C code `nsberror` is just a wrapper around `error`. So it makes sense to fix `error` when and if we find strings that fail there.

db48x: Yea, don't rewrite error!; it's the right tool for the job. I think we will need to implement our own string type to handle Emacs' internal encoding, but that can be done in a different bug. (See section 22.2 of the Emacs manual and section 32.1 of the Elisp manual.)

shaleh: :+1:

brotzeit: So I add a TODO note and we merge or wait until we have a better solution ?

shaleh: add a note and merge

brotzeit: Done.

brotzeit:
It seems there is no `%s` in rust and `std::fmt` doesn't work with pointers.

db48x:
I don't think that code was meant to be copied verbatim. It seems instead to be a sketch to indicate what he meant. The code you originally wrote was fine; I would simply use `git reflog` to retrieve it, then add a commit to it with the comment.

brotzeit:
So `(set-buffer (string-make-unibyte "ε"))` segfaults on remacs and not on emacs, as rust `str` is only 4 bytes instead of 5 ?

brotzeit:
I guess this is a case that can be ignored, but I don't like the idea that remacs segfaults and emacs doesn't...

shaleh:
We know we have string issues. As I said at the beginning, start writing tests. They can be ignored for now. But it will help once the next step starts.

birkenfeld:
We could also, you know, just *not* port `nsberror` for now :)

shaleh:
Well that rebase went poorly.... Sigh.

rust_src/src/buffers.rs
```diff
@@ -428,4 +428,15 @@ pub fn barf_if_buffer_read_only(position: Option<EmacsInt>) -> () {
     }
 }
 
+/// No such buffer error.
+#[no_mangle]
+pub extern "C" fn nsberror(spec: Lisp_Object) -> ! {
+    let spec = LispObject::from_raw(spec);
+    if let Some(s) = spec.as_string() {
+        error!("No buffer named {}", s);
```
shaleh: Yes, this calls fromtutf8_lossy behind the scenes. But frankly we have similar issues elsewhere so I see no reason to hold up the PR for it.

shaleh:
Code no longer segfaults.

brotzeit:
> Code no longer segfaults.

I can confirm that. So I will merge it. If I understood correctly, we have to change the encoding stuff anyway.

PR #467, opened 2017-11-08
Port lookup-key and lucid_event_type_list_p

shaleh:
Along the way, add a message_with_string macro, and an Iterator that walks
the cars of a cons object.

### DIFF
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -1,8 +1,12 @@
 //! Keymap support
 
-use lisp::{defsubr, LispObject};
 use remacs_macros::lisp_fn;
-use remacs_sys::{current_global_map as _current_global_map, Flookup_key};
+use remacs_sys::{current_global_map as _current_global_map, EmacsInt};
+use remacs_sys::{access_keymap, get_keymap, maybe_quit, Faref, Fevent_convert_list};
+use remacs_sys::CHAR_META;
+
+use keyboard::lucid_event_type_list_p;
+use lisp::{defsubr, LispObject};
 use threads::ThreadState;
 
 
@@ -17,13 +21,7 @@ fn local_key_binding(keys: LispObject, accept_default: LispObject) -> LispObject
     if map.is_nil() {
         LispObject::constant_nil()
     } else {
-        unsafe {
-            LispObject::from(Flookup_key(
-                map.to_raw(),
-                keys.to_raw(),
-                accept_default.to_raw(),
-            ))
-        }
+        lookup_key(map, keys, accept_default)
     }
 }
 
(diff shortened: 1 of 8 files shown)


birkenfeld:
Why is `CarIter` needed at all? `TailsIter` already yields conses, which you only need to call `.car()` on.

shaleh:
@birkenfeld this is why I included code. Look at the change in `posn_at_point`.

As for TailsIter, I would need to special case to get the first car. So the code would:

make sure is cons
get car
make iter
walk rest

shaleh:
Or how if what you want is every item in the list now there is a way to do it cleanly. I was able to use the All property of Iterators to reduce code to a simple `if it.all(predicate)` that used to be a bramble of if, as, get, unwrap.

birkenfeld:
@shaleh it's clear that the iterator form is much nicer. What I'm saying is that TailsIter already does what you need, just add `.car()`. No need to special case the first element. Plus, it already has detection of circular lists.

shaleh:
There is code all over the C tree like this;

    for (Lisp_Object tail = fonts; CONSP (tail); tail = XCDR (tail))
        {
        foo = XCAR(tail)
        }

No circular checks to be seen.

Yes, I know this looks like it shows your point about TailsIter and car. But I made the CarIter because it does not make ANY checks. If the item is not a cons the iter immediately returns None. If it hits a non cons cell it just stops. This is very different from the TailsIter. Because of this it is simple and fast to for the above for each loops. Just like it is in the C code. Both Iterators have a use. This is a clear following of Rust norms. I can put this in a block comment above the two iterators or I can add it to the contributing document. But there are a fair number of these for each loops left to port and it was my impression the goal of the Rust port was to make better code using Rust's expressiveness. The CarIter allows this more in more places than the TailsIter does.

birkenfeld:
> There is code all over the C tree like this;

These instances should probably be converted to the version with checks. Infinite loops in Emacs functions are pretty bad, especially when nothing calls `maybe_quit` inside that gives the user a chance to `C-g`.

> If the item is not a cons the iter immediately returns None.  If it hits a non cons cell it just stops. This is very different from the TailsIter.

Not at all, that's what the `iter_tails_safe` variant does.

Look, if you insist on having a variant that yields cars instead of conses, fair enough (although I still think that's of questionable advantage), but it's pretty easy to base it on TailsIter and profit from the added infinite loop safety.

rust_src/src/keymap.rs
```diff
@@ -40,4 +38,74 @@ fn current_global_map() -> LispObject {
...
+    let length = key.as_vector_or_string_length() as EmacsInt;
+    if length == 0 {
+        return LispObject::from(keymap);
+    }
+
+    let mut idx = 0;
+    loop {
+        let mut c = LispObject::from(unsafe {
+            Faref(key.to_raw(), LispObject::from_fixnum(idx).to_raw())
+        });
```
brotzeit: Would it be better to declare the variable binding outside of the loop ?

shaleh:
@birkenfeld thank you for the patience and explanations. I completely misread `TailsIter` the first time. I thought it was returning the cdr not the entire cons. Your position makes a lot more sense.

I updated the PR to have `CarIter` be a simple adapter over `TailsIter`. This should be a good middle ground. No repeated code but it still enables clean code where only the car is needed. Fair?

I prefer:

    if !it.all(|elt| elt.is_fixnum() || elt.is_symbol()) {
        return false;
    }

over

    if !it.all(|elt| elt.car().is_fixnum() || elt.car().is_symbol()) {
        return false;
    }

shaleh:
Triggering a rebuild

birkenfeld:
> Fair?

Fair! Looks nice now.

shaleh:
What about my use of the name `current` to mimic `Peekable`? Did I miss an idiom?

rust_src/src/lisp.rs
```diff
@@ -745,6 +763,30 @@ impl Iterator for TailsIter {
...
+    pub fn current(&self) -> LispObject {
+        self.tails.tail
+    }
+}
+
+impl Iterator for CarIter {
+    type Item = LispObject;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.tails.next().and_then(|c| Some(c.car())).or(None)
```
birkenfeld: That looks like it should just be `.map(|c| c.car())`?

rust_src/src/lists.rs
```diff
@@ -98,19 +98,13 @@ fn cdr_safe(object: LispObject) -> LispObject {
 #[lisp_fn]
 pub fn nthcdr(n: LispObject, list: LispObject) -> LispObject {
     let num = n.as_fixnum_or_error();
-    let mut tail = list;
+    let mut it = list.iter_tails();
```
birkenfeld: The whole rest of the function is now just `it.nth(num)`?

rust_src/src/lisp.rs
```diff
@@ -699,6 +711,12 @@ impl TailsIter {
         }
     }
 
+    pub fn current(&self) -> LispObject {
```
birkenfeld: Hmm, `current` is not an intuitive name for this. Mabye `rest`?

shaleh: Hmmm, maybe.

rust_src/src/lisp.rs
shaleh: Yes. I was not sure which read better.

rust_src/src/lists.rs
shaleh: Sounds right

shaleh:
Sorry for the messy PRs all. I had accidentally pulled from the wrong direction and it looks fine locally. I have updated them all.

rust_src/src/lists.rs
```diff
@@ -97,27 +97,21 @@ fn cdr_safe(object: LispObject) -> LispObject {
...
 }
 
 /// Return the Nth element of LIST.
 /// N counts from zero.  If LIST is not that long, nil is returned.
 #[lisp_fn]
 fn nth(n: LispObject, list: LispObject) -> LispObject {
-    car(nthcdr(n, list))
+    list.iter_cars()
+        .nth(n.as_fixnum_or_error() as usize)
+        .map_or(LispObject::constant_nil(), |c| c)
```
Wilfred: Nice refactoring :)

PR #468, opened 2017-11-08
Remove SYMBOL_NAME

shaleh:


rust_src/src/strings.rs
```diff
@@ -33,24 +33,13 @@ fn string_bytes(string: LispObject) -> LispObject {
...
     LispObject::from_bool(
-        s1.len_chars() == s2.len_chars() && s1.len_bytes() == s2.len_bytes() && unsafe {
-            libc::memcmp(
-                s1.data_ptr() as *mut c_void,
-                s2.data_ptr() as *mut c_void,
-                s1.len_bytes() as usize,
-            ) == 0
-        },
+        s1.len_chars() == s2.len_chars() && s1.len_bytes() == s2.len_bytes()
+            && s1.chars().zip(s2.chars()).all(|(c1, c2)| c1 == c2),
```
birkenfeld: That translation decodes the individual characters, which is much slower than the `memcmp`, and unnecessary here. But we have `as_slice`, so you should be able to compare the two slices without requiring `unsafe`.

BTW, this would otherwise be written with `Iterator::eq`.

birkenfeld: Also, comparing slices will make the `len_bytes()` check unnecessary.

shaleh: I was not sure I could trust Iterator::eq, I intended to test it.

Slice comparison does make more sense. I started off mimicking `string-lessp` and when I realized I did not need to I did not come back to this. Thanks for the catch.

shaleh:
hmm, `rebase -i` did not really clear up the branch history.

shaleh:
@birkenfeld mind flipping your review?

birkenfeld:
Not yet, sorry.

rust_src/src/strings.rs
```diff
@@ -33,25 +33,11 @@ fn string_bytes(string: LispObject) -> LispObject {
...
-                s2.data_ptr() as *mut c_void,
-                s1.len_bytes() as usize,
-            ) == 0
-        },
-    )
+pub fn string_equal(s1: LispObject, s2: LispObject) -> LispObject {
+    let s1 = LispObject::symbol_or_string_as_string(s1);
+    let s2 = LispObject::symbol_or_string_as_string(s2);
+
+    LispObject::from_bool(s1.as_slice() == s2.as_slice())
```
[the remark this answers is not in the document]
birkenfeld: Hmm, or compare `s1.is_multibyte() == s2.is_multibyte()`, which should avoid the need for a comment.

shaleh:
@birkenfeld thanks for the catch. This is why code review is a good thing.

shaleh:
Try build again

shaleh:
@birkenfeld the `is_multibyte` check did not work, tests failed. So I have put back in the length based checks that were there originally but kept the `as_slice` comparison.

birkenfeld:
Strange, but I don't have time to debug that right now. However, the `len_bytes` check is still useless because that's the length of the slice.

shaleh:
I left the size checks in place because the 2 length checks are more efficient than the slice comparison.

birkenfeld:
> I left the size checks in place because the 2 length checks are more efficient than the slice comparison.

Did you measure that? Or otherwise, what is the reason you think that?

shaleh:
No, I have not measured. Intuitively, simple numeric comparison should be more efficient than comparing the in memory representation of the strings.

birkenfeld:
Ah, but slices know their length, so the first thing in their comparison method is to check the lengths. Checking it twice will be optimized away at best.

[https://github.com/rust-lang/rust/blob/master/src/libcore/slice/mod.rs#L2531](It's actually quite easy to find in libcore.)

shaleh:
Oh agreed, I should remove the bytes len.

PR #470, opened 2017-11-10
Rollup till nov 9

shanavas786:
Managed to resolve conflicts, But build fails. Help, Help :)

@brotzeit Copied #425 to this. Could not cherry pick due to conflicts.


shaleh:
I am away for the weekend starting as soon as I close my laptop lid :-) I will pitch in either Sunday or Monday.

shaleh:
OH. This is easy. You need to run `cargo fmt` in rust_src. The math.rs file is not properly formatted. The imports are not orderer properly.

shaleh:
What can we do to make rollups easier?

Would preserving the original C code by wrapping ported functions in `#if 0 ... #endif` help?

brotzeit:
@shanavas786 No problem and thanks for doing this!

birkenfeld:
> Would preserving the original C code by wrapping ported functions in #if 0 ... #endif help?

That would be counterproductive. A conflict in a ported function is useful, because it shows you need to check whether the change needs to be ported.

Only the conflicts in stuff we removed (mostly changelogs and MSDOS support) are a nusiance.

shanavas786:
Found a regression. `(min 1.0 0.0e+NaN)` returns `0.0e+NaN` in emacs and `1.0` in remacs

shanavas786:
Hmm, It took around 2 hours to find that misnamed variable :-|

shanavas786:
Some tests fails on travis. All tests pass locally

Wilfred:
Awesome! Rollup always requires some tricky investigation.

The only test failure seems to be this:

```
Test perl-backend condition:
    (ert-test-failed
     ((should
       (eq 'flymake-error
	    (face-at-point)))
      :form
      (eq flymake-error flymake-warning)
      :value nil))
   FAILED  4/8  perl-backend
   passed  5/8  recurrent-backend
   passed  6/8  ruby-backend
   passed  7/8  warning-predicate-function-gcc
   passed  8/8  warning-predicate-rx-gcc

Ran 8 tests, 6 results as expected, 1 unexpected, 1 skipped (2017-11-10 18:37:14+0000)

1 unexpected results:
   FAILED  perl-backend
```

There's been some discussion on emacs-devel about updating flymake, trying to bring its featureset closer to flycheck. `perl-backend` is a test that was modified only recently: https://github.com/emacs-mirror/emacs/commit/58bb3462ee5

I suspect that we're missing perl or using a different version than the test suite expects?

shanavas786:
I am running perl v5.24.1 in my system.

And travis log has the following.
```
perlbrew list
  5.22 (5.22.0)
  5.22-extras (5.22.0)
  5.22-shrplib (5.22.0)
  5.22.0
  5.24-extras (5.24.0)
  5.24-shrplib (5.24.0)
  5.24 (5.24.2)
  5.24.2
```

Not sure which is the default.

shaleh:
What is the procedure for rollups? Are you using admin/gitmerge.el?

shanavas786:
On November 13, 2017 9:31:21 PM GMT+03:00, Sean Perry <notifications@github.com> wrote:
>What is the procedure for rollups? Are you using admin/gitmerge.el?

No, As remacs is a fork, I just added a remote to Emacs's repo, merged emacs/master to remacs/master and resolved conflicts.
-- 
ഷാനവാസ്
Sent from my Android device with K-9 Mail.

shaleh:
I can reproduce the Perl failure in docker by using trusty instead of xenial. Trusty is what Travis is using. I am investigating the perl failure currently.

shaleh:
I think this boils down to a perl version issue. My Trusty version in docker has perl 5.18. The file that flymake checks is test/lisp/progmodes/flymake-resources/test.pl. If you run `perl -W test.pl` the flymake test expected the output to be a warning followed by an error. But 5.18 emits a suggested after the first warning:

    $ perl -W test/lisp/progmodes/flymake-resources/test.pl
    Unquoted string "unknown" may clash with future reserved word at test/lisp/progmodes/flymake-   resources/test.pl line 2.
    Scalar value @arr[1] better written as $arr[1] at test/lisp/progmodes/flymake-resources/test.pl line 3.
    syntax error at test/lisp/progmodes/flymake-resources/test.pl line 4, at EOF
    Missing right curly or square bracket at test/lisp/progmodes/flymake-resources/test.pl line 4, at end of line
    Execution of test/lisp/progmodes/flymake-resources/test.pl aborted due to compilation errors.

shaleh:
This minor twiddle to the Perl code makes the test run:

    diff --git a/test/lisp/progmodes/flymake-resources/test.pl b/test/lisp/progmodes
    index 6f4f1cc..df71a87 100644
    --- a/test/lisp/progmodes/flymake-resources/test.pl
    +++ b/test/lisp/progmodes/flymake-resources/test.pl
    @@ -1,4 +1,4 @@
     @arr = [1,2,3,4];
     unknown;
    -my $b = @arr[1];
     [
    +my $b = @arr[1];

What this does is move the line with the suggestion after the error for the ERT code finds what it expects in the order it expects. We should probably send this back upstream.

shaleh:
Looking upstream, the current version of the test code moves to the end of the perl file and looks backwards. So that was their solution.

shaleh:
All @shanavas786 has to do is grab the commit that updates flymake-tests.el from upstream as a cherry pick. https://github.com/emacs-mirror/emacs/commit/0da08f2f8ef1946978f0974e9e8cdb87fece018c

shanavas786:
@shaleh Thanks a lot :).

src/gnutls.c
```diff
@@ -26,7 +26,7 @@ along with GNU Emacs.  If not, see <https://www.gnu.org/licenses/>.  */
 #include "coding.h"
```
shanavas786: gnutls test are faling. though all passes locally.

shaleh: I can reproduce that in a Trusty docker. Looking at it now.

shaleh: GNUTLS version numbering is the culprit.This commit https://github.com/emacs-mirror/emacs/commit/583995c62dd424775dda33d5134ce04bee2ae685 is what broke things.

The build checks for gnutls but does not enforce version >= 3. Trusty has 2.6 and that is what our docker build is running. So the elisp tests assume that GNUTLS works because `gnutls-available-p` is true but some of the functions are only in version 3 and there is no check for the supported version in the elisp. This is a bug upstream. Either the configure should fail or it should detect the GNUTLS version and let the user know their version is no longer supported. Barring this, the elisp code should have the ability to detect the version supported or at the very least check if the v3 functions are defined before using them.

shanavas786:
Great!!!,

Is this bug known upstream ?.

We can install v3 in Travis as a temporary fix until this bug gets resolved. right ?
-- 
ഷാനവാസ്
Sent from my Android device with K-9 Mail.

shaleh:
No we cannot install the newer gnutls. It is not an available package.

I do not know if the bug is known upstream. I guess not since it is 4 months old. We can work around it by either modifying the tests to check for the v3 API functions using `fboundp` or by updating the gnutls support to export a 'gnutls-v3-supported-p` and wrapping the test logic in that predicate. Either solution can be sent upstream as well.

shanavas786:
Well, https://lists.gnu.org/archive/html/bug-gnu-emacs/2017-11/msg00146.html gave me some insights. It seems [emacs-mirror@8c08e4](https://github.com/emacs-mirror/emacs/commit/c572e1f329583d40a6ae1a6cf493aba4a58c08e4) broke the tests.

shaleh:
Status?

shanavas786:
I played with `gnutls3-available-p`. But did not help much.

A little busy these days. Hope I would get some time this weekend. I can give access to my repo if any one likes to work on it.
-- 
ഷാനവാസ്
Sent from my Android device with K-9 Mail.

shaleh:
Don't sweat it. I read the rest of the test code. They actually do the right thing later in the code. So all we have to do is follow suit.

```
(when (gnutls-available-p).....
```

becomes

```
(when (memq 'feature (gnutls-available-p))....
```

where 'feature' is 'macs', 'ciphers', or 'digests'.

shaleh:
Should we close this?

shanavas786:
Done

PR #473, opened 2017-11-13
bump versions

brotzeit:


shaleh:
Can this wait for my `rust-toolchain` PR? That provides a mechanism to keep the Rust version in sync for tests and developers. Sadly there is no mechanism for `rustfmt`.

shaleh:
#460 has landed.

PR #474, opened 2017-11-13
Add elisp package for remacs developers

shaleh:


rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,34 @@
+;;; Package -- helpers for Remacs porting
+
+(defun remacs-helpers/make-rust-args-from-C-worker (input)
+  "Transform function C arguments INPUT into Rust style arguments."
+  (mapconcat (lambda (arg) (let* ((is-struct (lambda (s) (string= s "struct")))
+                                  (pieces (cl-remove-if is-struct (split-string (string-trim arg) " ")))
+                                  (name (s-append ":" (car (last pieces))))
+                                  (rest (butlast pieces)))
+                             (if (s-starts-with? "\*" name)
```
birkenfeld: the backslash is confusing, this is not a regex

rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,34 @@
...
+                                 (s-join " " (cons (s-chop-prefix "*" name)
+                                                   (cons "*" rest)))
```
birkenfeld: `*` -> `*mut`?

rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,34 @@
+;;; Package -- helpers for Remacs porting
```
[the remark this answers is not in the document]
shaleh: Yes it should.

rust_src/admin/remacs-helpers.el
shaleh: I wanted to be sure since the `s-*` functions try regex too

rust_src/admin/remacs-helpers.el
shaleh: I am ok with defaulting to mut or const.

shaleh:
This started as a macro I had defined and saved. I am open to improvements for sure.

shaleh:
@birkenfeld thank you for the review. My elisp skills are clunky at best.

shaleh:
Next up is one to convert simple enums or structs. Any takers :-)

shaleh:
Or wrap this logic and convert the function with return value.

rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,34 @@
...
+
+(defun remacs-helpers/make-rust-args-from-C-worker (input)
+  "Transform function C arguments INPUT into Rust style arguments."
+  (mapconcat (lambda (arg) (let* ((is-struct (lambda (s) (string= s "struct")))
+                                  (pieces (cl-remove-if is-struct (split-string (string-trim arg) " ")))
```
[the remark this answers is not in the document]
shaleh: I used `cl-remove-if` because it allowed the list of ellided words to grow. I could pull `is-struct` out, rename it to "ignored-type-modifiers" and use that.

rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,34 @@
...
+                                  (name (s-append ":" (car (last pieces))))
+                                  (rest (butlast pieces)))
+                             (if (s-starts-with? "\*" name)
+                                 (s-join " " (cons (s-chop-prefix "*" name)
+                                                   (cons "*" rest)))
+                               (s-join " " (cons name rest)))))
+             (split-string input ",")
+             ", "))
```
Wilfred: If you're using s.el, might I suggest using dash.el as well? `->>` is lovely, and lets you do things like this:

```emacs-lisp
(defun remacs-helpers/make-rust-args-from-C-worker (input)
  "Transform function C arguments INPUT into Rust style arguments.

For example, convert this:
int *foo, struct Bar b

into this:
foo: * int, b: Bar"
  (->> input
       (s-split ",")
       (-map #'s-trim)
       (--map (s-chop-prefix "struct " it))
       (--map (s-split " " it))
       (-map (-lambda ((type name))
               (if (s-starts-with-p "*" name)
                   (format "%s: * %s" (s-chop-prefix "*" name) type)
                 (format "%s: %s" name type))))
       (s-join ", ")))
```

rust_src/admin/remacs-helpers.el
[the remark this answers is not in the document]
shaleh: Highlight region, run command.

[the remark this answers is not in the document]
shaleh: I'd like to get it the point it works on the whole function definition. Then it could work from point.

rust_src/admin/remacs-helpers.el
shaleh: Neat. I was not familiar with dash.

rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,34 @@
...
+         (input (if stringInputP string (buffer-substring-no-properties from to)))
+         (output (remacs-helpers/make-rust-args-from-C-worker input)))
+    (if stringInputP
+        output
+      (save-excursion
+        (delete-region from to)
+        (goto-char from)
+        (insert output) )) ) )
+
+(provide 'remacs-helpers)
```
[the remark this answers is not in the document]
shaleh: Right. I ran a linter, will have updates.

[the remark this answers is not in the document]
brotzeit: Can you remove the whitespace between parentheses.

rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,52 @@
...
+  (->> input
+       (s-split ",")
+       (-map #'s-trim)
+       (--map (-remove 'remacs-helpers/ignored-type-part-p (s-split " " it)))
+       (-map (-lambda ((type name))
+               (if (s-starts-with-p "*" name)
+                   (format "%s: *mut %s" (s-chop-prefix "*" name) type)
+                 (format "%s: %s" name type))))
+       (s-join ", ")))
+
```
brotzeit: This doesn't work if the argument contains the `register` keyword.

rust_src/admin/remacs-helpers.el
```diff
@@ -0,0 +1,52 @@
...
+(defun remacs-helpers/make-rust-args-from-C (string &optional from to)
+  "Transform provided STRING or region indicated by FROM and TO into Rust style arguments."
+  (interactive
+   (if (use-region-p)
+       (list nil (region-beginning) (region-end))
+     (let ((bds (bounds-of-thing-at-point 'paragraph)) )
```
brotzeit: I think you need `(require 'thingatpt)` for `bounds-of-thing-at-point`.

rust_src/admin/remacs-helpers.el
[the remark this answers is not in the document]
shaleh: This is why I don't code in elisp. The style is in opposition to enjoyment.
Yeah, I will update.

rust_src/admin/remacs-helpers.el
shaleh: Yup. That is why I have the predicate to drop parts....

rust_src/admin/remacs-helpers.el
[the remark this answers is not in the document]
brotzeit: I don't think it's that bad. But it would be nice to have better formatting tools. Do you use paredit or something similiar ?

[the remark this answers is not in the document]
shaleh: My personal style is more in line with modern languages. I put end parens on their on line like braces.

shaleh:
Ok, this is working and more complete. At this point rather than nickel and dime minor changes I would like to see it merged and worked on further.

shaleh:
I added the setup and run for ERT because of the dash and s depends. If there is a better way let me know.

shaleh:
Only ugly piece is `void *foo`. There are some things I can't automagic :-)

rust_src/admin/remacs-helpers-tests.el
```diff
@@ -0,0 +1,21 @@
+;;; Code:
+
+(ert-deftest transmute-type-test ()
```
Wilfred: ert-deftest should be indented the same amount as defun.

If you execute `(require 'ert)` in your current Emacs instance, it should do the right thing :)

shaleh: I will give it a go.

PR #476, opened 2017-11-16
Port posn-at-x-y

shaleh:


shaleh:
Thoughts on `tool_bar_p` and `menu_bar_p`? Should I have made these using a `is_` form instead?

rust_src/src/keyboard.rs
```diff
@@ -60,14 +62,52 @@ pub fn posn_at_point(pos: LispObject, window: LispObject) -> LispObject {
...
+        x += w.left_edge_x();
+
+        if whole.is_nil() {
+            x += unsafe { window_box_left_offset(w.as_ptr(), TEXT_AREA) };
+        }
+
+        y = w.frame_pixel_y(y);
+    });
+
+    return LispObject::from(unsafe {
```
Wilfred: Return is unnecessary here.

rust_src/src/windows.rs
```diff
@@ -107,6 +117,43 @@ impl LispWindowRef {
...
 
+    /// The frame x-position at which the text (or left fringe) in
+    /// window starts. This does not include a left-hand scroll bar
+    /// if any.
+    #[inline]
+    pub fn left_edge_x(self) -> i32 {
+        self.frame().as_frame_or_error().internal_border_width() + self.left_pixel_edge()
+    }
+
+    /// The frame y-position at which window starts.
```
Wilfred: *at which the window

Likewise for the other docstrings.

src/keyboard.c
```diff
@@ -10714,47 +10714,6 @@ The elements of this list correspond to the arguments of
...
-    IMAGE (DX . DY) (WIDTH . HEIGHT))
-The `posn-' functions access elements of such lists.  */)
-  (Lisp_Object x, Lisp_Object y, Lisp_Object frame_or_window, Lisp_Object whole)
-{
-  CHECK_NUMBER (x);
-  /* We allow X of -1, for the newline in a R2L line that overflowed
-     into the left fringe.  */
-  if (XINT (x) != -1)
-    CHECK_NATNUM (x);
-  CHECK_NATNUM (y);
```
Wilfred: This errors if x is -2, and I don't think that's true of the Rust code.

src/window.c
```diff
@@ -239,6 +239,31 @@ wget_pseudo_window_p(struct window *w)
...
+          && (W) == XWINDOW (WINDOW_XFRAME (W)->menu_bar_window));
+#else
+/* No menu bar windows if X toolkit is in use.  */
+  return false;
+#endif
+}
+
+/* True if W is a tool bar window.  */
+bool
+WINDOW_TOOL_BAR_P(struct window *W)
```
Wilfred: I think the name is fine, but I'd use lower case now they're functions.

src/keyboard.c
shaleh: Good catch.

shaleh: I am curious how my solution will be received. I also considered and experimented with extending `LispNumber` but that  means constantly verifying the value is not a float. Which is no better.

shaleh:
Toggle build.

shaleh:
toggling the build

src/window.c
Wilfred: The C code only uses upper case for macros, doesn't it?

shaleh:
Fair points @Wilfred. New commit on its way.

PR #479, opened 2017-11-17
Set EMACS_CLASS to Remacs

shaleh:


birkenfeld:
Do we want this? It'll make some WM customizations made for Emacs not apply to Remacs, which is not ideal.

shaleh:
Are you aware of a window manager in common use that specifically looks for emacs? Typically window managers do this to work around bugs in applications. We should be resolving those issues not continuing to paper over them.

If a user has made customizations based on emacs's class those are easy fix. They either cut & paste their config or click the window.

today having remacs claim to be emacs causes issues when you run both of them. Which I do.

My opinion is we set the WM_CLASS and resolve any bugs. If the problems are too many or annoying we can always revert this.

birkenfeld:
> today having remacs claim to be emacs causes issues when you run both of them. Which I do.

What issues are those?

shaleh:
From my commit message:

    In X, the WM_CLASS is set to 'Remacs'. This enables alt-tab on Linux
    to differentiate between emacs and remacs.

I never see the Remacs red icon because I have C emacs running. There are these little issues such as alt-tab.

PR #486, opened 2017-11-18
use bindgen to generate most of remacs-sys at build-time

db48x:
This saves us most of the work of finding the declarations of C types
and functions and converting them into Rust. As the C side of Remacs
shrinks, the generated bindings will shrink.


rust_src/src/category.rs
```diff
@@ -11,7 +11,7 @@ use threads::ThreadState;
 #[lisp_fn]
 fn category_table_p(arg: LispObject) -> LispObject {
     LispObject::from_bool(arg.as_char_table().map_or(false, |table| {
-        LispObject::from(table.purpose).eq(LispObject::from(Qcategory_table))
+        LispObject::from(table.purpose).eq(LispObject::from(unsafe { Qcategory_table }))
```
db48x: You should be able to do this without turning it into a LispObject along the way.

PR #487, opened 2017-11-18
Port forward-char and backward-char

shaleh:
Add mechanism to set initial keybindings in Rust.


shaleh:
toggling build after merge

PR #489, opened 2017-11-19
Simplify `lisp_fn` attribute parsing using the `darling` crate

birkenfeld:


shaleh:
Yeah, but I missed remacs-macros.

shaleh:
@birkenfeld, put this in your config:

    (setq rust-format-on-save t)

Whenever you save it runs rustfmt on that file only. Quite handy.

shaleh:
The Windows build is failing because it cannot link `sys_open`.

birkenfeld:
Hmm, that `sys_open` thing is not good. Looks like the `remacs-lib` crate is not a good place to put this, which means we'd need a new crate...

(`remacs-macros`, as a `proc-macro` crate, is not allowed to export anything other than macros.)

birkenfeld:
As for `rust-format-on-save`, I have to figure out dir-locals for that, since I don't like autoformatting in other projects.

brotzeit:
Is it possible to run rustfmt as part of the compilation process ?

shaleh:
YUCK. Please no.

Possible and a good idea are not the same thing :-)

brotzeit:
I kind of expected this answer :)
But this is such a regular issue. Is there any reason or is this a "You just don't do it that way" thing. I'm very pragmatic ;)

birkenfeld:
Tests are passing now. r?

birkenfeld:
> But this is such a regular issue. Is there any reason or is this a "You just don't do it that way" thing. I'm very pragmatic ;)

Well, it's just moving the problem from "I haven't run `make rustfmt` before commit" to "I haven't run `make` before commit". Which is more unlikely, but still doesn't catch all cases.

And since it doesn't, it makes no sense to introduce the very dubious behavior of altering the source during compilation.

shaleh:
User has code loaded in emacs. `m-x compile`, failure due to a minor coding slip. Say a missing semicolon. ``C-x ` `` to jump to line. But that is not the line and emacs says the file changed?!?

This is why I suggested the `rust-format-on-save`. The buffer visibly changes. The developer can adapt to the changes. Then they can compile. I understand why @birkenfeld does not want to make this global for their config.

shaleh:
Is the intent that `remacs-util` can have more utility added to it as time goes on? Or should this be named `remacs-lisp_fn-helper` or something similar?

birkenfeld:
There might be more, yes. But I don't have a concrete example.

Given enough stuff ported from `-lib`, the lib crate might also be restructured into `bin` targets of the main crate.

shaleh:
Fair enough.

PR #491, opened 2017-11-21
Remove references on Pointer refs

shaleh:


rust_src/src/buffers.rs
```diff
@@ -23,42 +23,42 @@ pub type LispOverlayRef = ExternalPtr<Lisp_Overlay>;
...
 
     #[inline]
-    pub fn pt(&self) -> ptrdiff_t {
+    pub fn pt(self) -> ptrdiff_t {
         self.pt
     }
 
     #[inline]
-    pub fn beg_addr(&self) -> *mut c_uchar {
+    pub fn beg_addr(&mut self) -> *mut c_uchar {
```
birkenfeld: This should also become `self`. (You might say that this discards the mutability requirement, but it carries an ownership requirement which is stronger.)

rust_src/src/buffers.rs
```diff
@@ -125,34 +125,36 @@ impl LispBufferRef {
...
 
     #[inline]
-    pub fn fetch_multibyte_char(&self, n: ptrdiff_t) -> c_int {
+    pub fn fetch_multibyte_char(self, n: ptrdiff_t) -> c_int {
         let offset = if n >= self.gpt_byte() && n >= 0 {
             self.gap_size()
         } else {
             0
         };
+        let mut clone = self;
```
birkenfeld: Use `mut self` in the signature instead. But it'll become unnecessary after the other change.

rust_src/src/buffers.rs
shaleh: I like that the mutable reference lets the developer know that the use of this method will change the internals of the object. I almost made a `beg_addr_const` method since not every use needs a mutable pointer as shown later in your review. To my eye we are being too sneaky since that mutable pointer should have a lifetime attached to it.

birkenfeld: Well, it makes no sense to use different conventions here. This

birkenfeld: In any case, it's a truth in the garbage-collected, single-threaded world of Emacs Lisp objects that there is no need for the "mutable XOR shared" principle that Rust references and lifetimes enforce.

shaleh: Fair enough. This is one of those places I think remacs can do better. But I see your point too.

rust_src/src/hashtable.rs
```diff
@@ -165,6 +165,8 @@ impl<'a> Iterator for KeyAndValueIter<'a> {
     }
 }
 
+// The references are necessary to satisfy the lifetime requirements.
+// Otherwise the iterator would be have the copy which would be out of scope.
```
birkenfeld: Grammar here isn't quite obvious enough to me :)

rust_src/src/crypto/mod.rs
```diff
@@ -502,7 +502,7 @@ fn buffer_hash(buffer_or_name: LispObject) -> LispObject {
     if buffer.is_nil() {
         unsafe { nsberror(buffer_or_name.to_raw()) };
     }
-    let b = buffer.as_buffer().unwrap();
+    let mut b = buffer.as_buffer().unwrap();
```
birkenfeld: This isn't necessary anymore.

Unfortunately this file uses `#![allow(unused)]` which suppresses "unused mut" warnings too.

It should use `#![allow(dead_code)]` instead.

rust_src/src/hashtable.rs
shaleh: Haha. yeah good catch :-)

shaleh: Clearly I was interrupted while typing that.

DavidDeSimone:
LGTM 👍

shaleh:
Toggle full build

PR #492, opened 2017-11-21
lisp_fn exported functions should be public within rust too.

shaleh:
Enforce this by in the build process.
Also, remove export files if there are no longer exported functions.
This prevents odd compilation issues when the sole exported function
in a module has its #[lisp_fn] macro commented out to help debug a
compilation error.

### DIFF
--- a/rust_src/build.rs
+++ b/rust_src/build.rs
@@ -5,6 +5,7 @@ use std::fs::File;
 use std::io;
 use std::io::{BufRead, BufReader, Write};
 use std::path::PathBuf;
+use std::process;
 
 #[derive(Eq, PartialEq)]
 enum ParseState {
@@ -20,7 +21,7 @@ fn ignore(path: &str) -> bool {
 
 fn path_as_str(path: Option<&OsStr>) -> &str {
     path.and_then(|p| p.to_str())
-        .unwrap_or_else(|| panic!(format!("Cannot understand string: {:?}", path)))
+        .unwrap_or_else(|| panic!("Cannot understand string: {:?}", path))
 }
 
 fn get_function_name(line: &str) -> Option<String> {
@@ -31,14 +32,11 @@ fn get_function_name(line: &str) -> Option<String> {
         }
     }
 
-    None
+    panic!("Failed to find function name in: '{}'", line)
 }
 
 fn env_var(name: &str) -> String {
-    match env::var(name) {
-        Ok(value) => value,
-        Err(e) => panic!(format!("Could not find {} in environment: {}", name, e)),
-    }
+    env::var(name).unwrap_or_else(|e| panic!("Could not find {} in environment: {}", name, e))
 }
 
 fn source_file(modname: &str) -> String {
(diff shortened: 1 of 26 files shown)


rust_src/build.rs
```diff
@@ -110,8 +110,19 @@ where
             },
 
             ParseState::LispFn(name) => {
-                if line.starts_with("pub") || line.starts_with("fn") {
-                    if let Some(func) = name.or_else(|| get_function_name(&line)) {
+                let name = name.or_else(|| get_function_name(&line));
+
+                if line.starts_with("fn") {
+                    panic!(format!(
```
birkenfeld: You don't need `format!` in `panic!`, it supports format args out of the box.

rust_src/build.rs
```diff
@@ -110,8 +110,19 @@ where
...
+                        "
+{} is not public in {} module.
```
birkenfeld: ... in *the* ... module?

rust_src/build.rs
```diff
@@ -110,8 +110,19 @@ where
...
+lisp_fn functions are meant to be used from within remacs as well as in lisp code.
+",
+                        name.unwrap_or("<unknown>".to_string()),
```
birkenfeld: If `name` is still `None` (can that even happen?) the error isn't very helpful. Do we have a line number?

rust_src/build.rs
shaleh: Which one is it that doesn't? I know one of the macros complains when I give it format args. I thought it was panic.

rust_src/build.rs
shaleh: At the moment no. I could try and count.

shaleh: It is highly unlikely that the None case happens.

rust_src/build.rs
birkenfeld: I seem to remember something similar, but right now I don't find any. It might have been fixed some Rust versions ago.

shaleh: Good to know.

shaleh:
I am a lot happier with this commit. The error when a failure occurs is much easier to read.

rust_src/build.rs
```diff
@@ -31,14 +32,11 @@ fn get_function_name(line: &str) -> Option<String> {
         }
     }
 
-    None
+    panic!("Failed to find function name in: '{}'", line)
```
birkenfeld: function can now return `String` directly?

rust_src/build.rs
```diff
@@ -110,19 +110,29 @@ where
...
+                if line.starts_with("fn") {
+                    eprintln!(
+                        "
+`{}` is not public in the {} module at line {}.
+lisp_fn functions are meant to be used from Rust as well as in lisp code.",
+                        name,
+                        modname,
+                        lineno,
+                    );
+                    process::exit(1);
```
birkenfeld: Why use panic in some cases and process::exit in others?

shaleh: The panics are exceptions. The exits are user errors.

rust_src/build.rs
shaleh: Good point. Although I meant move the panic into the main line.

shaleh: Bah phones. Might not meant.

shaleh:
You can see what a build error looks like. Apparently my branch was not in sync.

shaleh:
toggling build

PR #502, opened 2017-11-24
Refresh base64 support

shaleh:
I would like to move decode to no allocation as well. However, the
only mechanism upstream uses a Vector which cannot be traded with C.
Once I get information about decode we can merge this.

### DIFF
--- a/rust_src/src/base64.rs
+++ b/rust_src/src/base64.rs
@@ -17,8 +17,9 @@ use strings::MIME_LINE_LENGTH;
 #[no_mangle]
 pub extern "C" fn base64_encode_1(
     from: *const c_char,
+    in_length: ptrdiff_t,
     to: *mut c_char,
-    length: ptrdiff_t,
+    out_length: ptrdiff_t,
     line_break: bool,
     multibyte: bool,
 ) -> ptrdiff_t {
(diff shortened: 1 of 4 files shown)


rust_src/src/base64.rs
```diff
@@ -157,16 +257,12 @@ fn base64_encode_string(string: LispObject, no_line_break: LispObject) -> LispOb
...
         string.sdata_ptr(),
-        encoded,
         length,
+        encoded,
+        allength,
         no_line_break.is_nil(),
         string.is_multibyte(),
     );
-
-    if encoded_length > allength {
```
DavidDeSimone: Is this check no longer worthwhile?

shaleh: Correct. The base64 code already checked it.

shaleh: I hope to update the upstream crate to export the computation to make our allocation better.

shaleh: Currently the upstream code panics if the size is off. I also hope to get a Result returning API accepted.

DavidDeSimone: Sounds good to me. 👍

shaleh:
While I work on the upstream code I think this round is ok to check in. Any complaints?

PR #503, opened 2017-11-25
Porting "hash_remove_from_table". Updating "as_natnum_or_error" API. Port weak_table fns. 

DavidDeSimone:
This PR ports "hash_remove_from_table" to Rust. As part of this process, I discovered a flaw in the "as_natnum_or_error" API. "hash_remove_from_table" in C uses XUINT, which returns an EMACS_UINT. However, in Rust, both as_fixnum, and as_natnum return EmacsInts. This means that certain hashes that are very large values may overflow, which will cause the hash comparison checks to fail. In order to fix this, I have updated as_natnum to return an EmacsUint. An audit of the codebase shows we have two usages of as_natnum* in Rust, and both should be unaffected by this change.


rust_src/src/lisp.rs
```diff
@@ -390,13 +390,32 @@ impl LispObject {
...
-            unsafe { self.to_fixnum_unchecked() }
+            Some(unsafe { self.as_natnum_unchecked() })
+        } else {
+            None
+        }
+    }
+
+    #[inline]
+    pub fn as_natnum_or_error(self) -> EmacsUint {
+        if let Some(num) = self.as_natnum() {
```
shaleh: Same as in the string review, how about `unwrap_or_else` like elsewhere in the code.

rust_src/src/hashtable.rs
```diff
@@ -327,4 +401,126 @@ fn define_hash_table_test(name: LispObject, test: LispObject, hash: LispObject)
...
+    // there is no more change.  This is necessary for cases like
+    // value-weak table A containing an entry X -> Y, where Y is used in a
+    // key-weak table B, Z -> Y.  If B comes after A in the list of weak
+    // tables, X -> Y might be removed from A, although when looking at B
+    // one finds that it shouldn't.
+
+    // This may look odd, but this is 'abusing' a Rust while loop to make
+    // a 'do-while' loop. All the logic is done in the conditional of the while loop,
+    // and the body is just empty.
+    let mut marked;
```
shaleh: Yuck. Why?

DavidDeSimone:
@birkenfeld That sounds reasonable to me. Based on that feedback, I'll revert the change to the natnum, and add an "as_unsigned" API for dealing with LispObject => EmacsUint conversions.

rust_src/src/lisp.rs
DavidDeSimone: I'll update it. 

As an aside, one thing I've been toying around in my head is a macro to auto generate all these as_*_or_error/as_*/as_*_unchecked APIs, since it's a lot of repetition, the the style of implementation differs across the different types. That obviously won't be part of this PR, but it's something I think is worth discussing in the future.

rust_src/src/hashtable.rs
DavidDeSimone: Haha I think I was trying to get a little to clever here. Once I saw this worked I jumped the gun a little. I can update this to be more idiomatic Rust.

rust_src/src/lisp.rs
shaleh: :+1:

I have been chewing at uniformity. But some kind of `make_type!` sounds like a plan.

rust_src/src/hashtable.rs
birkenfeld: At the very least it doesn't seem like you need the `let` outside the loop.

shaleh: We all have those moments.

A wise manager once taught me "Debugging is harder than coding. If you code to the edge of your skill by definition you cannot debug your own code". Save the cute code for when it really matters was an important lesson.

DavidDeSimone:
So I'm seeing some inconsistencies on the tests with this PR. I saw a random test failure on OSX, and now a segfault on linux. Talking over gitter it seems we have some known issues with travis on OSX at the moment. I have not been able to reproduce these issues locally. I'm going to close this PR for now and break it up into smaller parts so I can have more confidence in impact of these changes.

PR #504, opened 2017-11-26
Porting clear_string to Rust.

DavidDeSimone:
Includes a minor refactor of the as_string_* APIs, to make them a little more DRY. Adds new functions to LispStringRef, to replace some C macros used in clear_string.

### DIFF
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -932,11 +932,7 @@ impl LispObject {
 
     #[inline]
     pub fn as_string_or_error(self) -> LispStringRef {
-        if let Some(string) = self.as_string() {
-            string
-        } else {
-            wrong_type!(Qstringp, self)
-        }
+        self.as_string().unwrap_or_else(|| wrong_type!(Qstringp, self))
     }
 
     #[inline]
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -932,7 +932,8 @@ impl LispObject {
 
     #[inline]
     pub fn as_string_or_error(self) -> LispStringRef {
-        self.as_string().unwrap_or_else(|| wrong_type!(Qstringp, self))
+        self.as_string()
+            .unwrap_or_else(|| wrong_type!(Qstringp, self))
     }
 
     #[inline]
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -924,9 +924,7 @@ impl LispObject {
     #[inline]
     pub fn as_string(self) -> Option<LispStringRef> {
         if self.is_string() {
-            Some(LispStringRef::new(
-                unsafe { mem::transmute(self.get_untaggedptr()) },
-            ))
+            Some(unsafe { self.as_string_unchecked() })
         } else {
             None
         }
@@ -934,11 +932,16 @@ impl LispObject {
 
     #[inline]
     pub fn as_string_or_error(self) -> LispStringRef {
-        if self.is_string() {
-            LispStringRef::new(unsafe { mem::transmute(self.get_untaggedptr()) })
+        if let Some(string) = self.as_string() {
+            string
         } else {
             wrong_type!(Qstringp, self)
         }
+    }
+
+    #[inline]
+    pub unsafe fn as_string_unchecked(self) -> LispStringRef {
+        LispStringRef::new(mem::transmute(self.get_untaggedptr()))
     }
 }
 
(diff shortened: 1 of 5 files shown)


rust_src/src/lisp.rs
```diff
@@ -924,22 +924,25 @@ impl LispObject {
...
-            ))
+            Some(unsafe { self.as_string_unchecked() })
         } else {
             None
         }
     }
 
     #[inline]
     pub fn as_string_or_error(self) -> LispStringRef {
-        if self.is_string() {
```
shaleh: Why not emulate `as_window`

      self.as_string()
            .unwrap_or_else(|| wrong_type!(Qstringp, self))

rust_src/src/multibyte.rs
```diff
@@ -110,6 +110,34 @@ impl LispStringRef {
...
+        if self.size != 0 {
+            self.size_byte = -1;
+        }
+    }
+
+    /// This function does not allocate. It will not change the size of the data allocation.
+    /// It will only set the 'size' variable of the string, if it is safe to do so.
+    /// Replaces STRING_SET_CHARS from C.
+    pub fn set_num_chars(mut self, newsize: isize) {
+        assert!(if self.is_multibyte() {
```
shaleh: Is assert really what we want long term? The C code uses `eassert` which can be disabled. In Rust `assert!` is always on.

DavidDeSimone: So this was an intentional decision by me that I should have clarified in the PR description. 

Basically I was very torn if this function should be marked 'unsafe'. At first glance this function seemed like a footgun to me. It's name/semantics seem to imply it's doing an potential allocation, when it's actually it's just assigning to a variable, AND we only want that assignment to be possible in a very specific situation (what is checked for in the eassert). 

I did some basic profiling of "clear_string", in emacs, release remacs with this being a debug_assert, and remacs with this being an assert, and all 3 were almost identical. With that information, I felt that the best course of action was to make this an "assert" and mark the function as safe. 

If we would like to switch this to be a "debug_assert" to keep in line with the C code, I would like to insist that this function be marked unsafe.

rust_src/src/lisp.rs
DavidDeSimone: Sounds good, will update

rust_src/src/multibyte.rs
shaleh: So let's kill the function. It is called in exactly 3 places in the emacs source. I like marking this as "unsafe" until then.

DavidDeSimone: Sounds good to me. 👍

shaleh: One place is dired. There is an open PR to help finish.
One place is this PR. Natch.
One place is lread.c

rust_src/src/multibyte.rs
```diff
@@ -110,6 +110,35 @@ impl LispStringRef {
...
         unsafe { *self.const_data_ptr().offset(index) }
     }
+
+    /// (Somewhat) Replaces STRING_SET_UNIBYTE in C. That macro has the behavior that
+    /// if your string has size 0, it will replace your string variable with
+    /// 'empty_unibyte_string'. This function does not have that side-effect.
+    #[inline]
+    pub fn set_unibyte(mut self) {
+        if self.size != 0 {
+            self.size_byte = -1;
```
Wilfred: This is changing the behaviour, isn't it?

If we're no longer replacing an empty string, then elisp code could change behaviour because it could observe whether the string pointer has changed.

DavidDeSimone: I don't believe it should change behavior based on that macro definition. Basically, in the case that size == 0, it would assign `empty_string_unibyte` to the C layer Lisp_Object. Since SET_STRING_UNIBYTE(...) is the last call before we return Qnil, that side effect shouldn't matter. BUT it could change the behavior if people use this function in place of other places we use SET_STRING_UNIBYTE in the future, if they think this function is a 1:1 replacement. 

``` c
/* Mark STR as a unibyte string.  */
#define STRING_SET_UNIBYTE(STR)				\
  do {							\
    if (XSTRING (STR)->size == 0)			\
      (STR) = empty_unibyte_string;			\
    else						\
      XSTRING (STR)->size_byte = -1;			\
  } while (false)
```

shaleh: Why are you not emptying the string?

DavidDeSimone: In the case of `clear_string`, it doesn't matter, since we don't use STR after calling STRING_SET_UNIBYTE. The case that size == 0 doesn't mutate the underlying data of the untagged pointer, so it doesn't change the behavior.

Eh honestly I'm just not in love with what this macro is doing semantically, and this was an attempt to move away from it. In one case it sets a variable on your untagged pointer, in the other case it changes what your pointer is pointer to. 

I'm going to make an update to provide a function that will provide an inline function that does something similar to this macro. I think it will help avoid confusion and avoid errors in the future.

PR #505, opened 2017-11-27
[WIP] Add rollup helpers

shaleh:
My goal was to make it easier for people to do small rollups. This looks at a branch, and uses git cherry-pick to get a list of needed commits then applies them. But it tests the tree between each commit. So we know where our code breaks or where the upstream missed something.

### DIFF
--- a/rust_src/admin/rollups/both-modified
+++ b/rust_src/admin/rollups/both-modified
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+files=$(git status | grep 'both modified:' | cut -f2 -d:)
+
+if [ ! -z "$files" ]; then
+    echo $commit
+    git show $commit
+
+    for i in $files; do
+        emacsclient "$i"
+    done
+fi
+
+git cherry-pick --continue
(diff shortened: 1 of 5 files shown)


shaleh:
The "allow-empty" part is a total hack. But I do not know how to detect this state. `git status` only says it once. It works to keep the loop going though.

shaleh:
I used cherry-pick because it allowed testing between commits. Thoughts on this are welcomed.

shaleh:
I am hoping maybe we can automate this eventually. When it is only 20 or 30 commits it is not too bad.

shaleh:
Oh, one other bit. I had to disable the cleaning of elc files when `make clean` is run. Otherwise the build loop takes ages. This is a tweak to lisp/Makefile.in. If that is acceptable I will add it to the PR.

birkenfeld:
I'm not in favor of cherry-picks. They will duplicate history, blow up repo size, and make it harder to track already ported changes.

shaleh:
Fair points. But you did not offer any solutions. The rollups are taking too long. If nothing else maybe we should do smaller merges. A week of deltas at a time perhaps?

birkenfeld:
Yes, smaller merges would be nice. I'm not sure how much tooling is required; myself I would probably not trust it.

shaleh:
With merges I agree. The wrapping of cherry picks was manageable.

shaleh:
Abandoned.

PR #510, opened 2017-11-29
Refactor as_obj

shaleh:
Add `as_obj` to `ExternPtr` so all instances can be converted into a
`LispObject`.


shaleh:
Nevermind, we need to move as_obj down to the individual types.

PR #511, opened 2017-11-29
Latest Travis includes GNUTLS for OSX.

shaleh:
Remove our own attempt to install it which is causes build failures.


shaleh:
hmmmmm. No OSX tests are running. Looks like Travis is having issues.

PR #513, opened 2017-12-01
port frame-visible-p, frame-position, window-system

brotzeit:


rust_src/src/frames.rs
```diff
@@ -155,4 +156,27 @@ pub fn framep(object: LispObject) -> LispObject {
...
+/// On graphical displays, invisible frames are not updated and are
+/// usually not displayed at all, even in a window system's \"taskbar\".
+///
+/// If FRAME is a text terminal frame, this always returns t.
+/// Such frames are always considered visible, whether or not they are
+/// currently being displayed on the terminal.
+#[lisp_fn]
+pub fn frame_visible_p(frame: LispObject) -> LispObject {
+    let frame_ref = frame.as_live_frame_or_error();
+    if unsafe { fget_visible(frame_ref.as_ptr()) != 0 } {
```
shaleh: go ahead and add the `frame.is_visible()` and `is_iconified` methods and use them here. Then this function should look like the C function.

rust_src/src/frames.rs
```diff
@@ -155,4 +156,27 @@ pub fn framep(object: LispObject) -> LispObject {
...
+/// Return nil if FRAME was made invisible, via `make-frame-invisible'.
+/// On graphical displays, invisible frames are not updated and are
+/// usually not displayed at all, even in a window system's \"taskbar\".
+///
+/// If FRAME is a text terminal frame, this always returns t.
+/// Such frames are always considered visible, whether or not they are
+/// currently being displayed on the terminal.
+#[lisp_fn]
+pub fn frame_visible_p(frame: LispObject) -> LispObject {
+    let frame_ref = frame.as_live_frame_or_error();
```
shaleh: It is common elsewhere in the tree to just mask the incoming variable. If you add methods as I suggest this will be possible.

src/frame.c
```diff
@@ -5455,6 +5433,18 @@ fget_output_method(const struct frame *f)
   return f->output_method;
 }
 
+unsigned
```
shaleh: You could return bool here. SImplify code elsewhere. Or handle in the method. Either way.

rust_src/src/frames.rs
brotzeit: Forgot about that one.

brotzeit: I'm not sure I understand correctly. You mean shadow `frame` ?

shaleh: Yes

brotzeit: Hmmm. Is there really value in changing 
`let frame_ref = frame.as_live_frame_or_error();` to 
`frame = frame.as_live_frame_or_error();`

Also the type of `frame` is changed and I think the additional variable indicates it. And this is the way we do it in all the other functions.

brotzeit:
If we really want to use shadowing in this situation, I would prefer to add a guide line and change all functions where this is appropriate.

shaleh:
Fair enough. Clear up the rest.

shaleh:
It looks like you messed up a rebase

brotzeit:
I have tried to apply the recent changes. Definitely an improvement. I was already annoyed that the arguments always were LispObjects in lisp functions. Very nice solution.

rust_src/src/frames.rs
```diff
@@ -58,6 +73,16 @@ impl LispFrameRef {
...
     }
+
+    #[inline]
+    pub fn is_visible(self) -> bool {
+        unsafe { fget_visible(self.as_ptr()) }
+    }
+
+    #[inline]
+    pub fn is_iconified(self) -> bool {
+        unsafe { LispObject::from_bool(fget_iconified(self.as_ptr()) as BoolBF).is_not_nil() }
```
db48x: You should be able to do this without turning it into a LispObject along the way.

rust_src/src/frames.rs
```diff
@@ -183,4 +208,64 @@ pub fn framep(object: LispObject) -> LispObject {
...
+///  `ns' for an Emacs frame on a GNUstep or Macintosh Cocoa display,
+///  `pc' for a direct-write MS-DOS frame.
+///
+/// FRAME defaults to the currently selected frame.
+///
+/// Use of this function as a predicate is deprecated.  Instead,
+/// use `display-graphic-p' or any of the other `display-*-p'
+/// predicates which report frame's specific UI-related capabilities.
+#[lisp_fn(min = "0")]
+pub fn window_system(frame: Option<LispFrameRef>) -> LispObject {
```
db48x: consider returning `-> Option<LispSymbolRef>`

rust_src/src/frames.rs
```diff
@@ -183,4 +208,64 @@ pub fn framep(object: LispObject) -> LispObject {
...
+    let f = frame.map_or(selected_frame(), |f| f.as_lisp_obj());
+
```
shaleh: Is it worth preventing the double call to `framep`?

PR #515, opened 2017-12-02
port process-exit-status

brotzeit:


rust_src/src/process.rs
```diff
@@ -320,4 +325,20 @@ pub fn set_process_query_on_exit_flag(process: LispObject, flag: LispObject) ->
     flag
 }
 
+/// Return the exit status of PROCESS or the signal number that killed it.
+/// If PROCESS has not yet exited or died, return 0.
+#[lisp_fn]
+pub fn process_exit_status(process: LispObject) -> LispObject {
+    let p_ref = process.as_process_or_error();
```
shaleh: another place where shadowing the argument is a good choice.

rust_src/src/process.rs
```diff
@@ -320,4 +325,20 @@ pub fn set_process_query_on_exit_flag(process: LispObject, flag: LispObject) ->
...
+    if p_ref.raw_status_new() {
+        unsafe { update_status(p_ref.as_ptr()) };
+    }
+    let status = LispObject::from(p_ref.status);
+    if let Some(s) = status.as_cons() {
+        s.cdr().as_cons().unwrap().car()
```
shaleh: That `unwrap` will not have the same behaviour as the `eassert(CONSP)` from the C code.

brotzeit: I just realized that there is a macro 
`# define CONSP(x) lisp_h_CONSP (x)` 
and a function 
```
INLINE bool
(CONSP) (Lisp_Object x)
{
  return lisp_h_CONSP (x);
}
```

Now I'm consued :D
I also can't find `eassert(CONSP)`.

shaleh:
```
#define lisp_h_XCAR(c) XCONS (c)->car
#define lisp_h_XCDR(c) XCONS (c)->u.cdr
#define lisp_h_XCONS(a) \
   (eassert (CONSP (a)), (struct Lisp_Cons *) XUNTAG (a, Lisp_Cons))
```

Remember there isa space before the parens when grepping.

brotzeit: My bad. I only looked at `CONSP` and not at `XCAR` and `XCDR`....sorry.

lisp.h is really confusing...

shaleh: yes it is! No worries.

rust_src/src/process.rs
```diff
@@ -318,4 +323,20 @@ pub fn set_process_query_on_exit_flag(process: LispObject, flag: LispObject) ->
...
 
+/// Return the exit status of PROCESS or the signal number that killed it.
+/// If PROCESS has not yet exited or died, return 0.
+#[lisp_fn]
+pub fn process_exit_status(process: LispObject) -> LispObject {
+    let p_ref = process.as_process_or_error();
+    if p_ref.raw_status_new() {
+        unsafe { update_status(p_ref.as_ptr()) };
+    }
+    let status = LispObject::from_raw(p_ref.status);
```
brotzeit: Out of curiosity, this has worked with `from`, why did we change this again ?

rust_src/src/lisp.rs
```diff
@@ -948,6 +948,12 @@ impl LispObject {
         }
     }
 
+    #[inline]
+    pub fn xcons(self) -> LispCons {
+        debug_assert!(self.is_cons());
+        LispCons(self)
+    }
```
brotzeit: I think this is worth a method, but I can also remove it.

rust_src/src/process.rs
db48x: I changed what `::from` does; there are some good examples in `README.md`. In this case you can write `pub fn process_exit_status(process: LispProcessRef) -> LispObject` once someone writes an `impl From<LispObject> for LispProcessRef`. It's not a huge benefit for this function, since it immediately has to call an unported function with a raw ptr, but it's a large win overall.

rust_src/src/lisp.rs
shaleh: So this is `as_cons_unchecked`?

rust_src/src/process.rs
```diff
@@ -57,6 +57,11 @@ impl LispProcessRef {
         LispObject::from_raw(self.buffer)
     }
 
+    #[inline]
+    fn raw_status_new(&self) -> bool {
+        unsafe { LispObject::from_bool(pget_raw_status_new(self.as_ptr()) as BoolBF).is_not_nil() }
```
shaleh: why the cast to `BoolBF` when the function returns a `BoolBF`?

rust_src/src/process.rs
```diff
@@ -231,12 +236,12 @@ pub fn process_status(process: LispObject) -> LispObject {
...
-    if unsafe { pget_raw_status_new(p_ref.as_ptr()) } != 0 {
+    if p_ref.raw_status_new() {
         unsafe { update_status(p_ref.as_ptr()) };
     }
     let mut status = LispObject::from_raw(p_ref.status);
-    if let Some(c) = status.as_cons() {
-        status = c.car();
+    if status.is_cons() {
+        status = status.xcons().car();
     };
```
shaleh: So why is the `xcons` version better? You are still doing the `is_cons` check, possibly twice.

rust_src/src/process.rs
```diff
@@ -318,4 +323,20 @@ pub fn set_process_query_on_exit_flag(process: LispObject, flag: LispObject) ->
...
+#[lisp_fn]
+pub fn process_exit_status(process: LispObject) -> LispObject {
+    let p_ref = process.as_process_or_error();
+    if p_ref.raw_status_new() {
+        unsafe { update_status(p_ref.as_ptr()) };
+    }
+    let status = LispObject::from_raw(p_ref.status);
+    if status.is_cons() {
+        status.xcons().cdr().xcons().car()
+    } else {
```
shaleh: Would this read any better if you used one of the iterators? Something like `iter_cars().nth(2).unwrap()`?

brotzeit:
`XCONS` makes an assertion and I think we don't have an equivalent function in rust.

brotzeit:
So, do we want to keep the assertions from the c code or not ?

rust_src/src/process.rs
brotzeit: This doesn't work.

rust_src/remacs-sys/lib.rs
```diff
@@ -1043,7 +1043,7 @@ extern "C" {
 }
 
 extern "C" {
-    pub fn pget_raw_status_new(p: *const Lisp_Process) -> c_int;
+    pub fn pget_raw_status_new(p: *const Lisp_Process) -> BoolBF;
```
shaleh: Just use bool.

rust_src/src/process.rs
```diff
@@ -63,6 +63,11 @@ impl LispProcessRef {
         LispObject::from_raw(self.buffer)
     }
 
+    #[inline]
+    fn raw_status_new(&self) -> bool {
+        unsafe { LispObject::from_bool(pget_raw_status_new(self.as_ptr())).is_not_nil() }
```
shaleh: This can be simplified.

    unsafe { pget_raw_status_new(self.as_ptr()) }

rust_src/src/process.rs
```diff
@@ -322,4 +327,19 @@ pub fn set_process_query_on_exit_flag(mut process: LispProcessRef, flag: LispObj
...
+/// Return the exit status of PROCESS or the signal number that killed it.
+/// If PROCESS has not yet exited or died, return 0.
+#[lisp_fn]
+pub fn process_exit_status(process: LispProcessRef) -> LispObject {
+    if process.raw_status_new() {
+        unsafe { update_status(process.as_ptr()) };
+    }
+    let status = LispObject::from_raw(process.status);
+    if let Some(s) = status.as_cons() {
+        s.cdr().as_cons().unwrap().car()
```
shaleh: `car(s.cdr())` is what you want. You are correct in thinking this is not `XCAR`. But naked unwraps are not wise unless we are sure they will not fail.

One way to write this is:

    status.as_cons().map_or_else(|| LispObject::from(0),
                                 |cons| car(cons.cdr()))

src/process.c
```diff
@@ -7790,7 +7774,7 @@ returns non-`nil'.  */);
   defsubr (&Sprocess_attributes);
 }
 
-int
+bool_bf
```
shaleh: Does just `bool` work here? If not, maybe we could cast it?

src/process.h
```diff
@@ -310,7 +310,7 @@ INLINE_HEADER_END
 
 #endif /* EMACS_PROCESS_H */
 
-int
+bool_bf
```
shaleh: Same here.

`BoolBF` exposes an implementation detail that might change.

rust_src/remacs-sys/lib.rs
brotzeit: Why did we introduce BoolBF if we can use bool ?

shaleh: I believe because it was needed to access the bit fields. But there is no reason to expose that the underlying implementation is a bit field here.

brotzeit: It seems there were reasons for using BoolBF #430.

shaleh: Yes, there were reasons. But in this code there is no reason to export the value as a bit field. All we need is a boolean YES/NO.

brotzeit: `raw_status_new` is a bool_bf like the other struct members where we use BoolBF. What am I missing here ?

shaleh: Write a `bool_bf` but expose a `bool` to Rust. As far as Rust is concerned there is no difference between the two.

brotzeit: So this is like every other case where we use BoolBF. I would suggest that we stay consistent with it and change all occurrences then. It's already confusing enough for newcomers.

shaleh:
Are you coming back to this @brotzeit?

brotzeit:
Should be good now.

Issue #464, opened 2017-11-07
Verify that dynamic module loading works


shaleh:
The test included from emacs is not running. However, it is not running for me from the main emacs tree either. Shrug.

Compile `--with-modules` is supposed to enable it.

http://diobla.info/blog-archive/modules-tut.html is a nice over view.

DavidDeSimone:
So I verified this tonight, dynamic module loading works. However, my test module is written in C, based on the example linked in this issue. Next step... write a module in Rust >:)

shaleh:
:+1:

ubolonton:
In case you are interested, I've recently started a dynamic module in Rust here: https://github.com/ubolonton/magit-libgit2. It uses this binding https://github.com/ubolonton/emacs-module-rs, which itself contains 2 sample modules.

shaleh:
@ubolonton would you be able to look at DavidDeSimone's crate in PR #516. Does it make the work any easier? Harder?

ubolonton:
I tried my above modules with Remacs. They all worked, both loading and (basic) tests.

I'll check that PR.

brotzeit:
I can confirm that it works #1083 :sparkles:

brotzeit:
Ah it's about rust module.

PR #516, opened 2017-12-03
Add a 'remacs-modules' crate for building Emacs/Remacs modules in Rust.

DavidDeSimone:
I recently verified https://github.com/Wilfred/remacs/issues/464, by writing a small basic C module. It worked, but it got me thinking about writing emacs/remacs modules in Rust. In order to make this easier, I made a Rust crate that wraps the types provided by emacs-modules.h. The idea is that this crate can be used by emacs modules as a dependency. The good news is that these modules should be usable by both emacs AND remacs, so this can be useful for people who want to write modules in Rust but want them to work for both platforms. 

I decided to use `bindgen` to generate the code for emacs-modules.h. I like the idea of using bindgen where we can, and emacs-modules.h is fairly self contained. 

Some notes:

On OSX, Cargo produces .dylibs. By default, the shared_lib extension in remacs is .so on OSX. This can be changed, or you can work around it by manually calling (module-load "\<name\>"), but it is a little annoying.

### DIFF
--- a/rust_src/remacs-modules/lib.rs
+++ b/rust_src/remacs-modules/lib.rs
@@ -2,12 +2,12 @@ extern crate libc;
 extern crate remacs_sys;
 
 use remacs_sys::{emacs_env, emacs_runtime, emacs_value};
+use std::ffi::{IntoStringError, NulError};
 use std::ffi::CString;
 use std::mem;
 use std::ops::{Deref, DerefMut};
 use std::ptr;
 use std::slice;
-use std::ffi::{IntoStringError, NulError};
 
 #[repr(C)]
 pub struct EmacsRuntime(*mut emacs_runtime);
--- a/rust_src/remacs-modules/lib.rs
+++ b/rust_src/remacs-modules/lib.rs
@@ -5,10 +5,9 @@ use remacs_sys::{emacs_env, emacs_runtime, emacs_value};
 use std::ffi::CString;
 use std::mem;
 use std::ops::{Deref, DerefMut};
+use std::ptr;
 use std::slice;
-//use std::ptr;
-//use std::io::{Error, ErrorKind};
-//use std::ffi::IntoStringError;
+use std::ffi::{IntoStringError, NulError};
 
 #[repr(C)]
 pub struct EmacsRuntime(*mut emacs_runtime);
@@ -19,6 +18,26 @@ pub struct EmacsEnv(*mut emacs_env);
 #[repr(C)]
 #[derive(Clone, Copy)]
 pub struct EmacsValue(emacs_value);
+
+#[repr(C)]
+#[derive(Debug)]
+pub enum EmacsModuleError {
+    InvalidValue,
+    NulError(NulError),
+    IntoStringError(IntoStringError),
+}
+
+impl From<IntoStringError> for EmacsModuleError {
+    fn from(error: IntoStringError) -> EmacsModuleError {
+        EmacsModuleError::IntoStringError(error)
+    }
+}
+
+impl From<NulError> for EmacsModuleError {
+    fn from(error: NulError) -> EmacsModuleError {
+        EmacsModuleError::NulError(error)
+    }
+}
 
 pub type EmacsInt = ::remacs_sys::EmacsInt;
 pub type FuncallExit = ::remacs_sys::emacs_funcall_exit;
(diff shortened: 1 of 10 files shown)


rust_src/remacs-modules/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+        let mut args = vec![qfeat];
+        self.funcall(qprovide, &mut args);
+    }
+
+    pub fn funcall(&mut self, fun: EmacsValue, args: &mut Vec<EmacsValue>) -> EmacsValue {
+        let fnptr = self.funcall.unwrap();
+        let len = args.len() as libc::ptrdiff_t;
+        let mut raw_array: Vec<emacs_value> = args.iter().map(EmacsValue::to_raw).collect();
+        EmacsValue(unsafe { (fnptr)(self.0, fun.to_raw(), len, raw_array.as_mut_ptr()) })
+
```
shaleh: empty line

rust_src/remacs-sys/build.rs
```diff
@@ -196,7 +197,36 @@ fn generate_globals() {
...
+
+    let bindings = builder
+        .rustfmt_bindings(true)
+        .generate()
+        .expect("Unable to generate bindings");
+
+    let source = bindings.to_string();
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("modules.rs");
+    let file = File::create(out_path);
+    file.unwrap().write_all(source.as_bytes()).unwrap();
```
shaleh: Checkout the error handling I added to the top level build.rs. If this function returned a `Result` with `io::Error` you could use `?` handling to clarify this logic and avoid `unwrap` failures.

rust_src/remacs-modules/lib.rs
```diff
@@ -1,11 +1,14 @@
-extern crate remacs_sys;
 extern crate libc;
```
shaleh: Spacing and grouping matters. std, then external, then remacs, then local is the pattern the rest of the code follows.

DavidDeSimone:
Removing WIP tag from this.

rust_src/remacs-modules/lib.rs
```diff
@@ -0,0 +1,363 @@
+extern crate libc;
+extern crate remacs_sys;
+
+use remacs_sys::{emacs_env, emacs_runtime, emacs_value};
```
shaleh: Grouping here.

rust_src/remacs-modules/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+pub type intmax_t = ::libc::intmax_t;
+
+struct UserData {
+    func: Box<FnMut(&mut EmacsEnv, &mut Vec<EmacsValue>) -> EmacsValue>,
+}
+
+macro_rules! call {
+    ($self: ident, $func: ident) => (call!($self, $func,));
+    ($self: ident, $func: ident, $($arg:expr),*) => {{
+        let fnptr = $self.$func.unwrap();
```
shaleh: Do we want to handle errors on unwrap here?

ubolonton:
This is somewhat similar to https://github.com/ubolonton/emacs-module-rs. Perhaps we can combine the efforts somehow?

There are 2 big differences I noticed:
- This uses a single C function that routes to the corresponding Rust function based on data in the extra pointer. My version uses macro to generate different C functions. I'm not sure which approach is more favorable.
- My version converts between non-local exit and `Result` at the Emacs-module boundary. Module functions deal only with `Result`. Any thought on whether this is an appropriate strategy?

The most unclear issue to me is the ownership and lifetime of `EmacsValue`. From this doc https://phst.github.io/emacs-modules (the best one I found so far on the topic), it seems like:
- The lifetime of `EmacsValue` should be the that of the associated `EmacsEnv`. Module functions should only get `&mut EmacsValue`.
- `EmacsValue` should not implement `Clone`.
   + A special function `env.clone(&value)` should be used to get an owned version instead (which should call `make_global_ref`)
   + Their `Drop` implementation should call `free_global_ref`.

shaleh:
@ubolonton sorry for not seeing this earlier. I was deferring to @DavidDeSimone.

We have currently decided not to use `Result` in the remacs port because it disrupts the lisp stack when a failure occurs. With `result` the failure is no longer reported at the point it occurred but at the point it was noticed. Which is different from using `signal!`. If you look at the remacs code where you have `Result` as the return value from a `From` trait we use `Option` if the cast can fail or we explicitly use `as_foo_or_error()` which calls `signal!`. Which is similar to invoking `unwrap` on a `Result` that is not `Ok`.

I need to examine the code more closely and ponder before I can comment on whether the macro or all-in-one approach is better.

Your reading about the lifetimes matches my expectation.

shaleh:
I am closing this as it is getting more and more out of date.

@ubolonton if you have interest please open a new PR.

PR #517, opened 2017-12-03
Add a pre-commit hook

shaleh:


rust_src/admin/pre-commit
```diff
@@ -0,0 +1,47 @@
...
+
+trap finish EXIT
+
+set -eu
+
+git stash save -q --keep-index $STASH_NAME
+
+check_char='\xE2\x9C\x93'
+cross_char='\xE2\x9D\x8C'
+green='\033[0;32m'
```
db48x: this could be $(tput setaf 2), so that it's not specific to any particular terminal.

shaleh: Or we could drop it. Either way.

shaleh:
I dropped the colored output. It says success and FAIL! now. It should work everywhere. Support color blind people. The works :-)

shaleh:
Note the OSX failure despite me touching NO code. YAY :-1:

shaleh:
Toggling build.

PR #525, opened 2017-12-05
Update readme example to use latest LispObject APIs

Wilfred:
This isn't a perfect example, because we now have `LispObject::as_cons` which provides a safe API. Still, I think it illustrates the principle, and it's better than it was before.

Suggestions for better examples of genuine uses of pointer manipulation are welcome :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,26 +460,21 @@ If you can break memory safety by passing a valid value to a function,
 then it should be marked as `unsafe`. For example:
 
 ``` rust
-// This function is unsafe because it's dereferencing the car
-// of a cons cell. If `object` is not a cons cell, we'll dereference
-// an invalid pointer.
-unsafe fn XCAR(object: LispObject) -> LispObject {
-    (*XCONS(object)).car
-}
+impl LispObject {
+    // This function is unsafe because it's accessing a raw pointer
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -471,15 +471,15 @@ unsafe fn XCAR(object: LispObject) -> LispObject {
 // of XCAR: it only passes valid cons cells. We just use
 // unsafe blocks instead.
 fn car(object: LispObject) -> LispObject {
```
shaleh: We no longer need `unsafe`. Time for a new example!

    if let Some(cons) = object.as_cons() {
        cons.car()
    } else if object.is_nil() {
        LispObject::constant_nil()
    } else {
        wrong_type!(Qlistp, object)
    }

shaleh: There are not many "clean" examples of unsafe. We have wrapped clean APIs on a lot of it. Maybe an example showing access to `globals`? Or an example using the accessor functions to get data out of structs?

shaleh: multibyte.rs has a fair amount of slice creation and what not.

Wilfred: OK, I've updated the example to show dealing with strings, based on our wrapped safe APIs. I wanted to focus on pointers as a fundamental source of unsafety.

PR #526, opened 2017-12-06
Rollup 20171002 - 20171009

shaleh:


shaleh:
Rather than try and sync up with the recent changes I am going to make several smaller rollups.

PR #528, opened 2017-12-06
Fixing areas of the codebase where we were mismanaging FFI with Rust functions.

DavidDeSimone:
It looks like we were either a) using #[no_mangle] without an extern "C", and b) using LispObjects in extern "C" functions, which has caused issues with ffi in the past (see https://github.com/Wilfred/remacs/pull/496).

Once I cleaned up our FFI, several new unused_code warnings came up. I added some allow(dead_code) statements, since we will most likely want these functions available in the future.

### DIFF
--- a/rust_src/src/ffi.rs
+++ b/rust_src/src/ffi.rs
@@ -1,7 +1,4 @@
-//! Module that is used for FFI exports.
-/// Used for creating #[no_mangle] wrappers around functions defined in the top level rust
-/// crate for remacs. These calls should NOT be used in Rust directly,
-/// it is only for C compat.
+//! Module that is used for FFI exports.These calls should NOT be used in Rust directly.
 use data;
 use keyboard;
 use lisp::LispObject;
--- a/rust_src/src/ffi.rs
+++ b/rust_src/src/ffi.rs
@@ -1,10 +1,11 @@
 //! Module that is used for FFI exports.These calls should NOT be used in Rust directly.
+use remacs_sys::{Lisp_Object, Lisp_Window};
+
 use data;
 use keyboard;
 use lisp::LispObject;
 use lists;
 use math;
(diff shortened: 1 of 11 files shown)


shaleh:
I have lint additions to build.rs to catch these in the future. That PR will be up soon.

rust_src/src/ffi.rs
```diff
@@ -0,0 +1,57 @@
+//! Module that is used for FFI exports.
+/// Used for creating #[no_mangle] wrappers around functions defined in the top level rust
+/// crate for remacs. These calls should NOT be used in Rust directly,
+/// it is only for C compat.
+use data;
+use keyboard;
+use lisp::LispObject;
+use lists;
+use math;
+use remacs_sys::{Lisp_Object, Lisp_Window};
```
shaleh: You did not follow the pattern.

std
external crates
remacs other crates
local files

rust_src/src/lib.rs
```diff
@@ -74,6 +74,7 @@ mod threads;
 mod util;
 mod vectors;
 mod windows;
+mod ffi;
```
shaleh: Alphy order....

rust_src/src/ffi.rs
DavidDeSimone: Thanks for the reminder, I had forgotten we adopted that convention. Will update.

DavidDeSimone: Follow up: `cargo fmt` is insisting I order the crates in the way they are currently ordered. Moving remacs_sys to the top of the list triggers `cargo fmt` to complain.

shaleh: Look at the other files. As long as you put a space between the groups rustfmt will honor the grouping.

DavidDeSimone: ah, thats what I was missing

PR #529, opened 2017-12-06
Add explicit linting code.

shaleh:
Catch no_mangle without extern "C"
Catch LispObject arguments or returns for extern "C" functions

Part of #496 and PR #528

### DIFF
--- a/rust_src/build.rs
+++ b/rust_src/build.rs
@@ -15,6 +15,70 @@ enum ParseState {
     LispFn(Option<String>),
 }
 
+/// Exit with error $code after printing the $fmtstr to stderr
+macro_rules! fail_with_msg {
+    ($code:expr, $modname:expr, $lineno:expr, $($arg:expr),*) => {{
+        eprintln!("In {} on line {}", $modname, $lineno);
+        eprintln!($($arg),*);
+        process::exit($code);
+    }};
+}
+
+struct LintMsg {
+    modname: String,
+    lineno: i32,
+    msg: String,
+}
+
+impl LintMsg {
+    fn new(modname: &str, lineno: i32, msg: String) -> Self {
+        Self {
+            modname: modname.to_string(),
+            lineno: lineno,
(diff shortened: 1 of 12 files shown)


shaleh:
This build will fail until #528 is landed.

rust_src/build.rs
```diff
@@ -62,23 +126,24 @@ fn export_function(
     name: Option<String>,
     line: &str,
     modname: &str,
-    lineno: u32,
+    lineno: i32,
     msg: &str,
-) -> Option<String> {
+) -> Result<Option<String>, LintMsg> {
```
DavidDeSimone: Can we consolidate this into just a Result<String, LintMsg> instead of a Result<Option<String, ...> ? It feels unidiomatic to me in this case.

shaleh: /home/shaleh/repos/opensource/rust/src $ grep -r -- 'Result<Option' . | wc -l
70

shaleh: This is what a strong type system gives you. Falling back to an empty string is weaker and carries less semantic meaning.

DavidDeSimone: That is a good point, and I agree that there are many cases where Result<Option<T>, U> carries important semantic meaning. My question was more aimed at our specific usage. We use the result of export_function in two places, in one case, if our option is in the None case, we just panic anyway, implying an error condition. In the other case, we do nothing, which backs your point of the option having semantic meaning.
 
My personal philosophy is "aggressive minimalism", so I like to always question usages of type pattens like this, because I feel they can be intimidating to newer potential contributors. However, after thinking about it some more, I think the code is fine as is.

tdlr; question withdrawn 😃

shaleh: Please keep questioning. I approve of minimalism.

As for newbies, clear code following idioms is how we learn. Otherwise it is C with window dressing

shaleh:
Toggling build

PR #530, opened 2017-12-07
A round of clippy suggestions

shaleh:


rust_src/src/crypto/mod.rs
```diff
@@ -106,16 +106,14 @@ fn get_coding_system_for_buffer(
     if LispObject::from(unsafe { globals.f_Vcoding_system_for_write }).is_not_nil() {
         return LispObject::from(unsafe { globals.f_Vcoding_system_for_write });
     }
-    if LispObject::from(buffer.buffer_file_coding_system).is_nil() || LispObject::from(unsafe {
+    if (LispObject::from(buffer.buffer_file_coding_system).is_nil() || LispObject::from(unsafe {
```
birkenfeld: I would actually pull out this whole condition.

rust_src/src/vector_macros.rs
```diff
@@ -1,6 +1,6 @@
 macro_rules! offset_of {
     ($ty:ty, $field:ident) => {
-        &(*(0 as *const $ty)).$field as *const _ as usize
+        &(*(ptr::null() as *const $ty)).$field as *const _ as usize
```
birkenfeld: The `as *const $ty` should not be needed.

rust_src/src/vectors.rs
```diff
@@ -139,25 +139,28 @@ impl LispVectorRef {
...
         }
     }
 
     #[inline]
-    pub fn as_mut_slice(&self) -> &mut [LispObject] {
+    pub fn as_mut_slice(&mut self) -> &mut [LispObject] {
         unsafe {
             slice::from_raw_parts_mut(
-                mem::transmute::<_, *mut LispObject>(&self.contents),
+                &self.contents as *const [i64; 1] as *mut LispObject,
```
birkenfeld: Isn't that `i64` platform dependent?

shaleh: Why? Lisp_Object is a 64bit int.

rust_src/src/vector_macros.rs
shaleh: I am open to another way to get it to compile. Just `ptr::null` does not work the compiler insists it does not know the type.

rust_src/src/crypto/mod.rs
shaleh: Pull out and do what with it? Do you mean make some local variables?

rust_src/src/vectors.rs
birkenfeld: Even on 32bit platforms?

rust_src/src/vector_macros.rs
birkenfeld: Ah right. You could give a turbofish annotation. But that isn't more readable IMO.

rust_src/src/crypto/mod.rs
birkenfeld: Yep.

rust_src/src/vector_macros.rs
shaleh: I tried `ptr::nul<$ty>()` and it did not like it.

rust_src/src/vectors.rs
shaleh: I had thought so. But it is a `c_long`. Either way, a naked i64 is wrong there.

rust_src/src/vector_macros.rs
birkenfeld: It would be `ptr::null::<$ty>()`. As I said, not particularly an improvement here.

shaleh: Yeah, not much.

rust_src/src/crypto/mod.rs
```diff
@@ -418,7 +416,7 @@ fn _secure_hash(
         digest_size as EmacsInt
     };
     let digest = LispObject::from(unsafe { make_uninit_string(buffer_size as EmacsInt) });
-    let digest_str = digest.as_string_or_error();
+    let mut digest_str = digest.as_string_or_error();
```
Wilfred: This is interesting: if it wasn't mutable before, why make it mutable now?

rust_src/src/vectors.rs
```diff
@@ -139,25 +139,28 @@ impl LispVectorRef {
     pub fn as_slice(&self) -> &[LispObject] {
         unsafe {
             slice::from_raw_parts(
-                mem::transmute::<_, *const LispObject>(&self.contents),
+                &self.contents as *const [i64; 1] as *const LispObject,
+                //mem::transmute::<_, *const LispObject>(&self.contents),
```
Wilfred: Why is this commented out?

rust_src/src/vectors.rs
```diff
@@ -139,25 +139,28 @@ impl LispVectorRef {
...
+                //mem::transmute::<_, *mut LispObject>(&self.contents),
                 self.len(),
             )
         }
     }
 
     #[inline]
     pub unsafe fn get_unchecked(&self, idx: ptrdiff_t) -> LispObject {
-        ptr::read(mem::transmute::<_, *const LispObject>(&self.contents).offset(idx))
+        let tmp = &self.contents as *const [i64; 1] as *const LispObject;
```
Wilfred: i64 here too: does this mean clippy was misleading? If so, we should report a bug upstream as I'm sure they'd be interested :)

rust_src/src/crypto/mod.rs
shaleh: Look a few lines down `digest_str.as_mut_slice()`

rust_src/src/vectors.rs
shaleh: I  need to remove it. I had left it as reference while testing.

rust_src/src/vectors.rs
shaleh: Clippy appears confused by Lisp_Object. It appears to give us the final base Rust type `i64` instead of `EmacsInt` or `Lisp_Object`. I am working through making sure all of the naked i64s introduced are properly named.

shaleh: I will see about making a small test case and checking with the Clippy folk.

Issue #417, opened 2017-10-26
Port gap-position and gap-size to Rust


brotzeit:
```c
DEFUN ("gap-position", Fgap_position, Sgap_position, 0, 0, 0,
       doc: /* Return the position of the gap, in the current buffer.
See also `gap-size'.  */)
  (void)
{
  Lisp_Object temp;
  XSETFASTINT (temp, GPT);
  return temp;
}

DEFUN ("gap-size", Fgap_size, Sgap_size, 0, 0, 0,
       doc: /* Return the size of the current buffer's gap.
See also `gap-position'.  */)
  (void)
{
  Lisp_Object temp;
  XSETFASTINT (temp, GAP_SIZE);
  return temp;
}
```

brotzeit:
In order to port those functions, we need to access the struct members `gpt` and `gap_size`.
There's already a method called `gap_size` in buffers.rs. The method can be copied for `gpt`. You only have to change the name of the struct member.

The c macro `XSETFASTINT` can be ported with the rust method `from_natnum`. You can find examples in our rust codebase.

PR #531, opened 2017-12-07
Porting gap-position and gap-size to Rust

dcluna:
This should work for #417.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -53,6 +53,11 @@ impl LispBufferRef {
     }
 
     #[inline]
+    pub fn gap_position(self) -> ptrdiff_t {
+        unsafe { (*self.text).gpt }
+    }
+
+    #[inline]
     pub fn gap_size(&self) -> ptrdiff_t {
         unsafe { (*self.text).gap_size }
     }
--- a/rust_src/src/editfns.rs
+++ b/rust_src/src/editfns.rs
@@ -77,6 +77,21 @@ pub fn eolp() -> LispObject {
     )
 }
 
+/// Return the position of the gap, in the current buffer.
+/// See also `gap-size'.
+#[lisp_fn]
+pub fn gap_position() -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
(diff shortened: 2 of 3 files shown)


rust_src/src/buffers.rs
```diff
@@ -52,6 +52,11 @@ impl LispBufferRef {
         unsafe { (*self.text).gpt_byte }
     }
 
+    #[inline]
+    pub fn gap_position(&self) -> ptrdiff_t {
```
shaleh: Is the reference required here? If not leave it off.

rust_src/src/editfns.rs
```diff
@@ -77,6 +77,21 @@ pub fn eolp() -> LispObject {
     )
 }
 
+/// Return the position of the gap, in the current buffer.
+/// See also `gap-size'.
+#[lisp_fn]
+pub fn gap_position() -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
+    LispObject::from_natnum(buffer_ref.gap_position() as EmacsInt)
```
shaleh: natnum is a promise to be greater than or equal to zero. Will this always be true?

rust_src/src/buffers.rs
brotzeit: I think somebody mentioned, that we should do this for consistency reasons some time ago. @Wilfred ?

brotzeit: IMO we should reorganize our impls.

shaleh: I have a PR open to remove refs.#491

rust_src/src/editfns.rs
dcluna: Yes, the original uses `XSETFASTINT`, replaced by [this function](https://github.com/Wilfred/remacs/blob/master/rust_src/src/lisp.rs#L315-L316).

rust_src/src/buffers.rs
dcluna: I removed the reference for this function, so that it doesn't conflict with your PR.

rust_src/src/editfns.rs
shaleh: :+1:

PR #432, opened 2017-10-28
port overlay-lists

brotzeit:

PR #536, opened 2017-12-11
[WIP] Port overlay-recenter to Rust

### LINKED ISSUE #433: Port overlay-recenter to Rust
```c
DEFUN ("overlay-recenter", Foverlay_recenter, Soverlay_recenter, 1, 1, 0,
       doc: /* Recenter the overlays of the current buffer around position POS.
That makes overlay lookup faster for positions near POS (but perhaps slower
for positions far away from POS).  */)
  (Lisp_Object pos)
{
  ptrdiff_t p;
  CHECK_NUMBER_COERCE_MARKER (pos);

  p = clip_to_bounds (PTRDIFF_MIN, XINT (pos), PTRDIFF_MAX);
  recenter_overlay_lists (current_buffer, p);
  return Qnil;
}
```

dcluna:
This fixes #433.

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -546,9 +546,9 @@ pub struct Lisp_Buffer {
     // XXX in C, bitfield with two bools
     flags: u8,
 
-    overlays_before: *mut c_void,
-    overlays_after: *mut c_void,
-    overlay_center: ptrdiff_t,
+    pub overlays_before: *mut c_void,
+    pub overlays_after: *mut c_void,
+    pub overlay_center: ptrdiff_t,
 
     undo_list: Lisp_Object,
 }
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -1257,7 +1257,7 @@ extern "C" {
     pub fn window_box_left_offset(w: *const Lisp_Window, area: glyph_row_area) -> c_int;
     pub fn window_menu_bar_p(w: *const Lisp_Window) -> bool;
     pub fn window_tool_bar_p(w: *const Lisp_Window) -> bool;
-
+    pub fn recenter_overlay_lists(buffer: *mut Lisp_Buffer, pos: ptrdiff_t);
 }
 
 /// Contains C definitions from the font.h header.
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -175,10 +175,10 @@ pub struct LispOverlayRefIter(LispOverlayRef);
 impl Iterator for LispOverlayRefIter {
     type Item = LispOverlayRef;
     fn next(&mut self) -> Option<Self::Item> {
-        let next = LispOverlayRef::new((*self).0.next as *mut Lisp_Overlay);
-        if next.as_ptr().is_null() {
+        if (*self).0.as_ptr().is_null() {
             return None;
         }
+        let next = LispOverlayRef::new((*self).0.next as *mut Lisp_Overlay);
         Some(next)
     }
 }
@@ -385,7 +385,7 @@ pub fn recenter_overlay_lists(mut buf: LispBufferRef, pos: ptrdiff_t) {
                         None => set_buffer_overlays_after(buf, tail.as_ptr()),
                     }
                 }
-                _ => panic!("Unexpected value for toMoveAfter"),
+                _ => {}
             }
             tail = maybePrev.map_or(LispOverlayRef::new(ptr::null_mut()), |mut prev| {
                 LispOverlayRef::new(prev.as_mut())
@@ -394,7 +394,11 @@ pub fn recenter_overlay_lists(mut buf: LispBufferRef, pos: ptrdiff_t) {
             break;
         }
         maybePrev = Some(tail);
-        tail = LispOverlayRef::new(tail.next as *mut Lisp_Overlay);
+        tail = if tail.next.is_null() {
+            LispOverlayRef::new(ptr::null_mut())
+        } else {
+            LispOverlayRef::new(tail.next as *mut Lisp_Overlay)
+        }
     }
 
     maybePrev = None;
@@ -429,14 +433,18 @@ pub fn recenter_overlay_lists(mut buf: LispBufferRef, pos: ptrdiff_t) {
                         None => set_buffer_overlays_before(buf, tail.as_ptr()),
                     }
                 }
-                _ => panic!("Unexpected value for toMoveBefore"),
+                _ => {}
             }
             tail = maybePrev.map_or(LispOverlayRef::new(ptr::null_mut()), |mut prev| {
                 LispOverlayRef::new(prev.as_mut())
             });
         }
         maybePrev = Some(tail);
-        tail = LispOverlayRef::new(tail.next as *mut Lisp_Overlay);
+        tail = if tail.next.is_null() {
+            LispOverlayRef::new(ptr::null_mut())
+        } else {
+            LispOverlayRef::new(tail.next as *mut Lisp_Overlay)
+        }
     }
 
     (*buf).overlay_center = pos
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -5,7 +5,7 @@ use std::{mem, ptr};
 
 use remacs_macros::lisp_fn;
 use remacs_sys::{EmacsInt, Lisp_Buffer, Lisp_Object, Lisp_Overlay, Lisp_Type, Vbuffer_alist};
-use remacs_sys::{make_lisp_ptr, nsberror, recenter_overlay_lists, set_buffer_internal};
+use remacs_sys::{make_lisp_ptr, nsberror, set_buffer_internal};
 
 use lisp::{ExternalPtr, LispObject};
 use lisp::defsubr;
@@ -323,7 +323,7 @@ pub fn overlay_recenter(pos: LispObject) -> LispObject {
         pos.as_fixnum_coerce_marker_or_error(),
         ptrdiff_t::max_value(),
     );
-    unsafe { recenter_overlay_lists(ThreadState::current_buffer().as_mut(), p) };
+    unsafe { recenter_overlay_lists(ThreadState::current_buffer(), p) };
     LispObject::constant_nil()
 }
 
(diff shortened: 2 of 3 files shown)


rust_src/src/buffers.rs
```diff
@@ -312,6 +313,20 @@ pub fn overlay_end(overlay: LispObject) -> LispObject {
...
+/// for positions far away from POS).
+#[lisp_fn]
+pub fn overlay_recenter(pos: LispObject) -> LispObject {
+    let p: ptrdiff_t = clip_to_bounds(
+        ptrdiff_t::min_value(),
+        pos.as_fixnum_coerce_marker_or_error(),
+        ptrdiff_t::max_value(),
+    );
+    unsafe { recenter_overlay_lists(ThreadState::current_buffer().as_mut(), p) };
+    return LispObject::constant_nil();
```
shaleh: No need for `return` here. I know, hard habit to break.

rust_src/remacs-sys/lib.rs
```diff
@@ -1247,6 +1247,7 @@ extern "C" {
     pub fn window_box_left_offset(w: *const Lisp_Window, area: glyph_row_area) -> c_int;
     pub fn window_menu_bar_p(w: *const Lisp_Window) -> bool;
     pub fn window_tool_bar_p(w: *const Lisp_Window) -> bool;
+    pub fn recenter_overlay_lists(buffer: *mut Lisp_Buffer, pos: ptrdiff_t);
```
shaleh: Did you want to give porting this function a shot?

dcluna: Good idea. If I manage to rewrite the original in Rust, I'll try to benchmark both versions, too.

shaleh: :+1:

shaleh: `LispOverlayRef` is in rust_src/src/buffer.rs. That is a good place to start. The start and end methods replace the `OVERLAY_START` and `OVERLAY_END` calls in the C code.

shaleh:
We have a gitter channel for remacs. You can seek help there if you get stuck. It is also ok to mark the PR as [WIP] and push code that isn't working yet. There is also a label for "help wanted" if you get most of the way there but need help with a problem.

rust_src/remacs-sys/lib.rs
dcluna: While reading the source for this code, I found [this](https://github.com/Wilfred/remacs/blob/master/etc/TODO#L263), which suggests that `recenter_overlay_lists` will be reimplemented in another way. Do you think it still makes sense to rewrite it in Rust, given that?

shaleh: That is a emacs TODO, not ours. Just FYI.

If you look at the upstream tree there is a branch for hacking on this. https://github.com/emacs-mirror/emacs/commits/feature/noverlay/src

Note that once we have finished porting Emacs to Rust we will be free to reimplement these internals.

rust_src/src/buffers.rs
```diff
@@ -320,6 +335,57 @@ pub fn overlay_buffer(overlay: LispObject) -> LispObject {
...
 
+pub fn recenter_overlay_lists(mut buf: LispBufferRef, pos: ptrdiff_t) {
+    // let mut overlay: LispObject;
+    let (mut prev, mut tail, mut next): (
+        Option<LispOverlayRef>,
+        LispOverlayRef,
+        Option<LispOverlayRef>,
+    ) = (
+        None,
+        LispOverlayRef::new((*buf).overlays_before as *mut Lisp_Overlay),
```
dcluna: Does this allocate? Or just wraps a pointer with a better API?

rust_src/src/buffers.rs
```diff
@@ -320,6 +335,57 @@ pub fn overlay_buffer(overlay: LispObject) -> LispObject {
...
+        None,
+    );
+    loop {
+        if tail.as_ptr() as ptrdiff_t == 0 {
+            break;
+        }
+        let (beg, end, curNext) = (tail.start(), tail.end(), tail.next);
+        if end.as_fixnum_coerce_marker_or_error() as ptrdiff_t > pos {
```
dcluna: This is almost a translation from the [original](https://github.com/Wilfred/remacs/blob/aabf185e251f8586e1e4003af7fc388dcbf537e8/src/buffer.c#L3327), but it doesn't seem right to implement an Iterator here, due to performance concerns. Does that make sense?

rust_src/src/buffers.rs
shaleh: A `LispOverlayRef` is a Rust type that wraps a C pointer. So yes, an allocation is made but not for the C data. We treat the `ExternalPtr` wrappers as free in the Rust code.

rust_src/src/buffers.rs
shaleh: I say get it working then consider an iterator. If this code is only used in one place I would not worry about the iterator. Unless of course the iterator makes the code easier/better/faster.

rust_src/src/buffers.rs
shaleh: `new` in Rust is an idiom for an initializer. It does not mean a heap allocation has necessarily been made.

rust_src/src/buffers.rs
```diff
@@ -320,6 +335,57 @@ pub fn overlay_buffer(overlay: LispObject) -> LispObject {
     marker_buffer(marker)
 }
 
+pub fn recenter_overlay_lists(mut buf: LispBufferRef, pos: ptrdiff_t) {
+    // let mut overlay: LispObject;
```
shaleh: i don't know that creating these 3 variables at once is cleaner than 3 let statements.

rust_src/src/buffers.rs
```diff
@@ -320,6 +335,57 @@ pub fn overlay_buffer(overlay: LispObject) -> LispObject {
...
+        Option<LispOverlayRef>,
+        LispOverlayRef,
+        Option<LispOverlayRef>,
+    ) = (
+        None,
+        LispOverlayRef::new((*buf).overlays_before as *mut Lisp_Overlay),
+        None,
+    );
+    loop {
+        if tail.as_ptr() as ptrdiff_t == 0 {
```
shaleh: Is this a check for `ptr::null()'?

dcluna: Actually, this is the check for the condition in the [`for` loop](https://github.com/Wilfred/remacs/blob/aabf185e251f8586e1e4003af7fc388dcbf537e8/src/buffer.c#L3338), which indeed seems to be a `ptr::null()` check. How would you write this?

shaleh: `tail.is_null()`

shaleh: Instead of a `loop` why not use `while !tail.is_null()`? Loop means we have to really check for run away infinite loops.

rust_src/src/buffers.rs
```diff
@@ -336,43 +354,23 @@ pub fn overlay_buffer(overlay: LispObject) -> LispObject {
...
-            }
-        }
-        // TODO: check loop changes
-        prev = Some(tail);
-        tail = next.unwrap()
+    let before = LispOverlayRef::new((*buf).overlays_before as *mut Lisp_Overlay);
+    let start: (Option<LispOverlayRef>, Option<LispOverlayRef>) = (None, Some(before));
+    let toMove: Option<(Option<LispOverlayRef>, Option<LispOverlayRef>)> = before
+        .iter()
+        .scan(start, |pair, current: LispOverlayRef| {
```
shaleh: This feels like it wants to be a zip against itself. Could it be?

rust_src/src/buffers.rs
```diff
@@ -336,43 +354,23 @@ pub fn overlay_buffer(overlay: LispObject) -> LispObject {
...
+            Some((pair.1, Some(current)))
+        })
+        .find(|pair| {
+            (pair.1).unwrap().end().as_fixnum_coerce_marker_or_error() as ptrdiff_t > pos
+        });
```
shaleh: Why is `pair.1` an Option type? it is always a Some value as the unwrap demonstrates.

rust_src/src/buffers.rs
dcluna: Well, I was going for something like the [`each_cons`](https://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-each_cons) method from Ruby. I haven't seen the zip against itself pattern yet, but does it perform well?

shaleh: I worry about performance after I have working, obvious code.

dcluna:
@shaleh I'd like to add some assertions/Rust unit tests for the code that was changed. Could you point me to an example for that?

shaleh:
Make sure you have rebased on a current master. We recently landed a PR which allows elisp testing of Rust code.

What you do is make a file in test/rust_src/src/ with the name of the Rust file and '-tests.el' after it. So in your case `buffer-tests.el`. In this file add ERT based tests that verify the Rust code is doing what it should be doing. Create a buffer with some data, set the overlay so 15 characters in then shift it to be 30 characters in. Verify everything is where it should be. Then write more tests.

shaleh:
@brotzeit's work in #432 incorporates your work on adding an iterator. Once it lands this PR should be revisited.

shaleh:
Ok this PR is ready to have the cobwebs shaken free

rust_src/remacs-sys/lib.rs
```diff
@@ -546,9 +546,9 @@ pub struct Lisp_Buffer {
     // XXX in C, bitfield with two bools
     flags: u8,
 
-    overlays_before: *mut c_void,
-    overlays_after: *mut c_void,
-    overlay_center: ptrdiff_t,
+    pub overlays_before: *mut c_void,
+    pub overlays_after: *mut c_void,
+    pub overlay_center: ptrdiff_t,
```
brotzeit: I think these struct members were declared private for a reason.

dcluna:
I'll be unavailable to work on this for the next 2 months or so, so if anyone wants to take this up, no problem. Otherwise I'll get back to it later.

shaleh:
We look forward to your return. Thanks @dcluna for letting us know.

@brotzeit you want to take this one over? You should be able to push into this branch.

brotzeit:
@dcluna already spent time on this. This isn't urgent and we still have tons of stuff to port. I think we can wait until he has time to finish his work.

shaleh:
This PR is sufficiently stale it is probably more worthwhile to start fresh. I am closing it.

@dcluna if you have time to hack, you are welcome to try again.

PR #539, opened 2017-12-13
Lisp_Subr now knows whether a function is defined in C or Rust

### LINKED ISSUE #259: C-h f on a Rust primitive shows 'in C source code'
Jumping to definition works, however.

shaleh:
This allows `describe-function` to report whether a function is
defined in C or Rust.

Closes #259


shaleh:
Good question. I added the bit I needed to the end of the struct. So likely this is ok.

There is no reference to `Lisp_Subr` in the ML code. There is also no reference to Lisp_Subr in @DavidDeSimone 's PR #516 allowing the same from Rust.

Did I miss something?

shaleh:
From how I read the Emacs code, if you define a function within an external module the code treats it as if it was defined within Lisp code. So it will not try to wind back to the FFI module.

shaleh:
Talking in the Gitter channel I dug a little deeper.

External modules end up calling an exposed function `make_module_function` which they see as `make_function`. This creates a `Lisp_Module_Function` struct in Emacs which has a pointer to the user's function, arity info, etc. No `Lisp_Subr` is involved. Unless I am really missing something that struct is only created by either the `DEFUN` C macro or the Rust `lisp_fn` macro.

This means external code should never need to know about `Lisp_Subr`. If anyone has poked their head behind the curtain that far they will just have to recompile their code. No big deal.

Wilfred:
@db48x by the way, you should have received an invite to be a collaborator on the Remacs repo, so your approvals allow PRs to merged. It looks like you haven't accepted yet.

PR #541, opened 2017-12-14
implements LispObj::from(&str)

db48x:
This depends on #539, which introduces a call to build_string and all the attendant hoops, prompting me to write this.

### DIFF
--- a/rust_src/src/data.rs
+++ b/rust_src/src/data.rs
@@ -1,7 +1,5 @@
 //! data helpers
 
-use std::ffi::CString;
-
 use remacs_macros::lisp_fn;
 use remacs_sys::{Lisp_Misc_Type, Lisp_Type, PseudovecType};
 use remacs_sys::{Lisp_Subr_Lang_C, Lisp_Subr_Lang_Rust};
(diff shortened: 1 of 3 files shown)


shaleh:
Toggling build

db48x:
The From trait is basically the Rust convention for type conversions, and `from_str` would conflict with the method name for the FromStr trait, which is about parsing a string and returning some kind of object.

db48x:
Also, I was going to delete from_bool, from_fixnum, etc and make different From trait impls.

shaleh:
I thought we talked about that in a PR. Maybe it was in the Gitter channel. There was concern about `::from` working for all types if I recall. It might confuse the typing. Or maybe it was just concern over using `foo.into()` instead of `::from(foo)`.

Before you start the code, look at the history. `from_bool` and `from_float` are nearly a year old. So they are not a good indicator of current thinking.

db48x:
That's a good point; I haven't really looked at the history. Also, there's the slightly sticky problem of integers; sometimes an integer identifies a LispObject, and other times we might want a LispObject whose value is the integer.

Wilfred:
There was some discussion about `From` in #371.

shaleh:
Kicking the build

shaleh:
Your nemesis rustfmt strikes again!

PR #547, opened 2017-12-19
Automatically convert arguments to native Rust types when possible

db48x:
Not every type can currently be converted, but new types can be handled simply by adding an impl From<LispObject> for them.

Return values are not currently handled, but could very easily be done in the same way as arguments, by adding an impl From<foo> for LispObject for all the relevant types. Some discussion of that will be necessary, however.

The automatic conversions signal the usual "argument type error" if the incoming arguments are of incompatible types. Functions which aren't supposed to signal an error, or which need to handle multiple argument types, should simply keep using LispObject.

I've not touched every single lisp_fn that we have, but some of the ones that are unchanged should stay as they are. For example, ffloor and fceiling take floats but not fixnums, and hence still take LispObjects rather than EmacsDoubles. Others are untouched because they take a symbol|string, or a buffer|string, etc. Perhaps some of those can be handled later with more clever typing, if they come up often.


rust_src/src/keymap.rs
```diff
@@ -77,7 +77,7 @@ pub fn lookup_key(keymap: LispObject, key: LispObject, accept_default: LispObjec
         });
         idx += 1;
 
-        if c.is_cons() && lucid_event_type_list_p(c) {
+        if c.is_cons() && lucid_event_type_list_p(Option::<LispCons>::from(c)) {
```
shaleh: Isn't this `c.as_cons()`?

shaleh:
This PR reinvigorates the question about using `into()` instead of `::from()`.

rust_src/src/keymap.rs
db48x: well, technically I suppose it's c.as_cons_or_error()

db48x:
Yea, now that you mention it... calling `into` here might be a better choice than calling `from`, because it would eliminate the hack of mangling the type. I'll try it out.

shaleh:
Rustfmt bit you. Consider adding the pre-commit hook to catch this.

README.md
```diff
@@ -275,6 +275,40 @@ fn numberp(object: LispObject) -> LispObject {
 }
 ```
 
+Additinally, `lisp_fn` can automatically translate LispObjects passed
```
shaleh: Typo. `Additionally`

README.md
```diff
@@ -275,6 +275,40 @@ fn numberp(object: LispObject) -> LispObject {
...
+pub fn atan(y: EmacsDouble, x: Option<EmacsDouble>) -> EmacsDouble {
+    match x {
+        None => y.atan(),
+        Some(x) => y.atan2(x)
+    }
+}
+```
+
+The automatic translation signals a Lisp argument-type error if it
+sees an argument of the wrong type, so LispObjects are still the
```
shaleh: Comma splice. End the sentence at the comma and drop the 'so'.

rust_src/src/crypto/mod.rs
```diff
@@ -122,20 +123,21 @@ fn get_coding_system_for_buffer(
...
         /* If we still have not decided a coding system, use the
            default value of buffer-file-coding-system. */
-        return LispObject::from(buffer.buffer_file_coding_system);
+        return LispObject::from_raw(buffer.buffer_file_coding_system);
     }
-    let sscsf = LispObject::from(unsafe { globals.f_Vselect_safe_coding_system_function });
-    if fboundp(sscsf).is_not_nil() {
+    let sscsf =
+        LispObject::from_raw(unsafe { globals.f_Vselect_safe_coding_system_function });
+    if fboundp(LispSymbolRef::from(sscsf)).is_not_nil() {
```
shaleh: Again, `as_symbol_or_error`? Or maybe the right thing is to call `Ffboundp`. The leading `F` should take the LispObject and perform the transforms on it as if called from Lisp.

rust_src/src/ffi.rs
```diff
@@ -35,13 +35,17 @@ pub extern "C" fn arithcompare(
...
+        LispObject::from_raw(obj2),
+        comparison,
+    );
     result.to_raw()
 }
 
 #[no_mangle]
 pub extern "C" fn lucid_event_type_list_p(event: Lisp_Object) -> bool {
-    keyboard::lucid_event_type_list_p(LispObject::from(event))
+    keyboard::lucid_event_type_list_p(Option::<LispCons>::from(LispObject::from_raw(event)))
```
shaleh: `LispObject::from_raw(event).as_cons()` right?

rust_src/src/keyboard.rs
```diff
@@ -109,25 +109,25 @@ pub fn posn_at_x_y(
...
+        if first.eq(LispObject::from_raw(Qhelp_echo))
+            || first.eq(LispObject::from_raw(Qvertical_line))
+            || first.eq(LispObject::from_raw(Qmode_line))
+            || first.eq(LispObject::from_raw(Qheader_line))
+        {
+            return false;
+        }
 
-    let mut it = event.iter_cars_safe();
+        let mut it = event.as_obj().iter_cars_safe();
```
shaleh: Here is that inconsistency we talked about. This one is `as_obj` elsewhere it is `as_lisp_obj`

rust_src/src/crypto/mod.rs
db48x: No, `fboundp` checks if a symbol has a function definition. `boundp` checks to see if it has a value, but here we care that the value is also not nil.

db48x:
Sean, bisecting this failure shows that it only happens when using your implementation of `char-to-string`. I notice that `write-codepoint` and `CHAR_STRING` do not encode characters the same way; presumably that's the cause.

db48x:
Nevermind, I'm just dumb or sleep-deprived.

PR #548, opened 2017-12-20
Update icons with proper transparency

shaleh:


Dietr1ch:
I noted that the icons looked blurry, and confirmed that using the `-density 1536` flag improved them a lot

shaleh:
I ignored your `density` flag because I could not find good info on it. The examples I saw appeared to match the DPI of the image. Which is `90` for the SVG. But you are right, 1536 looks much better. I will update #548. Thank you again.

shaleh:
After fussing around and doing some research I gave up on ImageMagick and used rsvg directly. ImageMagick was delegating to it if the library was installed. This yields good looking output AND smaller images. So it is a win all the way around.

shaleh:
For anyone interested.

The density option basically fools the ImageMagick engine into handling the resizing in a more optimal way. Basically it is an arbitrary `$LARGE_VALUE` that tricks the rendering.

PR #325, opened 2017-09-24
port process stuff

brotzeit:

PR #549, opened 2017-12-20
Port thread-alive-p.

jaccarmac:
Address #422.

#325 is an older PR but I used it as a guide on how the thread functions are being ported. PR could probably use some work before getting merged so I've put what I have so far (simple stuff right now) for critique. Specifically it would be nice to avoid the mucking around with `std::ptr::null_mut`, I saw that the field I'm using it for is marked TODO in the thread struct anyway, not sure what future plans are on that.

And are the boolean conversions in match necessary for returning Lisp objects? I was surprised at the lack of a `From` impl for `bool` to `LispObject`.

### DIFF
--- a/rust_src/src/threads.rs
+++ b/rust_src/src/threads.rs
@@ -24,6 +24,11 @@ impl ThreadStateRef {
     pub fn name(self) -> LispObject {
         LispObject::from_raw(self.name)
     }
+
+    #[inline]
+    pub fn is_alive(self) -> bool {
+        !self.m_specpdl.is_null()
+    }
 }
 
 /// Return the name of the THREAD.
@@ -31,6 +36,12 @@ impl ThreadStateRef {
 #[lisp_fn]
 pub fn thread_name(thread: ThreadStateRef) -> LispObject {
     thread.name()
+}
+
+/// Return t if THREAD is alive, or nil if it has exited.
+#[lisp_fn]
+pub fn thread_alive_p(thread: LispObject) -> bool {
+    thread.as_thread_or_error().is_alive()
 }
 
 include!(concat!(env!("OUT_DIR"), "/threads_exports.rs"));
--- a/src/thread.c
+++ b/src/thread.c
@@ -854,18 +854,6 @@ or `thread-join' in the target thread.  */)
   return Qnil;
 }
 
-DEFUN ("thread-alive-p", Fthread_alive_p, Sthread_alive_p, 1, 1, 0,
-       doc: /* Return t if THREAD is alive, or nil if it has exited.  */)
-  (Lisp_Object thread)
-{
-  struct thread_state *tstate;
-
-  CHECK_THREAD (thread);
-  tstate = XTHREAD (thread);
-
-  return thread_alive_p (tstate) ? Qt : Qnil;
-}
-
 DEFUN ("thread--blocker", Fthread_blocker, Sthread_blocker, 1, 1, 0,
        doc: /* Return the object that THREAD is blocking on.
 If THREAD is blocked in `thread-join' on a second thread, return that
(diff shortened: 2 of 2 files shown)


rust_src/src/threads.rs
```diff
@@ -24,6 +25,14 @@ impl ThreadStateRef {
     pub fn name(self) -> LispObject {
         LispObject::from(self.name)
     }
+
+    #[inline]
+    pub fn alive_p(self) -> LispObject {
+        match self.m_specpdl == ptr::null_mut() {
+            true => LispObject::constant_nil(),
+            false => LispObject::constant_t(),
+        }
```
db48x: You could also just do an if/else, but this will work just fine.

db48x:
`From` trait impls are probably coming soon enough; see #547 for what that looks like.

rust_src/src/threads.rs
```diff
@@ -24,6 +25,14 @@ impl ThreadStateRef {
     pub fn name(self) -> LispObject {
         LispObject::from(self.name)
     }
+
+    #[inline]
+    pub fn alive_p(self) -> LispObject {
```
shaleh: Methods in the Rust code need to follow Rust naming standards instead of elisp standards. This method should be named `is_alive`.

While the match code below is valid, an if/else would match the rest of the code better.

Consider having this routine return a bool and performing the t/nil conversion in the `lisp_fn`. Then you do not need an if/else here.

rust_src/src/threads.rs
```diff
@@ -33,4 +42,10 @@ pub fn thread_name(thread: LispObject) -> LispObject {
     thread.as_thread_or_error().name()
 }
 
+/// Return t if THREAD is alive, or nil if it has exited.
+#[lisp_fn]
+pub fn thread_alive_p(thread: LispObject) -> LispObject {
```
shaleh: This has to follow the elisp convention and is fine as is.

rust_src/src/threads.rs
jaccarmac: Yeah, the if/else is definitely more idiomatic. On it.

rust_src/src/threads.rs
```diff
@@ -24,6 +25,11 @@ impl ThreadStateRef {
     pub fn name(self) -> LispObject {
         LispObject::from(self.name)
     }
+
+    #[inline]
+    pub fn is_alive(self) -> bool {
+        self.m_specpdl != ptr::null_mut()
```
jaccarmac: Would it be worth instantiating our own copy of `ptr::null_mut()` in one of the Remacs libs and using it to check nulls or is the use of `std::ptr` expected where null checks need to be performed?

shaleh: `ptr::null` and `null_mut` are the idiomatic wau of dealing with pointers instead of checking agsinst zero. It is analagous to `NULL` in C.

jaccarmac:
All right, in that case everything should be square at this point!

jaccarmac:
Followed Clippy's suggestion to remove the `std::ptr` reference with `is_null`.

Updated to the most recent master and abused the new `From` impls to eliminate the conditional.

rust_src/src/threads.rs
shaleh: Using the updated support in `lisp_fn` this can be simplified to:

    pub fn thread_alive_p(thread: ThreadStateRef) -> bool {
        thread.is_alive()
    }

You just need to add an `impl` for the `From` interface if there is not already one.

jaccarmac: Didn't realize `lisp_fn` had been changed as well, thanks for the heads-up!

jaccarmac:
Return type changed and the explicit `LispObject` conversion removed.

PR #551, opened 2017-12-24
Use the new `decode_config_slice` in base64 0.9.0

shaleh:
Add elisp tests.
Update Rust tests.

### DIFF
--- a/rust_src/src/base64.rs
+++ b/rust_src/src/base64.rs
@@ -1,6 +1,6 @@
 //! Base64 de- and encoding functions.
 
-use libc::{c_char, c_uchar, ptrdiff_t};
+use libc::{c_char, c_uchar};
 use std::ptr;
 use std::slice;
 
@@ -17,12 +17,12 @@ use strings::MIME_LINE_LENGTH;
 #[no_mangle]
 pub extern "C" fn base64_encode_1(
     from: *const c_char,
-    in_length: ptrdiff_t,
+    in_length: usize,
     to: *mut c_char,
-    out_length: ptrdiff_t,
+    out_length: usize,
     line_break: bool,
     multibyte: bool,
-) -> ptrdiff_t {
+) -> isize {
     let config = if line_break {
         // base64_crate::MIME, but with LF instead of CRLF
         base64_crate::Config::new(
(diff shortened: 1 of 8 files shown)


rust_src/src/base64.rs
```diff
@@ -243,10 +248,8 @@ bWFpbGluZyBsaXN0cywgc2VlIDxodHRwOi8vbGlzdHMuZ251Lm9yZy8+LgoK";
 #[lisp_fn(min = "1")]
 pub fn base64_encode_string(mut string: LispStringRef, no_line_break: bool) -> LispObject {
     // We need to allocate enough room for the encoded text
-    // We will need 33 1/3% more space, plus a newline every 76 characters(MIME_LINE_LENGTH)
-    // and then round up
     let length = string.len_bytes();
-    let mut allength: ptrdiff_t = length + length / 3 + 1;
+    let mut allength: ptrdiff_t = length * 4 / 3 + 4;
```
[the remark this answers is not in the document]
shaleh: I understand. Yeah, this should be a function.

The code is straight from the computation comments in the base64 crate. The Emacs math was not always good enough so there were checks for when the decoding or encoding exceeded the allocated block.

Rust is smart enough to not need the parens that C might require. The math is ((x * 4) / 3) + 4. And a similar but inverted version for the inverse operation.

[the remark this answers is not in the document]
shaleh: The upstream was loath to export this computation as part of the API. I wish they would have.

rust_src/src/base64.rs
```diff
@@ -271,19 +288,26 @@ pub fn base64_encode_string(mut string: LispStringRef, no_line_break: bool) -> L
 /// Base64-decode STRING and return the result.
 #[lisp_fn]
 pub fn base64_decode_string(mut string: LispStringRef) -> LispObject {
-    let length = string.len_bytes();
+    let length = (string.len_bytes() + 3) / 4 * 3;
```
db48x: use `compute_decode_size` here

rust_src/src/base64.rs
```diff
@@ -68,32 +68,37 @@ pub extern "C" fn base64_encode_1(
...
             // Decode non-ASCII bytes into UTF-8 pairs.
-            let s: String = decoded.iter().map(|&byte| byte as char).collect();
+            let s: String = decoded[..decoded_length]
+                .iter()
+                .map(|&byte| byte as char)
+                .collect();
+            let s_len = s.len();
             unsafe {
-                ptr::copy_nonoverlapping(s.as_ptr(), to as *mut u8, s.len());
+                ptr::copy_nonoverlapping(s.as_ptr(), to as *mut u8, s_len);
```
db48x: Can't we skip this copy? We should be able to decode directly into the `to` buffer...

shaleh: I had weird failures when I tried so I put it back. I will look at it again.

rust_src/src/base64.rs
shaleh: Thanks, I thought I had that one....

rust_src/src/base64.rs
shaleh: We do decode directly into `to`. It is named `decoded` since I had to make it a slice in Rust.

The issue is we have to handle the Emacs multibyte transcoding. The else hard does exactly as you suggest and simply uses `to`.

db48x:
looks like a merge from master would fix this.

shaleh:
Trying again

shanavas786:
@shaleh you can send these elisp tests upstream too.

PR #553, opened 2017-12-24
Update rustfmt config

shaleh:


shaleh:
Try again.

shaleh:
Correct, this gets rid of complaints from rustfmt about unknown config options.

PR #555, opened 2017-12-25
Move the Rust toolchain file to the top level.

jaccarmac:
rustup recognizes the tool even if it is in a parent directory and
away from Cargo.toml. Thus putting the toolchain file in the root
directory eliminates the need to manually manage the overridden
toolchain for Remacs.


shaleh:
You need to update .travis.yml

jaccarmac:
Indeed I do.

jaccarmac:
Hopefully that will sort things. Not totally sure on the nuance of the setup before, but I grepped the locations of `rust_src/rust-toolchain` so should be close.

PR #561, opened 2017-12-29
Generate docker files.

shaleh:


shaleh:
Yeah, it could be a little brittle. But the alternative was depending on a template system. Which felt like overkill.

The documentation is "cd into docker and run ./generate. Then check in the updated files." I will drop a README in.

PR #564, opened 2017-12-29
port make-sparse-keymap

brotzeit:


rust_src/src/keymap.rs
```diff
@@ -139,4 +139,25 @@ pub fn define_prefix_command(
...
+/// Construct and return a new sparse keymap.
+/// Its car is `keymap' and its cdr is an alist of (CHAR . DEFINITION),
+/// which binds the character CHAR to DEFINITION, or (SYMBOL . DEFINITION),
+/// which binds the function key or mouse event SYMBOL to DEFINITION.
+/// Initially the alist is nil.
+///
+/// The optional arg STRING supplies a menu name for the keymap
+/// in case you use it as a menu with `x-popup-menu'.
+#[lisp_fn(min = "0")]
+pub fn make_sparse_keymap(string: LispObject) -> LispObject {
```
db48x: You could use an Option<LispStringRef> here.

shaleh: :+1:

brotzeit: I'm not sure what to use here. I assume I should use `LispStringRef` when I see the argument is called string, but when I take a look at the docstring of `purecopy`

```Make a copy of object OBJ in pure storage.
Recursively copies contents of vectors and cons cells.
Does not copy symbols.  Copies strings without text properties.
```

it looks like some kind of special case to me ?

shaleh: I grepped both the Remacs lisp tree and my own installed packages from elpa et al.

The argument is always a string. `purecopy` takes all kinds of arguments and is used all over the tree. This is one of those moments where Lisp being typeless is at odds with our desire to have Rust use types.

Use `LispStringRef`. That matches the docstring and all code I have seen.

db48x: A cons cell probably wouldn't be so useful as a menu name anyway.

brotzeit: But then I can remove the `Fpurecopy` call ?

db48x: No, the input string presumably still needs to be copied.

shaleh: If you read the comments in and around the source code it is clear the `purecopy` is expected.

brotzeit: Unfortunately, I still don't get it ;P

Wilfred: The Emacs startup works roughly like this:

1. Create a tiny lisp interpreter. This is 'temacs'.
2. Allocate a big blob of memory. This is 'pure storage'.
3. Execute the base elisp files included in Emacs (e.g. simple.el).
4. Copy all the newly created elisp objects (strings, lists, etc) into the big blob of memory.
5. Write the big blob of memory along with temacs to a new binary. This is 'dumping', and the output is the `emacs` binary.

Does that make sense?

Since `make-sparse-keymap` calls `purecopy`, we know it is used in step 2. If we don't call `purecopy` I imagine step 5 file will fail during compilation.

A pure string is pretty much the same as a string, but it's immutable. You can't add properties to it, for example:

```emacs-lisp
;; Symbol names are often in pure storage if they are common symbols.
(add-text-properties 0 1 (list 'key 'value) (symbol-name 'message))
*** Eval error ***  Attempt to modify read-only object: "message"
```

brotzeit: @Wilfred Thanks!

I would like to document this somewhere as I can't find anything for `Vpurify_flag` or `purecopy`. But I'm not sure where to put it.

PR #565, opened 2017-12-30
Port barf-if-buffer-read-only

shaleh:


brotzeit:
I would say we merge this as it is and open an issue to discuss how we want to handle positiv EmacsInt ?

brotzeit:
The only problem I can see is that we throw errors for fixnums while emacs doesn't. But is it really an issue ?

rust_src/src/buffers.rs
```diff
@@ -367,4 +373,25 @@ pub fn set_buffer(buffer_or_name: LispObject) -> LispObject {
...
 
+/// Signal a `buffer-read-only' error if the current buffer is read-only.
+/// If the text under POSITION (which defaults to point) has the
+/// `inhibit-read-only' text property set, the error will not be raised.
+#[lisp_fn(min = "0")]
+pub fn barf_if_buffer_read_only(position: Option<EmacsInt>) -> () {
+    let pos = match position {
+        None => point(),
+        Some(n) => n,
+    };
```
db48x: This could be position.unwrap_or_else(point), but it's fine whichever way you think is clearer.

rust_src/src/buffers.rs
```diff
@@ -382,7 +382,7 @@ pub fn set_buffer(buffer_or_name: LispObject) -> LispObject {
 /// `inhibit-read-only' text property set, the error will not be raised.
 #[lisp_fn(min = "0")]
 pub fn barf_if_buffer_read_only(position: Option<EmacsInt>) -> () {
-    let pos = position.unwrap_or_else(point());
+    let pos = position.unwrap_or_else(|| point());
```
db48x: minor details :)

shaleh: Technically I could have passed just 'point' but I think the lambda reads better.

PR #567, opened 2017-12-30
Bring back emoji 👿

kbdluv:
Took this from here: https://lists.gnu.org/archive/html/emacs-devel/2016-06/msg00630.html

I don't know if Remacs wants to get into political territory (stand up for emojis ✊!) or not, but this change made my day =)

Please feel free to completely ignore this, and many thanks for Remacs!

Revert "Disable multicolor fonts on OS X since they are not supported on free systems"

This reverts commit 9344612d3cd164317170b6189ec43175757e4231.

### DIFF
--- a/src/macfont.m
+++ b/src/macfont.m
@@ -2358,9 +2358,9 @@ macfont_list (struct frame *f, Lisp_Object spec)
                   != (spacing >= FONT_SPACING_MONO)))
             continue;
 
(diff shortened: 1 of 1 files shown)


terlar:
I am not entirely sure of the implementation details, but it might be able to support it on free systems: https://github.com/eosrei/emojione-color-font

At least it is supported by mozilla applications. Guess this is not the forum for that though.

PR #573, opened 2017-12-31
apply recent changes

brotzeit:


rust_src/src/process.rs
```diff
@@ -310,10 +315,9 @@ pub fn process_query_on_exit_flag(process: LispObject) -> bool {
 /// exiting or killing a buffer if PROCESS is running.  This function
 /// returns FLAG.
 #[lisp_fn]
-pub fn set_process_query_on_exit_flag(process: LispObject, flag: LispObject) -> LispObject {
-    let p = process.as_process_or_error().as_mut();
+pub fn set_process_query_on_exit_flag(mut process: LispProcessRef, flag: LispObject) -> LispObject {
```
brotzeit: Is it ok to make the argument mutable in this case or do we want to avoid it completely ?

PR #576, opened 2018-01-01
port marker functions

brotzeit:


rust_src/src/marker.rs
```diff
@@ -102,4 +107,63 @@ pub fn set_point_from_marker(marker: LispMarkerRef) {
...
+
+/// Set the insertion-type of MARKER to TYPE.
+/// If ITYPE is non-nil, it means the marker advances when you insert text at it.
+/// If ITYPE is nil, it means the marker stays behind when you insert text at it.
+#[lisp_fn]
+pub fn set_marker_insertion_type(mut marker: LispMarkerRef, itype: LispObject) -> LispObject {
+    unsafe { mset_insertion_type(marker.as_mut(), itype.is_not_nil() as BoolBF) };
+    itype
+}
+
```
shaleh: `itype` is actually just a boolean then and this function returns a boolean.

rust_src/src/marker.rs
```diff
@@ -102,4 +107,63 @@ pub fn set_point_from_marker(marker: LispMarkerRef) {
...
+/// The optional argument ITYPE specifies the insertion type of the new marker;
+/// see `marker-insertion-type'.
+#[lisp_fn(min = "0")]
+pub fn copy_marker(marker: LispObject, itype: LispObject) -> LispObject {
+    if marker.is_not_nil() {
+        marker.as_fixnum_coerce_marker_or_error();
+    }
+    let new = unsafe { LispObject::from_raw(Fmake_marker()) };
+    let buffer_or_nil = if marker.is_marker() {
+        marker_buffer(marker)
```
shaleh: This no longer correct. `marker_buffer` now returns a `Option<Marker>`.

I think this should be:

buffer_or_nil = marker.as_marker().and_then(|m| m.buffer()).unwrap_or(LispObject::constant_nil(), |b| b.as_lisp_obj())

With proper formatting of course :-)

rust_src/src/marker.rs
brotzeit: No, it can be nil and non-nil. I also changed the docstring as it was wrong.

shaleh: Looking at the lisp code, t and nil are the usual parameters.

shaleh: Right. In Lisp NIL is false and anything else is true.

rust_src/src/marker.rs
brotzeit: Yeah, I have changed it. I should have made this one PR, sorry :)

shaleh: No wories, it is easy to fix.

shaleh:
Looks like transient network failure. Let's try again.

PR #581, opened 2018-01-03
Bump nightly and use the packaged rustfmt.

jaccarmac:
New nightly has it as a component! 🎆🎆🎆


jaccarmac:
Travis cache will need to be cleared before this gets merged.

shaleh:
I did clear the caches first.

jaccarmac:
Thanks much! New PRs will break it if we aren't careful as their config has a `cargo install --force` in it which will nuke the cache's versions of the `rustfmt` and `cargo-fmt` binaries.

shaleh:
What does being careful require? Is this a travis.yml fix?

shaleh:
Also, when I applied this locally and followed the instructions I no longer have a working rustfmt. This is because I have no binary in ~/.cargo/bin instead it is in ~/.rustup/toolchains/nightly-2018-01-2/bin which is not in my PATH. Did something fail locally or is the rustup path also supposed to be in my PATH?

jaccarmac:
Uninstall `rustfmt` globally and make sure your global nightly is up-to-date. If your default toolchain is stable and not nightly then temporarily set the default to nightly and update. That will let rustup put its own versions of `rustfmt` and `cargo-fmt` in `~/.cargo/bin`. They will link to the appropriate toolchain's versions of the formatting tools.

jaccarmac:
It is a travis.yml fix. If a PR with the old Travis version runs it will replace the binaries. So a cache clear will be necessary to fix that.

shaleh:
So where is the Travis PR?

jaccarmac:
It's in this one. Problem is since the old config uses `cargo install --force` it overrides any new config.

PR #584, opened 2018-01-03
Refactor frame handling

shaleh:
Break the core logic of `framep` out into a separate function. Use
this function internally since we already have `LispFrameRef` objects.

This allows `frame_live_p` to not duplicate the frame checks again in
`framep`.

`window-system` should just return a `LispObject`. There is no reason
to convert to a Symbol and immediately turn it back into a
`LispObject`.

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -5,15 +5,14 @@ use std::mem;
 use libc::c_int;
 
 use remacs_macros::lisp_fn;
-use remacs_sys::{selected_frame as current_frame, BoolBF, EmacsInt, Lisp_Frame};
+use remacs_sys::{selected_frame as current_frame, BoolBF, EmacsInt, Lisp_Frame, Lisp_Type};
 use remacs_sys::{fget_column_width, fget_iconified, fget_internal_border_width, fget_left_pos,
                  fget_line_height, fget_minibuffer_window, fget_output_method, fget_root_window,
                  fget_terminal, fget_top_pos, fget_visible, frame_dimension, Fcons, Fselect_window};
-use remacs_sys::{Qframe_live_p, Qicon, Qns, Qpc, Qt, Qw32, Qx};
+use remacs_sys::{Qframe_live_p, Qframep, Qicon, Qns, Qpc, Qt, Qw32, Qx};
 
 use lisp::{ExternalPtr, LispObject};
 use lisp::defsubr;
-use symbols::LispSymbolRef;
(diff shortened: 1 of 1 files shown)


rust_src/src/frames.rs
```diff
@@ -225,14 +227,19 @@ pub fn framep(object: LispObject) -> LispObject {
 /// use `display-graphic-p' or any of the other `display-*-p'
 /// predicates which report frame's specific UI-related capabilities.
 #[lisp_fn(min = "0")]
-pub fn window_system(frame: Option<LispFrameRef>) -> Option<LispSymbolRef> {
-    let f = frame.map_or(selected_frame(), |f| f.as_lisp_obj());
-    let window_system = framep(f);
+pub fn window_system(frame: Option<LispFrameRef>) -> LispObject {
+    let frame = frame.unwrap_or_else(|| selected_frame().as_frame_or_error());
```
brotzeit: Do we really need to check if selected_frame is a frame ? If we do I think an assertion would be better.

rust_src/src/frames.rs
```diff
@@ -225,14 +227,19 @@ pub fn framep(object: LispObject) -> LispObject {
 /// use `display-graphic-p' or any of the other `display-*-p'
 /// predicates which report frame's specific UI-related capabilities.
 #[lisp_fn(min = "0")]
-pub fn window_system(frame: Option<LispFrameRef>) -> Option<LispSymbolRef> {
-    let f = frame.map_or(selected_frame(), |f| f.as_lisp_obj());
-    let window_system = framep(f);
+pub fn window_system(frame: Option<LispFrameRef>) -> LispObject {
```
brotzeit: I've thought that we changed our return types and now use more explicit types. You didn't complain about this in the last PRs ?

brotzeit:
> window-system should just return a LispObject. There is no reason
> to convert to a Symbol and immediately turn it back into a
> LispObject.

I have asked myself the same thing. But I thought it's about indicating the return type.

shaleh:
The purpose of the recent changes is to avoid having the developer juggle types. If a function is always working on a Window the developer should not need to switch back and forth between the LispObject and the Window. If that Window is returned then it should be the return type.

In the case of this function the returned object is always used as a LispObject in the function so it might as well stay one and be returned appropriately.

Basically, write the function using the most natural types possible. That may be LispObject or that may be bool or something else. Then write the function definition to match. This means some functions will take 2 LispObjects and return another one. Others will take a LispFrameRef and a bool and return a LispWindowRef.

PR #587, opened 2018-01-04
remove buffer-has-markers-at ?

brotzeit:
This function is marked as obsolete, isn't used anywhere and doesn't work correctly.

### DIFF
--- a/src/emacs.c
+++ b/src/emacs.c
@@ -1455,7 +1455,6 @@ Using an Emacs configured with --with-x-toolkit=lucid does not have this problem
       syms_of_insdel ();
       /* syms_of_keymap (); */
       syms_of_macros ();
-      syms_of_marker ();
       syms_of_minibuf ();
       syms_of_process ();
       syms_of_search ();
--- a/src/marker.c
+++ b/src/marker.c
@@ -620,23 +620,6 @@ marker_byte_position (Lisp_Object marker)
   return m->bytepos;
 }
 
-DEFUN ("buffer-has-markers-at", Fbuffer_has_markers_at, Sbuffer_has_markers_at,
-       1, 1, 0,
-       doc: /* Return t if there are markers pointing at POSITION in the current buffer.  */)
-  (Lisp_Object position)
-{
-  register struct Lisp_Marker *tail;
-  register ptrdiff_t charpos;
-
-  charpos = clip_to_bounds (BEG, XINT (position), Z);
-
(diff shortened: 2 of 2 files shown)


Wilfred:
How doesn't it work correctly?

I'm reluctant to remove this, as it means there's elisp out there that would run on GNU Emacs but not Remacs. That said, I don't have any references to this function in my .emacs.d, and the only reference I can see on emacswiki (which has older elisp) is [just a function list](https://www.emacswiki.org/emacs/MonDocHelpUtilsDictionary).

If we do remove this, we should remove the reference in subr.el too.

brotzeit:
The function always returns t. I have no problem with porting this function, but it would behave differently.

shaleh:
Are you saying there is a bug in the code?

brotzeit:
I can't get the function to return nil.

Wilfred:
It works for me on GNU Emacs, at least.

```emacs-lisp
ELISP> (with-temp-buffer (buffer-has-markers-at (point-min)))
nil
ELISP> (with-temp-buffer (let ((m (make-marker))) (set-marker m (point-min)) (buffer-has-markers-at (point-min))))
t
```

brotzeit:
Interesting. Why does this not work without the macro ?

Wilfred:
Did you try it on a buffer that had lots of markers in it already? How did you test it?

brotzeit:
I have tried it in an empty scratch buffer with positiv and negativ numbers. But I just also tried it with `ielm` and it's the same result.

Wilfred:
I think moving point will add more markers to a buffer. Markers at the extreme edge of the buffer seem to be very common. If you pass a value less than `(point-min)` or greater than `(point-max)`, it's clipped to the limits before querying. That's why `(buffer-has-markers-at -1)` returns `t`.

`(buffer-has-markers-at (1+ (point-min)))` and `(buffer-has-markers-at (1- (point-max)))` both return `nil ` for me in the buffers I've tested in.

Wilfred:
I can't actually see a way to list all the markers in a buffer, at least from elisp. Given how limited introspection is, I'm even more inclined to leave this existing function in.

db48x:
That's a funny build error; let's try it again.

Wilfred:
Given that `buffer-has-markers-at` works as advertised, and it would be breaking compatibility with GNU Emacs, I think I'm going to decline this. Sorry.

PR #588, opened 2018-01-05
add function declaration for set_marker_internal

brotzeit:


Wilfred:
The test failure is really weird -- `format-properties` is complaining about arity, I wouldn't have expected that to be unreliable.

shaleh:
once more into the breach

shaleh:
And again....

Issue #582, opened 2018-01-03
window-system port changes spacemacs window colors


mateon1:
The colors in spacemacs windows changed (and are extremely broken) as of commit fc64d5a99ae0ebe603117cccbbdc49b0d3983aa7, found by git-bisect.

As of commit 812b84a241a07f1a2a65dd938efbc9691fc391e7, everything worked fine - notably, the background of the window was dark gray with bright gray text, but the next commit changed it to black text on white background with broken colors.

This only affects windowed mode, `src/remacs -nw` is unaffected.

shaleh:
Have you tried a97e31b612f26392519b1beec8596822ec428805?

shaleh:
When you open a window, enter `m-x ielm`. and then run `(framep (selected-frame))` and `(window-system (selected-frame))` what is the output?

mateon1:
I haven't tried a97e31b, but I have tried HEAD (7c8ed09f34caadf565531069d1e73fcfabf2fac4)
The following are on 7c8ed09f34caadf565531069d1e73fcfabf2fac4:
`(framep (selected-frame))` => `x`
`(window-system (selected-frame))` => `nil`
The following are on 812b84a241a07f1a2a65dd938efbc9691fc391e7:
[todo, waiting for build]

shaleh:
Perfect. Thank you. That was the confirmation I needed.

I am looking into this currently.

shaleh:
diff --git a/rust_src/src/frames.rs b/rust_src/src/frames.rs
    index f1b2eca..e38db9e 100644
    --- a/rust_src/src/frames.rs
    +++ b/rust_src/src/frames.rs
    @@ -9,11 +9,10 @@ use remacs_sys::{selected_frame as current_frame, BoolBF, Emac
    sInt, Lisp_Frame};
     use remacs_sys::{fget_column_width, fget_iconified, fget_internal_border_width, fget_left_pos,
                      fget_line_height, fget_minibuffer_window, fget_output_method, fget_root_window,
                      fget_terminal, fget_top_pos, fget_visible, frame_dimension, Fcons, Fselect_window};
    -use remacs_sys::{Qframe_live_p, Qicon, Qns, Qpc, Qt, Qw32, Qx};
    +use remacs_sys::{Qframe_live_p, Qframep, Qicon, Qns, Qpc, Qt, Qw32, Qx};
     
     use lisp::{ExternalPtr, LispObject};
     use lisp::defsubr;
[...]

PR #589, opened 2018-01-06
[WIP] Move from a macro to code generation

shaleh:
This is very early code. Please do not let the state of the code
interfere with the discussion about whether the idea is worthwhile. I
wanted to demonstrate it working but did not want to waste time if no
one approved.

The comment marker `// -> lisp_fn...` is definitely up for
discussion. I just picked something to get the ball rolling.

If this code is accepted maybe we look into moving it into an
external tool and not generating the code with each and every build.


shaleh:
If we move forward, I would really like to find a way to skip the casting in the C interface function if the Rust function uses `LispObject`.

birkenfeld:
Why is this considered? It seems strictly worse at the function definitition.

shaleh:
@birkenfeld why? Because of the use of a comment instead of a Rust macro? I am open to alternate syntax. Removing the arrow is like commenting out the line. But I am not rigid on this. I used a magic comment so that `rustfmt` and other code parses were not confused. Otherwise something like this was my first thought.

    @lisp_fn(min = "0")@

My inspiration was trying to debug the recent issue #582. It was impossible to get gdb to break on `framep`. In this case the usual tactic of breaking on `Fframep` was not helpful because it is called by other Rust routines. When I tried to break on a line number I ended up in a completely different function. You or I have worked on remacs enough to know some workarounds and the causes of these frustrations. What if you were a person interested in tackling their first issue? The barrier is higher than it should be. When there is more Rust than C this frustration will be amplified for all of us.

Plus, as I added in the comment above, this opens up the possibility of running the code generation stage once and checking the results in. Which will lead to faster compilation AND easier to debug code because there is no more magic hidden behind a macro.

shaleh:
Here is an alternative. I define an empty proc macro called `new_lisp_fn`. The macro exists solely to be a key for the code generator.

    diff --git a/rust_src/build.rs b/rust_src/build.rs
    index dd35f15..6403f62 100644
    --- a/rust_src/build.rs
    +++ b/rust_src/build.rs
    @@ -201,11 +201,11 @@ where
                         } else {
                             parse_state = ParseState::LispFn(None);
                         }
    -                } else if line.starts_with("// -> lisp_fn") {
    +                } else if line.starts_with("#[new_lisp_fn") {
                         parse_state = ParseState::LispFn2(None);
     
    -                    let begin = line.find(" lisp_fn").unwrap();
    -                    let start = begin + " lisp_fn".len();
    +                    let begin = line.find("new_lisp_fn").unwrap();
    +                    let start = begin + "new_lisp_fn".len();
                         if let Some(c) = line[start..].get(0..1) {
                             if c == "(" {
                                 if let Some(mut end) = line[start..].find(')') {
    diff --git a/rust_src/remacs-macros/lib.rs b/rust_src/remacs-macros/lib.rs
    index 88db1d9..4fb2c50 100644
    --- a/rust_src/remacs-macros/lib.rs
    +++ b/rust_src/remacs-macros/lib.rs
    @@ -129,6 +129,11 @@ pub fn lisp_fn(attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
         TokenStream::from_str(tokens.as_str()).unwrap()
     }
     
    +#[proc_macro_attribute]
    +pub fn new_lisp_fn(_attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
    +    fn_ts
    +}
    +
     struct CByteLiteral<'a>(&'a str);
     
     impl<'a> quote::ToTokens for CByteLiteral<'a> {
    diff --git a/rust_src/src/cmds.rs b/rust_src/src/cmds.rs
    index 20d644d..7f69ce4 100644
    --- a/rust_src/src/cmds.rs
    +++ b/rust_src/src/cmds.rs
    @@ -2,7 +2,7 @@
     
     use std::ffi::CString;
     
    -use remacs_macros::lisp_fn;
    +use remacs_macros::{lisp_fn, new_lisp_fn};
     use remacs_sys::{initial_define_key, scan_newline_from_point, set_point, set_point_both,
                      Fline_beginning_position, Fline_end_position};
     use remacs_sys::{Qbeginning_of_buffer, Qend_of_buffer, Qnil};
    @@ -161,7 +161,7 @@ pub fn end_of_line(n: Option<EmacsInt>) -> () {
     /// function will move point to the end of such a line and will count
     /// it as a line moved across, even though there is no next line to
     /// go to its beginning.
    -// -> lisp_fn(min = "0", intspec = "^p")
    +#[new_lisp_fn(min = "0", intspec = "^p")]
     pub fn forward_line(n: Option<EmacsInt>) -> EmacsInt {
         let count: isize = n.unwrap_or(1) as isize;
     
    diff --git a/rust_src/src/editfns.rs b/rust_src/src/editfns.rs
    index 325a357..9404cee 100644
    --- a/rust_src/src/editfns.rs
    +++ b/rust_src/src/editfns.rs
    @@ -2,7 +2,7 @@
     
     use libc::{c_uchar, ptrdiff_t};
     
    -use remacs_macros::lisp_fn;
    +use remacs_macros::{lisp_fn, new_lisp_fn};
     use remacs_sys::{EmacsInt, Fadd_text_properties, Fcons, Fcopy_sequence, Finsert_char,
                      Qinteger_or_marker_p, Qmark_inactive, Qnil};
     use remacs_sys::{buf_charpos_to_bytepos, globals, make_string_from_bytes, set_point_both};
    @@ -265,7 +265,7 @@ pub fn char_after(mut pos: LispObject) -> Option<EmacsInt> {
     /// Remaining arguments form a sequence of PROPERTY VALUE pairs for text
     /// properties to add to the result.
     /// usage: (fn STRING &rest PROPERTIES)
    -// -> lisp_fn(min = "1")
    +#[new_lisp_fn(min = "1")]
     pub fn propertize(args: &mut [LispObject]) -> LispObject {
         /* Number of args must be odd. */
         if args.len() & 1 == 0 {
    diff --git a/rust_src/src/fns.rs b/rust_src/src/fns.rs
    index 47271fb..6b91566 100644
    --- a/rust_src/src/fns.rs
    +++ b/rust_src/src/fns.rs
    @@ -1,6 +1,6 @@
     //* Random utility Lisp functions.
     
    -use remacs_macros::lisp_fn;
    +use remacs_macros::{lisp_fn, new_lisp_fn};
     use remacs_sys::{Fcons, Fmapc, Vautoload_queue};
     use remacs_sys::{Qfuncall, Qlistp, Qprovide, Qquote, Qsubfeatures, Qwrong_number_of_arguments};
     use remacs_sys::globals;
    @@ -91,7 +91,7 @@ pub fn provide(feature: LispSymbolRef, subfeature: LispObject) -> LispObject {
     /// See the common pitfall in info node `(elisp)Rearrangement' for an example
     /// of unexpected results when a quoted object is modified.
     /// usage: (quote ARG)
    -// -> lisp_fn(unevalled = "true")
    +#[new_lisp_fn(unevalled = "true")]
     pub fn quote(args: LispCons) -> LispObject {
         if args.cdr().is_not_nil() {
             xsignal!(
    diff --git a/rust_src/src/frames.rs b/rust_src/src/frames.rs
    index 94dba50..05987f5 100644
    --- a/rust_src/src/frames.rs
    +++ b/rust_src/src/frames.rs
    @@ -4,7 +4,7 @@ use std::mem;
     
     use libc::c_int;
     
    -use remacs_macros::lisp_fn;
    +use remacs_macros::{lisp_fn, new_lisp_fn};
     use remacs_sys::{selected_frame as current_frame, BoolBF, EmacsInt, Lisp_Frame, Lisp_Type};
     use remacs_sys::{fget_column_width, fget_iconified, fget_internal_border_width, fget_left_pos,
                      fget_line_height, fget_minibuffer_window, fget_output_method, fget_root_window,
    @@ -195,7 +195,7 @@ pub fn set_frame_selected_window(
     ///  `w32' for an Emacs frame that is a window on MS-Windows display,
     ///  `ns' for an Emacs frame on a GNUstep or Macintosh Cocoa display,
     /// See also `frame-live-p'.
    -// -> lisp_fn
    +#[new_lisp_fn]
     pub fn framep(object: LispObject) -> LispObject {
         object
             .as_frame()

db48x:
I agree that debugging is a little annoying at the moment, but I'm not sure it's that bad. You can break on `Fframep` or `remacs::frames::framep` and debug from there.

The real problem is the discoverability of the names; to know about Fframep you have to know about how Emacs arranged for C symbols to correspond to Lisp symbols, and to know about the Rust symbol you basically have to run `nm` (or remember that they're always namespaced, which I had forgotten).

birkenfeld:
The "empty proc macro" way is much better. I have an aversion to magic comments (there are enough "magic" parts to this whole story already).

shaleh:
I  will update the PR once I can convert the tree

shaleh:
The code in the build program continues to be a bit rough.

All `lisp_fn` exports are now handled by the new mechanism. If you have any problems trying out this branch run `cargo clean`. I hit snags with the macros crate until I cleaned.

db48x:
I feel like improving the macro would be better. I was looking through the `proc_macro` crate, and it already appears to keep track of the source locations for the tokens it parses. The `quote` crate, on the other hand, does not. They have different types to represent tokens and lists of tokens too. We then turn the `quote` crate's tokens into a string and parse it back into `proc_macro` tokens.

Incidentally, the final line of `lisp_fn` could be simplified to `tokens.parse().unwrap()`. That doesn't _help_ anything, but it makes the code a little nicer.

shaleh:
I have cleaned up the PR.

The build code is reasonable now. There are sure to be improvements there.

I have checked in the generated content. There is no magic.

In my opinion this is superior to the `lisp_fn` macro because GDB knows the proper line numbers and the developer has easy access to the source. Because we get proper line numbers from compiler errors when we make a mistake inside a `lisp_fn` function. Lastly, because all code is visible. No more magic.

I intend to update the build program to behave like `make` and not generate the export files if there are no relevant changes. This should help our compile times a little (yes, just a little).

shaleh:
Toggle build

brotzeit:
Does anybody know if macros provide better debugging possibilities soon ?

shaleh:
Closing in favor of #591.

PR #590, opened 2018-01-11
Port obarray support

shaleh:
Update the LispObarrayRef to match current style.

### DIFF
--- a/rust_src/src/obarray.rs
+++ b/rust_src/src/obarray.rs
@@ -2,9 +2,11 @@
 use libc;
 
 use remacs_macros::lisp_fn;
-use remacs_sys::{check_obarray, check_vobarray, intern_driver, make_unibyte_string, oblookup,
+use remacs_sys::{intern_sym, make_pure_c_string, make_unibyte_string, oblookup, Fmake_symbol,
                  Fpurecopy};
 use remacs_sys::{globals, Lisp_Object};
+use remacs_sys::{fatal_error_in_progress, initial_obarray};
+use remacs_sys::Qvectorp;
 
 use lisp::LispObject;
 use lisp::defsubr;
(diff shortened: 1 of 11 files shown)


rust_src/src/obarray.rs
```diff
@@ -13,22 +15,13 @@ use lisp::defsubr;
...
-    ///
-    /// This is a wrapper around the C function `check_obarray` instead of a
-    /// translation, so that fatal errors are handled in the same fashion.
-    pub fn from_object_or_error(object: LispObject) -> LispObarrayRef {
-        LispObarrayRef(LispObject::from_raw(unsafe {
-            check_obarray(object.to_raw())
-        }))
+    /// Return a reference to the Lisp variable `obarray`.
+    pub fn constant_obarray() -> LispObarrayRef {
+        LispObarrayRef(check_vobarray())
```
db48x: This is a great factorization, but the name confused me. Since this is checking a global variable rather than producing a constant value, perhaps `global_obarray` would be better?

rust_src/src/obarray.rs
```diff
@@ -99,13 +92,60 @@ impl LispObarrayRef {
...
+                tem.to_raw(),
+            ))
+        }
+    }
+}
+
+/// Get an error if OBARRAY is not an obarray.
+/// If it is one, return it.
+#[no_mangle]
+pub extern "C" fn check_obarray(obarray: Lisp_Object) -> Lisp_Object {
```
db48x: Why is this explicitly `extern "C"`? Let the `lisp_fn` macro generate the `extern "C" version of the function for you.

shaleh: Because it is needed in C land too. It is not called from elisp.

rust_src/src/obarray.rs
shaleh: I did not name it :-) I had left it alone for now. Let me give it another look.

rust_src/src/obarray.rs
shaleh: This is a `no_mangle` not a `lisp_fn`.

db48x: Oh right, it's not called from elisp. How many C callers are left?

shaleh: More than 11.

shaleh: Sorry that is the original tree. It is 7 after this PR.

db48x: That's not many, but it's enough that I won't make you guys port them all before merging this in.

shaleh: Good because the others are not as easy :-) The rest of this file gets deep into s-expression parsing, memory allocation calls, etc.

db48x: gotta tackle that sooner or later though :)

PR #595, opened 2018-01-14
escape quotes and backslashes when we use a byte literal in the lisp_fn macro

db48x:
otherwise, even a properly-escape quote or baskslash in the macro
invocation will cause us to emmit code that can't compile.

### DIFF
--- a/rust_src/remacs-macros/lib.rs
+++ b/rust_src/remacs-macros/lib.rs
@@ -7,6 +7,7 @@ extern crate quote;
 extern crate regex;
 extern crate remacs_util;
 extern crate syn;
+#[macro_use] extern crate lazy_static;
 
 use proc_macro::TokenStream;
 use regex::Regex;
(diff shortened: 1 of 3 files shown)


db48x:
bah

rust_src/remacs-macros/lib.rs
```diff
@@ -137,7 +139,11 @@ struct CByteLiteral<'a>(&'a str);
 
 impl<'a> quote::ToTokens for CByteLiteral<'a> {
     fn to_tokens(&self, tokens: &mut quote::Tokens) {
-        tokens.append(&format!(r#"b"{}\0""#, self.0));
+        let re = Regex::new(r#"["\\]"#).unwrap();
```
shaleh: The regex should be compiled in a lazy_static! so we do not pay for it in every function call. There are examples in the regex doc https://doc.rust-lang.org/regex/regex/index.html#example-avoid-compiling-the-same-regex-in-a-loop

PR #596, opened 2018-01-14
Port insert-char

db48x:


rust_src/remacs-sys/lib.rs
```diff
@@ -85,6 +86,38 @@ pub const INTMASK: EmacsInt = (EMACS_INT_MAX >> (INTTYPEBITS - 1));
...
+        EmacsInt::max_value() as ptrdiff_t,
+        libc::ptrdiff_t::max_value(),
+    ];
+    const arith_max: ptrdiff_t =
+        p[((p[1] - p[0]) >> ((8 * std::mem::size_of::<ptrdiff_t>()) - 1)) as usize];
+    const q: [ptrdiff_t; 2] = [(MOST_POSITIVE_FIXNUM - 1) as ptrdiff_t, arith_max];
+    q[((q[1] - q[0]) >> ((8 * std::mem::size_of::<ptrdiff_t>()) - 1)) as usize]
+}
+pub const BUF_BYTES_MAX: ptrdiff_t = buf_bytes_max();
+
```
shaleh: Why is this in remacs-sys and not in buffers,rs?

rust_src/src/editfns.rs
```diff
@@ -209,12 +211,68 @@ pub fn insert_byte(byte: EmacsInt, count: LispObject, inherit: LispObject) -> Li
...
+    //let buf = ThreadState::current_buffer();
+    //let multibyte: bool = LispObject::from_raw(buf.enable_multibyte_characters).is_not_nil().into();
+    let mut str = [0_u8; MAX_MULTIBYTE_LENGTH];
+    let len = write_codepoint(&mut str[..], character);
+
+    if BUF_BYTES_MAX / (len as isize) < (count as isize) {
+        unsafe { buffer_overflow() };
+    }
+    let mut n: ptrdiff_t = (count * (len as EmacsInt)) as ptrdiff_t;
+    let mut buffer = [0_i8; 4000];
```
shaleh: Where did 4k come from?

rust_src/remacs-sys/lib.rs
db48x: Because it's a global variable that we pulled in from the C side of things. It's merely an implementation detail that they wrote it as a macro, and that they put it in buffer.h.

rust_src/src/editfns.rs
```diff
@@ -209,12 +211,68 @@ pub fn insert_byte(byte: EmacsInt, count: LispObject, inherit: LispObject) -> Li
...
+/// called interactively, INHERIT is t.
+#[lisp_fn(min = "1", intspec = "(list (read-char-by-name \"Insert character (Unicode name or hex): \") (prefix-numeric-value current-prefix-arg) t))")]
+pub fn insert_char(character: Codepoint, count: Option<EmacsInt>, inherit: bool) {
+    let count = count.unwrap_or(1);
+    if count <= 0 {
+        return;
+    }
+
+    //let buf = ThreadState::current_buffer();
+    //let multibyte: bool = LispObject::from_raw(buf.enable_multibyte_characters).is_not_nil().into();
```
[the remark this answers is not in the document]
db48x: An oversight. The C code branches on multibyte to determine how to treat the character input, but I can't see how `write_codepoint` will do the wrong thing for a unibyte character, so I took the branch out. I just forgot to delete the commented-out code.

rust_src/src/editfns.rs
db48x: It's a magic constant from the original code. The original also uses `sizeof` to avoid repeating it the next two times, but that's not so useful for us.

rust_src/remacs-sys/lib.rs
shaleh: Right but why put it in the remacs-sys? At some point we are just going to copy it into buffers.rs so why not just do it now? remacs-sys is "temporary".

db48x: bad joke :)

When I discovered that it wasn't a real global variable I decided I would move it once I got it working. I actually woke up in the middle of the night with the idea for how to finally implement it in spite of the restrictions on constant functions, so I was probably just tired enough to forget.

db48x:
That's a weird one. Is that a failure during unexec? I'm just going to bounce it and look at it again tomorrow.

PR #598, opened 2018-01-18
Port if, and, or, cond, and progn

shaleh:
Move original eval.rs to eval_macros.rs.

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -1,7 +1,7 @@
 //! Generic Lisp eval functions
 
 use remacs_macros::lisp_fn;
-use remacs_sys::Qnil;
+use remacs_sys::{Qnil, Qt};
 use remacs_sys::eval_sub;
 
 use lisp::LispObject;
@@ -31,7 +31,7 @@ pub fn or(args: LispObject) -> LispObject {
 /// usage: (and CONDITIONS...)
 #[lisp_fn(min = "0", unevalled = "true")]
 pub fn and(args: LispObject) -> LispObject {
-    let mut val = Qnil;
+    let mut val = Qt;
 
     for elt in args.iter_cars_safe() {
         val = unsafe { eval_sub(elt.to_raw()) };
@@ -53,10 +53,10 @@ pub fn lisp_if(args: LispObject) -> LispObject {
     let cell = args.as_cons_or_error();
     let cond = unsafe { eval_sub(cell.car().to_raw()) };
 
-    if cond == Qnil {
-        progn(cell.cdr().as_cons_or_error().cdr())
-    } else {
+    if cond != Qnil {
         LispObject::from_raw(unsafe { eval_sub(cell.cdr().as_cons_or_error().car().to_raw()) })
+    } else {
+        progn(cell.cdr().as_cons_or_error().cdr())
     }
 }
 
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -1,91 +1,100 @@
-//! Generic Lisp eval functions and macros.
+//! Generic Lisp eval functions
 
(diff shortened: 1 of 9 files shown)


rust_src/src/eval.rs
```diff
@@ -1,91 +1,100 @@
...
-            ::remacs_sys::message_with_string($str.as_ptr() as *const ::libc::c_char,
-                                              $obj.to_raw(),
-                                              $should_log);
+/// Eval args until one of them yields nil, then return nil.
+/// The remaining args are not evalled at all.
+/// If no arg yields nil, return the last arg's value.
+/// usage: (and CONDITIONS...)
+#[lisp_fn(min = "0", unevalled = "true")]
+pub fn and(args: LispObject) -> LispObject {
+    let mut val = Qnil;
```
Wilfred: The default value should be Qt here. Would you mind adding a test too? `(and) => t`

rust_src/src/eval.rs
```diff
@@ -1,91 +1,100 @@
...
+/// Returns the value of THEN or the value of the last of the ELSE's.
+/// THEN must be one expression, but ELSE... can be zero or more expressions.
+/// If COND yields nil, and there are no ELSE's, the value is nil.
+/// usage: (if COND THEN ELSE...)
+#[lisp_fn(name = "if", c_name = "if", min = "2", unevalled = "true")]
+pub fn lisp_if(args: LispObject) -> LispObject {
+    let cell = args.as_cons_or_error();
+    let cond = unsafe { eval_sub(cell.car().to_raw()) };
+
+    if cond == Qnil {
```
Wilfred: This is putting the ELSE before the THEN. I think it's clearer with THEN first, which is what the C does.

rust_src/src/eval.rs
shaleh: Good catch on the `t` that is a little counter intuitive.

rust_src/src/eval.rs
shaleh: Will do

test/rust_src/src/eval-tests.el
```diff
@@ -0,0 +1,67 @@
...
+
+(require 'ert)
+
+(ert-deftest eval-tests--or-base ()
+  "Check (or) base cases"
+  (should (eq (or) nil))
+  (should (eq (or t) t))
+  (should (eq (or nil) nil))
+  (should (eq (or nil t) t))
+  (should (eq (or t nil) t)))
```
Wilfred: If we want to be thorough here, we could test non-booleans and evaluation order.

```
(or 1 2) => 1
(or t (error "oh no")) => t
```

That said, this is already much improved test coverage and I think the elisp tests would catch issues too.

test/rust_src/src/eval-tests.el
```diff
@@ -0,0 +1,67 @@
...
+(ert-deftest eval-tests--cond-base ()
+  "Check (cond) base cases"
+  (should (eq (cond) nil))
+  (should (eq (cond (t)) t))
+  (should (eq (cond (nil)
+                    (t))
+              t))
+  (should (eq (cond (t)
+                    (nil))
+              t)))
```
Wilfred: Clauses can take two forms with cond:

```
(cond (1)) => 1
(cond (1 2 3)) => 3
```

We could also test evaluation here too:

```
(cond (nil (error "oh no"))
  (t 1)) => 1
```

test/rust_src/src/eval-tests.el
shaleh: Good idea.

test/rust_src/src/eval-tests.el
shaleh: roger that

PR #599, opened 2018-01-19
Port prog1 and prog2

shaleh:
Add a `both` method to `LispCons` which retuns car and cdr. Also,
prog_ignore

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -2,10 +2,16 @@
 
 use remacs_macros::lisp_fn;
 use remacs_sys::{Qnil, Qt};
+use remacs_sys::Lisp_Object;
 use remacs_sys::eval_sub;
 
 use lisp::LispObject;
 use lisp::defsubr;
+
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ *   NOTE!!! Every function that can call EVAL must protect its args   *
+ *   and temporaries from garbage collection while it needs them.      *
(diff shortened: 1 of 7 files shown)


shaleh:
Comments on the naming of the `both` method? I could not think of a better name short of `car_and_cdr`.

rust_src/src/lisp.rs
```diff
@@ -1137,6 +1137,10 @@ impl LispCons {
         LispObject::from_raw(unsafe { (*self._extract()).cdr })
     }
 
+    pub fn both(self) -> (LispObject, LispObject) {
```
Wilfred: How about .as_tuple() or .destructure()?

rust_src/src/eval.rs
```diff
@@ -1,91 +1,138 @@
...
 }
 
-/// Macro to format a "wrong argument type" error message.
-macro_rules! wrong_type {
-    ($pred:expr, $arg:expr) => {
-        xsignal!(::remacs_sys::Qwrong_type_argument, LispObject::from_raw($pred), $arg);
-    };
+/// Evaluate BODY sequentially, discarding its value.
+#[no_mangle]
+pub extern "C" fn prog_ignore(body: Lisp_Object) {
```
Wilfred: Is this used elsewhere? I can only see one reference to it, so I'm wondering if we need it.

shaleh: It is used twice in eval.c and once in bytecode.c.

shaleh:
This PR builds on the previous PR. Once that one lands I can rebase this one.

PR #600, opened 2018-01-19
Port setq

shaleh:
Add support for a DEFSYM replacement in Rust. The dummy Rust macro is
called `def_lisp_sym`.

Add tests.

### DIFF
--- a/rust_src/remacs-lib/docfile.rs
+++ b/rust_src/remacs-lib/docfile.rs
@@ -131,7 +131,7 @@ pub extern "C" fn scan_rust_file(
             }
         } else if line.starts_with("def_lisp_sym!(") {
             lazy_static! {
-                static ref RE: Regex = Regex::new(r#"def_lisp_sym!\((.+),\s+"(.+)"\);"#).unwrap();
+                static ref RE: Regex = Regex::new(r#"def_lisp_sym!\((.+?),\s+"(.+?)"\);"#).unwrap();
             }
             let caps = RE.captures(line).unwrap();
             let name = CString::new(&caps[1]).unwrap();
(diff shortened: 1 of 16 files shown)


rust_src/remacs-lib/docfile.rs
```diff
@@ -130,6 +129,14 @@ pub extern "C" fn scan_rust_file(
                     "F", attr_props.name, docstring, docstring_usage
                 );
             }
+        } else if line.starts_with("def_lisp_sym!(") {
+            lazy_static! {
+                static ref RE: Regex = Regex::new(r#"def_lisp_sym!\((.+),\s+"(.+)"\);"#).unwrap();
```
Wilfred: This could capture too much if there's a sequence of def_lisp_sym. I suggest you make this regexp non-greedy, using `+?` instead of `+`.

rust_src/src/eval.rs
```diff
@@ -97,4 +106,84 @@ pub fn progn(body: LispObject) -> LispObject {
...
+
+        // Like for eval_sub, we do not check declared_special here since
+        // it's been done when let-binding.
+        // N.B. the check against nil is a mere optimization!
+        if unsafe { globals.f_Vinternal_interpreter_environment != Qnil } && sym.is_symbol() {
+            let binding = assq(
+                sym,
+                LispObject::from_raw(unsafe { globals.f_Vinternal_interpreter_environment }),
+            );
+            if let Some(binding) = binding.as_cons() {
```
Wilfred: The C code only checks that binding is non-nil here. However, since we call .set_cdr afterwards, I think checking that it's a cons is more robust :)

rust_src/remacs-lib/docfile.rs
shaleh: Fair catch. Update on the way.

PR #601, opened 2018-01-19
Port 'function'

shaleh:
Add tests too.

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -352,44 +352,6 @@ do_debug_on_call (Lisp_Object code, ptrdiff_t count)
    and temporaries from garbage collection while it needs them.
    The definition of `For' shows what you have to do.  */
 
-DEFUN ("function", Ffunction, Sfunction, 1, UNEVALLED, 0,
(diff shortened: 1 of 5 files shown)


src/eval.c
```diff
@@ -352,217 +352,6 @@ do_debug_on_call (Lisp_Object code, ptrdiff_t count)
...
-  Lisp_Object quoted = XCAR (args);
-
-  if (!NILP (XCDR (args)))
-    xsignal2 (Qwrong_number_of_arguments, Qfunction, Flength (args));
-
-  if (!NILP (Vinternal_interpreter_environment)
-      && CONSP (quoted)
-      && EQ (XCAR (quoted), Qlambda))
-    { /* This is a lambda expression within a lexical environment;
-	 return an interpreted closure instead of a simple lambda.  */
```
Wilfred: This is non-trivial code and I think these comments really help. Could we add them to the Rust too?

shaleh: Thank you. I had intended to restore them and failed to do so.

PR #602, opened 2018-01-20
Port defconst, special-variable-p

shaleh:
Add support in Lisp_Symbol. Port LOADHIST_ATTACH (as loadhist_attach).

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -251,7 +251,7 @@ pub fn function(args: LispObject) -> LispObject {
 }
 def_lisp_sym!(Qfunction, "function");
 
-// Make SYMBOL lexically scoped.
+/// Make SYMBOL lexically scoped.
 /// Internal function
 #[lisp_fn(name = "internal-make-var-non-special")]
 pub fn make_var_non_special(symbol: LispSymbolRef) -> bool {
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -1,15 +1,17 @@
 //! Generic Lisp eval functions
 
 use remacs_macros::lisp_fn;
-use remacs_sys::{Fcons, Fset};
-use remacs_sys::{QCdocumentation, Qclosure, Qfunction, Qlambda, Qnil, Qsetq, Qt,
-                 Qwrong_number_of_arguments};
+use remacs_sys::{Fcons, Fpurecopy, Fset, Fset_default};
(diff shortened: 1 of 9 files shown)


src/eval.c
```diff
@@ -767,62 +556,6 @@ usage: (defvar SYMBOL &optional INITVALUE DOCSTRING)  */)
...
-
-The optional DOCSTRING specifies the variable's documentation string.
-usage: (defconst SYMBOL INITVALUE [DOCSTRING])  */)
-  (Lisp_Object args)
-{
-  Lisp_Object sym, tem;
-
-  sym = XCAR (args);
-  Lisp_Object docstring = Qnil;
-  if (!NILP (XCDR (XCDR (args))))
```
Wilfred: I can't actually convince myself that this C is safe.

How do we know that `XCDR (args)` is a cons cell, so we can call `XCDR` again without crashing?

rust_src/src/eval.rs
```diff
@@ -1,91 +1,320 @@
...
+///
+/// The optional DOCSTRING specifies the variable's documentation string.
+/// usage: (defconst SYMBOL INITVALUE [DOCSTRING])
+#[lisp_fn(min = "2", unevalled = "true")]
+pub fn defconst(args: LispObject) -> LispSymbolRef {
+    let (sym, tail) = args.as_cons_or_error().as_tuple();
+
+    let mut docstring = LispObject::constant_nil();
+    if cdr(tail).is_not_nil() {
+        if cdr(cdr(tail)).is_not_nil() {
```
Wilfred: Unlike the C, I'm sure this is correct as `cdr` checks the type of its argument :+1:

src/eval.c
shaleh: We know that args is a cons cell because `defconst` is marked as 'unevalled'. So both the `XCAR` and first `XCDR` are safe. In fact, these were changed from calls to `Fcar` and `Fcdr` because the `XCAR` is faster. It is not clear to me that the second call to `XCDR` is similarly guaranteed to be safe.

Wilfred: Yep, it was the second XCDR I was worried about. Still, not a problem after this PR! :)

Wilfred: I played with this today, but I've failed to crash Emacs with it.

```
ELISP> (eval (cons 'defconst (cons 'x (cons 1 (cons 2 nil)))))
x
ELISP> (eval (cons 'defconst (cons 'x (cons 1 'not-a-cons))))
*** Eval error ***  Wrong type argument: listp, not-a-cons
```

I think because the arguments are unevalled, and we require at least two arguments, it's safe to call `XCDR(XCDR ...)`.

PR #603, opened 2018-01-22
Port let* aka letX

shaleh:


shaleh:
Yeah, the rebase needs to happen. Will work it out once the predecessors land.

shaleh:
The last commit is where the important bits are for this PR

shaleh:
Dropping in favor of the combined let PR

PR #604, opened 2018-01-24
Port let and let* aka letX

shaleh:


Wilfred:
Ooh, awesome patch! I've had a quick read through but I would like to go through more thoroughly. Let me know if any of my comments are unclear or wrong :)

PR #605, opened 2018-01-24
Port while

shaleh:


rust_src/src/eval.rs
```diff
@@ -460,6 +460,23 @@ pub fn lisp_let(args: LispCons) -> LispObject {
...
+/// The order of execution is thus TEST, BODY, TEST, BODY and so on
+/// until TEST returns nil.
+/// usage: (while TEST BODY...)
+#[lisp_fn(name = "while", c_name = "while", min = "1", unevalled = "true")]
+pub fn lisp_while(args: LispCons) -> LispObject {
+    let (test, body) = args.as_tuple();
+
+    while unsafe { eval_sub(test.to_raw()) } != Qnil {
+        unsafe { maybe_quit() };
+
```
db48x: I'm not a fan of having lots of blank lines in things, but this one here is the silliest.

shaleh: style difference. whitespace exists to make code easier to read. `maybe_quit` and `progn` body have nothing to do with each other so they are broke up.

shaleh:
Ah. Hm. I never considered people would do that. Yeah, I can switch it for a let binding.

shaleh:
Try again.

PR #606, opened 2018-01-24
Port macroexpand

shaleh:


shaleh:
I did not add new tests. I am open to suggestions.

shaleh:
There are some C functions for looking up aliases. But they all behave a little differently. I will double check there is not existing code we can massage.

As for the tests, those look reasonable. I will add them.

shaleh:
Because `macroexpand` takes an environment and reuses the temporary variables I am not seeing a close fit for a helper routine.

There are numerous places where the function of a symbol is investigated. For this `indirect_function` does a good job.

PR #607, opened 2018-01-25
Disable OSX until Travis CI is stable.

shaleh:


Wilfred:
We don't need to go this far. We could either:

* mark OS X as allowed to fail but still run it on Travis
* compile Remacs on Travis but don't bother running tests
* change the unreliable tests to not run on OS X, similar to #160

I think it's only a few tests that regularly fail. Have you seen broader failures?

shaleh:
The issue is that OSX is not loading at all. Not that the tests are always failing.

shaleh:
Look at my open PRs for the eval changes. They are all hung waiting for OSX to run.

shaleh:
Appears stable now. Dropping this PR.

PR #608, opened 2018-01-26
Port commandp

shaleh:
Add tests.

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -604,7 +604,7 @@ fn signal_error(msg: &str, arg: LispObject) -> ! {
 /// then strings and vectors are not accepted.
 #[lisp_fn(min = "1")]
 pub fn commandp(function: LispObject, for_call_interactively: bool) -> bool {
-    let mut if_prop = false;
+    let mut has_interactive_prop = false;
 
     let mut fun = indirect_function(function); // Check cycles.
     if fun.is_nil() {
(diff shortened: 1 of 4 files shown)


rust_src/src/eval.rs
```diff
@@ -585,4 +588,91 @@ fn signal_error(msg: &str, arg: LispObject) -> ! {
...
+            return true;
+        } else {
+            return if_prop;
+        };
+    } else if let Some(vl) = fun.as_vectorlike() {
+        // Bytecode objects are interactive if they are long enough to
+        // have an element whose index is COMPILED_INTERACTIVE, which is
+        // where the interactive spec is stored.
+        if vl.is_pseudovector(PseudovecType::PVEC_COMPILED)
+            && vl.pseudovector_size() > (COMPILED_INTERACTIVE as i64)
```
Wilfred: Is this assuming 64-bit? Should we say `COMPILED_INTERACTIVE as isize` instead?

test/rust_src/src/eval-tests.el
```diff
@@ -197,4 +197,26 @@
   (should (equal (macroexpand 'x)
                  'x)))
 
+(ert-deftest eval-tests--commandp-base ()
+  "Check (commandp) base cases"
+  (should (not (commandp 'commandp)))
+  (should (commandp 'query-replace))
+  (should (commandp "query-replace"))
+  (should (commandp "C-x b"))
```
Wilfred: We should probably add a test for vectors too.

rust_src/src/eval.rs
shaleh: Yes we should use `isize`. The compiler says 'i64' instead of 'EmacsInt' or 'isize'.

PR #610, opened 2018-01-28
Don't require OS X to build correctly

Wilfred:
We've had PRs just hang waiting for an OS X build on Travis, and we
also have a few flaky unit tests on OS X.

Run the OS X build, but don't fail the build if OS X fails. Mark the
build as green immediately after the other parts have succeeded.

This is an alternative approach to #607.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,3 +70,5 @@ jobs:
 notifications:
   fast_finish: true
   email: false
+  allow_failures:
+    - os: osx


Wilfred:
Ugh, this isn't working as expected:

apple_yellow

apple_running

I was expecting the build to be green even as OS X ran.

shaleh:
My problem was it was taking like 12 hours for the OSX to run. If it ever did. So while the Linux is all green the build is stuck. Travis is behaving better now.

Wilfred:
Given that Travis is currently behaving, I'm going to close this. We can always revisit this.

PR #614, opened 2018-01-31
Refactor list functions.

shaleh:


rust_src/src/lists.rs
```diff
@@ -114,22 +112,24 @@ pub fn nthcdr(n: EmacsUint, list: LispObject) -> LispObject {
 /// Return the Nth element of LIST.
 /// N counts from zero.  If LIST is not that long, nil is returned.
 #[lisp_fn]
-pub fn nth(n: EmacsInt, list: LispObject) -> LispObject {
+pub fn nth(n: EmacsUint, list: LispObject) -> LispObject {
```
Wilfred: What does this do for negative values? Currently, `(nth -100 some-list)` will return the first value of `some-list`.

PR #615, opened 2018-01-31
use tag_ptr instead of make_lisp_ptr

brotzeit:


rust_src/src/buffers.rs
```diff
@@ -1,12 +1,12 @@
 //! Functions operating on buffers.
 
-use libc::{self, c_int, c_uchar, c_void, ptrdiff_t};
+use libc::{self, c_int, c_uchar, ptrdiff_t};
 use std::{self, mem, ptr};
 
 use remacs_macros::lisp_fn;
 use remacs_sys::{EmacsInt, Lisp_Buffer, Lisp_Object, Lisp_Overlay, Lisp_Type, Vbuffer_alist,
                  MOST_POSITIVE_FIXNUM};
-use remacs_sys::{globals, make_lisp_ptr, set_buffer_internal, Fget_text_property};
```
shaleh: Do we want to remove `make_lisp_ptr` from remacs-sys as well?

brotzeit: I would rather add a comment that suggests to use `tag_ptr`.

PR #618, opened 2018-02-02
Port frame-pointer-visible-p from frame.c

Kyo91:
Ported over frame-pointer-visible-p as per issue #572.

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -8,7 +8,7 @@ use remacs_macros::lisp_fn;
 use remacs_sys::{selected_frame as current_frame, BoolBF, EmacsInt, Lisp_Frame, Lisp_Type};
 use remacs_sys::{fget_column_width, fget_iconified, fget_internal_border_width, fget_left_pos,
                  fget_line_height, fget_minibuffer_window, fget_output_method, fget_root_window,
-                 fget_terminal, fget_top_pos, fget_visible, frame_dimension, Fcons, Fselect_window};
+                 fget_terminal, fget_top_pos, fget_visible, frame_dimension, Fcons, Fselect_window, fget_pointer_invisible};
 use remacs_sys::{Qframe_live_p, Qframep, Qicon, Qns, Qpc, Qt, Qw32, Qx};
 
 use lisp::{ExternalPtr, LispObject};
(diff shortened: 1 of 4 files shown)


rust_src/src/frames.rs
```diff
@@ -279,4 +285,17 @@ pub fn frame_position(frame: LispObject) -> LispObject {
     }
 }
 
+/// Returns t if the mouse pointer displayed on FRAME is visible.
+/// Otherwise it returns nil. FRAME omitted or nil means the selected frame.
+/// This is useful when `make-pointer-invisible` is set
+#[lisp_fn(min = "0")]
+pub fn frame_pointer_visible_p(frame: LispObject) -> LispObject {
+    let frame_ref = frame_or_selected(frame);
+    if frame_ref.pointer_invisible() {
```
shaleh: You can actually return the bool directly. Just return `!frame_Ref.pointer_invisible()` directly and update the function definition to reflect `bool` instead of `LispObject`.

I really need to update the docs. Thanks for the reminder.

src/frame.h
```diff
@@ -715,6 +715,8 @@ bool
 fget_visible(const struct frame *f);
 bool_bf
 fget_iconified(const struct frame *f);
+bool_bf
+fget_pointer_invisible(const struct frame *f);
```
shaleh: Why is it `bool` in the C file and `bool_bf` in the header?

Kyo91: Changed header to bool (and make appropriate change in frames.rs as well). The documentation states:
```/* The type of bool bitfields.  Needed to compile Objective-C with
   standard GCC.  It was also needed to port to pre-C99 compilers,
   although we don't care about that any more.  */
```
So it sounds like bool_bf is legacy at this point.

rust_src/src/frames.rs
Kyo91: Fixed.

src/frame.h
shaleh: As discussed in brotzeit's PR there are reasons we need BoolBF. But there is no reason to export it from C to Rust here.

shaleh:
When you are closing an Issue, add that in the git commit message as a single line 'Closes #XXX'.

Issue #571, opened 2017-12-31
Port window-display-table


brotzeit:
```c
DEFUN ("window-display-table", Fwindow_display_table, Swindow_display_table,
       0, 1, 0,
       doc: /* Return the display-table that WINDOW is using.
WINDOW must be a live window and defaults to the selected one.  */)
  (Lisp_Object window)
{
  return decode_live_window (window)->display_table;
}
```

[decode_live_window](https://github.com/Wilfred/remacs/wiki/Porting-cookbook#functions)

PR #620, opened 2018-02-02
Port window-display-table (#571)

ckoparkar:
Tested it with:

```
(window-display-table)

;; https://www.gnu.org/software/emacs/manual/html_node/elisp/Display-Tables.html
(setq disptab (make-display-table))
(dotimes (i 32)
  (or (= i ?\t)
      (= i ?\n)
      (aset disptab i
            (vector (make-glyph-code ?^ 'escape-glyph)
                    (make-glyph-code (+ i 64) 'escape-glyph)))))
(aset disptab 127
      (vector (make-glyph-code ?^ 'escape-glyph)
              (make-glyph-code ?? 'escape-glyph)))

(set-window-display-table (get-buffer-window) disptab)

(window-display-table)
```

This is my very first PR. So there's a good chance that I might have missed something.
But I  successfully built Remacs, and the above snippet works the same on regular Emacs.
Should I also start porting `set-window-display-table` ?

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -478,6 +478,14 @@ pub fn minibuffer_window(frame: LispObject) -> LispObject {
     frame.minibuffer_window()
 }
 
+/// Return the display-table that WINDOW is using.
+/// WINDOW must be a live window and defaults to the selected one.
+#[lisp_fn(min = "0")]
+pub fn window_display_table(window: LispObject) -> LispObject {
+    let win = window_live_or_selected(window);
+    LispObject::from_raw(win.display_table)
+}
+
 pub fn window_wants_mode_line(window: LispWindowRef) -> bool {
     window.wants_mode_line()
 }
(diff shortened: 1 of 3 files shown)


shaleh:
Definitely give porting the set function a good.

Maybe give porting `window_display_table` a go as well.

shaleh:
See https://github.com/Wilfred/remacs/blob/master/test/file-organization.org

ckoparkar:
Great! I'll try and port the other display-table* functions as well.

Added the test-case to `test/rust_src/`. Verified that it passes with `make check`. Is this PR good to go then ?

shaleh:
:+1:

shaleh:
If you are closing an Issue, in the future add 'Closes #XXX' to the end of the Git commit.

ckoparkar:
Ah, ok. Thanks for merging this one @shaleh!

shaleh:
Welcome aboard @ckoparkar. Don't forget about the Gitter channel if you want to chat about porting.

PR #631, opened 2018-02-07
Port FUNCTIONP and Ffunctionp

shaleh:
Add tests.

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -16,7 +16,7 @@ use lisp::{defsubr, is_autoload};
 use lists::{assq, car, cdr, get, memq, put};
 use multibyte::LispStringRef;
 use obarray::loadhist_attach;
-use symbols::{symbol_function, LispSymbolRef};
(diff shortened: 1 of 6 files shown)


rust_src/src/eval.rs
```diff
@@ -707,4 +707,50 @@ pub fn autoload(
...
+    if let Some(sym) = obj.as_symbol() {
+        if fboundp(sym) {
+            obj = sym.get_indirect_function();
+
+            if let Some(cons) = obj.as_cons() {
+                if cons.car().eq_raw(Qautoload) {
+                    // Autoloaded symbols are functions, except if they load
+                    // macros or keymaps.
+                    let mut it = obj.iter_tails_safe();
+                    for _ in 0..4 {
```
Wilfred: This 4 is not obvious to me at all. Still, it's consistent with the C.

shaleh: Yeah, I agree. Definitely a wart to resolve.

Issue #629, opened 2018-02-06
make install fails (on fedora26)


DaanHoogland:
I'm a lurking user that has remacs as his standard emacs, on macos and fedora mostly. once in a while i pretend to develop and i have to re-install on my fedora, as I did last weekend. however the make install step failed with "error[E0554]: #![feature] may not be used on the stable release channel" I am using nightly but maybe not for one of the tools?
`$ src/remacs -Q`
starts a nice windows with a over featured editor but
`$ sudo make install`
gave me a very ugly
```
error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/dahn/prj/remacs/rust_src/remacs-sys/lib.rs:1:1
  |
1 | #![feature(const_size_of)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `remacs-sys`.
warning: build failed, waiting for other jobs to finish...
error: build failed
make[1]: *** [Makefile:610: ../rust_src/target/release/libremacs_lib.a] Error 101

make[1]:

 Leaving directory '/home/dahn/prj/remacs/src'
make: *** [Makefile:426: src] Error 2

```
I realize the problem might be between chair and keyboard, if so please help. it seems strange to me that a compilation error occurs during install and not during normal build.

shaleh:
I am looking at this currently.

DaanHoogland:
I may well close this myself soon; rustup installs the nightly version for the current user not available under sudo, which in turn is needed to install. remains the question why compile is needed during install.

DaanHoogland:
thanks @shaleh , I just did a sudo make and that didn't go as well as make so i tried sudo rustc --version showing part of the problem

shaleh:
AH. yes, the annoyance of per user Rust toolchain setup.

Once you are satisfied with a working solution would you mind submitting a PR for updating the documentation?

DaanHoogland:
of course, but can you shine a light as to why make install compiles even when make finished succesfully?

shaleh:
No, I cannot. I was just about to look at that.

shaleh:
hmmm.

If I run it like this (as me):

    $ ./configure --prefix=~/tmp/remacs-install --enable-rust-debug
    $ make -j8
    $ make install  # works because it is my path so permissions are OK

I do not see a second compilation. However....

    $ make -j8
    $ sudo su -
    $ cd back to repo
[...]

PR #632, opened 2018-02-07
issue #629 followup: instruction on cargo build and install usage

DaanHoogland:
@shaleh I found only refs to this in the INSTALL file so no changes in readme

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -33,10 +33,11 @@ some of the steps manually.  The more detailed description in the other
 sections of this guide will help you do that, so please refer to those
 sections if you need to.
 
-  0. Before you start: Remacs uses cargo for it's build process. if
-     you build as an unpriviledged user and then install as an
+  0. Before you start: Remacs uses cargo for its build process. if you
+     build as an unpriviledged user and then install as an
      administrator cargo will rebuild remacs. Make sure you have the
-     right nightly version installed for both users.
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -33,6 +33,11 @@ some of the steps manually.  The more detailed description in the other
 sections of this guide will help you do that, so please refer to those
 sections if you need to.
 
+  0. Before you start: Remacs uses cargo for it's build process. if
```
shaleh: `It's` is a conjunction. You want `its`.

INSTALL
```diff
@@ -33,6 +33,11 @@ some of the steps manually.  The more detailed description in the other
...
+     you build as an unpriviledged user and then install as an
+     administrator cargo will rebuild remacs. Make sure you have the
+     right nightly version installed for both users.
+
```
shaleh: Maybe mention `rust-toolchain` as the source of that version?

INSTALL
DaanHoogland: 👍 will tweak

INSTALL
DaanHoogland: hm, not sure how elaborate to be here,
should i just mention rustc or refer it?

shaleh: How about

"Make sure you have the right nightly version as found in the rust-toolchain file." 

Or something like that.

PR #634, opened 2018-02-08
Port autoload-do-load

shaleh:


rust_src/src/eval_macros.rs
```diff
@@ -76,6 +83,7 @@ macro_rules! error {
     }};
     ($fmtstr:expr, $($arg:expr),*) => {{
         let formatted = format!($fmtstr, $($arg),*);
+        #[allow(unused_unsafe)]
```
Wilfred: Why are these now needed?

shaleh: Because I have unsafe blocks calling the macro. The compiler complains that the unsafe in the macro is unused because it is in the external unsafe block. But clearly that is not always the case so we cannot remove the unsafe from the macro. Adding these markers allows the compiler to ignore the double unsafe.

PR #635, opened 2018-02-08
Port require

shaleh:
Also port un_autoload.
Make Qnil and friends const instead of static which allows them in
turn to be used in other statics.

### DIFF
--- a/src/fns.c
+++ b/src/fns.c
@@ -1858,125 +1858,6 @@ advisable.  */)
   return ret;
 }
 
-/* `require' and its subroutines.  */
-
-/* List of features currently being require'd, innermost first.  */
-
-static Lisp_Object require_nesting_list;
-
-static void
-require_unwind (Lisp_Object old_value)
-{
-  require_nesting_list = old_value;
-}
-
-DEFUN ("require", Frequire, Srequire, 1, 3, 0,
-       doc: /* If feature FEATURE is not loaded, load it from FILENAME.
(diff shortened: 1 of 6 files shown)


shaleh:
Well this is annoying. It runs here.

src/fns.c
```diff
@@ -3591,9 +3471,6 @@ compilation.  */);
   staticpro (&string_char_byte_cache_string);
   string_char_byte_cache_string = Qnil;
 
-  require_nesting_list = Qnil;
-  staticpro (&require_nesting_list);
```
Wilfred: You don't seem to be protecting the variable from GC on the Rust side.

I've found some docs here: https://www.gnu.org/software/emacs/manual/html_node/elisp/Writing-Emacs-Primitives.html

shaleh: Aye. This probably explains the failures in CI that I do not see locally.

I am preparing a PR with a rewrite of the build.rs that will also handle protected statics. The current build code shows my lack of experience coding in Rust. Once that PR is blessed I will update this PR.

shaleh:
Waiting on #641.

src/fns.c
Wilfred: I did see some interesting bugs when I prototyped porting `cons` (which I still need to fix). If you deliberately create a some large lists in a loop you can trigger a few GCs to flush out issues.

shaleh: Good suggestion. A similar large list is what is holding up the map PR.

PR #638, opened 2018-02-09
Upgrade of packages fails because catget is not in sync.

shaleh:


shaleh:
Sadly @db48x it looks like `catgets` is not a valid package despite being named in the error message.

birkenfeld:
This:
```
:: msys2-runtime and catgets are in conflict. Remove catgets? [y/N] 
```
looks to me like `catgets` is an old packages that doesn't exist anymore.

There's probably an option to `pacman` to answer that query with `y` by default, therefore preventing the error exit.

shaleh:
The `--noconfirm` option is being used currently. This selects the default. Which is NO in this case. In theory we could use the 'yes' command. But that might just fail on something else.

PR #640, opened 2018-02-13
Port the global-key-binding function

accidentalrebel:
Went ahead and tackled this one to familiarize myself with the porting. Thankfully, someone already did `local-key-binding` which guided me on how to do it.

Feel free to point out if I made any mistakes so I can correct accordingly. Thanks!


accidentalrebel:
Yeah, i was looking at the other functions in keymap.c and there are a lot that are easy enough to figure out. Thanks!

shaleh:
Do consider adding tests to test/rust_src/src/ if it makes sense.

PR #642, opened 2018-02-15
Ported use-global-map and use-local-map with tests

### LINKED ISSUE #570: Port use-global-map and use-local-map
```c

DEFUN ("use-global-map", Fuse_global_map, Suse_global_map, 1, 1, 0,
       doc: /* Select KEYMAP as the global keymap.  */)
  (Lisp_Object keymap)
{
  keymap = get_keymap (keymap, 1, 1);
  current_global_map = keymap;

  return Qnil;
}

DEFUN ("use-local-map", Fuse_local_map, Suse_local_map, 1, 1, 0,
       doc: /* Select KEYMAP as the local keymap.
If KEYMAP is nil, that means no local keymap.  */)
  (Lisp_Object keymap)
{
  if (!NILP (keymap))
    keymap = get_keymap (keymap, 1, 1);

  bset_keymap (current_buffer, keymap);

  return Qnil;
}
```

accidentalrebel:
Ported `use-global-map` and `use-local-map` from `keymap.c`. Closes #570.

Also made `keymaps-test.el`. For now it only tests `current-local-map`, `use-local-map`, `current-global-map`, and `use-global-map`. I plan to add more tests as I port more of the `keymap.c` functions.


brotzeit:
btw, you can  [close issues automatically](https://help.github.com/articles/closing-issues-using-keywords/) when a PR is merged.

test/rust_src/src/keymaps-test.el
```diff
@@ -0,0 +1,21 @@
...
+
+(ert-deftest keymap-global-tests ()
+  (let ((sample-keymap '(keymap
+                         (3 keymap
+                            ;; C-c C-z
+                            (26 . emacs-version))))
+        (backup-keymap (current-global-map)))
+    (use-global-map sample-keymap)
+    (should (equal (current-global-map) '(keymap (3 keymap (26 . emacs-version)))))
+    (use-global-map backup-keymap)))
```
shaleh: You should also test what happens when NIL is passed like you do in `use-local-keymap` tests.

accidentalrebel: I tried it out and it seems the original c function errors when passed with `nil`. So it seems I might need to handle `nil` on my ported function.

In situations when you find bugs in Emacs and fixed it in Remacs, do you guys also submit a fix to Emacs?

accidentalrebel:
> btw, you can close issues automatically when a PR is merged.

That's reay cool. Will give that a try by editing the post.

test/rust_src/src/keymaps-test.el
shaleh: `(wrong-type-argument keymapp nil)` is exactly the error the C code says should happen when a nil is passed. The C code does not have a bug.

Look at roughly line 37 of the function `get_keymap` in keymap.c.

shaleh: That said, when we find bugs we send them upstream. If you want to include a patch that is up to you. Including a test in your patch is a good idea.

accidentalrebel: Aha! I should have checked where the error was coming from. This means the check `(should-error (use-global-map nil))` should do the trick. Will be fixing and updating my pull request in a bit. Thanks!

accidentalrebel:
Updated my pull request. Added a test where `use-global-map` is passed with nil.

shaleh:
The way to close an Issue automatically is in the Git commit message, not in the PR comments. I will update this when I merge it.

PR #653, opened 2018-02-18
port marker functions

brotzeit:


brotzeit:
I don't see the problem...

shaleh:
On this plus side it fails for me too. So it is not just a Travis issue.

brotzeit:
I didn't look at the struct order. Probably I need to use getters and setters.

shaleh:
Ah, that would do it :-(

shaleh:
We really need to get @db48x 's PRs in

brotzeit:
I didn't use an equivalent to `live_buffer`. That was the problem.

PR #655, opened 2018-02-20
Ported keymap-prompt with Tests.

### LINKED ISSUE #649: Port keymap-prompt from keymap.c
This is the next function that I'll be porting.

```c
DEFUN ("keymap-prompt", Fkeymap_prompt, Skeymap_prompt, 1, 1, 0,
       doc: /* Return the prompt-string of a keymap MAP.
If non-nil, the prompt is shown in the echo-area
when reading a key-sequence to be looked-up in this keymap.  */)
  (Lisp_Object map)
{
  map = get_keymap (map, 0, 0);
  while (CONSP (map))
    {
      Lisp_Object tem = XCAR (map);
      if (STRINGP (tem))
	return tem;
      else if (KEYMAPP (tem))
	{
	  tem = Fkeymap_prompt (tem);
	  if (!NILP (tem))
	    return tem;
	}
      map = XCDR (map);
    }
  return Qnil;
}
```

Things that are new to me are `CONSP`, `XCAR`, `STRINGP`, and `XCDR`. I bet can be able to figure these out.

accidentalrebel:
Successfully ported `keymap-prompt` from `keymap.c` with tests. Closes #649.

- The code works but I am unsure if the style I used is correct. 
- I learned that returning inside a for loop doesn't work so I had to pass it to the `result` variable so it can be returned by the function.
- I've also separated the Ffoo imports from the other function imports.

Thanks!

### DIFF
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -2,9 +2,9 @@
 
 use remacs_macros::lisp_fn;
 use remacs_sys::{current_global_map as _current_global_map, globals, EmacsInt, CHAR_META};
-use remacs_sys::{access_keymap, get_keymap, maybe_quit, Fcons, Fevent_convert_list, Ffset,
-                 Fmake_char_table, Fpurecopy, Fset};
+use remacs_sys::{Fcons, Fevent_convert_list, Ffset, Fmake_char_table, Fpurecopy, Fset};
 use remacs_sys::{Qkeymap, Qnil};
+use remacs_sys::{access_keymap, get_keymap, maybe_quit};
 
 use data::aref;
 use keyboard::lucid_event_type_list_p;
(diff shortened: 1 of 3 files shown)


shaleh:
`return` requires semicolons.

    pub fn keymap_prompt(map: LispObject) -> LispObject {
        let map = unsafe { LispObject::from_raw(get_keymap(map.to_raw(), false, false)) };
    
        for elt in map.iter_cars_safe() {
            if elt.is_string() {
                return elt;
            } else if keymapp(elt) {
                let tem = keymap_prompt(elt);
                if tem.is_not_nil() {
                    return tem;
                }
            }
        }
    
        LispObject::constant_nil()
    }

accidentalrebel:
Oh.. Let me try that.

accidentalrebel:
I tried it and it works! Fixed and pushed.

accidentalrebel:
Awesome. And the branch can easily be deleted. Sweet.

PR #657, opened 2018-02-21
Support commenting out functions

shaleh:


shaleh:
If it also closes on that line the script won't care. If the comment extends for multiple lines that could cause problems but would be an odd style. For now this handles all of the comments in the current code plus any I have made in my own while hacking about.

PR #658, opened 2018-02-21
Birkenfeld's work on vectors from the 'maps' branch.

shaleh:


shaleh:
The diff is hard to read. You are better off looking at the final file.

PR #660, opened 2018-02-21
Ported keymap-parent and set-keymap-parent from keymap.c with tests

### LINKED ISSUE #656: Port keymap-parent from keymap.c
I'll be porting `keymap-parent` and `keymap_parent` from `keymap.c` next.

```c
/* Return the parent map of KEYMAP, or nil if it has none.
   We assume that KEYMAP is a valid keymap.  */

static Lisp_Object
keymap_parent (Lisp_Object keymap, bool autoload)
{
  Lisp_Object list;

  keymap = get_keymap (keymap, 1, autoload);

  /* Skip past the initial element `keymap'.  */
  list = XCDR (keymap);
  for (; CONSP (list); list = XCDR (list))
    {
      /* See if there is another `keymap'.  */
      if (KEYMAPP (list))
	return list;
    }

  return get_keymap (list, 0, autoload);
}

DEFUN ("keymap-parent", Fkeymap_parent, Skeymap_parent, 1, 1, 0,
       doc: /* Return the parent keymap of KEYMAP.
If KEYMAP has no parent, return nil.  */)
  (Lisp_Object keymap)
{
  return keymap_parent (keymap, 1);
}
```

The for loop line ` for (; CONSP (list); list = XCDR (list))` is a bit confusing at first but I think it's just a different way of iterating through the keymap elements.

### LINKED ISSUE #661: Port set-keymap-parent and keymap_memberp
Here are the next set of functions I'm going to port next:

```c
/* Check whether MAP is one of MAPS parents.  */
static bool
keymap_memberp (Lisp_Object map, Lisp_Object maps)
{
  if (NILP (map)) return 0;
  while (KEYMAPP (maps) && !EQ (map, maps))
    maps = keymap_parent (maps, 0);
  return (EQ (map, maps));
}

/* Set the parent keymap of MAP to PARENT.  */

DEFUN ("set-keymap-parent", Fset_keymap_parent, Sset_keymap_parent, 2, 2, 0,
       doc: /* Modify KEYMAP to set its parent map to PARENT.
Return PARENT.  PARENT should be nil or another keymap.  */)
  (Lisp_Object keymap, Lisp_Object parent)
{
  Lisp_Object list, prev;

  /* Flush any reverse-map cache.  */
  where_is_cache = Qnil; where_is_cache_keymaps = Qt;

  keymap = get_keymap (keymap, 1, 1);

  if (!NILP (parent))
    {
      parent = get_keymap (parent, 1, 0);

      /* Check for cycles.  */
      if (keymap_memberp (keymap, parent))
	error ("Cyclic keymap inheritance");
    }

  /* Skip past the initial element `keymap'.  */
  prev = keymap;
  while (1)
    {
      list = XCDR (prev);
      /* If there is a parent keymap here, replace it.
	 If we came to the end, add the parent in PREV.  */
      if (!CONSP (list) || KEYMAPP (list))
	{
	  CHECK_IMPURE (prev, XCONS (prev));
	  XSETCDR (prev, parent);
	  return parent;
	}
      prev = list;
    }
}
```

Notes:
- Need to check how to throw an error the _remacs way_ when porting the line `error ("Cyclic keymap inheritance");`
- New function calls I have not encountered before: XSETCDR, CONSP, CHECK_IMPURE

accidentalrebel:
Successfully ported `keymap_parent` and `keymap-parent` from keymap.c. Closes #656 .

Edit: Added port for `set-keymap-parent` and `keymap_memberp`. Closes #661.

Notes:
- Created a `keymap_parent_internal` function that  works with directly with `LispObject`. Similar to [this](https://github.com/Wilfred/remacs/pull/496#issuecomment-346266814).
- Both `keymap_parent` and `keymap-parent` uses `keymap_parent_internal`.
- Hopefully in the future, when all `keymap.c` functions are ported, we can remove the `extern C` function and then rename `keymap_parent_internal` to `keymap_parent`.


rust_src/src/keymap.rs
```diff
@@ -3,9 +3,11 @@
 use remacs_macros::lisp_fn;
 use remacs_sys::{current_global_map as _current_global_map, globals, EmacsInt, CHAR_META};
 use remacs_sys::{Fcons, Fevent_convert_list, Ffset, Fmake_char_table, Fpurecopy, Fset};
-use remacs_sys::{Qkeymap, Qnil};
 use remacs_sys::{access_keymap, get_keymap, maybe_quit};
```
shaleh: Should not be an empty line here

rust_src/src/keymap.rs
```diff
@@ -50,6 +52,33 @@ pub fn keymapp(object: LispObject) -> bool {
...
 }
 
+/// Return the parent map of KEYMAP, or nil if it has none.
+/// We assume that KEYMAP is a valid keymap.
+#[no_mangle]
+pub extern "C" fn keymap_parent(keymap: Lisp_Object, autoload: bool) -> Lisp_Object {
+    let map = keymap_parent_internal(LispObject::from_raw(keymap), autoload);
+    map.to_raw()
+}
+
```
shaleh: Why have the `_internal` function? It looks like if you leave `keymap_parent` returning a `Lisp_Object`. The `lisp_fn` can call it and convert it to `LispObject`.

rust_src/src/keymap.rs
```diff
@@ -50,6 +52,33 @@ pub fn keymapp(object: LispObject) -> bool {
...
+fn keymap_parent_internal(keymap: LispObject, autoload: bool) -> LispObject {
+    let mut list = LispObject::constant_nil();
+    let map = unsafe { LispObject::from_raw(get_keymap(keymap.to_raw(), true, autoload)) };
+    for elt in map.iter_tails_safe() {
+        list = elt.cdr();
+        if keymapp(list) {
+            return list;
+        }
+    }
```
shaleh: Could the `find` method of `Iterator` be used here instead?

src/keymap.c
```diff
@@ -3481,8 +3452,6 @@ be preferred.  */);
   where_is_cache = Qnil;
   staticpro (&where_is_cache);
   staticpro (&where_is_cache_keymaps);
-
```
shaleh: Please keep the empty line here.

accidentalrebel: Got it.

rust_src/src/keymap.rs
accidentalrebel: Got it.

rust_src/src/keymap.rs
accidentalrebel: After reviewing it, I think you are right. I guess my initial idea on why I did this is to avoid converting to LispObject back and forth.

rust_src/src/keymap.rs
accidentalrebel: Yes. And it seems that it would even be more concise. Will be refactoring this.

accidentalrebel:
I was able to finish porting `set-keymap-parent` and `keymap_memberp` functions (#661)  and combined it in this pull request (I wasn't sure if I should make a new pull request for this or not, if I was supposed to then please tell me so I can correct in the future).

rust_src/src/keymap.rs
```diff
@@ -73,6 +73,59 @@ pub fn keymap_parent_lisp(keymap: LispObject) -> LispObject {
...
+        // If we came to the end, add the parent in PREV.
+        if !list.is_cons() || keymapp(list) {
+            prev.check_impure();
+            prev.set_cdr(parent);
+            return parent;
+        }
+        prev = list.as_cons_or_error();
+    }
+}
+
```
shaleh: How about this instead.

        loop {
            list = prev.cdr();

            match list.as_cons() {
                None => break,
                Some(cons) => if keymapp(list) {
                    break;
                } else {
                    prev = cons;
                }
            }
        }
        prev.check_impure();
        prev.set_cdr(parent);
        return parent;

src/keymap.c
```diff
@@ -3452,8 +3402,7 @@ be preferred.  */);
...
```
shaleh: See those `staticpro`? That protects the variable from the garbage collector (aka GC). Checkout PR #641

src/keymap.c
```diff
@@ -75,17 +75,18 @@ Lisp_Object control_x_map;	/* The keymap used for globally bound
 				   bindings when spaces are not encouraged
 				   in the minibuf.  */
 
+/* Hash table used to cache a reverse-map to speed up calls to where-is.  */
+Lisp_Object where_is_cache;
+
```
shaleh: Need protection from GC, see below.

src/keymap.c
accidentalrebel: I see. So it seems I need to use `declare_GC_protected_static` to protect the variable. Alright!

shaleh: Correct. Remove the remaining C bits and update the Rust.

shaleh: Once you are confident in this PR, would you update the porting section of the wiki?

rust_src/src/keymap.rs
```diff
@@ -73,6 +73,59 @@ pub fn keymap_parent_lisp(keymap: LispObject) -> LispObject {
...
 }
 
+/// Check whether MAP is one of MAPS parents.
+#[no_mangle]
+pub extern "C" fn keymap_memberp(map: Lisp_Object, maps: Lisp_Object) -> bool {
+    let map = LispObject::from_raw(map);
+    let mut maps = LispObject::from_raw(maps);
+    if map.is_nil() {
+        return false;
+    }
```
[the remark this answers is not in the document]
accidentalrebel: I see what you mean. `lisp.rs` has a section with a label `// The three Emacs Lisp comparison functions.`. I'll fix and use the appropriate comparison function. Thanks!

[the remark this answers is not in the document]
shaleh: :+1:

rust_src/src/keymap.rs
```diff
@@ -73,6 +73,59 @@ pub fn keymap_parent_lisp(keymap: LispObject) -> LispObject {
...
+    while keymapp(maps) && map != maps {
+        maps = LispObject::from_raw(keymap_parent(maps.to_raw(), false));
+    }
+    map == maps
+}
+
+/// Modify KEYMAP to set its parent map to PARENT.
+/// Return PARENT.  PARENT should be nil or another keymap.
+#[lisp_fn]
+pub fn set_keymap_parent(keymap: LispObject, parent: LispObject) -> LispObject {
```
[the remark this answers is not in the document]
accidentalrebel: I'm sorry, I do not follow what you mean here.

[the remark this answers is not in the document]
shaleh: The next 3 lines define variables with no value long before they are used. This is not idiomatic Rust.

src/keymap.c
accidentalrebel: Noted.

rust_src/src/keymap.rs
shaleh: ok, maybe no 'return' :-)

rust_src/src/keymap.rs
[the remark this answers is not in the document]
accidentalrebel: Ah I see. So I need to find a way to only define them only when they are needed.

rust_src/src/keymap.rs
shaleh: The code you had checked `is_cons` twice. Once with explicit `is_cons` and once with `as_cons_or_error`. The original code did not have that error. Instead it implicitly claimed that prev would always be a `Lisp_Cons`. My code makes this explicit and removes the double check. If there is another way to write this cleanly I am open to it.

In our opinion one of the reasons the Rust port has value is it removes the need for the assertions because we can know what types are involved. At least some of the time. This chunk is a demonstration of why we do not usually need `XCDR`.

rust_src/src/keymap.rs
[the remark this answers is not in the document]
accidentalrebel: That makes sense. Got it!

rust_src/src/keymap.rs
accidentalrebel: I get what you mean. I'll be revisiting this later to fully so I can fully comprehend. Thanks!

shaleh: BTW, If the return lives in the `loop` that is ok. I only pulled it out to simplify the code I wrote.

accidentalrebel:
Alright. All requested changes have been fixed. 👍

src/keymap.c
shaleh: This creates another `where_is_cache` and `where_is_cache_keymaps`.

accidentalrebel: Oh. When i removed this before it complained that the variable is not declared. How do I expose the protected variable from Rust to C?

shaleh: You can't, it is static. You could export a function from Rust to access them. Or you could port the rest of the functions that use the variables.

accidentalrebel: I see, that makes sense. I'll see which option is more viable for me at the moment. Thanks!

accidentalrebel:
Alright. I've removed the declarations of `where_is_cache` and `where_is_cache_keymaps` from `keymap.c`. The declarations are now in `keymap.rs` but in order for it to be accessed by `keymap.c` I opted to use get and set functions. 

My hope is that when I've fully ported the functions that use the variables above to Rust, then I can remove the get and set functions completely. For the meantime, this is temporary.

rust_src/src/keymap.rs
```diff
@@ -16,6 +17,40 @@ pub fn Ctl(c: char) -> i32 {
...
+
+/// Allows the C code to get the value of `where_is_cache`
+#[no_mangle]
+pub extern "C" fn get_where_is_cache() -> Lisp_Object {
+    unsafe { where_is_cache }
+}
+
+/// Allows the C code to set the value of `where_is_cache`
+#[no_mangle]
+pub extern "C" fn set_where_is_cache(val: Lisp_Object) -> () {
```
shaleh: There is no need to indicate an empty return here. We do that for `lisp_fn` since ALL `lisp_fn` return something.

rust_src/src/keymap.rs
```diff
@@ -50,6 +85,86 @@ pub fn keymapp(object: LispObject) -> bool {
     map.is_not_nil()
 }
 
+/// Return the parent map of KEYMAP, or nil if it has none.
+/// We assume that KEYMAP is a valid keymap.
+#[no_mangle]
+pub extern "C" fn keymap_parent(keymap: Lisp_Object, autoload: bool) -> Lisp_Object {
+    let map = unsafe { LispObject::from_raw(get_keymap(keymap, true, autoload)) };
```
shaleh: My only concern here is both `find` and `last` will walk the list. So this will be a little less efficient than the previous hand coded loop. Do we care? I do not know.

rust_src/src/keymap.rs
db48x: The `-> ()` notation is optional even for functions marked with `lisp_fn`; they're equivalent.

shaleh: It is a style we have, not a Rust requirement.

accidentalrebel: Noted. Thanks!

rust_src/src/keymap.rs
```diff
@@ -50,6 +85,86 @@ pub fn keymapp(object: LispObject) -> bool {
...
+    } else if let Some(last) = map.iter_tails_safe().last() {
+        unsafe { get_keymap(last.cdr().to_raw(), false, autoload) }
+    } else {
+        Qnil
+    }
+}
+
+/// Return the parent keymap of KEYMAP.
+/// If KEYMAP has no parent, return nil.
+#[lisp_fn(name = "keymap-parent")]
```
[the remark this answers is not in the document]
accidentalrebel: Ah, yes. I've been wondering where `c_name` could be useful. I guess this is it.

rust_src/src/keymap.rs
accidentalrebel: Hmm. I think this would be a concern for really large keymaps. And keymaps can get really large sometimes. Actually my old implementation has a `find` and `tail` too. I'll try looking for a different approach, if I can come up with one.

rust_src/src/keymap.rs
```diff
@@ -50,6 +85,86 @@ pub fn keymapp(object: LispObject) -> bool {
...
+
+/// Modify KEYMAP to set its parent map to PARENT.
+/// Return PARENT.  PARENT should be nil or another keymap.
+#[lisp_fn]
+pub fn set_keymap_parent(keymap: LispObject, parent: LispObject) -> LispObject {
+    // Flush any reverse-map cache
+    unsafe {
+        set_where_is_cache(Qnil);
+        where_is_cache_keymaps = Qt;
+    }
```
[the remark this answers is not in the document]
accidentalrebel: Ah yes, this is an error on my part. I tested calling `set_where _is_cache` from Rust and forgot to change it back. Will fix.

rust_src/src/keymap.rs
accidentalrebel: So I changed the implementation to this:

```rust
#[no_mangle]
pub extern "C" fn keymap_parent(keymap: Lisp_Object, autoload: bool) -> Lisp_Object {
    let map = unsafe { LispObject::from_raw(get_keymap(keymap, true, autoload)) };
    let mut current = LispObject::constant_nil();
    for elt in map.iter_tails_safe() {
        current = elt.cdr();
        if keymapp(current) {
            return current.to_raw();
        }
    }
    unsafe { get_keymap(current.to_raw(), false, autoload) }
}
```

The idea is that it only needs to walk the map once with `map.iter_tails_safe` not unlike the previous implementation where it would go through `find` and `last`. Tell me what you think. Thanks!

accidentalrebel:
Made fixes and changes according to the reviews. Also updated the test functions so that it would follow the function naming conventions.

rust_src/src/keymap.rs
accidentalrebel: Also, I looked at my first implementation and it is similar to this new one that I made. So I guess, if efficiency is concerned, this would be the way to go. ;)

shaleh: Yeah, as much as I like the other version this is the better choice.

accidentalrebel:
What can I do to help push the review forward? Should I make a separate pull request with just the recent changes? I imagine that this would make the pull request cleaner and hopefully easier to review.

Thanks!

brotzeit:
> Should I make a separate pull request with just the recent changes?

I try to do it this way. But it can be annoying if the PR depends on your previous/not yet merged PR.

shaleh:
@accidentalrebel if you want to clean up the PR you can `git rebase -i` against upstream/master and a) remove any commit that is not part of this PR but from a previous PR (usually at the beginning of the commits list) and b) squash this PR's commits down to one commit with a clear commit message and a Closes # with the right Issue at the end.

accidentalrebel:
Alright. Renamed `keymap-test.el` to `keymap-tests.el`.

@shaleh I might do that next time. Last time I tried rebasing my local repo became a mess. Need time to properly learn about it so I can apply it in the future.

shaleh:
Like this:

    $ git checkout master
    $ git remote add upstream https://github.com/Wilfred/remacs
    $ git pull upstream master
    $ git checkout my-branch
    $ git rebase -i upstream/master
       remove commits that do not belong
       squash the commits into a clean set
    $ git push --force  # needed because you rewrote the history

shaleh:
I have git load all rebases in Emacs. `k` removes a commit, `s` squashes a commit. `C-c C-c` commits the rebase.

accidentalrebel:
Now that you've outlined it that way, it seems less daunting. Let me try it.

shaleh:
The 'remote add' only needs to be done once.

shaleh:
BTW, this little script I have found to be very useful. https://github.com/shaleh/git-bits/blob/master/both-modified

accidentalrebel:
Your step by step instructions are helpful (This [video](https://www.youtube.com/watch?v=f1wnYdLEpgI) helped a lot too) but there are so many conflicts that needs to be dealt with after running rebase that I still end up with a jumbled mess. 

For now, can we continue with the merging without doing the rebase? I'll be sure to do the rebase workflow next time on a smaller port that would, hopefully, be easier to rebase.

Thanks!

shaleh:
Yeah, no worries.

PR #662, opened 2018-02-22
add function current_buffer_bounds

brotzeit:


brotzeit:
> What is the rationale of this function?

I should have waited with this, but I decided to try this function when I saw this snippet in `set_marker_internal`:

```
      charpos = clip_to_bounds
	(restricted ? BUF_BEGV (b) : BUF_BEG (b), charpos,
	 restricted ? BUF_ZV (b) : BUF_Z (b));
      /* Don't believe BYTEPOS if it comes from a different buffer,
	 since that buffer might have a very different correspondence
	 between character and byte positions.  */
      if (bytepos == -1
	  || !(MARKERP (position) && XMARKER (position)->buffer == b))
	bytepos = buf_charpos_to_bytepos (b, charpos);
      else
	bytepos = clip_to_bounds
	  (restricted ? BUF_BEGV_BYTE (b) : BUF_BEG_BYTE (b),
	   bytepos, restricted ? BUF_ZV_BYTE (b) : BUF_Z_BYTE (b));
```
And I thought the function could also be used for other uses of `clip_to_bounds`.


> The code you replaced was clear and had no extra logic.

I think't it's not really clear for newcomers if they see something like this 
```

        let clipped_charpos = clip_to_bounds(cur_buf.begv, charpos as EmacsInt, cur_buf.zv);
        let clipped_bytepos = clip_to_bounds(cur_buf.begv_byte, bytepos as EmacsInt, cur_buf.zv_byte);
```

> Have you considered defining convenience methods on buffer ref instead? You could have a visible_byte_range method, for example.

I did. But if you remember I already complained half a year ago about this behaviour. Only the current buffer's positions are used, even if it's a different buffer. So I thought when only the current buffer matters, it would be somehow irritating if I make a method that seems to provide the possibility that it could be used on a buffer other than the current. In other words:  I don't want to provide the option to do it incorrectly(I already mentioned that this is not my opinion ;D).

rust_src/src/util.rs
```diff
@@ -13,3 +15,40 @@ pub extern "C" fn clip_to_bounds(lower: isize, num: EmacsInt, upper: isize) -> i
         upper
     }
 }
+
+/// Get boundaries of the current buffer.
+/// If BYTE is true, return byte positions.
+/// If RESTRICTED is true, limit the bounds by the buffer's visible part.
+pub fn current_buffer_bounds(byte: bool, restricted: bool) -> (isize, isize) {
```
db48x: I agree with Sean; this isn't great. Instead of boolean arguments, just split it into four functions. That makes them easy to understand, and lets them have names that mean something. Also, instead of "restricted", you could be more specific and call it "visible".

shaleh: I was thinking the same thing. Named functions would be more clear

brotzeit:
I will make a new PR containing set_marker_internal.

PR #666, opened 2018-02-25
port set_marker_internal

brotzeit:


rust_src/src/marker.rs
```diff
@@ -341,12 +337,121 @@ pub extern "C" fn marker_byte_position(marker: Lisp_Object) -> ptrdiff_t {
...
+    let buf = live_buffer(buffer.to_raw());
+    let mut m = marker.as_marker_or_error();
+
+    // Set MARKER to point nowhere if BUFFER is dead, or
+    // POSITION is nil or a marker points to nowhere.
+    if position.is_nil()
+        || position
+            .as_marker()
+            .map_or(false, |m| !m.buffer().is_some()) || !buf.is_some()
+    {
```
shaleh: `!is_some` is better written as `is_none`.

shaleh: Twice this code potentially converts to marker and then extracts the buffer. Would this not read better if that buffer were stored in a temporary?

rust_src/src/marker.rs
```diff
@@ -341,12 +337,121 @@ pub extern "C" fn marker_byte_position(marker: Lisp_Object) -> ptrdiff_t {
...
+        unsafe {
+            unchain_marker(m.as_mut());
+        }
+
+    // Optimize the special case where we are copying the position of
+    // an existing marker, and MARKER is already in the same buffer.
+    } else if position.as_marker().map_or(false, |p| p.buffer() == buf) && m.buffer() == buf {
```
shaleh: Consistency. This time around the marker is named `p` but it is named `m` in the previous if.

rust_src/src/marker.rs
```diff
@@ -341,12 +337,121 @@ pub extern "C" fn marker_byte_position(marker: Lisp_Object) -> ptrdiff_t {
...
+    {
+        bytepos = unsafe { buf_charpos_to_bytepos(buf.as_mut(), charpos) };
+    } else {
+        let beg = buffer_beg_byte(buf, restricted);
+        let end = buffer_end_byte(buf, restricted);
+        bytepos = clip_to_bounds(beg, bytepos as EmacsInt, end);
+    }
+    attach_marker(marker.as_mut(), buf.as_mut(), charpos, bytepos);
+}
+
```
shaleh: Would the following functions work better as methods on `LispBuffer`?

rust_src/src/marker.rs
shaleh: Good point about using a more specific type in the argument list.

I missed that marker and position could be in different buffers. Thank you.

rust_src/src/marker.rs
brotzeit: I thought it would be bad to only pass booleans to a function/method ?

rust_src/src/marker.rs
brotzeit: I changed this to `p` because I thought it might be confusing with 2 `m`s and this way it's clear that it belongs to `position`. Even if you can see that the closure shadows `m`, I think it's easier to read.

shaleh: Fair enough.

rust_src/src/marker.rs
shaleh: The boolean now is no clearer. `(self, bool)` or `(LispBufferRef, bool)` are basically the same.

One option would be to pass an Enum instead. `BUFFER_VISIBLE` or `BUFFER_NOT_VISIBLE` maybe?

rust_src/src/marker.rs
brotzeit: > A minor improvement would be to declare marker as a LispMarkerRef in the argument list, to convert automatically. Similarly, position could be an Option<LispMarkerRef>.

I thought this only works with lisp functions ?

shaleh: automagically yes. But you can do it explicitly.

shaleh: The point of a type system is to have strong fences.

rust_src/src/marker.rs
brotzeit: I've decided to use a bool because this way I don't have to check `restricted`. Or do I miss something ?

rust_src/src/marker.rs
brotzeit: > A minor improvement would be to declare marker as a LispMarkerRef in the argument list, to convert automatically. 

If I pass a `LispMarkerRef` I have to do the type checking in the functions that are calling `set_marker_internal` ? I don't know how to convert it automatically here.

> Similarly, position could be an Option<LispMarkerRef>.

`position` can be an nil, a marker or an `EmacsInt`. I'm not sure how to do this with an ` Option<LispMarkerRef>`.

rust_src/src/marker.rs
```diff
@@ -341,12 +337,125 @@ pub extern "C" fn marker_byte_position(marker: Lisp_Object) -> ptrdiff_t {
 /// whether BUFFER is a buffer object and return buffer pointer
 /// corresponding to BUFFER if BUFFER is live, or NULL otherwise.
 #[no_mangle]
-pub extern "C" fn live_buffer(buffer: Lisp_Object) -> *mut Lisp_Buffer {
-    let mut b = LispObject::from_raw(buffer).as_buffer_or_current_buffer();
+pub extern "C" fn live_buffer(buffer: Lisp_Object) -> Option<LispBufferRef> {
```
shanavas786: Is this still used in C

rust_src/src/marker.rs
```diff
@@ -341,12 +337,125 @@ pub extern "C" fn marker_byte_position(marker: Lisp_Object) -> ptrdiff_t {
...
-        b.as_mut()
+        Some(b)
+    } else {
+        None
+    }
+}
+
+impl LispObject {
+    pub fn has_buffer(self) -> bool {
+        self.as_marker().map_or(false, |m| m.buffer().is_none())
```
shanavas786: shoudn't it be is_some() ?

rust_src/src/marker.rs
```diff
@@ -341,12 +337,125 @@ pub extern "C" fn marker_byte_position(marker: Lisp_Object) -> ptrdiff_t {
...
+    position: LispObject,
+    buffer: LispObject,
+    restricted: bool,
+) -> LispObject {
+    let buf = live_buffer(buffer.to_raw());
+    let mut m = marker.as_marker_or_error();
+
+    // Set MARKER to point nowhere if BUFFER is dead, or
+    // POSITION is nil or a marker points to nowhere.
+    if position.is_nil() || position.has_buffer() || buf.is_none() {
```
shanavas786: negate position.has_buffer()

brotzeit:
Oops. Thanks @shanavas786. I will fix it.

brotzeit:
Ok I give up. I know the problem is the function `has_buffer` and therefore I assume it's the method `buffer`. But I don't understand what's the problem here...

brotzeit:
I'm sure that's the problem as I replaced the condition with a c function and it worked...

Issue #670, opened 2018-02-27
Error startup with init.el


guoxiangyang:
```commit c001681bd9eafe1bbfec226f56ab503810982bd0``
```
File mode specification error: (error avl-tree--dummyroot accessing a non-avl-tree-)               
error in process sentinel: org-element--cache-find: avl-tree--dummyroot accessing a non-avl-tree-  
error in process sentinel: avl-tree--dummyroot accessing a non-avl-tree-                           
scroll-down-command: Beginning of buffer                                                           
scroll-down-command: Beginning of buffer                                                           
```

```
gxy@local:~/work/remacs$ lsb_release  -a
LSB Version:    core-9.20160110ubuntu0.2-amd64:core-9.20160110ubuntu0.2-noarch:printing-9.20160110ubuntu0.2-amd64:printing-9.20160110ubuntu0.2-noarch:security-9.20160110ubuntu0.2-amd64:security-9.20160110ubuntu0.2-noarch
Distributor ID: LinuxMint
Description:    Linux Mint 18.3 Sylvia
Release:        18.3
Codename:       sylvia
```

shaleh:
@guoxiangyang have you confirmed that commit causes your issue? Could you please describe more of your configuration? Maybe provide a small test case?

shaleh:
Is this file current? https://github.com/guoxiangyang/.emacs.d/blob/master/init.el

guoxiangyang:
It's strange that I just receive the notification email right now.
The .emacs.d repo you see 8 days before is very old. I just push my current .emacs.d.

HTH

shaleh:
Yes, that will be helpful. Thank you.

aborn:
Set org-element-use-cache to nil

`(setq org-element-use-cache nil)`

brotzeit:
@guoxiangyang Does it work now ?

guoxiangyang:
It works now.

mario-bermonti:
> Set org-element-use-cache to nil
> 
> `(setq org-element-use-cache nil)`

This worked for me.

PR #671, opened 2018-02-27
Revert " port byte-to-position "

brotzeit:
Reverts Wilfred/remacs#669


shaleh:
Are you reverting this in response to #670? Have you confirmed it is the cause?

brotzeit:
I did this preemptively, because I know how long our builds can take.

shaleh:
On the issue I posted a link to the submitter's emacs.d. We should try to reproduce the error with that config and if possible write a test case for remacs. Your code works in the minor cases I have tried so far.

brotzeit:
It's an error in the process sentinel. I doubt his config is the reason for it.

PR #675, opened 2018-03-01
Ported get_keymap C function

### LINKED ISSUE #668: Port get_keymap from keymap.c
Eyeing to work on this next one from `keymap.c`.

```c
/* Check that OBJECT is a keymap (after dereferencing through any
   symbols).  If it is, return it.

   If AUTOLOAD and if OBJECT is a symbol whose function value
   is an autoload form, do the autoload and try again.
   If AUTOLOAD, callers must assume GC is possible.

   ERROR_IF_NOT_KEYMAP controls how we respond if OBJECT isn't a keymap.
   If ERROR_IF_NOT_KEYMAP, signal an error; otherwise,
   just return Qnil.

   Note that most of the time, we don't want to pursue autoloads.
   Functions like Faccessible_keymaps which scan entire keymap trees
   shouldn't load every autoloaded keymap.  I'm not sure about this,
   but it seems to me that only read_key_sequence, Flookup_key, and
   Fdefine_key should cause keymaps to be autoloaded.

   This function can GC when AUTOLOAD is true, because it calls
   Fautoload_do_load which can GC.  */

Lisp_Object
get_keymap (Lisp_Object object, bool error_if_not_keymap, bool autoload)
{
  Lisp_Object tem;

 autoload_retry:
  if (NILP (object))
    goto end;
  if (CONSP (object) && EQ (XCAR (object), Qkeymap))
    return object;

  tem = indirect_function (object);
  if (CONSP (tem))
    {
      if (EQ (XCAR (tem), Qkeymap))
	return tem;

      /* Should we do an autoload?  Autoload forms for keymaps have
	 Qkeymap as their fifth element.  */
      if ((autoload || !error_if_not_keymap) && EQ (XCAR (tem), Qautoload)
	  && SYMBOLP (object))
	{
	  Lisp_Object tail;

	  tail = Fnth (make_number (4), tem);
	  if (EQ (tail, Qkeymap))
	    {
	      if (autoload)
		{
		  Fautoload_do_load (tem, object, Qnil);
		  goto autoload_retry;
		}
	      else
	      	return object;
	    }
	}
    }

 end:
  if (error_if_not_keymap)
    wrong_type_argument (Qkeymapp, object);
  return Qnil;
}
```

This particular function uses `goto` statements so I reckon I'll be doing some work with restructuring the logic for this.

accidentalrebel:
Ported the `get_keymap` C function. Closes #668.

Notes:
- Other parts of `keymap.rs` needed to be updated by removing the now unneeded `unsafe` blocks.
- I had to also update `data.rs` which also uses `get_keymap`.
- Tested working by making temporary lisp tests.
- Initially used Rust labels but refactored it so there are no more nested loops.

### DIFF
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -125,6 +125,7 @@ pub extern "C" fn get_keymap_2(
                     }
                 }
             }
+            break 'end;
         }
     }
 
@@ -132,6 +133,24 @@ pub extern "C" fn get_keymap_2(
         wrong_type!(Qkeymapp, object);
     }
     Qnil
+}
+
+#[lisp_fn]
+pub fn get_keymap_2_lisp(
+    object: LispObject,
+    error_if_not_keymap: bool,
+    autoload: bool,
+) -> LispObject {
+    LispObject::from_raw(get_keymap_2(object.to_raw(), error_if_not_keymap, autoload))
+}
+
+#[lisp_fn]
+pub fn get_keymap_lisp(
+    object: LispObject,
+    error_if_not_keymap: bool,
+    autoload: bool,
+) -> LispObject {
+    LispObject::from_raw(unsafe { get_keymap(object.to_raw(), error_if_not_keymap, autoload) })
 }
 
 /// Construct and return a new keymap, of the form (keymap CHARTABLE . ALIST).
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -4,12 +4,14 @@ use remacs_macros::lisp_fn;
 use remacs_sys::{current_global_map as _current_global_map, globals, EmacsInt, Lisp_Object,
                  CHAR_META};
 use remacs_sys::{Fcons, Fevent_convert_list, Ffset, Fmake_char_table, Fpurecopy, Fset};
-use remacs_sys::{Qkeymap, Qnil, Qt};
+use remacs_sys::{Qautoload, Qkeymap, Qkeymapp, Qnil, Qt};
 use remacs_sys::{access_keymap, get_keymap, maybe_quit};
 
-use data::aref;
+use data::{aref, indirect_function};
+use eval::autoload_do_load;
 use keyboard::lucid_event_type_list_p;
 use lisp::{defsubr, LispObject};
+use lists::nth;
 use threads::ThreadState;
 
 #[inline]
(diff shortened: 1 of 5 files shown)


rust_src/src/keymap.rs
```diff
@@ -51,6 +53,90 @@ pub extern "C" fn set_where_is_cache_keymaps(val: Lisp_Object) {
...
+        if object.is_nil() {
+            break;
+        }
+
+        if object.is_cons()
+            && object
+                .as_cons()
+                .unwrap()
+                .car()
+                .eq(LispObject::from_raw(Qkeymap))
```
shaleh: if let Some(cons) = object.as_cons() {
        if cons.car().eq_raw(Qkeymap) {
            return object.to_raw();
        }
    }

rust_src/src/keymap.rs
```diff
@@ -51,6 +53,90 @@ pub extern "C" fn set_where_is_cache_keymaps(val: Lisp_Object) {
...
+        {
+            return object.to_raw();
+        }
+
+        let tem = indirect_function(object);
+        if tem.is_cons() {
+            if tem.as_cons()
+                .unwrap()
+                .car()
+                .eq(LispObject::from_raw(Qkeymap))
```
shaleh: Same as above.

rust_src/src/keymap.rs
```diff
@@ -51,6 +53,90 @@ pub extern "C" fn set_where_is_cache_keymaps(val: Lisp_Object) {
...
+
+            // Should we do an autoload?  Autoload forms for keymaps have
+            // Qkeymap as their fifth element.
+            if (autoload || !error_if_not_keymap)
+                && tem.as_cons()
+                    .unwrap()
+                    .car()
+                    .eq(LispObject::from_raw(Qautoload)) && object.is_symbol()
+            {
+                let tail = nth(4, tem);
```
shaleh: Again with `eq_raw`

rust_src/src/keymap.rs
```diff
@@ -51,6 +53,90 @@ pub extern "C" fn set_where_is_cache_keymaps(val: Lisp_Object) {
...
+/// `Fautoload_do_load` which can GC.
+#[no_mangle]
+pub extern "C" fn get_keymap(
+    object: Lisp_Object,
+    error_if_not_keymap: bool,
+    autoload: bool,
+) -> Lisp_Object {
+    let object = LispObject::from_raw(object);
+    let mut autoload_retry;
+    loop {
```
shaleh: This is a while loop based on the `autoload_rety` variable.

accidentalrebel: I initially wanted to use a `do while` loop but learned that Rust doesn't have one. Switched to this one. But I agree maybe a `while` loop is better.

rust_src/src/keymap.rs
accidentalrebel: This is quite useful! Thanks!

rust_src/src/keymap.rs
accidentalrebel: Got it.

rust_src/src/keymap.rs
accidentalrebel: Gotcha. The long chain should have been my clue to break it down.

accidentalrebel:
Alright. Fixed the requested changes. :)

Wilfred:
@accidentalrebel I've added you as a collaborator, so you now have the power to approve PRs as well if you fancy :)

PR #677, opened 2018-03-02
Port run hooks support

shaleh:


rust_src/.dir-locals.el
```diff
@@ -0,0 +1,3 @@
+((rust-mode
+  (racer-rust-src-path . "/home/shaleh/\.rustup/toolchains/nightly-2018-02-03-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src")))
+
```
Wilfred: I don't think you meant to check this in :)

rust_src/src/eval.rs
```diff
@@ -860,4 +861,115 @@ pub fn autoload_do_load(
...
+    run_hook_with_args_internal(args, funcall_nil)
+}
+
+fn funcall_nil(args: &mut [LispObject]) -> LispObject {
+    let mut obj_array: Vec<Lisp_Object> = args.iter().map(|o| o.to_raw()).collect();
+    unsafe { Ffuncall(obj_array.len() as isize, obj_array.as_mut_ptr()) };
+    LispObject::constant_nil()
+}
+
+/* Run the hook HOOK, giving each function no args.  */
```
Wilfred: For consistency, we should probably use `///` comments here.

rust_src/.dir-locals.el
shaleh: Indeed. I thought this was covered with gitignore. Bah!

rust_src/src/eval.rs
shaleh: right. Oops.

shaleh:
Yes, there are more hook functions to port.

PR #678, opened 2018-03-02
Handle clippy issues.

shaleh:


shaleh:
Hmm, let's try that again.

PR #680, opened 2018-03-02
port window functions

brotzeit:


rust_src/src/lisp.rs
```diff
@@ -899,6 +899,17 @@ impl From<LispWindowRef> for LispObject {
     }
 }
 
+impl From<LispObject> for Option<LispWindowRef> {
+    #[inline]
+    fn from(o: LispObject) -> Self {
```
shaleh: We pay for this check twice. Why not use as_window?

brotzeit: True. Maybe we should start introducing result types now ?

brotzeit:
@db48x is already working on this and we have several similiar functions that also need updating when we have a solution. This PR is fixing a bug so I think it would be good if we would merge it before applying the changes.

rust_src/src/windows.rs
```diff
@@ -622,4 +622,96 @@ pub extern "C" fn CURRENT_MODE_LINE_HEIGHT(mut window: LispWindowRef) -> i32 {
...
+/// MINIBUF t means include the minibuffer window, even if it isn't active.
+/// MINIBUF nil or omitted means include the minibuffer window only
+/// if it's active.
+/// MINIBUF neither nil nor t means never include the minibuffer window.
+#[lisp_fn(min = "0")]
+pub fn window_list(
+    frame: Option<LispFrameRef>,
+    minibuf: LispObject,
+    window: Option<LispWindowRef>,
+) -> LispObject {
```
shaleh: If you wrote the following `if` as a `match` you can avoid the `is_none` and the `unwrap`.

rust_src/src/windows.rs
```diff
@@ -622,4 +622,96 @@ pub extern "C" fn CURRENT_MODE_LINE_HEIGHT(mut window: LispWindowRef) -> i32 {
...
+        if let Some(f) = frame {
+            f.selected_window()
+        } else {
+            selected_window()
+        }
+    } else {
+        window.unwrap().as_lisp_obj()
+    };
+
+    let f_obj: LispObject;
```
shaleh: Why is this not assigned using the if/else?

rust_src/src/windows.rs
```diff
@@ -622,4 +622,96 @@ pub extern "C" fn CURRENT_MODE_LINE_HEIGHT(mut window: LispWindowRef) -> i32 {
...
+
+    unsafe {
+        LispObject::from_raw(window_list_1(
+            w_obj.to_raw(),
+            minibuf.to_raw(),
+            f_obj.to_raw(),
+        ))
+    }
+}
+
```
shaleh: What is with this function? You are not using it.

brotzeit: `window_list_one` ? It's lisp function ?

shaleh: I missed the name in the `lisp_fn` macro. Never mind.

brotzeit:
@Wilfred we already have similiar impls in lisp.rs for other types. I think @db48x is already working on a solution using result types.

IMO we should merge this PR first and then fix all occurrences of this issue at once.

shaleh:
We don't know when the Result experiment will land. If it does. Let's get this one right until then.

PR #682, opened 2018-03-06
Ported map_keymap and map-keymap functions with tests.

accidentalrebel:
Some notes that I am unsure of and need guidance with:
- The ported functions needed the `map_keymap_function_t` typedef declared in C, so what i did was I declared a similar type on the Rust side with:
```rust
type map_keymap_function_t =
    unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object, *const c_void);
```
- I used Rust `c_void` in place of C's `void`.
- I also used `ptr::null_mut()` in place of `NULL`.

As always, thanks for the help!

### DIFF
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -1,5 +1,9 @@
 //! Keymap support
 
+use std::ptr;
+
+use libc::c_void;
+
 use remacs_macros::lisp_fn;
 use remacs_sys::{current_global_map as _current_global_map, globals, EmacsInt, Lisp_Object,
                  CHAR_META};
@@ -11,11 +15,9 @@ use remacs_sys::{access_keymap, map_keymap_call, map_keymap_function_t, map_keym
 use data::{aref, indirect_function};
 use eval::autoload_do_load;
 use keyboard::lucid_event_type_list_p;
-use libc::c_void;
 use lisp::{defsubr, LispObject};
 use lists::nth;
 use obarray::intern;
-use std::ptr;
 use threads::ThreadState;
 
 #[inline]
@@ -275,9 +277,10 @@ pub extern "C" fn map_keymap(
     let mut map = LispObject::from_raw(get_keymap(map, true, autoload));
     while map.is_cons() {
         if let Some(cons) = map.as_cons() {
-            if keymapp(cons.car()) {
-                map_keymap(cons.car().to_raw(), fun, args, data, autoload);
-                map = cons.cdr();
+            let (car, cdr) = cons.as_tuple();
+            if keymapp(car) {
+                map_keymap(car.to_raw(), fun, args, data, autoload);
+                map = cdr;
             } else {
                 map = LispObject::from_raw(unsafe {
                     map_keymap_internal(map.to_raw(), fun, args, data)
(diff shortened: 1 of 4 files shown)


rust_src/src/keymap.rs
```diff
@@ -5,13 +5,17 @@ use remacs_sys::{current_global_map as _current_global_map, globals, EmacsInt, L
...
 use remacs_sys::{Fcons, Fevent_convert_list, Ffset, Fmake_char_table, Fpurecopy, Fset};
 use remacs_sys::{Qautoload, Qkeymap, Qkeymapp, Qnil, Qt};
-use remacs_sys::{access_keymap, maybe_quit};
+use remacs_sys::{access_keymap, map_keymap_call, map_keymap_function_t, map_keymap_internal,
+                 maybe_quit};
 
 use data::{aref, indirect_function};
 use eval::autoload_do_load;
 use keyboard::lucid_event_type_list_p;
+use libc::c_void;
```
shaleh: `libc` is an external crate. It belongs at the top, before remacs_sys.

rust_src/src/keymap.rs
```diff
@@ -5,13 +5,17 @@ use remacs_sys::{current_global_map as _current_global_map, globals, EmacsInt, L
...
 use lisp::{defsubr, LispObject};
 use lists::nth;
+use obarray::intern;
+use std::ptr;
```
shaleh: `std` should be the first.

rust_src/src/keymap.rs
```diff
@@ -258,6 +262,58 @@ pub fn keymap_prompt(map: LispObject) -> LispObject {
...
+pub extern "C" fn map_keymap(
+    map: Lisp_Object,
+    fun: map_keymap_function_t,
+    args: Lisp_Object,
+    data: *const c_void,
+    autoload: bool,
+) {
+    let mut map = LispObject::from_raw(get_keymap(map, true, autoload));
+    while map.is_cons() {
+        if let Some(cons) = map.as_cons() {
```
shaleh: `let (keymap, map_1) = cons.as_tuple();`

This gives you car and cdr at once.

rust_src/src/keymap.rs
accidentalrebel: I see. Got it.

rust_src/src/keymap.rs
accidentalrebel: So is the reason why it is first is because it is a Rust internal library?

This means, the arrangement should be:
Rust standard library > External crates > remacs_sys > everything else?

rust_src/src/keymap.rs
accidentalrebel: Ah , that is really nice. I'll use that. Thanks!

rust_src/src/keymap.rs
shaleh: correct. As documented in the contributing.

accidentalrebel: I see.

```markdown
Imports are grouped with an empty line between groups. This lets rustfmt keep them ordered. The grouping is:

    Rust crates
    external crates
    remacs crates
    modules within the current crate
```

I've read this before but it didn't register with me at that time. Now I know. Thanks!

accidentalrebel:
Oh. Does "LGTM" automatically approve the pull request? Kinda like this? https://github.com/lgtmco/lgtm

PR #683, opened 2018-03-07
merge

db48x:
I accidentally pushed the wrong buttons this morning, and this came out. There are a handfull of test failures; some fresh eyes on them would be nice.


db48x:
Yes, I shouldn't have tried to fix any of the tests last night; I was too tired. Fixing the missing close-paren (doh) is easy, but the test still doesn't pass...

shaleh:
The other failure is from 'loaddefs.el' but I do not see output from it. Check out its log.

db48x:
I think that's the same error.

shaleh:
Push what you have updated. I will look at it tonight.

db48x:
ok, I fixed the `find-func-tests-finds-function-def-in-rust-source-file` test.

db48x:
The flymake failures are all because `face-at-point` is failing, which is because `(thing-at-point 'symbol)` returns nil, which is because `(get 'symbol 'thing-at-point)` is returning nil. Someone is supposed to call `(set 'symbol 'thing-at-point 'foo)` at some point; gotta figure out why nobody ever does. `(thing-at-point 'number)` does work, however, so it's not completely broken...

db48x:
No, that's not quite right; `face-at-point` does generally seem to be working.

db48x:
I took a long look at the diff between Emacs and Remacs after this merge, and found a worrying number of differences that don't have anything to do with the changes we've introduced with Remacs. I don't know if they're there because this merge went wrong, or because previous merges were wrong, but the differences weren't anything that came up while I was adjudicating merge conflicts.

I'll have to do a diff between them at earlier points to find out.

PR #685, opened 2018-03-09
Port get file buffer

shanavas786:


shanavas786:
build fails with `Symbol’s function definition is void: nil`

am I missing something?

accidentalrebel:
Made a quick check so this is just a quick hunch:
Have you removed `DEFSYM (Qget_file_buffer, "get-file-buffer");`?

shaleh:
```def_lisp_sym!(Qget_file_buffer, "get-file-buffer");``` is the right thing to do after removing the `DEFSYM` from the C file. But that does not help the compilation.

File: rust_src/src/buffers.rs [hunk 1]
```diff
@@ -554,4 +560,44 @@ pub fn overlay_lists() -> LispObject {
...
+    verify_lisp_type!(filename, Qstringp);
+    let filename = unsafe { LispObject::from_raw(Fexpand_file_name(filename.to_raw(), Qnil)) };
+
+    // If the file name has special constructs in it,
+    // call the corresponding file handler.
+    let handler = unsafe {
+        LispObject::from_raw(Ffind_file_name_handler(filename.to_raw(), Qget_file_buffer))
+    };
+
+    if handler.is_nil() {
```
shaleh: You got this backwards. It is  supposed to be `is_not_nil`

shaleh:
This still fails to pass the tests. But it gets you closer @shanavas786 

```
/// Return the buffer visiting file FILENAME (a string).
/// The buffer's `buffer-file-name' must match exactly the expansion of FILENAME.
/// If there is no such live buffer, return nil.
/// See also `find-buffer-visiting'.
#[lisp_fn]
pub fn get_file_buffer(filename: LispObject) -> Option<LispBufferRef> {
    verify_lisp_type!(filename, Qstringp);
    let filename = unsafe { LispObject::from_raw(Fexpand_file_name(filename.to_raw(), Qnil)) };

    // If the file name has special constructs in it,
    // call the corresponding file handler.
    let handler = unsafe {
        LispObject::from_raw(Ffind_file_name_handler(filename.to_raw(), Qget_file_buffer))
    };

    if handler.is_not_nil() {
        let handled_buf = call_raw!(handler.to_raw(), Qget_file_buffer, filename.to_raw());
        handled_buf.as_buffer()
    } else {
        LispObject::from_raw(unsafe { Vbuffer_alist })
            .iter_alist_vals()
            .find(|buf| {
                let buf_filename = buf.as_buffer_or_error().filename();
                buf_filename.is_string() && string_equal(buf_filename, filename)
            })
            .and_then(|obj| obj.as_buffer())
    }
}

def_lisp_sym!(Qget_file_buffer, "get-file-buffer");
```

shaleh:
The tests are failing because the progmodes test depends on a function existing in C code. Every time we port that function we have to find another C function to take its place. Eventually, we need to fix this test to look at Rust too.

shanavas786:
Great, Thanks.
Is there any way to debug this failure ?

shaleh:
```
diff --git a/test/lisp/progmodes/elisp-mode-tests.el b/test/lisp/progmodes/elisp-mode-tests.el
index 0ef225a..e37a158 100644
--- a/test/lisp/progmodes/elisp-mode-tests.el
+++ b/test/lisp/progmodes/elisp-mode-tests.el
@@ -676,10 +676,10 @@ to (xref-elisp-test-descr-to-target xref)."
 
 ;; This needs to refer to a function defined in C.
 (xref-elisp-deftest find-defs-defun-c
-  (elisp--xref-find-definitions 'get-file-buffer)
+  (elisp--xref-find-definitions 'get-buffer-create)
   (list
-   (xref-make "(defun get-file-buffer)"
-             (xref-make-elisp-location 'get-file-buffer nil "src/buffer.c"))))
+   (xref-make "(defun get-buffer-create)"
+             (xref-make-elisp-location 'get-buffer-create nil "src/buffer.c"))))
 
 ;; FIXME: deftype
 ```

shaleh:
If you stare at the error and ponder it becomes obvious. i have tripped this error a few times :-)

shaleh:
Good to see you again @shanavas786

shaleh:
@shanavas786 I updated my comment with the refactored code. Your commit still has the fundamental error which is the handler check should be 'is_not_nil`.

shaleh:
oh, you meant the symbol is nil error.

Here is what I did. I put the C function back and renamed the Rust version to `rust_get_file_buffer`. Now I had a working version and a broken version. Then I compared the two versions. The use of `call!` was a good place to look since it goes back through the system.

shanavas786:
> Your commit still has the fundamental error which is the handler check should be 'is_not_nil`.

Oh that was a miss :-|

shanavas786:
Hope we are good to go :)

Issue #684, opened 2018-03-07
Port map_keymap_internal and map-keymap-internal from keymap.c


accidentalrebel:
Here's what I'll be working on next:

```c
/* Call FUN for every binding in MAP and stop at (and return) the parent.
   FUN is called with 4 arguments: FUN (KEY, BINDING, ARGS, DATA).  */
Lisp_Object
map_keymap_internal (Lisp_Object map,
		     map_keymap_function_t fun,
		     Lisp_Object args,
		     void *data)
{
  Lisp_Object tail
    = (CONSP (map) && EQ (Qkeymap, XCAR (map))) ? XCDR (map) : map;

  for (; CONSP (tail) && !EQ (Qkeymap, XCAR (tail)); tail = XCDR (tail))
    {
      Lisp_Object binding = XCAR (tail);

      if (KEYMAPP (binding))	/* An embedded parent.  */
	break;
      else if (CONSP (binding))
	map_keymap_item (fun, args, XCAR (binding), XCDR (binding), data);
      else if (VECTORP (binding))
	{
	  /* Loop over the char values represented in the vector.  */
	  int len = ASIZE (binding);
	  int c;
	  for (c = 0; c < len; c++)
	    {
	      Lisp_Object character;
	      XSETFASTINT (character, c);
	      map_keymap_item (fun, args, character, AREF (binding, c), data);
	    }
	}
      else if (CHAR_TABLE_P (binding))
	map_char_table (map_keymap_char_table_item, Qnil, binding,
			make_save_funcptr_ptr_obj ((voidfuncptr) fun, data,
						   args));
    }

  return tail;
}
```
```c
DEFUN ("map-keymap-internal", Fmap_keymap_internal, Smap_keymap_internal, 2, 2, 0,
       doc: /* Call FUNCTION once for each event binding in KEYMAP.
FUNCTION is called with two arguments: the event that is bound, and
the definition it is bound to.  The event may be a character range.
If KEYMAP has a parent, this function returns it without processing it.  */)
  (Lisp_Object function, Lisp_Object keymap)
{
  keymap = get_keymap (keymap, 1, 1);
  keymap = map_keymap_internal (keymap, map_keymap_call, function, NULL);
  return keymap;
}
```

Should be a straightforward port. There is however a reference to `voidfuncptr` which I have not encountered before. Seems to be a C type. Will find out soon.

shaleh:
[...]

PR #686, opened 2018-03-11
Ported map-keymap-internal and map_keymap_internal functions with tests.

accidentalrebel:
This took me awhile mostly because of the casting to `voidfuncptr` as stated [here](https://github.com/Wilfred/remacs/issues/684#issuecomment-371837539).

After a lot of trial and error and researching I found that the solution is to use `std::mem::transmute` as shown [here](https://doc.rust-lang.org/std/mem/fn.transmute.html#examples) which worked out great.

There is a caveat though:
> Turning a pointer into a function pointer. This is not portable to machines where function pointers and data pointers have different sizes.

Not sure if this concerns our project or not. By machines they may refer to different OSs? Should we be concerned?

Thanks!

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -1363,6 +1363,7 @@ pub struct lisp_time {
 
 pub type map_keymap_function_t =
     unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object, *const c_void);
+pub type c_function = unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object);
 
 extern "C" {
     pub static initialized: bool;
@@ -1389,6 +1390,12 @@ extern "C" {
     // Use LispObject::tag_ptr instead of make_lisp_ptr
     pub fn make_lisp_ptr(ptr: *const c_void, ty: Lisp_Type) -> Lisp_Object;
     pub fn Fmake_char_table(purpose: Lisp_Object, init: Lisp_Object) -> Lisp_Object;
+    pub fn map_char_table(
+        c_function: unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object),
+        function: Lisp_Object,
+        table: Lisp_Object,
+        arg: Lisp_Object,
+    );
     pub fn CHAR_TABLE_SET(ct: Lisp_Object, idx: c_int, val: Lisp_Object);
 
     pub fn aset_multibyte_string(array: Lisp_Object, idxval: EmacsInt, c: c_int);
@@ -1593,6 +1600,14 @@ extern "C" {
         args: Lisp_Object,
         data: *const c_void,
     ) -> Lisp_Object;
+    pub fn map_keymap_item(
+        fun: map_keymap_function_t,
+        args: Lisp_Object,
+        key: Lisp_Object,
+        val: Lisp_Object,
+        data: *const c_void,
+    );
+    pub fn map_keymap_char_table_item(args: Lisp_Object, key: Lisp_Object, val: Lisp_Object);
     pub fn map_keymap_call(
         key: Lisp_Object,
         val: Lisp_Object,
@@ -1613,6 +1628,12 @@ extern "C" {
         x: Lisp_Object,
         y: Lisp_Object,
         t: Time,
+    ) -> Lisp_Object;
+
+    pub fn make_save_funcptr_ptr_obj(
+        a: *const c_void,
+        b: *const c_void,
+        c: Lisp_Object,
     ) -> Lisp_Object;
 
     pub fn Fselect_window(window: Lisp_Object, norecord: Lisp_Object) -> Lisp_Object;
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -1595,18 +1595,6 @@ extern "C" {
     );
     pub fn STRING_BYTES(s: *const Lisp_String) -> ptrdiff_t;
     pub fn Fevent_convert_list(event_desc: Lisp_Object) -> Lisp_Object;
-    pub fn map_keymap_internal_test(
-        binding: Lisp_Object,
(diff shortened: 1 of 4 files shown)


rust_src/src/keymap.rs
```diff
@@ -317,6 +319,96 @@ pub fn map_keymap_lisp(function: LispObject, keymap: LispObject, sort_first: boo
...
+    let map = LispObject::from_raw(map);
+    if let Some(cons) = map.as_cons() {
+        let (car, cdr) = cons.as_tuple();
+        if car.eq_raw(Qkeymap) {
+            tail = cdr
+        } else {
+            tail = map
+        }
+    };
+
```
shaleh: Instead of `loop` you could use `while let Some(tail_cons) = tail.as_cons()`

rust_src/src/keymap.rs
```diff
@@ -317,6 +319,96 @@ pub fn map_keymap_lisp(function: LispObject, keymap: LispObject, sort_first: boo
...
+/// FUN is called with 4 arguments: FUN (KEY, BINDING, ARGS, DATA).  */
+#[no_mangle]
+pub extern "C" fn map_keymap_internal(
+    map: Lisp_Object,
+    fun: map_keymap_function_t,
+    args: Lisp_Object,
+    data: *const c_void,
+) -> Lisp_Object {
+    let mut tail = LispObject::constant_nil();
+    let map = LispObject::from_raw(map);
```
shaleh: Instead of defining tail and then updating it you could assign here.

```
let mut tail = match map.as_cons() {
    None => LispObject::constant_nil(),
    Some(cons) => {
        let (car, cdr) = cons.as_tuple();
        if car.eq_raw(Qkeymap) {
            cdr
        } else {
            map
        }
};
````

File: rust_src/src/keymap.rs (hunk 1 above)
Comment (accidentalrebel) [reply rust_src/src/keymap.rs]: Ah. I didn't know you could do the same thing with `while`. Nice. Will fix. Thanks!

File: rust_src/src/keymap.rs (hunk 2 above)
Comment (accidentalrebel) [reply rust_src/src/keymap.rs]: I see. I should get the hang of this kind of format soon. ;)

**Comment by accidentalrebel** [comment]
I also notice that Travis is failing due to clippy complaints. Will be sure to fix this too.

**Comment by accidentalrebel** [comment]
Fixed the requested changes and fixed the clippy errors in Travis too.

File: rust_src/src/keymap.rs [hunk 3]
```diff
@@ -317,6 +317,93 @@ pub fn map_keymap_lisp(function: LispObject, keymap: LispObject, sort_first: boo
     LispObject::constant_nil()
 }
 
+/// Call FUN for every binding in MAP and stop at (and return) the parent.
+/// FUN is called with 4 arguments: FUN (KEY, BINDING, ARGS, DATA).  */
```
brotzeit: You forgot to remove the `*/` at the end of the comment.

accidentalrebel: Whoops. I should make an elisp function to deal with conversion of comments. Thanks will fix.

rust_src/remacs-sys/lib.rs
```diff
@@ -1363,6 +1363,8 @@ pub struct lisp_time {
 
 pub type map_keymap_function_t =
     unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object, *const c_void);
+pub type c_function = unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object);
```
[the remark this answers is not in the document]
accidentalrebel: Good point. Hmm. I'll think up of something.

accidentalrebel:
Alright made another fix.

> If you do write some elisp to help with porting add it to https://github.com/Wilfred/remacs/blob/master/rust_src/admin/remacs-helpers.el

This is great. I will.

accidentalrebel:
Just giving this a slight nudge just in case it has been overlooked ;)

rust_src/src/keymap.rs
```diff
@@ -317,6 +316,92 @@ pub fn map_keymap_lisp(function: LispObject, keymap: LispObject, sort_first: boo
...
+                        map_keymap_char_table_item,
+                        Qnil,
+                        binding.to_raw(),
+                        make_save_funcptr_ptr_obj(funcptr, data, args),
+                    );
+                }
+            }
+        }
+
+        tail = tail_cons.cdr();
```
shaleh: Why are you assigning to tail here? This is what the Iterator is doing for you.

accidentalrebel: Ah. I ported this line without thinking that it was not needed anymore. Will fix.

accidentalrebel: I double checked the code and the reason for `tail = tail_cons.cdr()` is because `tail` is where the parent keymap is stored as a return for the function.

accidentalrebel: So just to reiterate this. The `tail = tail_cons.cdr()` cannot be removed as it is needed for the function to work properly.

Maybe to not make it confusing I can move this line further up after the for loop line. Hmm.

rust_src/src/keymap.rs
```diff
@@ -317,6 +316,92 @@ pub fn map_keymap_lisp(function: LispObject, keymap: LispObject, sort_first: boo
...
+            }
+        }
+    };
+
+    for tail_cons in tail.iter_tails_safe() {
+        let binding = tail_cons.car();
+        if binding.eq_raw(Qkeymap) {
+            break;
+        } else {
+            tail = tail_cons.cdr();
```
shaleh: ok, i see my confusion. This _looks_ like you are trying to affect the loop. Which can't happen because you are using an iterator. But what you are really trying to do is update tail for the eventual return. The right thing to do here is use a different variable name for the return value.

accidentalrebel: Makes sense. Alright ;)

accidentalrebel:
Added change: Renamed the return value to parent to avoid confusion.

PR #689, opened 2018-03-14
fix stuff

brotzeit:


rust_src/src/cmds.rs
```diff
@@ -58,7 +58,7 @@ fn move_point(n: LispObject, forward: bool) -> () {
 ///
 /// Depending on the bidirectional context, the movement may be to the
 /// right or to the left on the screen.  This is in contrast with
-/// \\[right-char], which see.
+/// [right-char], which see.
```
Wilfred: This isn't right. `\\[foo]` means that describe-function will show the keybinding for the interactive function foo. `[foo]` has no special meaning.

rust_src/src/process.rs
```diff
@@ -118,8 +118,8 @@ pub fn process_buffer(process: LispProcessRef) -> LispObject {
 /// This is the pid of the external process which PROCESS uses or talks to.
 /// For a network, serial, and pipe connections, this value is nil.
 #[lisp_fn]
-pub fn process_id(process: LispObject) -> Option<EmacsInt> {
-    let pid = unsafe { pget_pid(process.as_process_or_error().as_ptr()) };
```
Wilfred: Will this signal the same error if the argument is not a process?

shaleh: Yes. The way it works is `as_foo_or_error` is called if `Foo` is used and `as_foo` is called when `Option<Foo>` is used. All of this is managed by supplying both `from(LispObject) -> Foo` and `from(LispObject) -> Option<Foo>`.

shaleh:
Other than @Wilfred 's issues this looks good. Nice work.

shaleh:
Hmmm. It looks like the Windows environment is having problems finding DLLs. Maybe the environment was updated by AppVeyor?

PR #690, opened 2018-03-14
port char-equal

brotzeit:


brotzeit:
It's really annoying that I can't implement From<LispObject> for Codepoint because there's already an impl for `u32`. What's the reason for that ?

shaleh:
Yeah, I was wondering if maybe @brotzeit was using a different version of rustfmt.

brotzeit:
I'm using latest nightly. The last commit contains the related changes.

I'm not sure if we should use `rustfmt-preview`. As far as I know it only makes it possible to use nightly rustfmt versions with stable toolchain. But it seems travis doesn't care about the used version anymore ?

shaleh:
You need to either stick to the version we have documented or update the repo to the new version.

brotzeit:
Where is it documented ? I think it was removed when we switched to `rustfmt-preview`. And even if it's documented somewhere, we should make it obligatory again.

shaleh:
I removed the rustfmt changes and rebased.

PR #694, opened 2018-03-16
Ported copy-keymap function with tests.

### LINKED ISSUE #688: Port copy-keymap function from keymap.c
I'm planning to concentrate my efforts on clearing all DEFUN declarations for keymap.c.

Here's the next one I plan to port. It's fairly straightforward with most of the functions I've already come across in the past:

```c
DEFUN ("copy-keymap", Fcopy_keymap, Scopy_keymap, 1, 1, 0,
       doc: /* Return a copy of the keymap KEYMAP.

Note that this is almost never needed.  If you want a keymap that's like
another yet with a few changes, you should use map inheritance rather
than copying.  I.e. something like:

    (let ((map (make-sparse-keymap)))
      (set-keymap-parent map <theirmap>)
      (define-key map ...)
      ...)

After performing `copy-keymap', the copy starts out with the same definitions
of KEYMAP, but changing either the copy or KEYMAP does not affect the other.
Any key definitions that are subkeymaps are recursively copied.
However, a key definition which is a symbol whose definition is a keymap
is not copied.  */)
  (Lisp_Object keymap)
{
  Lisp_Object copy, tail;
  keymap = get_keymap (keymap, 1, 0);
  copy = tail = list1 (Qkeymap);
  keymap = XCDR (keymap);		/* Skip the `keymap' symbol.  */

  while (CONSP (keymap) && !EQ (XCAR (keymap), Qkeymap))
    {
      Lisp_Object elt = XCAR (keymap);
      if (CHAR_TABLE_P (elt))
	{
	  elt = Fcopy_sequence (elt);
	  map_char_table (copy_keymap_1, Qnil, elt, elt);
	}
      else if (VECTORP (elt))
	{
	  int i;
	  elt = Fcopy_sequence (elt);
	  for (i = 0; i < ASIZE (elt); i++)
	    ASET (elt, i, copy_keymap_item (AREF (elt, i)));
	}
      else if (CONSP (elt))
	{
	  if (EQ (XCAR (elt), Qkeymap))
	    /* This is a sub keymap.  */
	    elt = Fcopy_keymap (elt);
	  else
	    elt = Fcons (XCAR (elt), copy_keymap_item (XCDR (elt)));
	}
      XSETCDR (tail, list1 (elt));
      tail = XCDR (tail);
      keymap = XCDR (keymap);
    }
  XSETCDR (tail, keymap);
  return copy;
}
```

accidentalrebel:
Ported the copy-keymap function with tests. Closes #688.

### DIFF
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -1403,7 +1403,6 @@ extern "C" {
     );
     pub fn CHAR_TABLE_SET(ct: Lisp_Object, idx: c_int, val: Lisp_Object);
 
-    pub fn asetm_ultibyte_string(array: Lisp_Object, idxval: EmacsInt, c: c_int);
     pub fn Fcons(car: Lisp_Object, cdr: Lisp_Object) -> Lisp_Object;
     pub fn Fsignal(error_symbol: Lisp_Object, data: Lisp_Object) -> !;
     pub fn Fcopy_sequence(seq: Lisp_Object) -> Lisp_Object;
--- a/rust_src/remacs-sys/lib.rs
+++ b/rust_src/remacs-sys/lib.rs
@@ -1395,9 +1395,15 @@ extern "C" {
     // Use LispObject::tag_ptr instead of make_lisp_ptr
     pub fn make_lisp_ptr(ptr: *const c_void, ty: Lisp_Type) -> Lisp_Object;
     pub fn Fmake_char_table(purpose: Lisp_Object, init: Lisp_Object) -> Lisp_Object;
+    pub fn map_char_table(
+        c_function: unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object),
+        function: Lisp_Object,
+        table: Lisp_Object,
+        arg: Lisp_Object,
+    );
     pub fn CHAR_TABLE_SET(ct: Lisp_Object, idx: c_int, val: Lisp_Object);
 
-    pub fn aset_multibyte_string(array: Lisp_Object, idxval: EmacsInt, c: c_int);
+    pub fn asetm_ultibyte_string(array: Lisp_Object, idxval: EmacsInt, c: c_int);
     pub fn Fcons(car: Lisp_Object, cdr: Lisp_Object) -> Lisp_Object;
     pub fn Fsignal(error_symbol: Lisp_Object, data: Lisp_Object) -> !;
     pub fn Fcopy_sequence(seq: Lisp_Object) -> Lisp_Object;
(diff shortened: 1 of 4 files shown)


rust_src/remacs-sys/lib.rs
```diff
@@ -1395,9 +1395,15 @@ extern "C" {
...
+    pub fn map_char_table(
+        c_function: unsafe extern "C" fn(Lisp_Object, Lisp_Object, Lisp_Object),
+        function: Lisp_Object,
+        table: Lisp_Object,
+        arg: Lisp_Object,
+    );
     pub fn CHAR_TABLE_SET(ct: Lisp_Object, idx: c_int, val: Lisp_Object);
 
-    pub fn aset_multibyte_string(array: Lisp_Object, idxval: EmacsInt, c: c_int);
+    pub fn asetm_ultibyte_string(array: Lisp_Object, idxval: EmacsInt, c: c_int);
```
shaleh: TYPO!!

rust_src/src/keymap.rs
```diff
@@ -244,6 +245,70 @@ pub fn set_keymap_parent(keymap: LispObject, parent: LispObject) -> LispObject {
...
+            if elt.eq_raw(Qkeymap) {
+                elt = copy_keymap(elt); // This is a sub keymap.
+            } else {
+                let (elt_car, elt_cdr) = elt_cons.as_tuple();
+                elt = LispObject::from_raw(unsafe {
+                    Fcons(elt_car.to_raw(), copy_keymap_item(elt_cdr.to_raw()))
+                });
+            }
+        }
+
```
shaleh: This `if` implies `tail` could safely not be a cons cell. Which is not the case. The C code uses macros starting with 'X' to let you know they are not doing any type checking. Their intent is faster processing. This means one strategy would be to do this `as_cons().unwrap()`. Perhaps we could be a little smarter and use `unwrap_or_else(|| Error message)`??

rust_src/src/keymap.rs
```diff
@@ -244,6 +245,70 @@ pub fn set_keymap_parent(keymap: LispObject, parent: LispObject) -> LispObject {
...
+        if let Some(tail_cons) = tail.as_cons() {
+            tail_cons.set_cdr(list!(elt));
+            tail = tail_cons.cdr();
+            keymap = keymap_cons.cdr();
+        }
+    }
+
```
shaleh: Same goes for this if too.

rust_src/src/keymap.rs
accidentalrebel: That makes sense. I assumed that the goal was to ensure safety all the time, but I understand the need for moving type checking to the side for faster processing.

How about using `as_cons_or_error()` instead? I've used it before and it seems to be doing what we want.

accidentalrebel: Also, how does this look?

```rust
        } else if elt.is_cons() {
            let elt_cons = elt.as_cons_or_error();
            if elt.eq_raw(Qkeymap) {
                elt = copy_keymap(elt); // This is a sub keymap.
            } else {
                let (elt_car, elt_cdr) = elt_cons.as_tuple();
                elt = LispObject::from_raw(unsafe {
                    Fcons(elt_car.to_raw(), copy_keymap_item(elt_cdr.to_raw()))
                });
            }
        }

        let tail_cons = tail.as_cons_or_error();
        tail_cons.set_cdr(list!(elt));
        tail = tail_cons.cdr();
        keymap = keymap_cons.cdr();
```

accidentalrebel:
Another change :)

rust_src/src/keymap.rs
shaleh: It is fine for now. We may need to revisit this once we are at a point where performance matters.

shaleh:
kicking the build

shaleh:
Updated in #1087

PR #695, opened 2018-03-18
Lisp fn result

db48x:


db48x:
appveyor!

shaleh:
:+1:

db48x:
I agree that this is a little odd, but my thought was that once everything was ported to Rust, and everything was returning Results instead of calling `signal`, then the stack unwinding could go away. Everything would have the same semantics, but instead of a `longjmp` or whatever Emacs is currently doing, it would be all be explicit returns. The adapter logic we have to write to convert the Result into a call to signal would vanish as well.

Wilfred:
OK, so I've had a look at what Emacs actually does here. `Fsignal` (defined in eval.c) calls `signal_or_quit`. That code looks like this:

```c
  Vsignaling_function = Qnil;
  if (!NILP (error_symbol))
    {
      union specbinding *pdl = backtrace_next (backtrace_top ());
      if (backtrace_p (pdl) && EQ (backtrace_function (pdl), Qerror))
	pdl = backtrace_next (pdl);
      if (backtrace_p (pdl))
	Vsignaling_function = backtrace_function (pdl);
    }

  for (h = handlerlist; h; h = h->next)
    {
      if (h->type != CONDITION_CASE)
	continue;
      clause = find_handler_clause (h->tag_or_ch, conditions);
      if (!NILP (clause))
	break;
    }
```

So essentially what's happening here is that we're unwinding the elisp stack and calling whatever we find. Elisp does not use the C stack, it has its own, which is why the C code can have functions that don't return.

I can't see any way to reconcile this with using `Option`. This is also an implementation detail that is exposed to the Emacs FFI, so the current approach with signal is a public API.

db48x:
It seems like it would be exactly the same either way. When we get the `Result` we get the `error_symbol` out of it, then inspect the Lisp stack to find the signaling function and the condition handlers, just as it does now. I guess we'd have to leave the `setjmp` calls in, just so that we can handle the existing C api, but that doesn't mean that we ourselves must use `longjmp` to handle errors.

pipcet:
I'm also thinking about which calling convention is best for us, though I'm thinking about garbage collection more than error handling. For precise garbage collection, what we really want is for C code to have a:

```rust
pub fn Fcons(result: &mut LispObject, car: & LispObject, cdr: & LispObject) -> ???;
```

signature: we should avoid creating LispObjects because that requires us to root them, but references/pointers to LispObjects are fine. More importantly, we cannot use compound function calls a la

```rust
Fcons(Fcar(x), Fcdr(x))
```

because GC might strike after the call to `Fcar` and we never rooted its return value.

SpiderMonkey uses a `bool` return type where I have `???` above, and propagates exceptions as `false`. We could use an error object of some kind instead, if we're careful about handling GC while we're unwinding after an error; or we could simply use void/no return type and keep using setjmp.

Rust's metaprogramming facilities would make it relatively easy to write
```rust
   return Fcons(Fcar(x), Fcdr(x));
```

and have that expanded into
```rust
   let mut tmp1 = RootedLispObject::new();
   Fcar(&mut tmp1, x);
   let mut tmp2 = RootedLispObject::new();
   Fcdr(&mut tmp2, x);
   Fcons(ret, tmp1.handle(), tmp2.handle());
```

but, if we're rewriting code anyway, why not rewrite it to be conservative about what we can and cannot do with `LispObject`s? The code is more verbose, but also more readable and less ambiguous (it's clear which of `Fcar` and `Fcdr` is called first), and it should optimize to the old code when conservative stack rooting is used.

I'm playing around with something like this, and in particular trying to figure out to what extent Rust can make invalid code produce compile time errors. (I think, in Rust terms, the most important thing is that `LispObject` should not implement the `Copy` trait, but we want stronger guarantees than that).

brotzeit:
Out of curiosity, do you think this will be possible eventually ?

shaleh:
Closing as the code is bit rotting.

Let's examine this again when remacs is fully a Rust app.

Issue #644, opened 2018-02-16
Port process-inherit-coding-system-flag


brotzeit:
```c
DEFUN ("process-inherit-coding-system-flag",
       Fprocess_inherit_coding_system_flag, Sprocess_inherit_coding_system_flag,
       1, 1, 0,
       doc: /* Return the value of inherit-coding-system flag for PROCESS.
If this flag is t, `buffer-file-coding-system' of the buffer
associated with PROCESS will inherit the coding system used to decode
the process output.  */)
  (register Lisp_Object process)
{
  CHECK_PROCESS (process);
  return XPROCESS (process)->inherit_coding_system_flag ? Qt : Qnil;
}
```

mgyucht:
I'd like to try this one out!

PR #696, opened 2018-03-18
Port process-inherit-coding-system-flag to Rust

mgyucht:
# What changed?
I ported `process-inherit-coding-system-flag` to Rust. This addresses issue #644.

# How is it tested?
I ran the following code in Remacs:
```
(defvar myproc (start-process "test" nil "ls"))
(equal nil (process-inherit-coding-system-flag myproc)) ;; => t
(set-process-inherit-coding-system-flag myproc t)
(equal t (process-inherit-coding-system-flag myproc)) ;; => t
```

I can add this to a proper test in test/rust_src/src/process-tests.el if that's useful.

### DIFF
--- a/rust_src/src/process.rs
+++ b/rust_src/src/process.rs
@@ -3,8 +3,9 @@
 use remacs_macros::lisp_fn;
 use remacs_sys::{EmacsInt, Lisp_Process, Lisp_Type, Vprocess_alist};
 use remacs_sys::{current_thread, get_process as cget_process, pget_kill_without_query, pget_pid,
-                 pget_raw_status_new, pset_kill_without_query, send_process,
-                 setup_process_coding_systems, update_status, Fmapcar, STRING_BYTES};
+                 pget_raw_status_new, pget_process_inherit_coding_system_flag,
+                 pset_kill_without_query, send_process, setup_process_coding_systems,
+                 update_status, Fmapcar, STRING_BYTES};
 use remacs_sys::{QCbuffer, Qcdr, Qclosed, Qexit, Qlistp, Qnetwork, Qopen, Qpipe, Qrun, Qserial,
                  Qstop};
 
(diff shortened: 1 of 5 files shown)


rust_src/src/process.rs
```diff
@@ -327,4 +328,16 @@ pub fn waiting_for_user_input_p() -> bool {
     unsafe { (*current_thread).m_waiting_for_user_input_p != 0 }
 }
 
+/// Return the value of inherit-coding-system flag for PROCESS. If this flag is
+/// t, `buffer-file-coding-system` of the buffer associated with process will
+/// inherit the coding system used to decode the process output.
+#[lisp_fn]
```
shaleh: Because you immediately call a `as_foo_or_error` function and do not use the LispObject again you should use the `foo` type instead of the base LispObject. In this case, LispProcessRef. For the return you can just specify `bool` and avoid the explicit conversion.

mgyucht: OK, sounds good.

mgyucht:
Addressed your comments! In the future, I'll write the tests first and then work on the port itself. I'm curious how changing the type from LispObject to LispProcessRef still works. Does the lisp_fn macro take care of this somehow?

mgyucht:
Oh, nevermind, I saw the `impl From<LispObject> for LispProcessRef` in lisp.rs.

shaleh:
Yes, as you saw the macro simply calls `foo.into()` to convert from the LispObject into the more specific type and then does the same on the return value to convert it into a LispObject. This way the individual functions are clear on the types expected and we do not have constant 'as_foo` calls.

Be aware that `as_foo_or_error` is used when the LispFoo type is requested. This means the function will immediately fail if the typing is wrong. Be sure that the function being ported has this semantic before using the type. Sometimes an Option<LispFoo> is a better choice. In others, LispObject is ok and we just check the type and immediately return without erroring.

shaleh:
I have dropped my previous review. Once the new tests been fixed we can merge this.

shaleh:
@mgyucht checkout the "Closes" I added to the git commit message. Whenever your PR closes or references an Issue you should add it to the commit message. Not the first line but in the body.

PR #698, opened 2018-03-22
Port buffer-local-value to rust

shanavas786:


rust_src/src/buffers.rs
```diff
@@ -590,6 +591,21 @@ pub fn get_file_buffer(filename: LispObject) -> Option<LispBufferRef> {
...
 
+/// Return the value of VARIABLE in BUFFER.
+/// If VARIABLE does not have a buffer-local binding in BUFFER, the value
+/// is the default binding of the variable.
+#[lisp_fn(name = "buffer-local-value", c_name = "buffer_local_value")]
+pub fn buffer_local_value_lisp(variable: LispObject, buffer: LispObject) -> LispObject {
+    let result =
+        LispObject::from_raw(unsafe { buffer_local_value(variable.to_raw(), buffer.to_raw()) });
+
+    if result.eq(LispObject::from_raw(Qunbound)) {
```
shaleh: You can use `result.eq_raw(Qunbound)`

shanavas786: good catch. thanks.

shaleh: We added it a little while ago. It definitely simplifies checks in the code.

PR #699, opened 2018-03-22
Port fetch_buffer_markers, get_truename_buffer and record_buffer_markers

shanavas786:


rust_src/src/buffers.rs
```diff
@@ -562,6 +620,84 @@ pub fn overlay_lists() -> LispObject {
...
+fn get_truename_buffer_1(filename: LispObject) -> LispObject {
+    LispObject::from_raw(unsafe { Vbuffer_alist })
+        .iter_alist_vals()
+        .find(|buf| {
+            let buf_truename = buf.as_buffer_or_error().truename();
+            buf_truename.is_string() && string_equal(buf_truename, filename)
+        })
+        .and_then(|obj| obj.as_buffer())
+        .into()
+}
```
shaleh: Would it make send to add `.map(|buf| buf.as_buffer_or_error())` before the `find`? This would make it so we only do the conversion once. If we do this a bunch it may even make sense to emulate the `FOR_EACH_LIVE_BUFFER` and have an explicit `Buffer` iterator.

shanavas786: well, I would prefer a generic `AlistValIter<T>` so that we can use `type BufferIter = AlistValIter<LispBufferRef>`.

shaleh: That sounds reasonable.

shanavas786:
Can anyone have a look here ?.

implementing iterator for `BufferIter` works fine

shaleh:
It is possible you are running afoul of the issue where Rust does not actually treat `type` synonyms as real types. So `LispBufferRef` and `LispFooRef` are really `ExternalPtr<T>`.

shaleh:
The macro is smart. Good choice!

shanavas786:
**macro rules** :)

shanavas786:
Are we good to go ?

shanavas786:
Thanks for the review

PR #703, opened 2018-03-27
Make emacs-version report Remacs

nickdrozd:
I was going to file some bug reports and I realized the version information was inaccurate.

As an aside, how can I get Remacs to load this change? I made the change to `remacs/lisp/version.el`, but `emacs-version` is loaded from `remacs/nextstep/Remacs.app/Contents/Resources/lisp/version.el.gz`. Is there a fast and/or easy way to get the changed file to recompile or whatever?

### DIFF
--- a/lisp/version.el
+++ b/lisp/version.el
@@ -62,7 +62,7 @@ Don't use this function in programs to choose actions according
 to the system configuration; look at `system-configuration' instead."
   (interactive "P")
   (let ((version-string
-         (format "GNU Emacs %s (build %s, %s%s%s%s)%s"
+         (format "Remacs %s (build %s, %s%s%s%s)%s"
                  emacs-version
                  emacs-build-number
 		 system-configuration


shaleh:
`make install` is what copies the files into the app. Look in the `nextstep` directories for more details.

db48x:
Also, in this case you could use `C-x C-e` to execute just the defun that you modified, which would get your modification into your running copy of Emacs/Remacs without having to restart it.

We should probably also consider changing the email address so that bug reports from Remacs come to us rather than going to the Emacs folks.

db48x:
closing and reopening this so that the CI can retry it

Wilfred:
Agreed, this is worth doing. Code checking for version compatibility should be using `emacs-major-version` and `emacs-minor-version` anyway.

We might want to add a `remacs-version` and `remacs-commit-id` at some point. This is similar to the pypy project, which reports a Python version as e.g. 3.6 (affecting compatibility), but still has a separate version for their project.

PR #705, opened 2018-04-09
Don't tell users to override the Rust toolchain to nightly.

### LINKED ISSUE #704: Compilation error in `lisp_fn`
This line in remacs-macros/lib.rs is giving me a compilation error on current nightly:
```
    vec![tokens.parse().unwrap(), fn_ts].into_iter().collect()
```
The error is
```
a collection of type `proc_macro::TokenStream` cannot be built from an iterator over elements of type `proc_macro::TokenStream`
```
I'm very new to the Rust world, but the workaround that's described in the comment above that code, of putting `#fn_item` into the `quote!` and simply using `tokens.parse().unwrap()`, appears to work. I don't know whether error messages for that approach are still broken, though.

pipcet:
Fixes #704.


Wilfred:
I've taken the liberty of merging despite the failing AppVeyor build -- I don't think it could possibly be related.

PR #710, opened 2018-04-11
Update syn (to 0.13.1) and quote (to 0.5), adjust macro definitions.

pipcet:
I'd like to play around with the metaprogramming bits, and I found it hard to find documentation for the old quote/syn versions we are using. So I upgraded, and tried to adjust the code accordingly.

My first real PR, so please be gentle :-). Totally willing to make adjustments, of course. (It would be nice to avoid the `syn::TypeFoo(syn::Type::Foo {` stuff, for starters.

### DIFF
--- a/rust_src/remacs-macros/lib.rs
+++ b/rust_src/remacs-macros/lib.rs
@@ -130,6 +130,10 @@ pub fn lisp_fn(attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
         }
     };
 
+    // we could put #fn_item into the quoted code above, but doing so
+    // drops all of the line numbers on the floor and causes the
+    // compiler to attribute any errors in the function to the macro
+    // invocation instead.
     tokens.append_all(fn_item.into_tokens());
     tokens.into()
 }
--- a/rust_src/remacs-macros/lib.rs
+++ b/rust_src/remacs-macros/lib.rs
@@ -13,8 +13,8 @@ extern crate remacs_util;
 extern crate syn;
 
 use proc_macro::TokenStream;
-use regex::Regex;
 use quote::ToTokens;
+use regex::Regex;
 
 mod function;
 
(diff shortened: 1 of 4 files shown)


rust_src/remacs-macros/lib.rs
```diff
@@ -129,14 +130,8 @@ pub fn lisp_fn(attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
...
 
-    // we could put #fn_item into the quoted code above, but doing so
-    // drops all of the line numbers on the floor and causes the
-    // compiler to attribute any errors in the function to the macro
-    // invocation instead.
-    // Note: TokenStream has a FromIterator trait impl that converts
-    // an iterator over Token{Stream,Tree,Node}s into a single
-    // TokenStream; collect() calls that impl for us.
-    vec![tokens.parse().unwrap(), fn_ts].into_iter().collect()
+    tokens.append_all(fn_item.into_tokens());
```
shaleh: Please leave a comment telling future readers why the `fn_item` tokens are being appended.

pipcet: Thanks for catching that, you're absolutely correct, of course.

shaleh: No problem. I dropped my -1. Now we wait for @db48x to add their +1

rust_src/remacs-macros/lib.rs
```diff
@@ -13,13 +13,14 @@ extern crate remacs_util;
...
 use proc_macro::TokenStream;
+use quote::ToTokens;
 use regex::Regex;
 
 mod function;
 
 #[proc_macro_attribute]
 pub fn lisp_fn(attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
-    let fn_item = syn::parse_item(&fn_ts.to_string()).unwrap();
+    let fn_item = syn::parse(fn_ts).unwrap();
```
db48x: This is the line that I love the most!

db48x:
FWIW, I didn't write this originally, I was simply the last person to touch it.

PR #714, opened 2018-04-12
Allow lisp_fn slice arguments to be immutable

pipcet:
This allows slice references to be the sole argument for a lisp_fn function without requiring them to be mutable. I think there's a case to be made for banning mutable slice references entirely, but `run_hook_with_args_internal` actually does modify its argument array.


shaleh:
@pipcet rebase instead of merging. Yes, you need to `git push --force` but the results are better in the long run.

pipcet:
@shaleh thanks! Is this okay?

shaleh:
Looks good @pipcet

PR #715, opened 2018-04-13
Restore proper line number attributions for lisp_fn errors.

pipcet:
#710 didn't actually preserve those, since it parsed `fn_item` back into tokens. The explicit type annotation appears to be required.

It would be nice to add a test for this.

### DIFF
--- a/rust_src/remacs-macros/lib.rs
+++ b/rust_src/remacs-macros/lib.rs
@@ -13,14 +13,13 @@ extern crate remacs_util;
 extern crate syn;
 
 use proc_macro::TokenStream;
-use quote::ToTokens;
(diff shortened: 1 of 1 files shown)


Wilfred:
Woops, meant to approve.

shaleh:
A demo/test program in the macro's crate which has a purposeful compilation error should suffice. The test would be compile and save output. Compare that to a expected output.

pipcet:
How would that work? `rustc` requires special parameters, and `cargo` would have to use an extra crate, or at least a feature, and be very slow, right?

shaleh:
@pipcet this one is no longer needed, right?

PR #716, opened 2018-04-13
Restore proper line number attributions for lisp_fn errors

pipcet:
This is like #715, but includes testing code to make sure a `compile_error!` is attributed to itself rather than the macro.

### DIFF
--- a/.travis-cerrors.sh
+++ b/.travis-cerrors.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# This is a standalone script so we terminate as soon as anything
+# errors. See https://github.com/travis-ci/travis-ci/issues/1066
+set -e
+set -x
+
+export PATH=$PATH:~/.cargo/bin
+
+RUSTFMT_CONFIG_DIR=$DIR/rust_src
+
+echo "Checking compile errors"
+cd "$DIR/rust_src"
+cargo build --features compile-errors 2> err.out || true
+cat err.out
+grep -q 'compile_error!("error 001");' err.out
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ rust:
 stages:
   - rustfmt
   - test
+  - cerrors
 
 jobs:
   include:
@@ -36,7 +37,7 @@ jobs:
       script:
         - ./autogen.sh && ./configure
         - make -j 3 && echo '==> Running tests' && make check
-    - 
+    -
       <<: *FULL_TEST
       os: osx
       env:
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -66,6 +67,12 @@ jobs:
       os: osx
       before_script:
         - travis_wait rustup install $(cat rust-toolchain)
+    - stage: cerrors
+      os: linux  # No need to run this everywhere
+      before_script:
+        - travis_wait rustup install $(cat rust-toolchain)
+        - travis_wait rustup component add rustfmt-preview
```
shaleh: Why bother with rustfmt for this part of the build?

.travis-cerrors.sh
```diff
@@ -0,0 +1,18 @@
...
+
+export PATH=$PATH:~/.cargo/bin
+
+RUSTFMT_CONFIG_DIR=$DIR/rust_src
+
+echo "Checking compile errors"
+cd "$DIR/rust_src"
+cargo build --features compile-errors 2> err.out || true
+cat err.out
+grep -q 'compile_error!("error 001");' err.out
```
Wilfred: Are you expecting this script to exit with a non-zero exit code in some circumstances? I can't see where that could happen.

pipcet: (`grep` exits with an error code when no hit is found)

pipcet: Would you prefer explicit `|| exit 1` code? I think that's easier to read, but tried not to modify the other script except where necessary.

pipcet:
Okay, I think the newly-rebased commit, if it passes, is something we should merge: it makes tests take twice as long, but fixing the `remacs` crate to compile cleanly without requiring a full `make` of Emacs is best left for another PR.

shaleh:
I would like to try an alternative before we accept this.

shaleh:
I wrote a small crate to test this instead of adding another build to the main configuration. It should fail much faster.

pipcet:
@shaleh That sounds awesome. I think the magic invocation is `make -C src ../etc/DOC` for skipping most of the build, but obviously a more general solution would be great.

shaleh:
It will be about 10 - 12 hours until I get back to this.

.travis-cerrors.sh
Wilfred: I thought that `grep -q` always returned a zero exit code, but I was wrong. `grep -q` only changes the exit code when there's an error.

That said, could we simplify this to using the exit code of `cargo build` instead?

shaleh:
Well then. Trying to make a minimal crate that loads `lisp_fn` requires both remacs_sys and lisp.rs. @pipcet it looks like your intuition was right about collapsing `remacs_sys` into the code base but  maybe we should cut it and lisp.rs into types and function exports. The types could be a remacs_types crate which would be available for loading.

.travis-cerrors.sh
pipcet: > That said, could we simplify this to using the exit code of cargo build instead?
I don't think so. The idea was to make sure we got an error message originating from the `compile_error!` line, rather than one attributed to the macro (which is quite unhelpful). Thus the grep.

pipcet:
@shaleh, can you explain why you want a remacs_types crate? I think there are two cons: fewer warnings, and the inability to extend types in a different crate. I don't see any advantages, though I'm sure there must be some. Thus my preference for using a module rather than a crate.

shaleh:
I am not saying I _want_ one. I am saying that one is required to get a minimal test crate setup as I was intending.

```
#[lisp_fn]
pub fn foo() {
    compile_error!("oops");
}
```

That minimal bit of code requires `Lisp_Object`, `LispObject`, `xmalloc`, `ExternalPtr`, and more.

Issue #706, opened 2018-04-09
Can Lisp_Object and LispObject be unified?


pipcet:
I've started looking around the code a little, and the first thing that I'm wondering about is whether all this `from_raw` and `to_raw` stuff is strictly necessary.

There are currently something like 558 calls to these functions, not counting the ones that get autogenerated by `#[lisp_fn]`. Is that really worth the (nice, certainly) isolation of having the types live in separate crates?

Maybe there's a deeper difference between the types that I'm missing (if so, in my defense, it's not very obvious from the names), but I've played around with merging the remacs_sys crate into the remacs crate, and that appears to enable me to remove most of the calls.

Is there a better/Rustier way?

shaleh:
We have kept them separate to catch mistakes. We are using the type system as a safety net.

pipcet:
Sorry for being dense, but what kind of mistake? If Lisp*Objects were merged, treating one as the other wouldn't be a mistake.

The one use case I can think of is to drop the Copy or Clone traits for LispObjects, but I'm not convinced that's actually a good idea.

shaleh:
Basically we treat `Lisp_Object` as meaning "for C" and `LispObject` as meaning "for Rust". This catches mistakes during porting.

shaleh:
Deeper reason.

`typedef EMACS_INT Lisp_Object;`

in Rust a typedef is indistinguishable from the aliased object. So basically `Lisp_Object` looks like a `int64`. So we need to make a struct type in Rust.

pipcet:
> So we need to make a struct type in Rust.

No argument there. Do we need two struct types, though?
```
pub struct Lisp_Object(EmacsInt);
pub struct LispObject(Lisp_Object);
```
is what's currently in the code. My proposal is to use
```
pub struct LispObject(EmacsInt);
```
and drop the Lisp_Object.

Wilfred:
The problem here is that the C code has multiple different definitions for `Lisp_Object`.

It can be:

```c
typedef EMACS_INT Lisp_Object;
```

or it can be a struct wrapping an int. There is actually a config option to toggle whether a struct is used. (I imagine Remacs won't build if you change this flag.)

To make complicate matters further, structs have different calling conventions on 32-bit. I'm on my phone, but if you look at my comment on the repr(transparent) RFC I give an example.

We've not had any 32-bit issues or PRs submitted in some time, and I think it's a shrinking use case.

[...]

PR #562, opened 2017-12-29
Make Lisp_Object a proper struct type

shaleh:

PR #718, opened 2018-04-16
Turn remacs_sys into a module (rather than a crate)

pipcet:
This is related to #706. I'd like to merge `Lisp_Object` and `LispObject`. That appears to require turning `remacs_sys` into a module that's part of the `remacs` crate, but I think doing so is a good idea anyway, so this is an independent PR.

Keeping `remacs-sys` a separate crate means duplication, fewer warnings, and introduces a notion of independence that simply isn't appropriate.

`remacs-sys` is meant to shrink as code is translated, so it's not "Rust bindings for Emacs". Any attempt to build mismatched versions of `remacs` and `remacs-sys` is bound to fail, and that's a good argument for not keeping them in separate crates. As the `LispObject` case demonstrates, it's reasonable to want to extend Emacs C types using `remacs` functions; usually we're avoiding that by defining part of the C interface in `remacs` anyway, but that's just a lucky accident.

Rust analyzes unused functions, at least, on a crate basis, so we cannot hope to ever enable these warnings if we keep `remacs-sys` separate.

As far as I can make out, the obvious safety/isolation arguments aren't valid because Rust supports namespacing on a module level.

Again, it is my hope to combine the `Lisp*Object`s and that's my immediate reason for this PR, but I think it would be helpful to discuss this separately: even if you disagree that combining the `Lisp*Object`s is something remacs should do, it's something that we *should be able to do* easily, and that we're not is a sign something's wrong with our crate system.

If you're making the safety counterargument, I would appreciate it, if possible, if you could provide actual examples of code that becomes unsafe if the separation into crates is dropped. I know that dropping the separation would make code valid that currently isn't, but it doesn't follow that the newly-valid code is also unsafe.


rust_src/README.md
```diff
@@ -1,10 +1,10 @@
 # remacs-sys
 
-`remacs-sys` is a Rust library containing all the [Foreign Function Interface (FFI)](https://doc.rust-lang.org/book/first-edition/ffi.html) declarations of structs and functions that are defined in C. Then, they can be used in the Rust code. C implementations are in `remacs/src`.
+`remacs-sys` is a Rust module containing all the [Foreign Function Interface (FFI)](https://doc.rust-lang.org/book/first-edition/ffi.html) declarations of structs and functions that are defined in C. Then, they can be used in the Rust code. C implementations are in `remacs/src`.
 
 ## Example: Use `make_float` C function in Rust
 
-First, let's have a look how `make_float` is defined historically in the C code. Then, how one can re-use it in Rust.
+First, let's have a look how `make_float` is defined historically in the C code. Then, how one can re-use imt in Rust.
```
shaleh: Typo. `s/imt/it/`

rust_src/README.md
```diff
@@ -48,10 +48,10 @@ use libc::{c_double, /* ... */ };
 // ...
 ```
 
-Second, `Lisp_Object` is an alias to `EmacsInt`:
+Second, `Lisp_Object` is a struct containing an EmacsInt:
```
shaleh: It currently is an alias.

rust_src/README.md
```diff
@@ -48,10 +48,10 @@ use libc::{c_double, /* ... */ };
...
 
 ```rust
-pub type Lisp_Object = EmacsInt;
+pub struct Lisp_Object(EmacsInt);
```
shaleh: Not true yet.

rust_src/README.md
pipcet: Thanks

rust_src/README.md
pipcet: Hmm. Not in Wilfred/master, unless I'm missing something:
```
rust_src/remacs-sys/lib.rs:pub struct Lisp_Object(EmacsInt);
```

rust_src/README.md
pipcet: Again, that's what I'm seeing in a fresh checkout.

rust_src/README.md
Wilfred: As of #562 it's not an alias.

Wilfred:
Getting better compiler warnings and simplifying our datatypes seems very worthwhile to me :+1: 

I'm open to getting rid of `emacs-sys` and `Lisp_Object` (I will try to review properly tomorrow). We've iterated on the design of Remacs internals, and I think the codebase is getting cleaner over time. Nonetheless, some design decisions don't work out. We don't have any epic safety arguments in favour of a separate crate, for example.

rust_src/README.md
shaleh: Yeah, my bad. That was even my PR too. Sigh. I have swapped out remacs for the work stack.

PR #720, opened 2018-04-16
Fix a compiler warning.

pipcet:
I introduced a compiler warning, which this PR fixes.

I don't understand why it passed the tests, though. Should there be a `strict` feature in `remacs-macros/Cargo.toml`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
         - sudo apt install -y texinfo libgif-dev libxpm-dev
         - travis_wait rustup install $(cat rust-toolchain)
       script:
-        - ./autogen.sh && ./configure
+        - ./autogen.sh && RUSTFLAGS="-Dwarnings" ./configure
         - make -j 3 && echo '==> Running tests' && make check
     - 
       <<: *FULL_TEST
@@ -59,7 +59,7 @@ jobs:
       before_script:
         - travis_wait rustup install $(cat rust-toolchain)
       script:
-        - ./autogen.sh && ./configure --without-makeinfo --with-x=no --with-ns=no --without-gconf --without-gsettings
+        - ./autogen.sh && RUSTFLAGS="-Dwarnings" ./configure --without-makeinfo --with-x=no --with-ns=no --without-gconf --without-gsettings
         - make -j 3 && echo '==> Running nowindow tests' && make check
     -
       <<: *NO_WINDOW_SYSTEM_TEST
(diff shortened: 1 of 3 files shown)


shaleh:
If you want to add the compiler warning for Travis go ahead.

pipcet:
Is this okay? It looks like CARGO_FLAGS isn't used for building dependencies, but RUSTFLAGS is. I suppose this could bite us if an external dependency builds with a warning, but that's not supposed to happen, right?

.travis.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         - sudo apt install -y texinfo libgif-dev libxpm-dev
         - travis_wait rustup install $(cat rust-toolchain)
       script:
-        - ./autogen.sh && ./configure
+        - ./autogen.sh && ./configure RUSTFLAGS="-D warnings"
```
shaleh: Traditionally, environment variables are specified before the command. So `RUSTFLAGS="-D warnings" ./configure`.

shaleh:
If it causes problems we can deal with it then.

.travis.yml
pipcet: It's not an environment variable, it's (meant to be) a variable used for the autoconf mechanisms, which is specified as a command line argument after `./configure`. But, anyway, it doesn't work because we force RUSTFLAGS to be empty in configure.ac.

configure.ac
```diff
@@ -5191,8 +5191,6 @@ LIBS=$SAVE_LIBS
 
 # Add Rust variables, libs and utilities
 
-RUSTFLAGS=
-
```
shaleh: There is no need to remove this. That whole point of the environment variable is to override this.

pipcet: Hmm. Restoring the two lines appears to break things here, and RUSTFLAGS is once again empty if they're present. I've probably misunderstood what you were suggesting.

shaleh:
```
diff --git a/configure.ac b/configure.ac
index 15a3de99e6..965b262fa7 100644
--- a/configure.ac
+++ b/configure.ac
@@ -5190,8 +5190,7 @@ CFLAGS=$SAVE_CFLAGS
 LIBS=$SAVE_LIBS

 # Add Rust variables, libs and utilities
-
-RUSTFLAGS=
+RUSTFLAGS=$RUSTFLAGS

 AC_SUBST(RUSTFLAGS)
```

The issue is we do not apply the environment variable to the script variable before we call `AC_SUBST`.

shaleh:
```
$ RUSTFLAGS="-Dwarning" ./configure --without-makeinfo
$ grep RUSTFLAGS Makefile
RUSTFLAGS=-Dwarning
```

PR #721, opened 2018-04-18
Porting overlay-properties

nitishch:
Ported `overlay-properties` function.

I have a question. I added `LispOverlayRef::plist` function to be consistent with other attributes in `Lisp_Overlay`. But didn't use it in my `overlay_properties` method because I need a `Lisp_Object`. If I use the function, we'd have `overlay.plist().to_raw()`. Given that `plist()` converts `Lisp_Object` to `LispObject`, it seemed bad to convert convert `LispObject` back to `Lisp_Object`. Is this ok or should I use the `plist` function?


db48x:
That's a good question. Eventually `copy-sequence` will get ported, and at that time it will take a LispObject as an argument and return a LispObject. `overlay-properties` can be changed so that it calls the `plist` method.

nitishch:
You mean to change it now or that then it can be changed?

PR #722, opened 2018-04-23
Porting directory-name-p

nitishch:
I don't have a Windows system so I didn't run the tests in Windows. 

The [documentation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Directory-Names.html) for `directory-name-p` mentions only Windows and MS-DOS as systems with `\` as path-separator. So the test checks only for these two systems.

### DIFF
--- a/rust_src/src/fileio.rs
+++ b/rust_src/src/fileio.rs
@@ -9,7 +9,7 @@ use multibyte::LispStringRef;
 /// Return non-nil if NAME ends with a directory separator character.
 #[lisp_fn]
 pub fn directory_name_p(name: LispStringRef) -> bool {
-    if name.len_bytes() <= 0 {
+    if name.len_bytes() == 0 {
         return false;
     }
 


rust_src/src/fileio.rs
```diff
@@ -0,0 +1,18 @@
+//! Functions to deal with files
+
+use lisp::defsubr;
+use multibyte::LispStringRef;
+use remacs_macros::lisp_fn;
+use std::path;
+
+/// Return non-nil if NAME ends with a directory separator character.
+#[lisp_fn]
+pub fn directory_name_p(name: LispStringRef) -> bool {
```
shaleh: The more common style in the tree is:

    if name.len_bytes() <= 0 {
        return false;
    }

    let b = name.byte_at(name.len_bytes() - 1);
    b as char == path::MAIN_SEPARATOR

rust_src/src/fileio.rs
```diff
@@ -0,0 +1,18 @@
+//! Functions to deal with files
+
+use lisp::defsubr;
```
shaleh: Please be sure to read our contributing documentation. Your imports do not match our style.

rust_src/src/fileio.rs
nitishch: Changed as suggested.

rust_src/src/fileio.rs
nitishch: Reordered as per the guidelines.

rust_src/src/fileio.rs
nitishch: No. It doesn't make sense. Changed.

shaleh: :+1: Sorry about that.

PR #724, opened 2018-05-01
Update from upstream GNU Emacs

Wilfred:
This brings a few commits over from upstream Emacs. The previous common commit:

```
$ git merge-base origin/master upstream/master    
c0af83b6ccf2dab9a515dd7f52eb9d4500275ae3
```

looks like it was previously on a branch that merged into master. I've tried to take a commit that was clearly on the `master` branch, to make things simpler in future. I think this is why I had a few merge conflicts in files that haven't changed in Remacs, such as lcms.c.

This only brings in a few extra commits, but I hope this makes future merges easier.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,27 @@
+Language: Cpp
+BasedOnStyle: LLVM
+AlignEscapedNewlinesLeft: true
+AlwaysBreakAfterReturnType: TopLevelDefinitions
+BreakBeforeBinaryOperators: All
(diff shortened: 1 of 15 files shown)


Wilfred:
Looks like AppVeyor failed on the last build on master too: https://ci.appveyor.com/project/Wilfred/remacs/build/1.0.948

I suspect it's not honouring our rust version. We should fix that properly, but updating to the latest rustc might also help.

PR #727, opened 2018-05-06
Merge GNU Emacs changes up until November 2017

Wilfred:
This brings Remacs in line with changes in GNU Emacs up until November
2017.

Since Remacs started life as a fork of the emacs-26 branch and only
later moved to being based on master, a number of files conflicted due
to differences between emacs26 and master. In these cases, where files
have not been modified on the Remacs side, I've ensured the final
state of the file matches GNU Emacs master as of commit aa75424672.

There were several other sources of conflicts:

ChangeLog files: Remacs has removed these, as they duplicate
information in commit messages. ChangeLog files continue to be updated
on the GNU Emacs side, but remain deleted in Remacs. Likewise other
deleted documentation files that don't apply to Remacs.

MS-DOS: Remacs has removed DOS support, and GNU Emacs has made patches
to logic specific to DOS. I've taken these patches into Remacs, as
they're harmless and it makes resolving conflicts easier.

OS X tests: Remacs skips some tests on OS X that are prone to failing
non-deterministically. Some of those tests have seen other changes, so
I've re-applied the skip logic.

Movement optimisation: 66d37175ecac41dfb2 was a minor optimisation to
Fchar_after in C, although this function is pure Rust in Remacs. It's
not clear that the Rust logic would benefit from this change, so I've
skipped it.

Macro autoloading: bc9300ac5ed3bdf52a fixed autoload_do_load to better
handle errors during macro autoloading. I've applied an equivalent
patch to eval.rs.

Fmin/Fmax and NaN: 5f28f0db73c03b98b2 fixed a bug with NaN handling on
the min and max functions. This has already been patched in the
emacs26 branch and we applied it in f40f21ebf3c87b242a. However, the
commit on the C side removed an assert which was still present in
minmax_driver in the Rust implementation, so I've removed it.

Docstrings: eab051991e3698fae1 updated a docstring in a primitive
function that has been ported to Rust, so I've updated Remacs'
Ffunctionp docstring to match.


Wilfred:
Prefer #729 over this, as there are still some conflicts and it's a pain to fix everything in one epic merge.

PR #730, opened 2018-05-07
Merge Lisp_Object and LispObject into a new LispObject type.

### LINKED ISSUE #706: Can Lisp_Object and LispObject be unified?

pipcet:
Closes #706.

This is the mechanical rename part of the change, leaving out the significant clean-up that's possible once `Lisp_Object` is removed.

### DIFF
--- a/rust_src/build.rs
+++ b/rust_src/build.rs
@@ -295,12 +295,6 @@ impl<'a> ModuleParser<'a> {
                 self.lineno,
                 "'no_mangle' functions exported for C need 'extern \"C\"' too.".to_string(),
             ))
-        } else if line.contains(": LispObject") || line.contains("-> LispObject") {
-            Err(LintMsg::new(
-                &self.info.name,
-                self.lineno,
-                "functions exported to C must use 'Lisp_Object' not 'LispObject'".to_string(),
-            ))
         } else {
             Ok(())
         }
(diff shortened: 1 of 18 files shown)


Wilfred:
Looks like the build is failing due to an unused import.

Wilfred:
The build failure is really weird: it's complaining about unavailable linux packages. I wonder if the default distro version has changed on Travis?

pipcet:
https://github.com/Wilfred/remacs/compare/master...pipcet:lispo2?expand=1 has some of the cleanups that are possible after the `LispObject`/`Lisp_Object` merge. I'm not sure whether simple accessor functions should be removed, too, so I left them in for now.

shaleh:
@pipcet now update the docs!!

PR #732, opened 2018-05-09
port the DEFUN concat of fns.c

purplejacket:
usage: (concat &rest SEQUENCES)

### DIFF
--- a/rust_src/src/fns.rs
+++ b/rust_src/src/fns.rs
@@ -260,7 +260,7 @@ def_lisp_sym!(Qrequire, "require");
 /// Each argument may be a list, vector or string.
 /// The last argument is not copied, just used as the tail of the new list.
 /// usage: (append &rest SEQUENCES)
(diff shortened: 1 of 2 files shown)


shaleh:
We have been trying to either a) write tests or b) migrate tests.

purplejacket:
BTW, I intend to write tests for these last few changes, but those tests are failing because the C language concat function of fns.c hasn't been ported yet.  I want to port all the callers of this function first as it's a rather long and complicated function, then port that one, then write test coverage.

shaleh:
We have been writing all tests in elisp at this point. `remacs/test/src` matches the C source tree and `remacs/test/rust_src` matches the Rust src.

shaleh:
And yes, Travis appears to be making updates and we need to catch up.

Wilfred:
@purplejacket I've added you as a collaborator, so you can also review PRs if you wish :) :sparkles:

PR #733, opened 2018-05-11
port internal_self_insert to rust

db48x:


shaleh:
Sorry I was doing family stuff.

My only code critique is a style one and it is somewhat of a personal preference although consistency is good.

I find `foo != Qnil` or `foo == Qnil` to be preferable to `LispObject::from(foo).is_not_nil()` or `LispObject::from(foo) == Qnil`.

PR #734, opened 2018-05-11
minor improvements to the .gdbinit file

db48x:


Wilfred:
Since the gdb settings don't relate to the build ir tests, I'm going to merge this as-is rather than rebasing first.

PR #348, opened 2017-10-07
Port subset of dired.c to rust

gaak99:
I have the following ported from dired.c

 - directory_files_internal
 - directory-files
 - directory-files-and-attributes
 - file-attributes
 - file-attributes-lessp
 - system-users

Not ported from dired.c:

 - file_name_completion and related funcs
--it's a beast and worthy of a seperate proj

 - system-groups
--it appears no rust equiv (in libc anyways) of Linux's getgrent(3).

Note:
 - only tested on Linux (only OS avail to me at the moment)
 - 'make' and test/ 'make check' clean (dired&tramp seem to be well covered here)
 - Note tramp basic ssh usage seems badly broken in 27.0.50.
   FWIW I tracked it to find-file-name-handler not matching expected tramp regex.
   The new code here fails same way and passes make check tramp suite so I asssume it's ok.
 - Multibyte regex maching of multibyte filenames works in directory-files.
   Added a small recipe at bottom of dired_rust.rs to test this by hand.
 - rustfmt clean
 - docstrings added for lisp funcs

PR #738, opened 2018-05-13
Rust dired systemusers v2

gaak99:
Here's the first (of 3 probably) split-up of PR #348 ...

It's make/check and rustfmt clean and the lisp function and docstring tested by hand on Linux. Also the Fuser_real_login_name fallback block was tested (temp commented out prev block).

The prev version -- a few months ago -- was tested on MacOS too. This code is pretty simple so likely will build/work on MacOS (Windows we'll see too).

### DIFF
--- a/rust_src/src/dired.rs
+++ b/rust_src/src/dired.rs
@@ -1,12 +1,14 @@
 //! Lisp functions for making directory listings.
 
+use std::ffi::CStr;
+
 use libc;
-use lisp::LispObject;
-use lisp::defsubr;
-use lists::list;
+
 use remacs_macros::lisp_fn;
 use remacs_sys::Fuser_real_login_name;
-use std::ffi::CStr;
+
+use lisp::{defsubr, LispObject};
+use lists::list;
 
 /// Return a list of user names currently registered in the system.
 /// If we don't know how to determine that on this platform, just
--- a/rust_src/src/dired.rs
+++ b/rust_src/src/dired.rs
@@ -1,0 +1,39 @@
+//! Lisp functions for making directory listings.
+
+use libc;
+use lisp::LispObject;
+use lisp::defsubr;
+use lists::list;
+use remacs_macros::lisp_fn;
+use remacs_sys::Fuser_real_login_name;
+use std::ffi::CStr;
+
+/// Return a list of user names currently registered in the system.
+/// If we don't know how to determine that on this platform, just
+/// return a list with one element, taken from `user-real-login-name'.
+#[lisp_fn]
+pub fn system_users() -> LispObject {
+    let mut done = false;
+    let mut unames = Vec::new();
+
+    while !done {
+        let pw: *mut libc::passwd = unsafe { libc::getpwent() };
+        if pw.is_null() {
+            done = true;
+        } else {
+            let pwnam_string =
+                unsafe { CStr::from_ptr((*pw).pw_name).to_string_lossy().into_owned() };
+            let pwnam_str = &*pwnam_string;
+            unames.push(LispObject::from(pwnam_str));
+        }
+    }
+    unsafe { libc::endpwent() };
+
+    if unames.len() < 1 {
+        unames.push(LispObject::from_raw(unsafe { Fuser_real_login_name() }));
+    }
+
+    list(&mut unames)
+}
+
+include!(concat!(env!("OUT_DIR"), "/dired_exports.rs"));
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -55,6 +55,7 @@ mod chartable;
 mod cmds;
 mod crypto;
 mod data;
+mod dired;
 mod dispnew;
 mod editfns;
 mod eval;
--- a/rust_src/src/remacs_sys.rs
+++ b/rust_src/src/remacs_sys.rs
@@ -1724,7 +1724,7 @@ extern "C" {
         all_frames: Lisp_Object,
     ) -> Lisp_Object;
     pub fn buffer_local_value(variable: Lisp_Object, buffer: Lisp_Object) -> Lisp_Object;
-
+    pub fn Fuser_real_login_name() -> Lisp_Object;
 }
 
 /// Contains C definitions from the font.h header.
--- a/src/dired.c
+++ b/src/dired.c
@@ -1038,27 +1038,6 @@ Comparison is in lexicographic order and case is significant.  */)
 }
 
 
-DEFUN ("system-users", Fsystem_users, Ssystem_users, 0, 0, 0,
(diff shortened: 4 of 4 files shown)


shaleh:
This looks good. I'd like to see the appveyor branch land so the Windows side is tested too.

rust_src/src/dired.rs
```diff
@@ -0,0 +1,41 @@
...
+    let mut done = false;
+    let mut unames = Vec::new();
+
+    while !done {
+        let pw: *mut libc::passwd = unsafe { libc::getpwent() };
+        if pw.is_null() {
+            done = true;
+        } else {
+            let pwnam_string =
+                unsafe { CStr::from_ptr((*pw).pw_name).to_string_lossy().into_owned() };
```
Wilfred: I think this is a memory leak. You need to call `build_string` so the elisp GC can clean up the strings once they're unused.

rust_src/src/dired.rs
```diff
@@ -0,0 +1,41 @@
...
+            let pwnam_str = &*pwnam_string;
+            unames.push(LispObject::from(pwnam_str));
+        }
+    }
+    unsafe { libc::endpwent() };
+
+    if unames.len() < 1 {
```
Wilfred: Use is_empty

rust_src/src/dired.rs
```diff
@@ -0,0 +1,41 @@
...
+        unames.push(LispObject::from_raw(unsafe { Fuser_real_login_name() }));
```
Wilfred: Is this equivalent? The C code is using a variable that the user can set, but this Rust version is calling a function.

shaleh:
Toggling the build.

shaleh:
@gaak99 are you closing #348?

rust_src/src/dired.rs
gaak99: I did not know that -- thanks!

rust_src/src/dired.rs
gaak99: Ok.

rust_src/src/dired.rs
gaak99: Yes, the function will return the current value of the variable.

I tried the variable first but it does seem to be visible (in src/editfns.o) to rust. nm(1) does not output it fwiw. So anyways it looked like I would have to make a change to editfns.c so was not worth it.

db48x: In the C code a variable reference like `Vfoo` is really a macro that expands to `globals.f_Vfoo`, so in the Rust code we're currently just writing `globals.f_Vfoo` whenever we need to access the variable. The `globals` is still a global variable, but in the future it could be a thread-local variable or something.

gaak99:
@shaleh done (#348 closed) -- thanks for the reminder.

gaak99:
Getting an 'undefined reference __CxxFrameHandler3' on Windows 10.  Looks to be a symbol from a Windows dll (info gleaned from a quick web search).

My Windows sys was upgraded last week to Version	10.0.17134 Build 17134. What version is running in Appveyor? (is it listed in the build output?)

Fresh remacs clone:
```
   Compiling remacs-util v0.1.0 (file:///C:/projects/remacs/rust_src/remacs-util)
   Compiling remacs-lib v0.1.0 (file:///C:/projects/remacs/rust_src/remacs-lib)
    Finished release [optimized] target(s) in 113.67 secs
  CCLD     etags.exe
../rust_src/remacs-lib/target/release/remacs_lib.lib(std-5708efc2c093aa05.std0.rcgu.o):(.xdata[$cppxdata$_ZN33_$LT$alloc..arc..Arc$LT$T$GT$$GT$9drop_slow17hfd2fb3bf13678e29E]+0x10): undefined reference to `__CxxFrameHandler3'
...
../rust_src/remacs-lib/target/release/remacs_lib.lib(std-5708efc2c093aa05.std0.rcgu.o):(.xdata[$cppxdata$_ZN4core3ptr13drop_in_place17h2431536efcd84b2cE]+0x10): more undefined references to `__CxxFrameHandler3' follow
collect2.exe: error: ld returned 1 exit status
make[1]: *** [Makefile:401: etags.exe] Error 1
make[1]: Leaving directory '/c/projects/remacs/lib-src'
make: *** [Makefile:413: lib-src] Error 2
```

shaleh:
Relevant? https://social.msdn.microsoft.com/Forums/Windows/en-US/894b459c-f278-433d-8759-cfd435da4f2c/undefined-references-that-appear-during-linking-of-library-as-a-dll-but-not-as-a-static-library-on?forum=vssmartdevicesnative

shaleh:
https://www.appveyor.com/docs/build-environment/ is info about available environments. The appveyor.yml in the tree is where you change things if needed. You can add any changes to this file to this PR and they will be immediately applied to the PR.

gaak99:
Thanks for the links.  Looks like Appveyor is running Windows Server 2016 and my sys is running Windows 10 Home. 

FWIW I checked my ./configure results w/Appveyor ones and except for libxml2 they looks to be the same. (libxml2 is installed but could not find it. I can try to reinstall it)

But so far a clean baseline build on my sys fails.  At this point my guess it's the diff versions of Windows but still working on it.

In the meantime I'll push my latest code and yall can review it and let's see the Appveyor results now that I added Windows support.

shaleh:
Could you set the code up so that the original C code is used for Windows and the Rust dired code for the unix-like systems? Then we could focus on making Windows work.

gaak99:
I have an Appveyor acct now so have a puncher's chance of getting Windows code working.

Just pushed some build/rustfmt fixes.

And I just got a clean build on Appveyor I think.

(yesterday I still got that (very)surprising benchmark-tests fail and not quite sure why it's gone today)

shaleh:
@gaak99 the benchmark test fails intermittently. If that is your only Windows failure we will treat it as a success. When I looked last it appears the benchmark fails to run properly since both times are `0.0` and the code insists on a greater than check.

gaak99:
Looks like the benchmark-tests false positive was the cause of the Appveyor fail.  Anything else I need to do?

rust_src/src/dired.rs
```diff
@@ -0,0 +1,54 @@
...
+use lisp::{defsubr, LispObject};
+#[cfg(not(windows))]
+use lists::list;
+
+fn get_user_real_login_name() -> LispObject {
+    unsafe { globals.f_Vuser_real_login_name }
+}
+
+#[cfg(windows)]
+fn get_users() -> LispObject {
```
shaleh: Should this not return a one element list to act like the other `get_users`?

gaak99: Yes it should -- good catch -- I just pushed a fix and added a test to catch it and similar problems.

PR #739, opened 2018-05-14
port DEFVAR_{LISP,LISP_NOPRO,BOOL,INT} macros to Rust

db48x:
and finally clear out cmds.c!


rust_src/remacs-lib/docfile.rs
```diff
@@ -137,6 +138,24 @@ pub extern "C" fn scan_rust_file(
...
+            let kind = match kindstr {
+                "lisp" => LISP_OBJECT,
+                "lisp_nopro" => LISP_OBJECT,
+                "bool" => BOOLEAN,
+                "int" => EMACS_INTEGER,
+                _ => panic!("oops"),
+            };
+            let fvname = &caps[2];
+            assert!(fvname.starts_with("f_V"));
+            let vname = CString::new(fvname.splitn(2, "_").nth(1).unwrap()).unwrap();
```
shaleh: Any chance we could move to using the '?' operator here?

`let vname = CString::new(fvname.splitn(2, "_").nth(1)?)?;`

rust_src/src/eval_macros.rs
```diff
@@ -129,7 +129,95 @@ macro_rules! def_lisp_sym {
...
+            };
+            unsafe { ::remacs_sys::defvar_lisp(&mut o_fwd,
+                                               $lname.as_ptr() as *const i8,
+                                               &mut ::remacs_sys::globals.$fvname);
+            }
+            unsafe { ::remacs_sys::globals.$fvname = $value; }
+        }
+    };
+}
+#[macro_export]
```
shaleh: One, shouldn't there be an empty line between these macros?
two, can we get a comment point to more info on nopro/pro?

rust_src/remacs-lib/docfile.rs
db48x: Then it would have to return a `Result`.

shaleh: Right. I wonder if that might be better long term for this particular code.

db48x: It might be! I'll think about preparing a second pull request with that kind of thing in it.

admin/find-gc.el
```diff
@@ -56,7 +56,7 @@ Each entry has the form (FUNCTION . FUNCTIONS-IT-CALLS).")
     "term.c" "cm.c" "emacs.c" "keyboard.c" "macros.c"
     "keymap.c" "sysdep.c" "buffer.c" "filelock.c"
     "insdel.c" "marker.c" "minibuf.c" "fileio.c"
-    "dired.c" "cmds.c" "casefiddle.c"
+    "dired.c" "casefiddle.c"
```
Wilfred: Woah, I had no idea this file existed. We should look at updating it to look in Rust. I've created #747.

rust_src/remacs-lib/docfile.rs
```diff
@@ -137,6 +138,24 @@ pub extern "C" fn scan_rust_file(
...
+                static ref RE: Regex = Regex::new(r#"defvar_(lisp)!\((.+?),\s+"(.+?)",\s+(.+?)\);"#).unwrap();
+            }
+            let caps = RE.captures(line).unwrap();
+            let kindstr = &caps[1];
+            let kind = match kindstr {
+                "lisp" => LISP_OBJECT,
+                "lisp_nopro" => LISP_OBJECT,
+                "bool" => BOOLEAN,
+                "int" => EMACS_INTEGER,
+                _ => panic!("oops"),
```
Wilfred: Could we say why we're panicking in the message here? It helps future folks if they hit it :)

rust_src/src/eval_macros.rs
```diff
@@ -129,7 +129,100 @@ macro_rules! def_lisp_sym {
...
+                intvar: &mut ::remacs_sys::globals.$fvnaem,
+            };
+            unsafe { ::remacs_sys::defvar_int(&mut o_fwd,
+                                              $lname.as_ptr() as *const i8,
+                                              &mut ::remacs_sys::globals.$fvname); }
+            unsafe { ::remacs_sys::globals.$fvname = $value; }
+        }
+    };
+}
+// think about this later, since it uses offsetof
```
Wilfred: I'm not a fan of commented-out code, it tends to bit-rot. Could we move this to an issue instead?

db48x: sure

rust_src/remacs-lib/docfile.rs
db48x: I knew I forgot something

rust_src/src/cmds.rs
```diff
@@ -534,6 +534,10 @@ pub extern "C" fn rust_syms_of_cmds() {
     def_lisp_sym!(Qoverwrite_mode_binary, "overwrite-mode-binary");
     def_lisp_sym!(Qexpand_abbrev, "expand-abbrev");
     def_lisp_sym!(Qpost_self_insert_hook, "post-self-insert-hook");
+
+    /// Hook run at the end of `self-insert-command'.
+    /// This is run after inserting the character.
```
Wilfred: Do we extract docs from variables yet? In other words, does `C-h v post-self-insert-hook` work after this change?

I don't think it should hold up this PR if we don't do this yet though.

rust_src/src/eval_macros.rs
```diff
@@ -129,6 +129,84 @@ macro_rules! def_lisp_sym {
...
+/// These are used in the syms_of_FILENAME functions.
+///
+/// An ordinary (not in buffer_defaults, per-buffer, or per-keyboard)
+/// lisp variable is actually a field in `struct emacs_globals'.  The
+/// field's name begins with "f_", which is a convention enforced by
+/// these macros.  Each such global has a corresponding #define in
+/// globals.h; the plain name should be used in the code.
+///
+/// E.g., the global "cons_cells_consed" is declared as "int
+/// f_cons_cells_consed" in globals.h, but there is a define:
```
Wilfred: This isn't true of the Rust code, we use .f_Vfoo directly. We should update this paragraph or just remove it.

rust_src/src/remacs_sys.rs
```diff
@@ -1278,9 +1279,1303 @@ extern "C" {
...
+#[derive(Debug, Copy)]
+pub struct image_cache {
+    pub buckets: *mut *mut image,
+    pub images: *mut *mut image,
+    pub size: isize,
+    pub used: isize,
+    pub refcount: isize,
+}
+#[test]
+fn bindgen_test_layout_image_cache() {
```
shaleh: bindgen?

db48x: Yep. I used bindgen to generate them, since typing them all in manually would be annoying and quite error-prone.

Obviously at some point we'll actually use bindgen to generate these bindings during the build rather than checking them in, but this will do for now.

rust_src/src/cmds.rs
db48x: Alas, we do not.

rust_src/src/remacs_sys.rs
shaleh: ah. Thanks.

PR #743, opened 2018-05-15
Rust's ranges are exclusive of the last value

### LINKED ISSUE #742: ~ is not self-insert-command anymore?
Remacs from git 15.05.2018 (63d355d5fdf6b1f5edce1439169595a917559c61 )

RUST_BACKTRACE=1 src/remacs -q

Type: ~
Result: '~ is undefined'.
Expected: ~ inserted into the buffer.

db48x:
Just like the C code we ported it from.

In my defense, I just blindly copied this code from Sean, and didn't
double-check it. Fixes #742


db48x:
Technically you did.

PR #745, opened 2018-05-15
add form-feeds for pagination to cmds.rs

db48x:
This replicates the pagination present in cmds.c which this file
replaces.

I had meant to do this when I removed cmds.c, but I guess I forgot.


Wilfred:
@db48x it looks like this conflicts with our rustfmt build step. What do you think? Do you think the form-feeds are worth it? I imagine we could tweak the rustfmt configuration to allow this. I don't take advantage of them, do you?

db48x:
I'm going to close this for now. I'll revisit it in the future if https://github.com/rust-lang-nursery/rustfmt/issues/2757 pans out.

PR #753, opened 2018-05-16
Fix obarray checks in intern and mapatoms

### LINKED ISSUE #731: `suggest` crashes Remacs
Remacs crashes abruptly when running the [`suggest` package](https://github.com/wilfred/suggest.el). `suggest` takes some inputs and a desired output and attempts to find some Elisp code that will produce that output from those inputs. Here is an example:

```
;; Inputs (one per line):
5

;; Desired output:
nil

;; Suggestions:

(not 5) ;=> nil
(listp 5) ;=> nil
(consp 5) ;=> nil
(floatp 5) ;=> nil
(arrayp 5) ;=> nil
```

I'm looking for a function that takes 5 and returns `nil`, and those are `suggest`'s suggestions. That works fine on Remacs, and so do certain other queries. On the other hand, changing the `nil` to `t` and running the query causes Remacs to crash without warning, whereas on GNU Emacs it produces the expected output:

```
;; Inputs (one per line):
5

;; Desired output:
t

;; Suggestions:

(oddp 5) ;=> t
(atom 5) ;=> t
(natnump 5) ;=> t
(numberp 5) ;=> t
(integerp 5) ;=> t
```

I haven't figured out how to characterize the input/output sets that crash it, but here are some more examples.

Works:
```
;; Inputs (one per line):
'(1 2 3)
'(4 5 6)

;; Desired output:
'(1 2 3 4 5 6)
```

Crashes:
```
;; Inputs (one per line):
'(1 2 3)
'(4 5 6)

;; Desired output:
'((1 2 3) (4 5 6))
```

Works:
```
;; Inputs (one per line):
'(1)

;; Desired output:
'()
```

Works:
```
;; Inputs (one per line):
'()

;; Desired output:
'()
```

Crashes:
```
;; Inputs (one per line):
'()

;; Desired output:
'(1)
```

When I say "crashes", I mean the Remacs window vanishes immediately and a few seconds later an Apple dialog box comes up that says "Remacs quit unexpectedly." Occasionally it will freeze for a few seconds before crashing.

These crash triggers are pretty reliable, at least for me. I can post a crash report if it would be helpful.

`Remacs 26.0.60 (build 4, x86_64-apple-darwin16.7.0, NS appkit-1504.83 Version 10.12.6 (Build 16G29)) of 2018-05-07`

Wilfred:
We rely on check_obarray to verify that a value is an obarray. That
function assumed the argument was a vector (it used an unchecked
functions inappropriately).

This caused intern and mapatoms to crash when called with non-obarray
arguments.

Also fix an issue with intern where it assumed a non-vector means it
should use the global obarray, whereas it should only the global
obarray when called with nil.

Add tests for all these cases.

Fixes #731.

### DIFF
--- a/rust_src/src/obarray.rs
+++ b/rust_src/src/obarray.rs
@@ -97,8 +97,8 @@ pub extern "C" fn check_obarray(obarray: LispObject) -> LispObject {
         return obarray;
     }
 
-    let obarray = LispObject::from_raw(obarray);
-    let v = unsafe { obarray.as_vectorlike_unchecked().as_vector() };
+    // A valid obarray is a non-empty vector.
+    let v = obarray.as_vector();
     if v.map_or(0, |v_1| v_1.len()) == 0 {
         // If Vobarray is now invalid, force it to be valid.
         if LispObject::from_raw(unsafe { globals.f_Vobarray }).eq(obarray) {
@@ -107,7 +107,7 @@ pub extern "C" fn check_obarray(obarray: LispObject) -> LispObject {
         wrong_type!(Qvectorp, obarray);
     }
 
-    obarray.to_raw()
+    obarray
 }
 
 #[no_mangle]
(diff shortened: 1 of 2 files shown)


Wilfred:
Bouncing build.

Wilfred:
Some strange rustdoc failure on the Cerrors stage.

Wilfred:
Hm, so the travis-ci/pr build passed, but the travis-pr/push build failed. My understanding ([according to Stack Overflow](https://stackoverflow.com/a/41534988/509706)) is that the pr build tests the code after doing a trial merge.

I've rebased on latest master, to see if that helps fix the Cerrors build.

Issue #752, opened 2018-05-16
Cannot compile using newer nightly due to proc macro issue


darkanthey:
This happened with the latest git master repo. 
I used rust nightly-2018-05-01. This rust works well before.

`Compiling remacs-lib v0.1.0 (file:///home/xxx/remacs_src/rust_src/remacs-lib)
error: custom attribute panicked
   --> src/base64.rs:341:1
    |
341 | #[lisp_fn(min = "1")]
    | ^^^^^^^^^^^^^^^^^^^^^
    |
    = help: message: Invalid lisp_fn attribute: Unexpected literal type `string`
error: aborting due to previous error`

shaleh:
We saw this on the Windows appveyor builds when the wrong compiler was being installed. Make sure the Rust compiler version run for this build matches the version in `rust-toolchain`. The command `rustup` allows you to specify a version for just one project using an override. You do not have to switch the compiler used for other Rust projects.

shaleh:
ok. This is actually an issue with the latest nightly. Something is changing in proc macro land. We need to get a handle on it and perhaps make some noise.

nitishch:
@shaleh The fix that I talked about in Gitter channel is after including your changes in #759. So I guess this and #759 have to be merged in a single branch.

PR #754, opened 2018-05-16
Document how to set the correct rust toolchain

Wilfred:
This should help with #752.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ more Emacs-y.
 
         cd /path/to/remacs
         rustup install `cat rust-toolchain`
+        rustup override set `cat rust-toolchain`
 
 2. You will need a C compiler and toolchain. On Linux, you can do
    something like `apt-get install build-essential automake`. On


shaleh:
You should not have to do this. Just running make at the top of tree will detect and install the correct toolchain.

shaleh:
We removed it from the documentation because of the confusion. Maybe we should add it as a separate optional step. Or maybe this only happens on Windows?

Wilfred:
I found I needed to do this after 'rustup update'. I'll dig further.

jaccarmac:
```
darweshi :: ~/src/src/github.com/Wilfred/remacs » rustc --version
rustc 1.25.0-nightly (29c8276ce 2018-02-07)
darweshi :: ~/src/src/github.com/Wilfred/remacs » cd
darweshi :: ~ » rustc --version
rustc 1.28.0-nightly (952f344cd 2018-05-18)
darweshi :: ~ »
```

Works for me as well, really weird that you're encountering an issue with it.

Wilfred:
Further confusing matters, overrides don't apply to child directories:

```
$ rustc --version
rustc 1.25.0-nightly (29c8276ce 2018-02-07)
~/projects/remacs 
$ cd rust_src 
~/projects/remacs/rust_src 
$ rustc --version
rustc 1.26.0-nightly (517f24025 2018-03-31)
```

This makes doing `cargo check` less convenient.

shaleh:
@Wilfred, does moving the `rust-toolchain` file into `rust_src` help?

jaccarmac:
Do you have any manual overrides setup? Last I checked child directories were fine, but I'll check again on my machine in a couple hours.

jaccarmac:
```
darweshi :: ~/src/src/github.com/Wilfred/remacs » rustc --version
rustc 1.25.0-nightly (29c8276ce 2018-02-07)
darweshi :: ~/src/src/github.com/Wilfred/remacs » cd rust_src 
darweshi :: ~/src/src/github.com/Wilfred/remacs/rust_src » rustc --version
rustc 1.25.0-nightly (29c8276ce 2018-02-07)
darweshi :: ~/src/src/github.com/Wilfred/remacs/rust_src » rustup override list
no overrides
darweshi :: ~/src/src/github.com/Wilfred/remacs/rust_src » rustup --version
rustup 1.11.0 (e751ff9f8 2018-02-13)
darweshi :: ~/src/src/github.com/Wilfred/remacs/rust_src »
```

I would make sure you aren't accidentally setting up conflicting overrides. Beyond that it may just be works-on-my-machine and in that case I'm sorry for the red herring.

Wilfred:
You're probably right. I'll reopen this if I can reliably reproduce.

shaleh:
We might want to document that setting the override is not necessary and can cause problems if we update the rust-toolchain file.

PR #759, opened 2018-05-21
Get OSX building with new Alloc interface.

### LINKED ISSUE #725: New Rust compiler breaks OSX build
The build dies trying to create the remacs binary.

shaleh:
The interface to the Alloc trait is changing. This PR gets us building
again. This is probably unstable ground for bit.

Closes #725.


shaleh:
Sadly, this build is broken due to #752

Issue #758, opened 2018-05-21
"Invalid function: lamba" in .emacs file


jmelesky:
The following portion of my `.emacs` file is causing the message "Invalid function: lambda" on remacs startup, and has been for some weeks.

It runs without complaint on emacs 25.3 (and several spot-checked earlier versions).

    (setq
     emacs-lisp-mode-hook '(lambda () (font-lock-mode 1))
     c-mode-hook '(lambda () (font-lock-mode 1))
     c++-mode-hook '(lambda () (font-lock-mode 1))
     html-mode-hook '(lambda () (font-lock-mode 1))
     shell-script-mode-hook '(lambda () (font-lock-mode 1))
     makefile-mode-hook '(lambda () (font-lock-mode 1))
     mutt-mode-hook '(lambda () (font-lock-mode 1))
     perl-mode-hook '(lambda () (font-lock-mode 1))
     cperl-mode-hook '(lambda () (font-lock-mode 1))
     c++-c-mode-hook '(lambda () (font-lock-mode 1))
     tex-mode-hook '(lambda () (font-lock-mode 1))
     shell-mode-hook '(lambda () (font-lock-mode 1))
     dired-mode-hook '(lambda () (font-lock-mode 1))
     compilation-mode-hook '(lambda () (font-lock-mode 1)))

Happy to provide any further detail required.

(Edited to add: There's a very simple workaround in my case, which is remove the code, as it's old and no longer needed. The behavior discrepancy is troubling, though, which is why I mention it.)

Wilfred:
Could you start remacs with `--debug-init` to get a traceback?

shaleh:
You can also but this in a .el file and test for the regression using the `git bisect` tools in `rust_src/admin`.

nitishch:
Smaller reproducible case:

```elisp
    (setq temp-hook '(lambda () (setq c 1)))
    (run-hooks 'temp-hook)
```

jmelesky:
I finally get a chance to try and capture debug output, and it turns out y'all have already fixed the issue!

Thank you much. Closing this, as it's been fixed in master.

PR #760, opened 2018-05-23
run-hooks should accept a symbol evaluating to a function

nitishch:
Addresses #758.

Cause of the issue:

When a hook's value is a single function, C behaviour is to call that function. To check if the value is a function, C code uses `FUNCTIONP`. There was a mistake at this part in ported Rust code and because of that `'(lambda () (+ 1 1))` is treated as a list of functions and this leads to the error `Invalid function: lambda`

Since `run-hooks` takes a symbol as an argument, I couldn't use `let` construct in tests. I used `setq` and `makunbound`. Is this an acceptable usage?

### DIFF
--- a/test/rust_src/src/eval-tests.el
+++ b/test/rust_src/src/eval-tests.el
@@ -236,4 +236,15 @@
   (let ((f (lambda () nil)))
     (should (functionp f))))
 
+(ert-deftest eval-tests--run-hooks-base ()
+  ;; We can't use let construct here because run-hooks takes a symbol
+  ;; as an argument and because of lexical binding, that symbol's
+  ;; value will always be nil
+  (setq function-hook '(lambda () (+ 1 1)))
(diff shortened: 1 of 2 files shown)


test/rust_src/src/eval-tests.el
```diff
@@ -236,4 +236,15 @@
...
+(ert-deftest eval-tests--run-hooks-base ()
+  ;; We can't use let construct here because run-hooks takes a symbol
+  ;; as an argument and because of lexical binding, that symbol's
+  ;; value will always be nil
+  (setq function-hook '(lambda () (+ 1 1)))
+  (setq list-of-functions '((lambda () (+ 2 2))
+                            (lambda () (+ 3 3))))
+  (run-hooks 'function-hook 'list-of-functions)
+  (makunbound 'function-hook)
+  (makunbound 'list-of-functions))
```
Wilfred: Nice test, it still tidies up afterwards :+1:

PR #762, opened 2018-05-27
Start porting "casefiddle.c" (case conversion functions).

rocky:
Should be doing at least the other simple functions in the `CaseAction` enumeration


rust_src/src/casefiddle.rs
```diff
@@ -0,0 +1,26 @@
+//! Case conversion functions.
+use lisp::defsubr;
```
shaleh: https://github.com/Wilfred/remacs/blob/master/CONTRIBUTING.md#remacs-style

rust_src/src/remacs_sys.rs
```diff
@@ -1369,6 +1377,8 @@ extern "C" {
     pub static Vprocess_alist: LispObject;
     pub static Vrun_hooks: LispObject;
 
+    pub fn casify_object(case_action: CaseAction, object: LispObject) -> LispObject;
```
shaleh: Typically we add to the bottom of the list of exported functions.

rust_src/src/casefiddle.rs
```diff
@@ -0,0 +1,26 @@
...
+use lisp::LispObject;
+use remacs_macros::lisp_fn;
+use remacs_sys::{casify_object, CaseAction};
+
+/// Convert argument to upper case and return that.
+/// The argument may be a character or string.  The result has the same type.
+/// The argument object is not altered--the value is a copy.  If argument
+/// is a character, characters which map to multiple code points when
+/// cased, e.g.  , are returned unchanged.
+/// See also `capitalize', `downcase' and `upcase-initials'.  */
```
nitishch: Remove the final `*/`

nitishch: Also, the character ﬁ is missing from the comment.

rust_src/src/remacs_sys.rs
```diff
@@ -1748,7 +1758,7 @@ pub mod font {
 
 #[cfg(test)]
 macro_rules! offset_of {
-    ($ty: ty, $field: ident) => {
+    ($ty:ty, $field:ident) => {
```
shaleh: You are failing rustfmt because you likely have a newer rustfmt installed. This section should not have changed.

rust_src/src/casefiddle.rs
rocky: Good catch. Should be removed now in a01ea26755e6f1f2f106d7481f6d382c02953c5c

rocky: Another good catch. That is also in a01e26

rust_src/src/remacs_sys.rs
rocky: Ok. it is moved now. If I got that wrong, feel free just to move to the right place. Thanks.

rust_src/src/remacs_sys.rs
rocky: So many "standards" to choose from! How do I get the right rustmt installed? 

```
$ rustup component add rustfmt-preview 
info: component 'rustfmt-preview' for target 'x86_64-unknown-linux-gnu' is up to date
```

Ok good. Then I try: 
```
$ (cd rust_src && cargo fmt -- --write-mode overwrite) 
error: no such subcommand: `fmt`
```
I guess I have the wrong version of cargo? If that is the case, how do I get the right one so I can run with the right version of rustfmt? Thanks

rocky:
> Did you forget to git add the ERT test?

Indeed I did. That should be added now. (I make lots of mistakes, so I write lots of tests and work on debuggers.)

rocky:
I've removed the WIP on this. Unless other have other ideas, my view is early and often merge requests. 

Here is the bigger picture... I'll finish the enumerations that `casify_object` handles. I may then try as I have time to convert some of the easier-to-convert C functions to rust. 

But really my long-term goal is to rewrite the bytecode interpreter in Rust so that we can track frame structures when there is an error or so that it can be debugged. That is a very very long-term project, that I have somewhat mapped out in my head but it is vaporware right now.

rust_src/src/remacs_sys.rs
nitishch: First section in README says how to get proper version of Rust for this project. I guess you didn't run `rustup` from the Remacs root.

To format all the files, run `make rustfmt`.

nitishch:
About the ERT file, generally tests are put in `test/rust_src/src`. I guess that's because we don't have to touch existing upstream Emacs tests.

rust_src/src/casefiddle.rs
```diff
@@ -1,18 +1,19 @@
 //! Case conversion functions.
-use lisp::defsubr;
 use lisp::LispObject;
+use lisp::defsubr;
 use remacs_macros::lisp_fn;
```
shaleh: remacs_macros belongs on its own line before the local modules. See other files in the directory for examples.

shaleh:
I am a fan of lots of testing too.

We have been migrating existing elisp tests from the `test/src/ subtree` to the `test/rust_src subtree` as we port code.

rocky:
With respect to: 
> About the ERT file, generally tests are put in test/rust_src/src. I guess that's because we we wouldn't have to touch existing upstream Emacs tests.

That makes sense. However in this situation, tests were already there to cover the changes, so none were added. 

With respect to:
> We have been migrating existing elisp tests from the `test/src/` subtree to the `test/rust_src` subtree as we port code.

This doesn't make sense here right now as most of the tests are for the C code and some are for the Rust code. The existing C tests aren't exactly the way I'd like to see them from simple primitives to more complicated situations, but then I give people a lot of leeway in how they want to operate.  Right now I don't want to go down the rabbit hole of separating and rewriting the existing Emacs Lisp tests for casefiddle.  If however others would like to do that, great. The bottom line is that the existing tests cover the changed code.

rocky:
> Some of it is just old practices.

If that's the case (pun unintended), wouldn't GNU Emacs folks like to know about and fix these things too? 

IMHO if you take from GNU Emacs you should also consider giving back to it.

rocky:
For what it is worth, I posted the question of to the [emacs developer's maling list](https://lists.gnu.org/archive/html/emacs-devel/2018-05/msg00793.html) and the believe that what they had is as intended and correct. 

Given this, I don't think remacs should be changing the GNU Emacs decision, especially without public decision and not without documenting this decision describing why.

shaleh:
It is git. Nothing is permanent. Demonstrate why my change is incorrect beyond "not what Emacs does" and we can change it again. No big deal. But if the argument is only doing what Emacs did that is not sufficient.

Does it render wrong? Does it look bad in some language / font / whatever? Does it make the text harder to read? Search around?

rocky:
I'm pointing out what's been discussed. 

> Demonstrate why my change is incorrect beyond

Let's go the other way: demonstrate why your change is correct. Keep in mind your opinion is just your opinion.

Wilfred:
Thanks for your PR @rocky! The casefiddle function was the first place I managed to segfault GNU Emacs, so it's great to see it moving to Rust :)

Regarding docstrings, it's pretty common for us to port C functions where the docstring was briefly written when the function was first implemented, and never subsequently updated. There are plenty of docstrings where the wording could improve or don't fully follow emacs conventions. We have an issue to make them better.

Whilst I'd personally like to change the conventions in sone places, there's too much elisp inside and outside of emacs for that to be feasible.

The convention is ASCII in docstrings (adding curly quotes was painful and you now have to escape `` ` `` and `'` in docstring code samples), but I don't object to however you want to do spacing around hyphens. Ideally we'd go with the most common practice used in the codebase. You mentioned that comint uses this style too, for example.

We have had recent PRs that fix typos in docstrings whilst porting. I agree these could also be upstreamed, but no obligation :)

shaleh:
The front page of the project says this:

```
There's a difference between the idea of Emacs and the current implementation of Emacs. Forking allows us to explore being even more Emacs-y.
```

In this code base we are making no assumptions that what is present today is what we want tomorrow. Everyone should feel like they can question the code and style. We stand on 30+ years of Emacs history. Every one of us appreciates that. But sometimes a decision was made in the past that is no longer true. We do send bugs and potentially solutions back upstream. But other changes we make here that the Emacs team cannot. The Emacs tree is slow to change and has its warts because of it. All projects that live to a certain age will.

The process of review here is simple. If a change is merged and someone sees a problem, they open a new PR. In that PR we can talk about the problem and the solution. A consensus will be reached. At that point the PR will either be merged or closed depending on what was decided. It is all in git. Everything is ephemeral.

rocky:
I've wasted too much time on this. The facts are there. The remacs community can decide what it wants to do, but its decisions do have consequences. My suggestion to @shaleh is simple: when you make a mistake just admit and move on.

db48x:
I might be sticking my nose in where it isn't wanted, but I'd like to help if I can.

I'm certain that Sean didn't mean any harm with his change to the comments; it's not uncommon for us to step in with a commit like that when we are reviewing a pull request and want to avoid an extra round-trip with the author, especially with minor or procedural things like the order of imports (which we only do in order to try to minimize merge conflicts). On the other hand, I think that this should only be done for things that a reviewer would deny an approval for.

It's also not uncommon for us to change the comments in the code we're porting; sometimes they're wrong or outdated, or they would become wrong after our port, or we just feel that the wording could be improved. On the other hand, I agree with you that changing the typography or style of comments is probably unwarranted. Maybe we could come up with a better style, but it seems unlikely.

I would be happy to approve a pull request from you which puts the comments back how they were. I also hope that this minor disagreement doesn't put you off of your researches; we need all the help we can get, and anyone willing to look into the bytecode interpreter is especially welcome.

Wilfred:
> As another example of this kind of disparity there's this dogged adherence to a particular version of rustftmt which under certain occasions messes up in installation. Contrast the strictness of using rustfmt with the looseness in understanding under what conditionst rustfmt messes up, or how to fix it that I found on the gitter channel. The fact that rustfmt can't make up its mind about Rust style in an upward compatible way would be cause for me to stop insisting on using it.

@rocky this is my fault. `rustfmt` is supposed to make lives easier: it formats for you, and everything is consistent. We've fallen behind the latest upstream rustfmt, which causes surprised for people. You can run `rustup install $(cat rust-toolchain) && rustup component add rustfmt-preview && cd rust_rust && cargo fmt` to fix your code.

We've fallen behind in our rust toolchain versions. Whilst upgrading `rustc` has been more complex, we should look at moving to the latest rustfmt. I agree this is a nasty gotcha at the moment, and we should work to improve this.

rocky:
Thanks for the information and for working to improve this aspect. 

I'm not opposed at all to using formatters and neat tools that improve things. It's the dogged adherence to them when they make aren't working quite right. When that happens, I'd immediately drop the CI requirement. You can always go in _after_ things are copacetic and reformat everything. 

When a tool is not quite functioning properly all you are doing is raising the barrier for contributions and frustration level for people who might contribute. 

This is a little off topic, but  is "standard" formatting really that important it needs to be checked on _every_ commit instead of perhaps issuing a warning? 

Personally I don't understand the _obsession_ with "standard" formatting.  (I note however that a lot of this discussion has been over spacing around dashes, but to me the change was more about antagonism and lack of respect than it was about style). 

Consider all the variation in natural language with how to express something. Leaving as side grammar issues like do you drop the subject in an imperative, and just focus at the word level, some people use contractions and others don't (or is it "do not"? - and "can't can be "can not" and "cannot").  And then you throw in all the different dialects and ways of speaking. And yet, people can easily cope with it. 

If it really were the case that  a particular kind of camel case, or snake case for classes of identifiers were universally obviously superior, then more programming languages would agree better on this. But they don't, so Java uses camelCase while C uses snake_case. And you probably can find ardent defenders of why each one is superior. 

But even more amusing is the fact that rustfmt changes its decisions. 

Yes, I understand that a consistent style is more pleasant to read. I find that true myself.

But as a long time open-source developer and contributor, I have just had to tolerate  many different styles of formatting, because the contribution was more important than the coding style used.

So I submit that those who are so put off by style things haven't worked with programmers, some of them brilliant, or have had to switch between several programming languages.

The last time I looked Larry Wall's Perl coding style wasn't exactly what the Perl critic recommends. And if you look at how David Korn formats his code it is different from the way most POSIX programmers do. Should we tell them to fix up their coding style? I don't think so. Instead I think as programmers should be a little more tolerant.

Issue #717, opened 2018-04-15
re-builder freezes buffers


cjohansson:
When I start `M-x` `re-builder` in a clean remacs instance (started with `src/remacs -Q`) all buffers freezes, however it seems that the application is responsive in some way, the mini-buffer is updating on events and pop-ups are opened when I force quit the application but other than that it's frozen.

cjohansson:
Update, I don't get this issue if I run remacs with `src/remacs -Q -nw` so it seems to be related to the GUI mode. I use macOS 10.13.4

shaleh:
```
(defun my/test-issue ()
  ;; (interactive)
  (select-window (or (get-buffer-window "*my-test*")
                       (progn
                         (split-window (selected-window) (- (window-height) 4)))))
  (switch-to-buffer (get-buffer-create "*my-test*")))
```

Above is the minimal test case to reproduce. `M-: (my/test-issue)` or uncomment the interactive line.

cjohansson:
Yeah I can confirm I get the same issue with your test case

shaleh:
This does _not_ happen on my Linux system. I just built from master to confirm.

cjohansson:
Strange, it seems like the issue is related to running `(select-window)` before `(switch-to-buffer)`. Running them independently works.

**These examples work:**

```
(defun my/test-issue ()
  (select-window (or (get-buffer-window "*my-test*")
                       (progn
                         (split-window (selected-window) (- (window-height) 4)))))
  (let (buffer (get-buffer-create "*my-test*"))
    ;; (switch-to-buffer buffer)
    ))
(my/test-issue)
```

```
(defun my/test-issue ()
  ;; (select-window (or (get-buffer-window "*my-test*")
  ;;                      (progn
  ;;                        (split-window (selected-window) (- (window-height) 4)))))
  (let (buffer (get-buffer-create "*my-test*"))
    (switch-to-buffer buffer)
    ))
(my/test-issue)
```

**This hangs on line with** `(switch-to-buffer)`:

```
(defun my/test-issue ()
  (select-window (or (get-buffer-window "*my-test*")
                       (progn
                         (split-window (selected-window) (- (window-height) 4)))))
  (let (buffer (get-buffer-create "*my-test*"))
    (switch-to-buffer buffer)
    ))
(my/test-issue)
```

shaleh:
Is it the select or the split-window?

cjohansson:
It seems like it's the split-window, same issue with this:

```
[...]

PR #763, opened 2018-05-28
Temporary fix for MacOS re-builder freeze

cjohansson:
Resolves [#717](https://github.com/Wilfred/remacs/issues/717) even though this solution is not optimal. The original GNU Emacs C-code has the != 0 logic but for some reason it doesn't work in remacs, the glyph matrix seems to hold invalid values. This bug that seems to happen when a small window is being opened, like when launching re-builder.

(I moved these changes to from master to a separate branch in my forked repository, that's why I closed the old pull request and created a identical one.)

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -70,7 +70,7 @@ impl LispWindowRef {
 
         if mode_line_height >= 0 {
             mode_line_height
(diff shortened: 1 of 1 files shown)


shaleh:
Resolved.

PR #766, opened 2018-05-30
Add checks for eassert and emacs_abort in pre-commit hook

nitishch:
I guess not formatting code with `rustfmt` is a common beginner mistake. Hence added some explanatory text.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,8 @@ Thank you for contributing to Remacs!
   you have the correct version by using rustup (it will use the
   correct version based on our [toolchain
   file](https://github.com/rust-lang-nursery/rustup.rs#the-toolchain-file))
-  and running `rustup component add rustfmt-preview`
+  and running `rustup component add rustfmt-preview`. See
+  [Helpful Hooks](#helpful-hooks) for help to automatate this verification.
 * Add docstrings to your Rust functions `/// This function does ...`
 * _Really_ great PRs include tests. See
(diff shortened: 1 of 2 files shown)


nitishch:
Why do we have to verify the `rustfmt` version? Won't it pick the one specified in the toolchain file automatically? Checking if rustfmt is installed makes sense to me.

shaleh:
We were reminded of the hook because a couple of people were using a newer version of rustfmt. Which led to PRs we had to deny until they had rolled back their version.

nitishch:
Yeah. Right. I was in a hurry yesterday and forgot about those issues. I did not understand the cause of those issues though. Why was `rustup` not ensuring the right version there?

shaleh:
I am not sure either. My concern is that rustup is not installing a rustfmt for each version. Which means the developer needs to install rustfmt from within the remacs tree to get the right version and this is likely the wrong version for other work they are doing. Yuck.

The procedural macro changes are holding us from moving forward.

shaleh:
Ok, I just tried this again and things have improved.

The rustfmt version is bound to the version of nightly in use. So the user can have multiple rustfmt's installed.

shaleh:
I am going to merge this PR as is and make a separate one to remove the version checks which have bit rotted.

Wilfred:
@nitishch I've added you as a collaborator, so you now have the power to approve other PRs if you wish :)

nitishch:
Thanks @Wilfred :)

Issue #726, opened 2018-05-05
Active window isn't highlighted


nickdrozd:
GNU Emacs:
<img width="1280" alt="screen shot 2018-05-05 at 12 44 54 pm" src="https://user-images.githubusercontent.com/17630138/39665430-27822b8e-5062-11e8-958f-f93409437cf0.png">

Remacs:
<img width="1280" alt="screen shot 2018-05-05 at 12 44 59 pm" src="https://user-images.githubusercontent.com/17630138/39665434-33cf2158-5062-11e8-98f9-b5aab582f9f1.png">

In the GNU Emacs picture, the active window has its status bar highlighted, while in the Remacs picture it doesn't. This is actually a pretty big usability issue for me.

GNU Emacs: `GNU Emacs 25.2.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F1911)) of 2017-04-21`

Remacs: `Remacs 26.0.60 (build 2, x86_64-apple-darwin16.7.0, NS appkit-1504.83 Version 10.12.6 (Build 16G29)) of 2018-05-05`

(I'm assuming this is a Remacs issue rather than an Emacs version issue, but of course I could be wrong.)

shaleh:
Confirmed on OSX. Both terminal and GUI exhibit this behavior.

shaleh:
I tried out `highlight-focus` and the highlight does change when the window becomes the selected window. However the status bar does not.

nitishch:
I hit this issue in Linux too.

shaleh:
Please try again with current master.

PR #767, opened 2018-05-30
Fix current mode line height.

### LINKED ISSUE #744: Themes are busted
1. Start Remacs with `-Q` and do `eval-expression`: `(load-theme 'light-blue)`. For me, this causes the whole window to fill up with light blue, after which Remacs freezes. Other themes freeze up too, but only some. `(load-theme 'wheatgrass)`, for instance, doesn't seem to cause any problems. (You can run `customize-themes` and play around with the option presented to see which ones work and which ones don't.)

2. Start Remacs with `-Q` and do `eval-expression`: `(load-theme 'manoj-dark)`. Now do `eval-expression` and start typing `(load-theme`. For me, Remacs freezes at this point, apparently when it tries to load the function signature. Sometimes if I wait long enough I can `C-g` and regain control, but usually it just remains frozen. I use `manoj-dark`, so this is a problem for me.

`Remacs 26.0.60 (build 8, x86_64-apple-darwin16.7.0, NS appkit-1504.83 Version 10.12.6 (Build 16G29)) of 2018-05-14`

shaleh:
First, reorg the code to be more clear.
Second, use bindgen generated code to access the C structs. The
underlying bug was due to mishandling of the glyph_row array.

Thanks to @cjohansson for the debugging help.

Closes #744, #726, #717


cjohansson:
This sounds terrific, great job. I haven't been able to use Remacs as my main editor since these bugs emerged

PR #772, opened 2018-06-02
Update toolchain

Wilfred:
This isn't the very latest version, but recent nightlies have caused
problems for our OS X build. Start with a modest update to exercise
the build.


shaleh:
I am working on the OSX issue. @nitishch figured out the solution to the procedural macro problem. So within a week or so we can likely update to the current nightlys. Assuming they get the Alloc bits nailed down. There is a PR open that they are trying to land.

shaleh:
https://github.com/rust-lang/rust/pull/51241 is the one I am watching.

PR #774, opened 2018-06-02
Don't run a full build at the C errors stage

Wilfred:
All we're checking is that we get a compile error when we should. We
don't need a full build for this, and it's adding ~20 minutes to our
build time.


.travis.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         - sudo apt install -y texinfo libgif-dev libxpm-dev
         - travis_wait rustup install $(cat rust-toolchain)
         - ./autogen.sh && ./configure --without-makeinfo --with-x=no --with-ns=no --without-gconf --without-gsettings --with-gif=no
-        - make -j 3
+        - cd src && make globals.h && cd ..
```
shaleh: What you want is `make -C src globals.h`

PR #775, opened 2018-06-02
Port process-type

### LINKED ISSUE #737: Port process-type
```c
DEFUN ("process-type", Fprocess_type, Sprocess_type, 1, 1, 0,
       doc: /* Return the connection type of PROCESS.
The value is either the symbol `real', `network', `serial', or `pipe'.
PROCESS may be a process, a buffer, the name of a process or buffer, or
nil, indicating the current buffer's process.  */)
  (Lisp_Object process)
{
  Lisp_Object proc;
  proc = get_process (process);
  return XPROCESS (proc)->type;
}
```

leemachin:
Closes #737 

Left out a test for serial processes because I'm not sure how to do that without depending on something like socat, but there's enough there to prove it works :)


rust_src/src/process.rs
```diff
@@ -263,6 +268,15 @@ pub fn process_status(process: LispObject) -> LispObject {
     status
 }
 
+/// Return the connection type of PROCESS.
+/// The value is either the symbol `real', `network', `serial', or `pipe'.
+/// PROCESS may be a process, a buffer, the name of a process or buffer, or
+/// nil, indicating the current buffer's process.
+#[lisp_fn]
+pub fn process_type(process: LispProcessRef) -> LispObject {
+    process.process_type()
```
nitishch: The argument to `process-type` can either be a process/buffer/string/nil. When you define the argument as of type `LispProcessRef`, it accepts only process argument.

leemachin: Had a feeling it wasn't so simple ;) thanks!

shaleh: Indeed. You need the call to `get_process` or its equivalent.

leemachin:
You might guess I'm fairly fresh to Rust so if there's a more idiomatic approach to what I've done here I will gladly accept the feedback. Other than that, it should behave as expected now.

shaleh:
Do not stress being new to Rust. We are happy to help people learn.

rust_src/src/process.rs
```diff
@@ -263,6 +268,32 @@ pub fn process_status(process: LispObject) -> LispObject {
     status
 }
 
+/// Return the connection type of PROCESS.
+/// The value is either the symbol `real', `network', `serial', or `pipe'.
+/// PROCESS may be a process, a buffer, the name of a process or buffer, or
+/// nil, indicating the current buffer's process.
+#[lisp_fn]
+pub fn process_type(process: LispObject) -> LispObject {
+    let mut p;
```
shaleh: Whenever you mark a variable as mutable you should ask why. If the algorithm really requires updates that is expected. In the code below you can use the fact that if/else is also an expression which returns a value.

rust_src/src/process.rs
```diff
@@ -263,6 +268,32 @@ pub fn process_status(process: LispObject) -> LispObject {
...
+    };
+
+    if p.is_nil() {
+        return p;
+    }
+
+    let p_ref = p.as_process_or_error();
+
+    p_ref.process_type()
+}
```
shaleh:
```
let process = if process.is_string() {
    let p = get_process(process);
    if p.is_nil() {
        get_buffer_process(process)
    } else {
        unsafe { cget_process(process) }  // no more to_raw
   }
};

if process.is_nil() {
    process
} else {
    let p_ref = process.as_process_or_error();
    p_ref.process_type()
}
```

shaleh: Overwriting the function argument with the transformed version of it is also a common idiom.

shaleh: The original code uses a single function called `get_process`. Why not port it using the code above as a beginning? I see there is a `get_process` in this Rust file already. You can rename it to `get_process_lisp` and set the name to be `get_process` in the lisp_fn macro so it still works in C.

leemachin:
Makes sense 👍 saw a few places where this logic exists (with very subtle variations in some places, e.g. handling of processes when the string can refer to either a process name or a buffer name) so it makes sense to abstract it out and do a bit of cleanup here.

rust_src/src/process.rs
leemachin: Great :) I thought the `mut` was weird, just distracted myself a bit with the nested conditionals.

rust_src/src/process.rs
```diff
@@ -84,15 +89,36 @@ impl LispProcessRef {
...
+/// current buffer.
+pub fn get_process(process: LispObject) -> LispProcessRef {
+    let process = if process.is_string() {
+        let p = lisp_get_process(process);
+        if p.is_nil() {
+            get_buffer_process(process)
+        } else {
+            p
+        }
+    } else if process.is_nil() {
```
shaleh: That should not be an else. It should be a stand alone if.

leemachin: Good spot.

leemachin:
Back to work today so progress on finishing this off might be a little slower, but I've more or less grasped what needs to be done.

shaleh:
No worries, it will be waiting for you. :smile:

You just need to be ready to rebase.

rust_src/src/process.rs
```diff
@@ -84,15 +89,52 @@ impl LispProcessRef {
...
+    }
+
+    if proc_or_buf.is_buffer() {
+        let proc = get_buffer_process(proc_or_buf);
+        if proc.is_nil() {
+            panic!("Buffer {} has no process", name.as_string_or_error())
+        }
+        proc
+    } else {
+        // check process
```
shaleh: Should this not return the process?

leemachin: There's some stuff in the original C function that does some extra validation (there are bits and pieces to see if the process is actually running or not, it seems). I just need to figure out what exactly that is.

shaleh:
brotzeit has picked this up in a new PR.

PR #776, opened 2018-06-03
port set-process-filter

brotzeit:


rust_src/src/process.rs
```diff
@@ -273,14 +282,65 @@ pub fn set_process_buffer(process: LispObject, buffer: LispObject) -> LispObject
...
+    // for example when doing something like
+    //
+    // (setq process (start-process ...))
+    // (debug)
+    // (set-process-filter process ...)
+    if filter.is_nil() {
+        filter = Qinternal_default_process_filter;
+    }
+    unsafe { pset_filter(p_ref.as_mut(), filter) }
+
```
shaleh: This is not a clear check. What is really being checked is really `not == -1` right?

rust_src/src/process.rs
```diff
@@ -273,14 +282,65 @@ pub fn set_process_buffer(process: LispObject, buffer: LispObject) -> LispObject
...
+    if p_ref.infd() >= 0 {
+        unsafe { set_process_filter_masks(p_ref.as_ptr()) };
+    }
+
```
shaleh: And now you have used them twice. Why not make them functions? They do not need to be public.

rust_src/src/process.rs
shaleh: And yes, I realize this is inherited. But we can do better in the Rust code.

PR #777, opened 2018-06-03
clear marker.c

brotzeit:
I don't know how to port the debug stuff. Is there already anything similiar in the rust code base. How can I build emacs with `MARKER_DEBUG` ?


db48x:
add `CFLAGS=-DMARKER_DEBUG` when you configure

db48x:
Looks pretty good. For the `byte_char_debug_check`, you can change `src/marker.c` so that the function is always defined, and then use a cfg attribute to add or remove the calls based on a configuration option: https://doc.rust-lang.org/reference/attributes.html#conditional-compilation

brotzeit:
I've already tried it, but I can't figure out how I can access values in `CFLAGS`. It seems to work differently to the other cfg options we're already using like `NS_IMPL_GNUSTEP`.

db48x:
You would have to add something to the Makefile that adds it to the `RUSTFLAGS` whenever it's in the `CFLAGS`. I wouldn't bother; just let the user supply the `RUSTFLAGS`.

shaleh:
And document it maybe?

shaleh:
I see you hit a rustfmt snag. Why is that? rustup will load the right rustfmt for the rust compiler in use in a directory. Just run `rustup component add rustfmt-preview` from within the remacs directory.

brotzeit:
I just didn't run rustfmt again. I'm waiting for #781.

Wilfred:
#781 has landed, does this PR need anything else?

shaleh:
rebasing it looks like

rust_src/src/marker.rs
```diff
@@ -462,4 +491,316 @@ impl LispBufferRef {
...
+        // cache the correspondence by creating a marker here.
+        // It will last until the next GC.
+        // But don't do it if BUF_MARKERS is nil;
+        // that is a signal from Fset_buffer_multibyte.
+        if record && buffer_ref.markers().is_some() {
+            unsafe {
+                build_marker(b, best_below, best_below_byte);
+            }
+        }
+        // byte_char_debug_check(b, best_below, best_below_byte);
```
Wilfred: Rather than commenting out these calls, you could change `byte_char_debug_check` to return a bool and then do `debug_assert!(byte_char_debug_check(b, best_below, best_below_byte`. What do you think?

rust_src/src/marker.rs
```diff
@@ -462,4 +491,316 @@ impl LispBufferRef {
...
+#[no_mangle]
+pub extern "C" fn clear_charpos_cache(b: *mut Lisp_Buffer) -> () {
+    unsafe {
+        if cached_buffer == b {
+            cached_buffer = ptr::null_mut();
+        }
+    }
+}
+
+// /// Count the markers in buffer BUF.
```
Wilfred: Why are these commented out?

brotzeit:
Ok 2 questions.

I can't build remacs after removing marker.c
```
make[1]: Leaving directory '/home/brotzeit/remacs/lib-src'
make -C src VCSWITNESS='$(srcdir)/../.git/logs/HEAD' all
make[1]: Entering directory '/home/brotzeit/remacs/src'
make[1]: *** No rule to make target 'marker.c', needed by 'gl-stamp'.  Stop.
```

What do I have to change in the Makefile ? I can't find anything regarding marker.c

Next question. What's the best way to define c globals like `cached_charpos` in rust ? It seems the recommended way is using `lazy_static!`.

db48x:
See the gl-stamp rule in src/Makefile. It depends on `GLOBAL_SOURCES`, which is defined from `base_obj` and a few other things. Remove `marker.o` from `base_obj`.

A `lazy_static!` is only initialized once, and then has a fixed value thereafter. What you want for `cached_charpos` is a mutable global variable, because various functions in this file set it whenever they find a charpos. This is not a very idiomatic Rust design, but at first glance I doubt that there's much you can do about it. You might combine the four variables, `cached_charpos`, `cached_bytepos` `chached_buffer`, and `cached_modiff` into a single struct since they're always set at the same time, but that doesn't really help that they're mutable globals.

brotzeit:
I've seen some posts about mutable lazy_statics. For example [this one](https://users.rust-lang.org/t/how-can-i-use-mutable-lazy-static/3751).

db48x:
I've been thinking about it, and I suspect that there's a better way to go. `cached_charpos` and `cached_bytepos` are only used by `buf_charpos_to_bytepos` and `buf_bytepos_to_charpos`. They only use the cached values when they're working on the same buffer as `cached_buffer`:

```Rust
  if (b == cached_buffer && BUF_MODIFF (b) == cached_modiff)
    CONSIDER (cached_charpos, cached_bytepos);
```

Why not put the cached values on the buffer object instead of in global variables? Then they would just be `b.cached_modiff`, `b.cached_charpos`, etc.

brotzeit:
Ok. Out of curiosity, why is there also a global variable `cached_modiff` ? This value should always be the same value as modiff in `cached_buffer` or am I missing something ?

db48x:
```C
    EMACS_INT modiff;		/* This counts buffer-modification events
				   for this buffer.  It is incremented for
				   each such event, and never otherwise
				   changed.  */
```

So the `cached_modiff` is just there to invalidate the cached bytepos/charpos values when the buffer is modified. Basically, if `cached_bytepos` was saved when the buffer's `modiff` was 42, then it's valid as long as the buffer's `modiff` stays at 42.

brotzeit:
I'll better wait until tomorrow for this ;D

brotzeit:
I took some time to think about it :D
We still need `cached_buffer` right ? If I understand this correctly cached buffers can be overwritten without being cleared by `clear_charpos_cache`.

```

        unsafe {
            cached_buffer = b;
            buffer_ref.cached_modiff = buffer_ref.modifications();

            buffer_ref.cached_charpos = best_below;
            buffer_ref.cached_bytepos = best_below_byte;
        }
```

I noticed that the header of `tabulated-list-mode` disappeared but I couldn't figure out why.

brotzeit:
So you would add a field like `bool is_cached` to the struct ?

db48x:
That would work, or you could make the cached values be Option<Whatever>. Or, since you'll always have both a bytepos and a charpos or you'll have neither, it could be an Option<(...)>, or the bytepos/charpos pair could be a struct of their own, etc. Whatever is best, without making this patch impossible to write.

brotzeit:
Hmmm I don't know how to put an Option type in a c struct ? I would simply put the values on the buffer struct for now. We can change it when we port the buffer struct to rust ?

db48x:
Yes, that's definitely a refinement that can wait for the future.

What I had in mind was reserving the right amount of space in the C struct so that the Option would fit, and then making sure that only Rust code ever read or wrote those fields.

shaleh:
Or just use a pointer. Which is all an option really is.

Wilfred:
Looks like Travis didn't find the marker tests:

```
Ran 2605 tests, 2560 results as expected, 45 skipped
1 files did not contain any tests:
  src/marker-tests.log
```

Friendly ping @brotzeit :). Are you still interested in this, and is there any way we can help? Is it just the `tabulated-list-mode` header disappearing the only issue outstanding?

brotzeit:
Ah sorry I was frustrated because I couldn't fix it ;)
I will fix the comment and see if there are other problems than the `tabulated-list-mode` issue.

brotzeit:
```
error: unnecessary `unsafe` block
   --> src/editfns.rs:345:20
    |
345 |         pos_byte = unsafe { buf_charpos_to_bytepos(buffer_ref.as_mut(), p) };
    |                    ^^^^^^ unnecessary `unsafe` block
```

This unsafe block doesn't exist anymore...

brotzeit:
It seems to work. The `tabulated-list-mode` header is also displayed.

rust_src/src/buffers.rs
```diff
@@ -168,6 +174,49 @@ impl LispBufferRef {
         unsafe { (*self.text).z }
     }
 
+    // Same as the BUF_BYTE_ADDRESS c macro
+    /// Return the address of character at byte position BYTE_POS in buffer BUF.
+    #[inline]
+    pub fn buf_byte_address(self, byte_pos: isize) -> c_uchar {
```
shaleh: Hmmm, it looks like there is a repeated bit of code here:

```
fn pos_with_range(self, pos) -> isize {
    if pos >= self.gpt_byte() {
        self.gap_size()
    } else {
        0
    }
}
```

This can be used here and in the `dec_pos` method below. Likely in other spots too.

rust_src/src/marker.rs
```diff
@@ -505,4 +553,308 @@ impl LispBufferRef {
...
+
+    // We find in best_above and best_above_byte
+    // the closest known point above CHARPOS,
+    // and in best_below and best_below_byte
+    // the closest known point below CHARPOS,
+    //
+    // If at any point we can tell that the space between those
+    // two best approximations is all single-byte,
+    // we interpolate the result immediately.
+
```
shaleh: There is likely a better name for this macro.

src/lisp.h
```diff
@@ -559,6 +559,8 @@ extern bool initialized;
 /* Defined in Rust.  */
 extern double extract_float (Lisp_Object);
```
shaleh: ??

shaleh:
I left some comments. I do not have time at the moment to pull and test locally.

brotzeit:
@harryfei  Thanks for the hint. I will take a look at it and fix the permission.

PR #781, opened 2018-06-06
Bindgen redux

db48x:
This patch touches a lot of stuff, because the bindings are very
exact. Pointers are mutable by default, unions that we were just
ignoring are now there (see Lisp_Float and similar things), some
constants became enums, some enums became constants, some things have
worse names (pvec_type, for instance), etc. Many things are much
better, such as bitfields and the vectors and vectorlike objects.

The only major downside to this patch is the compile times. It adds 8
minutes to the build time (on my machine, anyway). Four minutes while
building libremacs to generate the bindings, then another four minutes
to generate the bindings again when building the Rust tests. It's a
separate binary, you see, so it has a separate output directory and
everything. Fun times. I'll look for a way to improve that.


shaleh:
I know how diligently you have worked on this. Thank you.

.dir-locals.el
```diff
@@ -1,14 +1,20 @@
 ((nil . ((tab-width . 8)
          (sentence-end-double-space . t)
-         (fill-column . 70)))
+         (fill-column . 70)
+         (projectile-project-configure-cmd . "CFLAGS='-O0 -g3' ./configure --enable-checking='yes,glyphs' --enable-check-lisp-object-type --enable-rust-debug")
+         (projectile-project-compilation-cmd . "make -j16")
+         (projectile-project-test-cmd . "make check")
+         (projectile-project-run-cmd . "./src/remacs")))
```
Wilfred: Could you remove your changes to this file? They seem specific to you (my little laptop wouldn't benefit from `-j16`, for example :) ).

.travis.yml
```diff
@@ -67,12 +70,14 @@ jobs:
...
       before_script:
         - travis_wait rustup install $(cat rust-toolchain)
+        - travis_wait rustup component add rustfmt-preview
     - stage: cerrors
       os: linux  # No need to run this everywhere
       before_script:
         - sudo apt update
         - sudo apt install -y texinfo libgif-dev libxpm-dev
         - travis_wait rustup install $(cat rust-toolchain)
+        - travis_wait rustup component add rustfmt-preview
```
Wilfred: Why have you added rustfmt to these stages of the build?

rust_src/build.rs
```diff
@@ -594,7 +593,552 @@ fn generate_globals() {
...
+                .blacklist_type("Lisp_Object")
+                .blacklist_type("emacs_globals")
+                .blacklist_type("Q.*") // symbols like Qnil and so on
+                .blacklist_type("USE_LSB_TAG")
+                .blacklist_type("VALMASK")
+                .blacklist_type("PSEUDOVECTOR_FLAG")
+                // these two are found by bindgen on mac, but not linux
+                .blacklist_type("EMACS_INT_MAX")
+                .blacklist_type("VAL_MAX")
+                //.blacklist_type("staticpro")
```
Wilfred: Why is this one commented out?

rust_src/build.rs
```diff
@@ -594,7 +593,552 @@ fn generate_globals() {
...
+                // https://github.com/rust-lang-nursery/rust-bindgen/issues/1318
+                .blacklist_type("max_align_t")
+                // these enums are better as simple constants
+                .constified_enum("EMACS_INT_WIDTH")
+                .constified_enum("BOOL_VECTOR_BITS_PER_CHAR")
+                .constified_enum("BITS_PER_BITS_WORD")
+                // these enums all meet the requirements to also be Rust enums (in theory). 90% of
+                // these are present in struct fields, return values, etc; we don't have much
+                // choice but to inspect each one or just hope that they're correct; guess which
+                // I've done?
```
Wilfred: Could you be a little more definitive here? Something like "these seem to be sensible, but I haven't closely checked them" perhaps.

rust_src/build.rs
```diff
@@ -594,7 +593,552 @@ fn generate_globals() {
...
+                .rustified_enum("vertical_scroll_bar_type")
+                .rustified_enum("window_part")
+                .rustified_enum("x_display_info__bindgen_ty_1")
+                .rustified_enum("z_group")
+                // the others are either unchecked, or definitely would cause undefined behavior
+                .default_enum_style(bindgen::EnumVariation::ModuleConsts);
+
+            //builder
+            //    .dump_preprocessed_input()
+            //    .expect("Unable to dump preprocessed bindings");
```
Wilfred: Debugging leftovers?

rust_src/src/buffers.rs
```diff
@@ -610,7 +619,7 @@ pub fn set_buffer(buffer_or_name: LispObject) -> LispObject {
     if !buf.is_live() {
         error!("Selecting deleted buffer");
     };
-    unsafe { set_buffer_internal(buf.as_mut()) };
+    unsafe { set_buffer_internal_1(buf.as_mut()) };
```
Wilfred: I presume this is doing the same thing?

rust_src/src/chartable.rs
```diff
@@ -17,22 +15,32 @@ pub type LispCharTableRef = ExternalPtr<Lisp_Char_Table>;
...
 pub struct LispSubCharTableAsciiRef(ExternalPtr<Lisp_Sub_Char_Table>);
 
-fn chartab_idx(c: isize, depth: isize, min_char: isize) -> isize {
+fn chartab_size(depth: i32) -> usize {
+    match depth {
+        0 => 1 << CHARTAB_SIZE_BITS::CHARTAB_SIZE_BITS_0 as isize,
+        1 => 1 << CHARTAB_SIZE_BITS::CHARTAB_SIZE_BITS_1 as isize,
+        2 => 1 << CHARTAB_SIZE_BITS::CHARTAB_SIZE_BITS_2 as isize,
+        3 => 1 << CHARTAB_SIZE_BITS::CHARTAB_SIZE_BITS_3 as isize,
+        _ => panic!("oops"),
```
Wilfred: Perhaps "invalid depth for chartab"?

rust_src/src/cmds.rs
```diff
@@ -377,14 +377,14 @@ fn internal_self_insert(mut c: Codepoint, n: usize) -> EmacsInt {
...
-    let previous_char = if current_buffer.enable_multibyte_characters.is_not_nil() {
+    let previous_char = if current_buffer.enable_multibyte_characters_.is_not_nil() {
         preceding_char() as Codepoint
     } else {
         unibyte_to_char(preceding_char() as Codepoint)
     };
-    if current_buffer.abbrev_mode.is_not_nil() && synt != syntaxcode::Word
-        && current_buffer.read_only.is_not_nil() && current_buffer.pt > current_buffer.begv
-        && unsafe { syntax_property(previous_char as libc::c_int, true) } == syntaxcode::Word
+    if current_buffer.abbrev_mode_.is_not_nil() && synt != syntaxcode::Sword
```
Wilfred: Word and Syntax are different -- is this a change in behaviour or renaming the constant?

rust_src/src/functions.rs
```diff
@@ -20,27 +19,31 @@ use lisp::LispObject;
...
-    pub fn make_string(s: *const c_char, length: ptrdiff_t) -> LispObject;
-    pub fn make_unibyte_string(s: *const c_char, length: ptrdiff_t) -> LispObject;
-}
+//#[mock]
+//extern "C" {
+//    pub fn Fcons(car: LispObject, cdr: LispObject) -> LispObject;
+//    pub fn Fsignal(error_symbol: LispObject, data: LispObject) -> !;
+//    pub fn make_string(s: *const c_char, length: isize) -> LispObject;
+//    pub fn make_unibyte_string(s: *const c_char, length: isize) -> LispObject;
+//}
```
Wilfred: We should remove this code and the mock derive dependency if we're no longer using it.

rust_src/src/functions.rs
```diff
@@ -84,3 +88,35 @@ macro_rules! assert_t {
 macro_rules! assert_nil {
     ($arg: expr) => {{ assert!($arg == ::lisp::LispObject::constant_nil()); }};
 }
+
+#[cfg(test)]
+#[allow(unused_variables)]
+#[allow(dead_code)]
+#[no_mangle]
+pub extern "C" fn Fcons(car: LispObject, cdr: LispObject) -> LispObject {
+    return Qnil;
```
Wilfred: I love this mock cons implementation. It gives you something of the right type!

rust_src/src/vectors.rs
```diff
@@ -346,20 +337,21 @@ impl LispBoolVecRef {
...
+        unsafe { self.set_unchecked(idx, b) }
     }
 
     pub unsafe fn set_unchecked(&mut self, idx: usize, b: bool) {
-        let limbp = self.as_mut_byte_ptr().offset(idx as isize / 8);
+        let limb = self.as_mut_slice()
+            .get_unchecked_mut(idx / BITS_PER_BITS_WORD as usize) as *mut usize;
         if b {
-            *limbp |= 1 << (idx % 8)
+            *limb |= 1 << (idx % BITS_PER_BITS_WORD as usize)
```
Wilfred: At some point we might want to remove our usage of BITS_PER_BITS_WORD. I don't think Remacs (or even GNU Emacs) will run on a machine where a byte isn't 8 bits.

src/.gdbinit
```diff
@@ -73,6 +73,12 @@ end
 # We save and restore print_output_debug_flag to prevent the w32 port
 # from calling OutputDebugString, which causes GDB to display each
 # character twice (yuk!).
+# NOTE(db48x): when debugging with rr, we can't set arbitrary
+# variables because that could cause the replay to diverge from the
+# recording. However, if we call a function instead, rr will fork a
+# copy of the process and then discard any changes that function
+# made. Thus instead of simply setting pring_output_debug_flag, we
```
Wilfred: Typo: print

rust_src/src/vectors.rs
db48x: BITS_PER_BITS_WORD is actually 64; a `bits_word` is generally a tyepdef for a `long`.

rust_src/src/cmds.rs
db48x: bindgen uses the C names, so the enum variants are Sword, Spunct, Sfoo, Sbar, etc. The `S` is for syntax, presumably.

rust_src/src/buffers.rs
db48x: yes, exactly the same.

.travis.yml
db48x: The bindings get run through rustfmt. Technically I suppose it's optional, but if we ever have to actually look at the bindings to figure out a build error, then we will definitely want them to be formatted nicely.

db48x:
It may still fail on Windows, if clang picks up the MSVC includes instead of it's own. If so, then we need to find a way to run the build without MSVC in the PATH/INCLUDE/etc variables

PR #787, opened 2018-06-14
Port case-table-p

shaleh:
Include tests.


shaleh:
Thanks @db48x this was a quick and easy port due to your work with bindgen.

PR #507, opened 2017-11-28
Merge emacs-26 from 2017-09-23 - 2017-10-01

shaleh:

PR #792, opened 2018-06-25
Merge upstream GNU Emacs master branch

Wilfred:
This brings Remacs up to commit f71569ac82fb21.

There were only two noteworthy conflicts:

1. We added an ifdef for HAVE_WINDOW_SYSTEM in frame.c in commit
b8c1b004b500925. Upstream made a few minor changes to iconify-frame,
so I've reapplied the the ifdef.

2. etc/NEWS had a conflict regarding a comment on file-name-base. This
comment isn't present in GNU Emacs at f71569ac82fb21, so I've likewise
removed it in Remacs.


Wilfred:
@shaleh I'd be interested on your thoughts as to whether b8c1b004b500925 is still necessary. You added it as part of #507.

shaleh:
@Wilfred I would need to look at it more closely. I will let you know.

PR #793, opened 2018-06-30
Port window-dedicated-p and set-window-dedicated-p

### LINKED ISSUE #746: Port set-window-dedicated-p
```c
DEFUN ("set-window-dedicated-p", Fset_window_dedicated_p,
       Sset_window_dedicated_p, 2, 2, 0,
       doc: /* Mark WINDOW as dedicated according to FLAG.
WINDOW must be a live window and defaults to the selected one.  FLAG
non-nil means mark WINDOW as dedicated to its buffer.  FLAG nil means
mark WINDOW as non-dedicated.  Return FLAG.

When a window is dedicated to its buffer, `display-buffer' will refrain
from displaying another buffer in it.  `get-lru-window' and
`get-largest-window' treat dedicated windows specially.
`delete-windows-on', `replace-buffer-in-windows', `quit-window',
`quit-restore-window' and `kill-buffer' can delete a dedicated window
and the containing frame.

As a special case, if FLAG is t, mark WINDOW as "strongly" dedicated to
its buffer.  Functions like `set-window-buffer' may change the buffer
displayed by a window, unless that window is strongly dedicated to its
buffer.  If and when `set-window-buffer' displays another buffer in a
window, it also makes sure that the window is no more dedicated.  */)
  (Lisp_Object window, Lisp_Object flag)
{
  wset_dedicated (decode_live_window (window), flag);
  return flag;
}
```

fangyuanziti:
Port `window-dedicated-p` and `set-window-dedicated-p`.

Closed #746

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -235,11 +235,6 @@ impl LispWindowRef {
                 || (self.contents().as_buffer_or_error().header_line_format_).is_not_nil())
             && self.pixel_height() > height
     }
-
-    #[inline]
-    pub fn dedicated(self) -> LispObject {
-        self.dedicated
-    }
 }
 
 pub type LispGlyphMatrixRef = ExternalPtr<glyph_matrix>;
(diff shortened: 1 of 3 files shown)


src/window.h
```diff
@@ -495,6 +495,12 @@ wset_next_buffers (struct window *w, Lisp_Object val)
   w->next_buffers = val;
 }
 
+INLINE void
+wset_dedicated (struct window *w, Lisp_Object val)
+{
```
db48x: You can actually just get rid of this accessor function, now that everything has a proper binding.

rust_src/src/windows.rs
```diff
@@ -235,6 +235,11 @@ impl LispWindowRef {
                 || (self.contents().as_buffer_or_error().header_line_format_).is_not_nil())
             && self.pixel_height() > height
     }
+
+    #[inline]
+    pub fn dedicated(self) -> LispObject {
+        self.dedicated
```
db48x: Just access the field directly.

src/window.h
fangyuanziti: Fixed already

rust_src/src/windows.rs
fangyuanziti: Fixed

PR #799, opened 2018-07-02
Rust dired file-attributes-lessp

gaak99:
I've got a new&improved file-attributes (and directory-files) that works good on not(windows) but am seeing some Windows problems. And my ability to debug non-trivial stuff on Windows is limited (as I'm not able to build Remacs on my Windows sys yet).

But in the meantime here is file-attributes-lessp.  Seems not to cause any more ```make check``` fails on Linux and Windows (Appveyor). Was tested earlier on MacOS and I assume it's ok but lets see Travis results.

### DIFF
--- a/rust_src/src/dired.rs
+++ b/rust_src/src/dired.rs
@@ -14,7 +14,7 @@ use strings::string_lessp;
 
 /// Return t if first arg file attributes list is less than second.
 /// Comparison is in lexicographic order and case is significant.
-#[lisp_fn(min = "2")]
+#[lisp_fn]
 pub fn file_attributes_lessp(f1: LispObject, f2: LispObject) -> LispObject {
     LispObject::from_bool(string_lessp(car(f1), car(f2)))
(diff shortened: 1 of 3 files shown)


rust_src/src/dired.rs
```diff
@@ -9,7 +9,15 @@ use remacs_sys::build_string;
 use remacs_sys::globals;
 
 use lisp::{defsubr, LispObject};
-use lists::list;
+use lists::{car, list};
+use strings::string_lessp;
+
+/// Return t if first arg file attributes list is less than second.
+/// Comparison is in lexicographic order and case is significant.
+#[lisp_fn(min = "2")]
```
shaleh: You only need the "min" if it is possible to call this function with a different number of arguments than specified to the Rust function. Otherwise the compilation will set this up correctly.

gaak99: Good catch!

PR #800, opened 2018-07-04
Try to find Rust code in find-function-C-source

nickdrozd:
Most Emacs functions are written in Lisp, with a few primitives like
display functions and basic arithmetic being written in a lower-level
language. In GNU Emacs that language is C, while in Remacs both C and
Rust are used.

Now consider the function `find-function-C-source'. In GNU Emacs, this
is what is used to find the source code for primitive functions, since
primitive functions are written in C. So is it a function for finding
the source code for primitive functions in general, or just those
written in C? In GNU Emacs that question is academic, but in Remacs it
needs to be answered. This commit takes the broad view that the
function is really intended to find all primitive functions, not just
those written in C.

As evidence, consider this perfectly good piece of real-life code:

  (if primitive-p
      (cdr (find-function-C-source sym path nil))
    (cdr (find-function-search-for-symbol sym nil path)))

This code looks for `sym' in the C directory if it's primitive and
elsewhere otherwise. But in Remacs, primitives can be defined in
Rust. In that case, this code will look for the function in C, fail to
find it, and raise an error. The author probably did not mean to
exclude Rust primitives from the search, so the function ought to be
rewritten to capture its intended use and avoid doing anything
unexpected.

It would be great if `find-function-C-source' could just be renamed to
something more accurate like `find-function-primitive-source', but for
compatibility with existing code it will be better to shim in a simple
dispatch on the file name, and decide there whether to look for the
function in Rust or C.

### DIFF
--- a/lisp/emacs-lisp/find-func.el
+++ b/lisp/emacs-lisp/find-func.el
@@ -242,6 +242,15 @@ return the symbol's function definition."
 
 (defun find-function-C-source (fun-or-var file type)
   "Find the source location where FUN-OR-VAR is defined in FILE.
+TYPE should be nil to find a function, or `defvar' to find a variable.
+
+Despite its name, this function will also look in the Rust directory."
+  (if (string-suffix-p ".rs" file)
+      (find-function-rust-source fun-or-var file)
+    (find-function--C-source fun-or-var file type)))
+
+(defun find-function--C-source (fun-or-var file type)
+  "Find the source location where FUN-OR-VAR is defined in FILE.
 TYPE should be nil to find a function, or `defvar' to find a variable."
   (let ((dir (or find-function-C-source-directory
                  (read-directory-name "Emacs C source dir: " nil nil t))))


nickdrozd:
Maybe my description was too oblique. This change makes @Wilfred 's `helpful` package work for Rust functions.

db48x:
This change doesn't seem to be necessary, `find-function-search-for-symbol` already calls both `find-function-C-source` and `find-function-rust-source` as needed.

nickdrozd:
Well, here's the situation as far as I can tell:

The expression `(helpful-callable '+)`

1. works fine in GNU Emacs,
2. does not work in Remacs, and
3. works fine in Remacs with this change.

The line where this issue arises is
https://github.com/Wilfred/helpful/blob/master/helpful.el#L1165

Now, maybe you want to say "Oh, well he used the wrong function there,
he should be using `find-function-search-for-symbol` instead". Maybe
that's true, but ultimately code that works there needs to work here,
and this seems to be a simple (if slightly inelegant) way to make that
happen in this case.

But then again, maybe I'm missing something obvious. @Wilfred, do you
have any thoughts?

Wilfred:
Sorry for the delayed response.

I added the explicit call to `find-function-C-source` in Wilfred/helpful@518d24b76bc6e4d55c8b2c288a33faf67100b101. The problem with `find-function-search-for-symbol` is that it doesn't find the position of primitives:

```
ELISP> (find-function-search-for-symbol 'if nil "~/src/emacs/src/eval.c")
(#<buffer eval.c>)

ELISP> (find-function-C-source 'if "~/src/emacs/src/eval.c" nil)
(#<buffer eval.c> . 11001)
```

I'm generally very hesitant to modify elisp in Remacs, but perhaps that's the best approach here. `find-function-C-source-directory` is the critical variable for an Emacs instance to find the source code, and that has C in its name. The only other symbol containing `-C-` in `find-func.el` is `find-function-C-source`.

I would be inclined to define `find-function-primitive-source` and `find-function-primitive-source-directory`, ensure they handle Rust, and make `find-function-C-source` and `find-function-C-source-directory` aliases. That way the symbol names reflect their purpose, but we remain compatible with code that works today.

db48x:
That's pretty odd. This does work:

```
(find-function-search-for-symbol 'if nil "src/eval.c")
(#<buffer eval.c> . 11001)
```

In Remacs this works in exactly the same way:
```
(find-function-search-for-symbol 'if nil "rust_src/src/eval.rs")
(#<buffer eval.rs> . 2658)
```

If you pass in a longer path to `find-function-search-for-symbol`, then it never even calls `find-function-C-source`.

I'm not sure it's obvious at all what the "right" thing to do is. I still think that `find-function-search-for-symbol` is the right function for external code to call, but it seems poorly documented, and the code is not at all as obvious as it first appeared to me. To be honest, I find it a little weird that the caller has to specify the filename at all. I can understand the type, since a symbol could be either a function or a variable or both.

Wilfred:
Agreed, the APIs aren't obvious at all. Apparently `find-function-search-for-symbol` expects a relative path.

I do think we may need to return Rust paths from `find-function-C-source` at some point. However, since this is my elisp project, I've fixed it there instead. I'm always reluctant to make elisp changes in Remacs.

Since this should now be fixed in Helpful, I'm going to close this. Thanks for raising this @nickdrozd, I think this has been a valuable discussion that we might well revisit in future :)

Issue #789, opened 2018-06-21
Windows build?


gaak99:
Now for my Windows build problem ...

My up to date master is failing on Appveyor
```
thread 'main' panicked at 'Unable to generate bindings: ()', libcore\result.rs:945:5
make[1]: *** [Makefile:610: ../rust_src/target/release/remacs_lib.lib] Error 101
```
known issue? (or is my config hosed)

db48x:
It very likely couldn't find clang. Install it from http://clang.llvm.org/.

db48x:
wrong button :)

gaak99:
There's a ton of Visual Studio errors too so I suspect the Appveyor default, Visual Studio 2015, is not correct now. I'm trying VS 2017 now...

db48x:
MSVC is not required at all.

db48x:
That's clang emitting error messages that contain the names of header files from MSVC. Clang can't deal with those headers, because they contain MSVC-specific extensions. The solution is to keep the MSVC headers out of your include path.

gaak99:
good info, thanks @db48x. 

I was able to reset
```
HEAD is now at 72f9ca0 Don't run a full build at the C errors stage (#774)
```
to a place where make check is clean (cept for benchmark test false positive). So a least I can test my new junk.

brotzeit:
#970

PR #805, opened 2018-07-10
Rust port of dired's file-attributes

gaak99:
Here's my second swing at ```file-attributes```.

- not(windows)
  - all Rust except
    - file modes - nice small-medium size separate project
    - symbolic links - ```std::os::fs::symlink_metadata``` does not provide several of the fields needed
  - ```make check``` clean

- windows:
  - except for entry point all C because ```std::os::windows::fs::MetadataExt``` does not provide several of the fields needed
  - to get a useful make check on Appveyor I had to reset master back several commits -- see https://github.com/Wilfred/remacs/issues/789 -- given that it seems to be ok

### DIFF
--- a/rust_src/src/dired.rs
+++ b/rust_src/src/dired.rs
@@ -1,15 +1,455 @@
 //! Lisp functions for making directory listings.
 
 #[cfg(not(windows))]
-use libc;
+use libc::{c_char, c_long, endpwent, getgrgid, getpwent, getpwuid, group, passwd,
+           timespec as c_timespec};
+
+#[cfg(not(windows))]
+use std::ffi::OsStr;
+#[cfg(not(windows))]
+use std::io;
+#[cfg(not(windows))]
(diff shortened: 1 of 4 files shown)


gaak99:
@db48x true that. (and thanks for the review)

Anybody else feel strongly either way?

I was curious if/how much cleaner Rust code would be compared to C/cpp for multi-platform code and starting to get some feel for it now at least.

shaleh:
I am ok with merging this. But yeah, we should maybe make a mini crate for unix like or windows dired and load the right one

db48x:
I wouldn't have to be a crate, just a pair of modules that dired.rs can then import from.

shaleh:
Fair enough

gaak99:
Sounds like a win. Lemme give it a shot...

gaak99:
New br/pr for unix/windows split so closing this one.

PR #807, opened 2018-07-13
port char-before

brotzeit:


rust_src/src/editfns.rs
```diff
@@ -317,6 +317,38 @@ pub fn preceding_char() -> EmacsInt {
...
+    let pos_byte: isize;
+
+    if pos.is_nil() {
+        pos_byte = buffer_ref.pt_byte;
+    } else if let Some(m) = pos.as_marker() {
+        pos_byte = m.bytepos_or_error();
+        if pos_byte <= buffer_ref.begv_byte || pos_byte > buffer_ref.zv_byte {
+            return None;
+        }
+    } else {
```
shaleh: Why use coerce marker here? You just proved it is not a marker in the previous check? Either add a comment or just use `as_fixnum`.

brotzeit:
I'm surprised this is the first time you're asking me for tests ;)
I will add some.

rust_src/src/editfns.rs
brotzeit: I just realized, that I used coerce for the error message.

shaleh:
I am asking everyone who does not have tests. The project is in a good place for writing them.

PR #812, opened 2018-07-15
there's no longer any need for these accessors, now that we have bindgen

db48x:


shaleh:
We may need to find ways to take less time.

For right now, I think we can disable the "no window system" compilation so we only build twice.

db48x:
I don't know. It didn't exceed the overall time limit, but instead stalled and didn't produce any output for 10 minutes. Dropping one of the build steps wouldn't fix that.

db48x:
Looks like `python-tests` takes a long time on my computer as well. I don't know how long yet, however.

db48x:
In fact, it looks like it's deadlocked. The remacs process has called `read` and is waiting, while both python processes have called `select` and are waiting.

db48x:
Hrm, that time it didn't deadlock. There are half a dozen errors though, all of them "Error reading from stdin".

brotzeit:
Any chance this is related to #801 ?

db48x:
It shouldn't be, unless I did something wrong. Why do you ask?

brotzeit:
Because that PR fixed process related issues.

PR #813, opened 2018-07-18
port set-window-point

brotzeit:


rust_src/src/windows.rs
```diff
@@ -760,4 +761,39 @@ pub fn window_use_time(window: LispObject) -> LispObject {
...
+/// Make point value in WINDOW be at position POS in WINDOW's buffer.
+/// WINDOW must be a live window and defaults to the selected one.
+/// Return POS.
+#[lisp_fn]
+pub fn set_window_point(window: LispObject, pos: LispObject) -> LispObject {
+    let mut w = window_live_or_selected(window);
+
+    // Type of POS is checked by Fgoto_char or set_marker_restricted ...
+    if w == selected_window().as_window_or_error() {
+        let mut old_buffer = ThreadState::current_buffer();
```
Wilfred: Why is this old_buffer rather than current_buffer?

brotzeit: I often take stuff from the c code without thinking about it :)

Issue #804, opened 2018-07-06
build.rs exports and #[cfg(...)]


gaak99:
Ran into a bug or at least a limitation in the generation of c_exports.rs by build.rs...

Example code
```
#[cfg(windows)]
#[no_mangle]
pub extern "C" fn file_attributes_internal(//snip) -> LispObject {
//snip
}
```
```make``` on linux
```
error[E0432]: unresolved import `dired::file_attributes_internal`
   --> /home/gb/warez/remacs-final/remacs/rust_src/target/release/build/remacs-3030dc6b627ee1fc/out/c_exports.rs:113:9
    |
113 | pub use dired::file_attributes_internal;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `file_attributes_internal` in `dired`. Did you mean to use `file_attributes_c_internal`?
```
Looks like the ```#[cfg(windows)]``` is not respected in build.rs (a quick look at the code confirms it).

Ditto for ```#[lisp_fn]```.

**Comment by gaak99** [comment]
Probably most of the time the exported fn will be on both windows/not(windows) but I have some code that is only for one of them.

**Comment by shaleh** [comment]
Good catch. Yep. I totally missed that. For now, can you add a dummy function?

**Comment by gaak99** [comment]
Giving it a try now...

BTW my first thought of a good quick fix would be if there is a rustc opt to expand a subset (cfg here) of the macros but the only related opt I could find expands all the macros
```
rustc -- -Z unstable-options --pretty=expanded
```
I then hacked up build.rs to filter my symbols which works fine but def just a temp workaround.

gaak99:
Yea I'm able now to work around it (so no build.rs hacks needed to filter my symbols).

PR #814, opened 2018-07-18
Rust port of dired's file-attributes (unix/windows split)

gaak99:
As discussed in https://github.com/Wilfred/remacs/pull/805.

Code is def more readable without excessive cfg macros per sys type.


src/dired.c
```diff
@@ -47,8 +47,35 @@ extern int is_slow_fs (const char *);
...
+						  Lisp_Object);
+#ifdef WINDOWSNT
+Lisp_Object file_attributes_c(Lisp_Object, Lisp_Object);
+
+/*
+ * Workaround for
+ * https://github.com/Wilfred/remacs/issues/804
+ */
+static int
+dummy_file_attributes_rust_internal(void) {
```
db48x: I don't see why you need this function.

gaak99: Given the windows build and https://github.com/Wilfred/remacs/issues/804 ...

```file_attributes_rust_internal``` is exported  (it really does not need to be for windows but #804).

I believe -- forget now if I ran into this or guessed I would -- since ```file_attributes_rust_internal``` was exported from rust dired it needed to be called at least once from another module else build would complain/fail.

And the dummy_ func wrapper is just a way to make it more clear it's not doing anything useful.

Comment (db48x) [reply src/dired.c]: What error do you get if you remove `dummy_file_attributes_rust_internal`?

Comment (shaleh) [reply src/dired.c]: Was this related to the build.rs not parsing #cfg correctly? So it may not be needed.

File: rust_src/src/dired.rs [hunk 2]
```diff
@@ -1,16 +1,74 @@
 //! Lisp functions for making directory listings.
 
-#[cfg(not(windows))]
-use libc;
+#[cfg(unix)]
+include!("dired_unix.rs");
```
db48x: My other question is about these includes. Was there no way to use modules? I suspect it would be nicer.

gaak99: Good pt -- just did not think of it -- switching to modules now.

src/dired.c
gaak99: @db48x here's an example of the type of error
```
#[no_mangle]
pub extern "C" fn file_attributes_foo() -> bool {
    true
}
```
which gives
```
Compiling remacs v0.1.0 (file:///home/gb/warez/remacs-final/remacs/rust_src)

 warning: function is never used: `file_attributes_foo`
   --> src/dired.rs:78:1
    |
 78 | pub extern "C" fn file_attributes_foo() -> bool {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: #[warn(dead_code)] on by default
```
This was a quick example on Linux. Given the state of the windows build it's hard to repro the real thing.

gaak99: @shaleh yes it's ```cfg``` not able to restrict an exported fn (#804) to just one of unix/windows. What not needed? (the dummy_ fn for windows?, support for windows anymore in current dev projects?)

gaak99: After the switch to importing unix/windows as modules I saw an easy way to remove the #804 workaround -- including the dummy_ C func -- next push will have it removed.

PR #821, opened 2018-07-23
Add passing and failing subr-lang tests

nickdrozd:
`(describe-function 'rename-buffer)` errors, apparently due to the
inability of `subr-lang` to handle `rename-buffer`.


nickdrozd:
Should I open an issue separate from this, or do the failing tests suffice as a record of the issue?

shaleh:
An open issue would be good.

PR #826, opened 2018-07-27
Port process-coding-system, process-thread, current-thread, and thread--blocker.

### LINKED ISSUE #750: Port process-coding-system
```c
DEFUN ("process-coding-system",
       Fprocess_coding_system, Sprocess_coding_system, 1, 1, 0,
       doc: /* Return a cons of coding systems for decoding and encoding of PROCESS.  */)
{
  CHECK_PROCESS (process);
  return Fcons (XPROCESS (process)->decode_coding_system,
		XPROCESS (process)->encode_coding_system);
}
```

leungbk:
Closes #750 and #797.

I'm not sure if I ported `current-thread` correctly. Should the return type be `ThreadStateRef` instead of `LispObject`? I wrote `LispObject` only because that is what `LispObject::tag_ptr()` returns.

### DIFF
--- a/rust_src/src/threads.rs
+++ b/rust_src/src/threads.rs
@@ -5,7 +5,7 @@ use std::mem;
 use libc;
 
 use remacs_macros::lisp_fn;
-use remacs_sys::{current_thread, thread_state, SPECPDL_INDEX};
+use remacs_sys::{current_thread as current_thread_pointer, thread_state, Lisp_Type, SPECPDL_INDEX};
 
 use buffers::LispBufferRef;
 use lisp::{ExternalPtr, LispObject};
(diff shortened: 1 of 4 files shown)


shaleh:
If you can type `return thing.into()` while the return type is LispObject then you can replace LispObject with the type of thing and just return thing.

leungbk:
Thanks for your response. I've edited the submission.

rust_src/src/threads.rs
```diff
@@ -53,4 +62,21 @@ pub fn thread_alive_p(thread: ThreadStateRef) -> bool {
     thread.is_alive()
 }
 
+/// Return the current thread.
+#[lisp_fn]
+pub fn current_thread() -> LispObject {
+    ThreadState::current_thread().as_lisp_obj()
+}
```
db48x: You can make the `lisp_fn` macro do more of the work by adding a `From` impl:

```
impl From<ThreadStateRef> for LispObject {
    pub fn from(th: ThreadStateRef) -> LispObject {
        th.as_lisp_obj()
    }
}
```

However, this can be done later; I won't hold this patch up because of it.

PR #827, opened 2018-07-30
Port point-marker, point-min-marker, and point-max-marker.

### LINKED ISSUE #796: Port point-min-marker and point-max-marker
```c
DEFUN ("point-min-marker", Fpoint_min_marker, Spoint_min_marker, 0, 0, 0,
       doc: /* Return a marker to the minimum permissible value of point in this buffer.
This is the beginning, unless narrowing (a buffer restriction) is in effect.  */)
  (void)
{
  return build_marker (current_buffer, BEGV, BEGV_BYTE);
}

DEFUN ("point-max-marker", Fpoint_max_marker, Spoint_max_marker, 0, 0, 0,
       doc: /* Return a marker to the maximum permissible value of point in this buffer.
This is (1+ (buffer-size)), unless narrowing (a buffer restriction)
is in effect, in which case it is less.  */)
  (void)
{
  return build_marker (current_buffer, ZV, ZV_BYTE);
}
```

leungbk:
Closes #796.

### DIFF
--- a/rust_src/src/marker.rs
+++ b/rust_src/src/marker.rs
@@ -168,6 +168,27 @@ pub fn point_marker() -> LispObject {
     }
 }
 
+/// Return a marker to the minimum permissible value of point in this buffer.
+/// This is the beginning, unless narrowing (a buffer restriction) is in effect.
+#[lisp_fn]
+pub fn point_min_marker() -> LispObject {
+    unsafe {
+        let cur_buf = ThreadState::current_buffer().as_mut();
+        build_marker(cur_buf, (*cur_buf).begv, (*cur_buf).begv_byte)
+    }
+}
+
+/// Return a marker to the maximum permissible value of point in this buffer.
+/// This is (1+ (buffer-size)), unless narrowing (a buffer restriction)
+/// is in effect, in which case it is less.
+#[lisp_fn]
+pub fn point_max_marker() -> LispObject {
+    unsafe {
+        let cur_buf = ThreadState::current_buffer().as_mut();
+        build_marker(cur_buf, (*cur_buf).zv, (*cur_buf).zv_byte)
+    }
+}
+
 /// Set PT from MARKER's clipped position.
 #[no_mangle]
 pub extern "C" fn set_point_from_marker(marker: LispObject) {
--- a/rust_src/src/marker.rs
+++ b/rust_src/src/marker.rs
@@ -5,8 +5,9 @@ use std::mem;
 use std::ptr;
 
 use remacs_macros::lisp_fn;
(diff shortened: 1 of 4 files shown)


leungbk:
I rebased earlier and have tried to avoid using the accessors. I have several questions:

1) Where is `.set_marker_type()` defined for the `LispMarkerRef` type? I've searched with ripgrep for this method and can find two invocations but no definition.

2) Why is it that `LispMarkerRef::set_bytepos and charpos()` take `&mut self`, while `LispMarkerRef::set_buffer and next()` take `mut self`?

3) Why is it that the compiler reports the existence of a `need_adjustment` method here when there is none defined?
```
error[E0615]: attempted to take value of method `need_adjustment` on type `&mut lisp::ExternalPtr<remacs_sys::Lisp_Marker>`
  --> src/marker.rs:89:14
   |
89 |         self.need_adjustment = need;
   |              ^^^^^^^^^^^^^^^
   |
   = help: maybe a `()` to call it is missing?

If you want more information on this error, try using "rustc --explain E0615"
```

shaleh:
Solid questions. I can understand your confusion. We recently switched to using a tool called `bindgen` to help use automatically integrate with C code instead of writing most of it by hand. The resulting Rust code ends up in a file called `bindings.rs` in your target directory under `rust_src`. Mine for instance is here at the moment. `./rust_src/target/debug/build/remacs-dfaf22b48507735e/out/bindings.rs`

`need_adjustment` is in this file. It looks like this:

```
    #[inline]
    pub fn need_adjustment(&self) -> bool_bf {
        unsafe { ::std::mem::transmute(self._bitfield_1.get(30usize, 1u8) as u8) }
    }
```

db48x:
And you didn't find it using ripgrep because ripgrep honors the `.gitignore` files, and thus it didn't search the `target` directory.

leungbk:
Thanks for your responses. It seems to build properly now.

rust_src/src/marker.rs
```diff
@@ -127,6 +128,67 @@ pub fn marker_buffer(marker: LispMarkerRef) -> Option<LispBufferRef> {
...
+        obj
+    }
+}
+
+/// Return value of point, as a marker object.
+#[lisp_fn]
+pub fn point_marker() -> LispObject {
+    unsafe {
+        let cur_buf = (*current_thread).m_current_buffer;
+        build_marker(cur_buf, (*cur_buf).pt, (*cur_buf).pt_byte)
```
db48x: I missed this the first time, but you can simplify this by using `ThreadState::current_buffer()`.  Looks pretty good, otherwise.

db48x:
Thanks!

PR #834, opened 2018-08-02
Rust port of directory-files

gaak99:
On Unix it's all Rust here except for some support functionality.

On Windows all C except for entry point.

### DIFF
--- a/rust_src/src/dired.rs
+++ b/rust_src/src/dired.rs
@@ -3,12 +3,53 @@
 use remacs_macros::lisp_fn;
 
 #[cfg(unix)]
-use dired_unix::{file_attributes_intro, get_users};
+use dired_unix::{directory_files_and_attributes_intro, directory_files_intro,
+                 file_attributes_intro, get_users};
 #[cfg(windows)]
 use dired_windows::{file_attributes_intro, get_users};
 use lisp::{defsubr, LispObject};
 use lists::car;
 use strings::string_lessp;
+
+/// Return a list of names of files in DIRECTORY.
+/// There are three optional arguments:
+/// If FULL is non-nil, return absolute file names.  Otherwise return names
+///  that are relative to the specified directory.
+/// If MATCH is non-nil, mention only file names that match the regexp MATCH.
+/// If NOSORT is non-nil, the list is not sorted--its order is unpredictable.
+///  Otherwise, the list returned is sorted with `string-lessp'.
+///  NOSORT is useful if you plan to sort the result yourself.
+#[lisp_fn(min = "1")]
+pub fn directory_files(
+    directory: LispObject,
+    full: LispObject,
+    match_re: LispObject,
+    nosort: LispObject,
+) -> LispObject {
+    directory_files_intro(directory, full, match_re, nosort)
+}
+
+/// Return a list of names of files and their attributes in DIRECTORY.
+/// There are four optional arguments:
+/// If FULL is non-nil, return absolute file names.  Otherwise return names
+///  that are relative to the specified directory.
+/// If MATCH is non-nil, mention only file names that match the regexp MATCH.
+/// If NOSORT is non-nil, the list is not sorted--its order is unpredictable.
+///  NOSORT is useful if you plan to sort the result yourself.
+/// ID-FORMAT specifies the preferred format of attributes uid and gid, see
+/// `file-attributes' for further documentation.
+/// On MS-Windows, performance depends on `w32-get-true-file-attributes',
+/// which see.
+#[lisp_fn(min = "1")]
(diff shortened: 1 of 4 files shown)


rust_src/src/dired_unix.rs
```diff
@@ -95,6 +100,270 @@ impl LispObjectExt for LispObject {
...
+            files_attrs: Vec::new(),
+        }
+    }
+
+    // read_dir() does not return .&.. so bolt them on
+    fn add_dots(&mut self) {
+        let d = String::from(".");
+        let dd = String::from("..");
+        let mut df = d.clone(); // dot final name
+        let mut ddf = dd.clone();
```
db48x: This has quite a lot of cloning. I don't see that any of the clones get written to, so there should be a way to avoid them.

rust_src/src/dired.rs
```diff
@@ -3,13 +3,54 @@
...
+///  NOSORT is useful if you plan to sort the result yourself.
+#[lisp_fn(min = "1")]
+pub fn directory_files(
+    directory: LispObject,
+    full: LispObject,
+    match_re: LispObject,
+    nosort: LispObject,
+) -> LispObject {
+    directory_files_intro(directory, full, match_re, nosort)
+}
```
db48x: Could this be simplified? Perhaps something like this:

```
#[lisp_fn(min="1")]
pub use dired_unix::directory_files_intro as directory_files;
```

rust_src/src/dired_unix.rs
```diff
@@ -95,6 +100,270 @@ impl LispObjectExt for LispObject {
...
+    recomp: *mut re_pattern_buffer,
+}
+
+impl RegEx {
+    fn new(match_re: LispObject) -> Self {
+        Self {
+            // MATCH_RE might be a flawed regular expression.  Rather than
+            // catching and signaling our own errors, we just call
+            // compile_pattern to do the work for us.
+            recomp: unsafe { compile_pattern(match_re, null_mut(), Qnil, false, true) },
```
db48x: This is a pretty good start on a regex library.

rust_src/src/dired_unix.rs
```diff
@@ -95,6 +100,270 @@ impl LispObjectExt for LispObject {
     }
 }
 
+struct DirFiles {
+    directory: String,
+    full: bool,
+    match_re: LispObject,
+    nosort: bool,
+    attrs: bool,
```
db48x: I don't think I'm alone in disliking inverted booleans, or boolean arguments. These could easily be enums.

rust_src/src/dired_unix.rs
```diff
@@ -95,6 +100,270 @@ impl LispObjectExt for LispObject {
...
+        }
+    }
+
+    // Assumes expand_file_name haz already run on
+    // self.directory and it will drop >1 trailing '/'.
+    // So here make sure just one '/' between dir&file.
+    fn get_full(&self, file: String) -> String {
+        let last_char = self.directory.as_str().chars().last().unwrap();
+        if last_char == '/' {
+            self.directory.clone() + &file.clone()
```
db48x: `&file.clone()` is definitely unneeded, as the `Add` trait impl copies from its right-hand argument into its left-hand argument. See https://doc.rust-lang.org/std/string/struct.String.html#impl-Add%3C%26%27a%20str%3E for details.

rust_src/src/dired_unix.rs
```diff
@@ -95,6 +100,270 @@ impl LispObjectExt for LispObject {
...
+        } else {
+            self.directory.clone() + "/" + &file.clone()
```
db48x: Same here.

rust_src/src/dired_unix.rs
gaak99: Agreed -- I got a little clone happy -- I'll cleanup all unnecessary clones.

rust_src/src/dired_unix.rs
gaak99: Sounds good. Lemme give it a shot.

rust_src/src/dired_unix.rs
```diff
@@ -361,7 +371,100 @@ pub fn directory_files_and_attributes_intro(
...
+            FullPath::Yes
+        },
+        if match_re.is_nil() {
+            None
+        } else {
+            Some(match_re)
+        },
+        if nosort.is_nil() {
+            SortFNames::Yes
+        } else {
```
db48x: It's a little wordy, but so much more understandable. Thank you!

db48x:
I'm going to merge this in spite of the failure on mac, because I've seen the epg tests fail myself and it's all to do with calling out to gpg.

PR #838, opened 2018-08-08
Fix file permission of crypto/mod.rs

harryfei:
Sorry, forget this file in #837.
All the rust file's (in `rust_src/src`) permission is correct now.

Besides, we may need a way to check those undesired changes automatically.


harryfei:
Can anyone merge this?

Wilfred:
Sorry for the delay!

PR #857, opened 2018-09-18
port some minibuffer functions

brotzeit:


rust_src/src/minibuf.rs
```diff
@@ -65,4 +69,51 @@ pub fn minibuffer_prompt() -> LispObject {
...
+
+    let end = field_end(Some(beg), false, None);
+    let buffer_end = ThreadState::current_buffer().zv as EmacsInt;
+    if end == buffer_end && get_char_property(beg, Qfield, LispObject::constant_nil()).is_nil() {
+        beg
+    } else {
+        end
+    }
+}
+
```
shaleh: Do it makes sense to collapse these two into an internal function? The only difference is the true/false for adding text properties. I know the C did not do it. Should we?

shaleh:
Looks good. Are there tests hitting it?

rust_src/src/minibuf.rs
brotzeit: Hmmm. I'm not sure if this is worth an additional function.

brotzeit:
I don't know how to write tests for functions that open a prompt.

PR #867, opened 2018-09-24
Docker updates

peterdelong:
Update docker configuration to use the appropriate rust toolchain. Also update the README


docker/fedora/Dockerfile
```diff
@@ -21,6 +21,6 @@ RUN yum install -y \
 ENV PATH "/root/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
shaleh: Is there some way we can read in the `rust-toolchain` so we do not have to constantly update this?

shaleh: Or do we go to a Dockerfile.in setup?

shaleh:
@peterdelong ??

docker/fedora/Dockerfile
peterdelong: I'm not at all familiar with Docker but the way it's set up right now is that we have a base `Dockerfile.in` that is used to build the other two. It reads in the toolchain version when the build script is run

shaleh: got it.

PR #870, opened 2018-09-25
Port copy_overlays to rust

shanavas786:


rust_src/src/buffers.rs
```diff
@@ -820,6 +820,57 @@ pub fn force_mode_line_update(all: LispObject) -> LispObject {
...
+        } else {
+            tail = Some(overlay_new.as_overlay_or_error());
+            result = Some(overlay_new.as_overlay_or_error());
+        }
+    }
+
+    if let Some(mut res) = result {
+        res.as_mut()
+    } else {
+        ptr::null_mut() as *mut Lisp_Overlay
```
brotzeit: I think you don't need to cast here.

shanavas786: thanks. updated

shanavas786:
Conflicts resolved

rust_src/src/buffers.rs
```diff
@@ -871,6 +871,57 @@ pub fn force_mode_line_update(all: LispObject) -> LispObject {
...
+            .set_insertion_type(mk_start.insertion_type());
+
+        let mk_end = overlay.end().as_marker_or_error();
+        let end = build_marker(buffer, mk_end.charpos_or_error(), mk_end.bytepos_or_error());
+        end.as_marker_or_error()
+            .set_insertion_type(mk_end.insertion_type());
+
+        let overlay_new = unsafe { build_overlay(start, end, Fcopy_sequence(overlay.plist())) };
+
+        if result.is_some() {
```
shaleh: Why is this different from the C code? next and tail should be the same righ?

shaleh: Overall this use of Option just feels clunky.

rust_src/src/buffers.rs
```diff
@@ -871,6 +871,57 @@ pub fn force_mode_line_update(all: LispObject) -> LispObject {
...
 
+/// Return a list of overlays which is a copy of the overlay list
+/// LIST, but for buffer BUFFER.
+#[no_mangle]
+pub extern "C" fn copy_overlays(
+    buffer: *mut Lisp_Buffer,
+    list: *mut Lisp_Overlay,
+) -> *mut Lisp_Overlay {
+    if list == ptr::null_mut() {
+        return list;
```
brotzeit: I think we don't need this NULL check since the function returns NULL if there are no overlays ? Btw we use `is_null` in this situation.

rust_src/src/buffers.rs
```diff
@@ -871,6 +871,57 @@ pub fn force_mode_line_update(all: LispObject) -> LispObject {
...
+    }
+
+    let overlays = LispOverlayRef::from_ptr(list as *mut c_void)
+        .unwrap_or_else(|| panic!("Invalid overlay reference."));
```
brotzeit: I'm not sure whether this is important, but the C code checks the markers of each overlay. Maybe we should too ?

brotzeit: We could put the assertions in the overlay iterator.

rust_src/src/buffers.rs
brotzeit: I realized you have to distinguish between NULL pointers and invalid overlay references.

rust_src/src/buffers.rs
shanavas786: > Why is this different from the C code? next and tail should be the same righ?

Type of tail and next are different here.

> Overall this use of Option just feels clunky.

True, `C` null checks `tail` to make decisions. I couldn't find a way other than `Option` to that.

shaleh: I will put some thought to it. This is likely a pattern we need to repeat. We might as well get it right now.

shanavas786:
Refactored. @shaleh @brotzeit can you guys have a look ?

rust_src/src/buffers.rs
```diff
@@ -1131,6 +1131,57 @@ pub unsafe fn per_buffer_idx(offset: isize) -> isize {
...
+        return list;
+    }
+
+    let mut result = ptr::null_mut();
+
+    let overlays_iter = LispOverlayRef::from_ptr(list as *mut c_void)
+        .unwrap_or_else(|| panic!("Invalid overlay reference."))
+        .iter();
+
+    let _ = overlays_iter.fold(None, |mut tail: Option<LispOverlayRef>, overlay| {
```
shaleh: The start and end logic do the same work. One potential simplification is to make a function out of it. Then the body of the fold becomes:

    let start = duplicate_marker(overlay.start.as_marker_or_error());
    let end = duplicate_marker(overlay.end.as_marker_or_error());
    let mut overlay_new = build_overlay(...);
    match tail {
        Some(mut tail_ref) => tail_ref.next = overlay_new.as_mut(),
        None => result = overlay_new.as_mut(),
    }
    tail = Some(overlay_new);
    tail

shaleh: Note the update to tail is pulled out of the `match` because it is the same in both arms. `match` is more idiomatic here too.

shaleh: `duplicate_marker` could be defined as a local lambda instead of making a full function.

shanavas786: Thanks. Done

shaleh: Before, the callback was too long and unwieldy. I was going to suggest using a loop instead. Your update is much easier to read. Incorporating `tail` in the fold argument is perfect.

shaleh:
@brotzeit your turn to review.

PR #871, opened 2018-09-25
port terminal-name

brotzeit:


shanavas786:
permission of some files has been changed from `100644 → 100755`

src/terminal.c
```diff
@@ -460,21 +430,6 @@ DEFUN ("terminal-list", Fterminal_list, Sterminal_list, 0, 0, 0,
...
-       doc: /* Return the name of the terminal device TERMINAL.
-It is not guaranteed that the returned value is unique among opened devices.
-
-TERMINAL may be a terminal object, a frame, or nil (meaning the
-selected frame's terminal). */)
-  (Lisp_Object terminal)
-{
-  struct terminal *t = decode_live_terminal (terminal);
-
-  return t->name ? build_string (t->name) : Qnil;
```
Wilfred: I can't see the null check of the name in the Rust. Have I missed it, or do we need to add it?

brotzeit: This is already checked in `decode_terminal` with `return t && t->name ? t : NULL;`. I used the method `is_live` in rust.

rust_src/src/terminal.rs
```diff
@@ -0,0 +1,83 @@
...
+    let t;
+    if let Some(mut term) = terminal.as_terminal() {
+        t = term.as_mut();
+    } else if let Some(frame) = terminal.as_frame() {
+        t = frame.terminal;
+    } else {
+        t = ptr::null_mut();
+    }
+
+    if let Some(term_ref) = LispTerminalRef::from_ptr(t as *mut c_void) {
```
shanavas786: I guess this causes segfault
We should check for null before transmuting to `LispTerminalRef`

shanavas786: We could assert that pointer is not null in `from_prt` methods too

brotzeit: Oh, I thought this is handled by `as_ref`.

rust_src/src/terminal.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Return the terminal object specified by TERMINAL.  TERMINAL may
+/// be a terminal object, a frame, or nil for the terminal device of
+/// the current frame.  If TERMINAL is neither from the above or the
+/// resulting terminal object is deleted, return NULL.
+#[no_mangle]
+pub extern "C" fn decode_terminal(mut terminal: LispObject) -> *mut Lisp_Terminal {
+    if terminal.is_nil() {
+        terminal = selected_frame();
+    }
+
```
shaleh: Why not assign to `t` here?

rust_src/src/terminal.rs
shaleh: `ptr.as_ref` handles whether the pre is null. That is why the return value is an Option type.

shaleh: https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref

brotzeit:
Why is this suddenly fixed ?

shaleh:
@brotzeit I think I have this one resolved. I updated with the style we have been using lately. `LispLiveTerminal` will be used for many of the soon to be ported terminal functions.

brotzeit:
Ok. I can't approve =)

shaleh:
I can but I shouldn't.

PR #879, opened 2018-09-28
don't skip the doc strings of kboard or per_buffer variables

### LINKED ISSUE #876: Variables defined in Rust aren't documented properly
Here's what I get running `(describe-variable 'last-command)`:

```
last-command’s value is ‘quit-window’
It is a terminal-local variable; global value is the same.

  This variable may be risky if used as a file-local variable.

Documentation:
Not documented as a variable.

[back]
```

`last-command` is defined in `keyboard.rs` along with a few other variables that suffer the same problem.

Here's what it looks like in GNU Emacs:

```
last-command is a variable defined in ‘keyboard.c’.
Its value is ‘eval-expression’
It is a terminal-local variable; global value is the same.

  This variable may be risky if used as a file-local variable.

Documentation:
The last command executed.
Normally a symbol with a function definition, but can be whatever was found
in the keymap, or whatever the variable ‘this-command’ was set to by that
command.

The value ‘mode-exit’ is special; it means that the previous command
read an event that told it to exit, and it did so and unread that event.
In other words, the present command is the event that made the previous
command exit.

The value ‘kill-region’ is special; it means that the previous command
was a kill command.

‘last-command’ has a separate binding for each terminal device.
See Info node ‘(elisp)Multiple Terminals’.
```

`Remacs 26.0.60 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.22.30) of 2018-09-25`

db48x:
The logic here was inverted. We don't want to report per keyboard or
per buffer variables as globals, but we do still want to put their
docstrings into the DOC file.

Fixes #876

### DIFF
--- a/rust_src/remacs-lib/docfile.rs
+++ b/rust_src/remacs-lib/docfile.rs
@@ -161,32 +161,33 @@ pub extern "C" fn scan_rust_file(
             add_global(SYMBOL, name.as_ptr(), 0, value.as_ptr());
         } else if line.starts_with("defvar_") {
             // defvar_lisp!(f_Vpost_self_insert_hook, "post-self-insert-hook", Qnil);
+            // defvar_kboard!(Vlast_command_, "last-command");
             lazy_static! {
                 static ref RE: Regex =
-                    Regex::new(r#"defvar_(.+?)!\((.+?),\s+"(.+?)",\s+(.+?)\);"#).unwrap();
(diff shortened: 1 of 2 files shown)


nickdrozd:
You forgot to update the failing test :)

But https://github.com/Wilfred/remacs/pull/880 also messes with those tests, so it would probably be easier to wait until that goes in. Or we can do it the other way around.

PR #882, opened 2018-09-29
autogen.sh should check rust-toolchain

gaak99:
After the upgrade to ```nightly-2018-09-09``` I noticed autogen.sh output
```
Your system has the required rust installation for building remacs.
```
but I had not installed ```nightly-2018-09-09``` yet and the build failed (as expected).

I took a look at autogen.sh and it should (I assume) check against rust-toolchain (instead of just doing the basic ```nightly``` check it does now).

Anyways, here's a patch that works for me on Ubuntu 16.04.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -30,9 +30,8 @@
 
 ### Code:
 
-## Used when remacs is dependent on nightly
-## Will print an error message telling people they need nightly rust
-require_nightly=true
+## Rust tool chain required version path
+rust_toolchain_vers_path="./rust-toolchain"
 
 ## Tools we need:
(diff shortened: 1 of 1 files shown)


autogen.sh
```diff
@@ -30,9 +30,8 @@
 
 ### Code:
 
-## Used when remacs is dependent on nightly
-## Will print an error message telling people they need nightly rust
-require_nightly=true
+## Rust tool chain required version path
+rust_toolchain_vers_path="./rust-toolchain"
```
Wilfred: Doesn't this assume that you're running autogen.sh from the root?

Issue #863, opened 2018-09-22
[Proposal] Delete all x-toolkit except `gtk3`.


harryfei:
Currently, Emacs has quite a few x-toolkit choices for Linux platform.
* gtk
* gtk2
* gtk3
* lucid
* motif
* no-toolkit

Shall we just keep the `gtk3` and remove all the other `x-toolkit`? The related code will be simpler and cleaner. I believe it will be helpful for our porting.

db48x:
It certainly would simplify things, but... I don't think we should do it.

shaleh:
Can you give reasons why not?

Which of the toolkits are in use on modern Linux?

shaleh:
I can see the argument for keeping gtk2 and 3. Lucid, gtk1, and motif are ancient.

Wilfred:
I did actually build a GUI Emacs on Fedora 7 last year without gtk, but it wasn't as nice and I chased down the missing gtk bits.

I think this is reasonable. As long as it doesn't cause problems on other platforms, it would certainly make the code simpler.

harryfei:
AFAIK, Emacs use win32 raw api to implement it's gui.

Wilfred:
It will also increase the number of conflicts when we merge updates from GNU Emacs, because any change to the removed backends will conflict.

It's still a nice simplification I think.

harryfei:
I will work on this during the next days if no more opposite opinion.

harryfei:
Should we delete the `gtk2` or not?

shaleh:
Gtk2 is  likely still viable.

db48x:
> @db48x before we tell Harry to present a PR, do you have concerns?

I don't think we should do it because it would be quite difficult to undo it later, and because it might not make our job any easier (merge conflicts, as mentioned), and because we shouldn't delete things merely to make our job more convenient.

brotzeit:
I remember that we removed DOS support last year. Did this make merges more complicated ?

Wilfred:
Removing DOS slightly increased conflicts, but the DOS logic is more coupled to the Windows logic so there's more code churn. Most of these old toolkits rarely see any changes, so I'm not terribly concerned about increasing conflicts. The most impactful change was removing changelog files, which always conflict (trivial to fix though).

Removing these toolkits makes the code smaller, easier to understand and easier to port. It also removes a ton of kludges, such as:

```c
#include <X11/CoreP.h>		/* foul, but we need this to use our own
[...]

PR #885, opened 2018-09-30
delete `lucid` and `motif` x-toolkits

harryfei:
see #863

### DIFF
--- a/src/doprnt.c
+++ b/src/doprnt.c
@@ -503,7 +503,7 @@ esprintf (char *buf, char const *format, ...)
   return nbytes;
 }
 
-#if HAVE_MODULES || (defined HAVE_X_WINDOWS && defined USE_X_TOOLKIT)
+#if HAVE_MODULES
 
 /* Format to buffer *BUF of positive size *BUFSIZE, reallocating *BUF
    and updating *BUFSIZE if the buffer is too small, and otherwise
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
       os: linux
       before_script:
         - sudo apt update
-        - sudo apt install -y texinfo libgif-dev libxpm-dev
+        - sudo apt install -y texinfo libgif-dev libxpm-dev libgtk-3-dev
         - travis_wait rustup install $(cat rust-toolchain)
         - travis_wait rustup component add rustfmt-preview
       script:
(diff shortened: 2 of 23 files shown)


shaleh:
This looks like a win to me.

I'd yank Xaw3d too.

src/doprnt.c
```diff
@@ -503,7 +503,7 @@ esprintf (char *buf, char const *format, ...)
   return nbytes;
 }
 
-#if HAVE_MODULES || (defined HAVE_X_WINDOWS && defined USE_X_TOOLKIT)
+#if HAVE_MODULES
```
Wilfred: This looks wrong to me. Shouldn't we still consider `defined HAVE_X_WINDOWS`?

harryfei: I understand your confusions. The problem is that `GTK` is not an `X_TOOLKIT` in Emacs' logic.

* If we build with `gtk3`
  * `HAVE_WINDOW_SYSTEM` is true
  * `HAVE_X_WINDOWS` is **true**
  * `HAVE_X_11` is true
  * `USE_X_TOOLKIT` is **false**
  * `USE_GTK` is true
  * `HAVE_GTK3` is true

* If we build with `lucid`
  * `HAVE_WINDOW_SYSTEM` is true
  * `HAVE_X_WINDOWS` is **true**
  * `HAVE_X_11` is true
  * `USE_X_TOOLKIT` is **true**
  * `USE_LUCID` is true

In the vanilla Emacs, using `gtk3` toolkit doesn't trigger the compilation of those codes.
If we change the line to `#if HAVE_MODULES || defined HAVE_X_WINDOWS`, those codes will be compiled when using `gtk3` toolkit (which isn't expected).

shaleh: Are you sure about that:

```
#### Report on what we decided to do.
#### Report GTK as a toolkit, even if it doesn't use Xt.
#### It makes printing result more understandable as using GTK sets
#### toolkit_scroll_bars to yes by default.
if test "${HAVE_GTK}" = "yes"; then
  USE_X_TOOLKIT="$USE_GTK_TOOLKIT"
fi
```

harryfei: Yes, I am sure.

I checked the config.h after building with the multiple toolkits.  You can check it too.

The lines you show here is just correct the `printing result` of configure.ac. It doesn't set the `USE_X_TOOLKIT` macro to `true` for c codes.

shaleh: This is a great example of why I am happy to see us cut off the old code.

shaleh: Then the Xaw3d should go too as it depends on X_Toolkit

harryfei: Deleted Xaw3d.

PR #889, opened 2018-10-02
port delete-overlay and delete-all-overlays

brotzeit:


rust_src/src/buffers.rs
```diff
@@ -890,6 +891,42 @@ pub extern "C" fn build_overlay(
     }
 }
 
+/// Delete the overlay OVERLAY from its buffer.
+#[lisp_fn]
```
shaleh: Can you not use `LispOverlayRef` here instead of explicitly converting the overlay?

rust_src/src/buffers.rs
```diff
@@ -890,6 +891,42 @@ pub extern "C" fn build_overlay(
...
+
+        // When deleting an overlay with before or after strings, turn off
+        // display optimizations for the affected buffer, on the basis that
+        // these strings may contain newlines.  This is easier to do than to
+        // check for that situation during redisplay.
+        if windows_or_buffers_changed != 0 && Foverlay_get(overlay, Qbefore_string).is_not_nil()
+            || Foverlay_get(overlay, Qafter_string).is_not_nil()
+        {
+            buf_ref.set_prevent_redisplay_optimizations_p(true);
+        }
```
shaleh: So is this the return value? It is kind of hidden.

rust_src/src/buffers.rs
```diff
@@ -890,6 +891,42 @@ pub extern "C" fn build_overlay(
     }
 }
 
+/// Delete the overlay OVERLAY from its buffer.
+#[lisp_fn]
+pub fn delete_overlay(overlay: LispObject) -> LispObject {
+    let mut ov_ref = overlay.as_overlay_or_error();
+    let mut buf_ref = match marker_buffer(ov_ref.start.as_marker_or_error()) {
+        Some(b) => b,
```
shaleh: Could this return an `Option`? Or is `LispObject` more clear?

rust_src/src/buffers.rs
shaleh: Please put an explicit `Qnil` at the end. Because I could not tell that from the code.

shaleh:
Re-running checks

brotzeit:
Should be ok now.

PR #890, opened 2018-10-02
Port car-less-than-car

nickdrozd:
This is a strange function, right? It could easily be written in Lisp,
but I guess it's implemented as a primitive for efficiency. It's
mostly used as a predicate in calls to `sort'.

### DIFF
--- a/rust_src/src/fileio.rs
+++ b/rust_src/src/fileio.rs
@@ -3,8 +3,17 @@ use std::path;
 
 use remacs_macros::lisp_fn;
 
-use lisp::defsubr;
+use lisp::{defsubr, LispCons};
+use math::{arithcompare, ArithComparison};
 use multibyte::LispStringRef;
+
+/// Return t if (car A) is numerically less than (car B).
+#[lisp_fn]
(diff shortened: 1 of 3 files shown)


rust_src/src/fileio.rs
```diff
@@ -3,9 +3,22 @@ use std::path;
 
 use remacs_macros::lisp_fn;
 
-use lisp::defsubr;
+use lisp::{defsubr, LispObject};
+use math::{arithcompare, ArithComparison};
 use multibyte::LispStringRef;
 
+/// Return t if (car A) is numerically less than (car B).
+#[lisp_fn]
```
shaleh: Can't you specify each as a `LispCons` here?

nickdrozd: Ah, I didn't know that was possible. Done.

shaleh: Arguments and return values are handled by the lisp_fn macro.

brotzeit:
@Wilfred nickdrozd needs collaborator rights.

Wilfred:
@brotzeit I don't know how I overlooked that! :cry: 

@nickdrozd added you as a collaborator, so you can merge PRs etc :)

PR #896, opened 2018-10-04
rustup-toolchain will be installed when running ./autogen.sh

rofrol:
```bash
remacs$ ./autogen.sh
Checking Rust toolchain install ...
info: syncing channel updates for 'nightly-2018-09-09-x86_64-unknown-linux-gnu'
info: latest update on 2018-09-09, rust version 1.30.0-nightly (0198a1ea4 2018-09-08)
info: downloading component 'rustc'
info: downloading component 'rust-std'
info: downloading component 'cargo'
info: downloading component 'rust-docs'
info: installing component 'rustc'
info: installing component 'rust-std'
info: installing component 'cargo'
info: installing component 'rust-docs'
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,6 @@ more Emacs-y.
    rustup will use that version as long as you are in the Remacs
    repository
 
-        cd /path/to/remacs
-        rustup install `cat rust-toolchain`
-
 2. You will need a C compiler and toolchain. On Linux, you can do
    something like `apt install build-essential automake clang`. On
    macOS, you'll need Xcode.
--- a/README.md
+++ b/README.md
@@ -142,11 +142,9 @@ more Emacs-y.
 ### Requirements
 
 1. You will need
-   [Rust installed](https://www.rust-lang.org/en-US/install.html). Remacs
-   uses unstable Rust features, so you will need to use nightly. The
-   file `rust-toolchain` indicates the correct version to install;
-   rustup will use that version as long as you are in the Remacs
-   repository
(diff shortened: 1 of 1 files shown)


shaleh:
autogen.sh is really not the right place for this is it? autogen.sh is just meant to create the autotools files so configure can be run.

rofrol:
I just wanted to point that this note in Readme is redundant as this command will get correct toolchain

https://github.com/Wilfred/remacs/blob/a171e148b2887d78a5da94845dd813b455f79cc8/autogen.sh#L140

brotzeit:
I think this part should say something about not changing the project's toolchain since it is automatically taken from the file rust-toolchain.

rofrol:
@brotzeit I think it says something similar:

>Remacs uses unstable Rust features, so you will need to use nightly. The file `rust-toolchain` indicates the correct version to install; rustup will use that version as long as you are in the Remacs repository

brotzeit:
IMO the important part here is that it is handled automatically.

rofrol:
Something like:

>Remacs uses unstable Rust features, so you will need to use nightly. The file `rust-toolchain` indicates the correct version to install; rustup will automatically install and use that version.

brotzeit:
IMO that's good. But I would add at the end something like "don't override the project's toolchain manually". AFAIK this breaks the automatic installation.

shaleh:
Correct, a manual override should have an old school BLINK tag style *DONT DO IT, ITS A TRAP!*

PR #897, opened 2018-10-04
insert-buffer-substring

brotzeit:


rust_src/src/editfns.rs
```diff
@@ -799,4 +800,58 @@ pub fn char_equal(c1: LispObject, c2: LispObject) -> bool {
...
+///
+/// Point and before-insertion markers move forward to end up after the
+/// inserted text.
+/// Any other markers at the point of insertion remain before the text.
+///
+/// If the current buffer is multibyte and BUFFER is unibyte, or vice
+/// versa, strings are converted from unibyte to multibyte or vice versa
+/// using `string-make-multibyte' or `string-make-unibyte', which see.
+#[lisp_fn(min = "1")]
+pub fn insert_buffer_substring(buffer: LispObject, beg: LispObject, end: LispObject) -> LispObject {
```
shaleh: So asking for `buffer: LispBuffer` is not enough?

db48x: No, it also accepts a string containing a buffer name. We could make a type that would stand in for either one, though. Some kind of enum, perhaps, with it's own From impl.

shaleh: That sounds useful as this is a recurring pattern.

brotzeit: Can we open an issue for it and merge this PR ? I assume you want the same for processes ?

db48x: Oh yea, it doens't actually need to hold up this pr.

rust_src/src/editfns.rs
```diff
@@ -799,4 +800,58 @@ pub fn char_equal(c1: LispObject, c2: LispObject) -> bool {
...
+    }
+
+    let mut cur_buf = ThreadState::current_buffer();
+    unsafe {
+        set_buffer_internal_1(buf_ref.as_mut());
+        update_buffer_properties(b, e);
+        set_buffer_internal_1(cur_buf.as_mut());
+        insert_from_buffer(buf_ref.as_mut(), b, e - b, false)
+    };
+    Qnil
```
db48x: Oh, btw. You don't have to return nil, you can just not return anything at all.

rust_src/src/editfns.rs
shaleh: Agreed, that was not my intent.

shaleh: I have a draft in the works BTW, once I have it working I will put it up for comment.

rust_src/src/editfns.rs
brotzeit: Hmmm. I'm not sure we should do this.

brotzeit: I will merge this. If we decide we return nothing instead of `Qnil`, I will change this for all functions. But I want this to be consistent.

brotzeit:
`thread '<unnamed>' panicked at 'assertion failed: charpos <= bytepos', src/marker.rs:298:13`

Where did this suddenly come from ?

brotzeit:
Don't rerun the build. This is really odd...

shaleh:
Agreed. It ran fine locally.

shaleh:
@brotzeit all clear now?

brotzeit:
I wanted to leave this open for the backtrace of the bug. But I guess this isn't worth much.

PR #903, opened 2018-10-06
rearrange stuff

brotzeit:


rust_src/src/buffers.rs
```diff
@@ -367,22 +367,12 @@ impl LispBufferRef {
 impl LispBufferRef {
     #[inline]
     pub fn overlays_before(&self) -> Option<LispOverlayRef> {
-        let p = self.overlays_before;
-        if p == ptr::null_mut() {
-            None
-        } else {
-            Some(ExternalPtr::new(p))
-        }
+        unsafe { self.overlays_before.as_ref().map(|m| mem::transmute(m)) }
```
nickdrozd: `|m| mem::transmute(m))` can be replaced with just `mem::transmute`, since it's already a function of one variable.

(It would be great if we could get Clippy working, because it finds stuff like that automatically.)

nickdrozd: Same below

brotzeit: I get this
```
370 |         unsafe { self.overlays_before.as_ref().map(mem::transmute) }
    |                                                ^^^ expected an `FnOnce<(&remacs_sys::Lisp_Overlay,)>` closure, found `unsafe extern "rust-intrinsic" fn(_) -> _ {std::intrinsics::transmute::<_, _>}`
```

nickdrozd: Oh, nevermind then :)

nickdrozd: Clippy also has the benefit of preventing many eyeball review errors.

brotzeit:
Can anybody approve. I think this should be ok.

PR #905, opened 2018-10-07
Port remaining casefiddle Lisp functions

nickdrozd:
A lot of work is still being done in C, but all the exposed Lisp stuff is in Rust.

### DIFF
--- a/rust_src/src/casefiddle.rs
+++ b/rust_src/src/casefiddle.rs
@@ -3,7 +3,7 @@ use remacs_macros::lisp_fn;
 
 use lisp::defsubr;
 use lisp::LispObject;
-use remacs_sys::{case_action, casify_object, casify_word};
+use remacs_sys::{case_action, casify_object, casify_region_nil, casify_word};
 
 /// Convert argument to capitalized form and return that.
 /// This means that each word's first character is converted to either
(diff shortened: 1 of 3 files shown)


src/casefiddle.c
```diff
@@ -482,89 +467,17 @@ casify_region (enum case_action flag, Lisp_Object b, Lisp_Object e)
...
-  Lisp_Object bounds = Qnil;
-
-  if (!NILP (region_noncontiguous_p))
-    {
-      bounds = call1 (Fsymbol_value (intern ("region-extract-function")),
-		      intern ("bounds"));
-
-      while (CONSP (bounds))
-	{
-	  casify_region (CASE_UP, XCAR (XCAR (bounds)), XCDR (XCAR (bounds)));
```
Wilfred: I'm not actually sure this C code was correct. It calls XCAR twice, but it's only checked that the outer value is a cons cell. Not that it's a problem for us :)

rust_src/src/casefiddle.rs
```diff
@@ -36,4 +96,132 @@ pub fn upcase(object: LispObject) -> LispObject {
...
+    if region_noncontiguous_p.is_nil() {
+        unsafe { casify_region_nil(action, beg, end) }
+    } else {
+        let mut bounds = call!(
+            symbol_value(intern("region-extract-function")),
+            intern("bounds")
+        );
+
+        while bounds.is_cons() {
+            let cons = bounds.as_cons_or_error();
```
Wilfred: We've just checked that it is a cons, do we need the or_error version here?

brotzeit: Maybe something like `while let Some(c) = bounds.as_cons()`.

nickdrozd: I thought something was ugly about that. Updated.

PR #906, opened 2018-10-08
Port xml defuns

nickdrozd:


nickdrozd:
I have no knowledge or opinion at all about that :)

PR #907, opened 2018-10-08
Port casify_word

steinuil:
I actually wanted to port the casify lisp functions, but somebody beat me to it so I tried doing `casify_word`.

I'm not sure I'm using the correct functions to deal with this, so a review would be much appreciated.

### DIFF
--- a/rust_src/src/casefiddle.rs
+++ b/rust_src/src/casefiddle.rs
@@ -7,12 +7,38 @@ use keymap::Ctl;
 use lisp::defsubr;
 use lisp::LispObject;
 use lists::put;
-use remacs_sys::{case_action, casify_object, casify_region_nil, casify_word};
-use remacs_sys::{control_x_map, initial_define_key, meta_map};
+use remacs_sys::EmacsInt;
+use remacs_sys::{case_action, casify_object, casify_region, casify_region_nil};
+use remacs_sys::{control_x_map, initial_define_key, meta_map, scan_words, set_point};
 use remacs_sys::{Qdisabled, Qnil, Qt};
+use threads::ThreadState;
 
 use obarray::intern;
 use symbols::symbol_value;
+
+fn casify_word(flag: case_action, words: EmacsInt) -> LispObject {
+    let buffer_ref = ThreadState::current_buffer();
+
+    let far_end = match unsafe { scan_words(buffer_ref.pt, words) } {
+        0 => if words <= 0 {
+            buffer_ref.begv
+        } else {
+            buffer_ref.zv
+        },
+        n => n,
+    };
+
+    let new_pos = unsafe {
+        casify_region(
+            flag,
+            LispObject::from(buffer_ref.pt),
+            LispObject::from(far_end),
+        )
+    };
+
+    unsafe { set_point(new_pos) };
+    Qnil
+}
 
 /// Convert argument to capitalized form and return that.
 /// This means that each word's first character is converted to either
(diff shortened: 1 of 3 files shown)


rust_src/src/casefiddle.rs
```diff
@@ -4,16 +4,46 @@ use std::ffi::CString;
...
+
+    let new_pos: ptrdiff_t = unsafe {
+        casify_region(
+            flag,
+            LispObject::from(buffer_ref.pt),
+            LispObject::from(far_end),
+        )
+    };
+
+    let _ = goto_char(LispObject::from(new_pos));
```
brotzeit: The c macro `SET_PT` calls `set_point`.

rust_src/src/casefiddle.rs
```diff
@@ -4,16 +4,46 @@ use std::ffi::CString;
...
 
+use editfns::goto_char;
 use obarray::intern;
 use symbols::symbol_value;
 
+fn casify_word(flag: case_action, object: LispObject) -> LispObject {
+    let words = object.as_fixnum_or_error();
+    let buffer_ref = ThreadState::current_buffer();
+
+    let far_end: ptrdiff_t = unsafe { scan_words(buffer_ref.pt, words) };
```
brotzeit: I think you don't have to provide the type here and we use `isize` in rust.

rust_src/src/casefiddle.rs
```diff
@@ -4,16 +4,46 @@ use std::ffi::CString;
...
+    let far_end = match far_end {
```
brotzeit: Maybe you shouldn't shadow this variable and make it mutable instead.

steinuil: Personally I prefer not to make variables mutable unless it's really necessary. Maybe I could move the function call inside the match clause if you prefer that?

steinuil:
Updated.

rust_src/src/casefiddle.rs
```diff
@@ -7,13 +7,37 @@ use keymap::Ctl;
 use lisp::defsubr;
 use lisp::LispObject;
 use lists::put;
-use remacs_sys::{case_action, casify_object, casify_region_nil, casify_word};
+use remacs_sys::{case_action, casify_object, casify_region_nil};
+use remacs_sys::{casify_region, scan_words, set_point};
```
brotzeit: We try to keep the import order consistent. You can take a look at the order in the other files. I think all non lisp functions should be in one import.

rust_src/src/casefiddle.rs
```diff
@@ -7,13 +7,37 @@ use keymap::Ctl;
...
 use obarray::intern;
 use symbols::symbol_value;
 
+fn casify_word(flag: case_action, object: LispObject) -> LispObject {
+    let words = object.as_fixnum_or_error();
+    let buffer_ref = ThreadState::current_buffer();
+
+    let far_end = match unsafe { scan_words(buffer_ref.pt, words) } {
+        0 if words <= 0 => buffer_ref.begv,
+        0 => buffer_ref.zv,
```
brotzeit: I think this is easier to read if you put an if else in the first match arm.

steinuil:
And updated again.

rust_src/src/casefiddle.rs
```diff
@@ -7,13 +7,39 @@ use keymap::Ctl;
...
-use remacs_sys::{control_x_map, initial_define_key, meta_map};
+use remacs_sys::{case_action, casify_object, casify_region, casify_region_nil};
+use remacs_sys::{control_x_map, initial_define_key, meta_map, scan_words, set_point};
 use remacs_sys::{Qdisabled, Qnil, Qt};
+use threads::ThreadState;
 
 use obarray::intern;
 use symbols::symbol_value;
 
+fn casify_word(flag: case_action, object: LispObject) -> LispObject {
```
shaleh: Can't words be put in the arg list? The type is `EmacsInt`.

steinuil:
Done. I've also had to change the types of the arguments of the lisp functions.

PR #910, opened 2018-10-09
delete `no` option for `--with-x-toolkit`

harryfei:
After merging #885, the x-toolkit has only four options now.
 * no
 * yes or gtk (means gtk2 or gtk3)
 * gtk2
 * gtk3

This PR delete the `no` option. After merging this, the `HAVE_X_WINDOWS` and
`USE_GTK` will equal in value. They will just be same in Remacs.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -312,13 +312,12 @@ AC_ARG_WITH([x-toolkit],[AS_HELP_STRING([--with-x-toolkit=KIT],
  [use an X toolkit (KIT one of: yes or gtk, gtk2, gtk3, no)])],
 [	  case "${withval}" in
 	    y | ye | yes )	val=gtk ;;
-	    n | no )		val=no  ;;
 	    g | gt | gtk  )	val=gtk ;;
 	    gtk2  )	val=gtk2 ;;
 	    gtk3  )	val=gtk3 ;;
 	    * )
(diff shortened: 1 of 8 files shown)


harryfei:
Should I remove `HAVE_X_WINDOWS` completely? It's just the same as `USE_GTK` after merging this.

shaleh:
Let's keep it for now. It is not causing a problem and will make merges even more complex.

configure.ac
```diff
@@ -312,13 +312,12 @@ AC_ARG_WITH([x-toolkit],[AS_HELP_STRING([--with-x-toolkit=KIT],
  [use an X toolkit (KIT one of: yes or gtk, gtk2, gtk3, no)])],
 [	  case "${withval}" in
 	    y | ye | yes )	val=gtk ;;
-	    n | no )		val=no  ;;
 	    g | gt | gtk  )	val=gtk ;;
 	    gtk2  )	val=gtk2 ;;
 	    gtk3  )	val=gtk3 ;;
 	    * )
```
shaleh: Why are we dropping the `no` support? Emacs should be able to build with a GUI.

harryfei: * `--with-x-toolkit=no` means building with X raw API (There is no x toolkit such as lucid).
* `--with-x-window=no` means building without GUI (That will be an terminal Emacs).

shaleh: :+1:

PR #911, opened 2018-10-10
Make all NS drawing be done from drawRect

cjohansson:
See bug#31904 and bug#32812.

* src/nsterm.m (ns_update_begin): Don't lock focus, only clip if there
is already a view focused.
(ns_update_end): Don't mess with view focusing any more.
(ns_focus): Only clip drawing if there is already a focused view,
otherwise mark area dirty for later drawing.  Renamed ns_clip_to_rect.
All callers changed.
(ns_unfocus): Don't unfocus the view any more.  Renamed
ns_reset_clipping. All callers changed.
(ns_clip_to_row): Update to match ns_clip_to_rect.
(ns_clear_frame):
(ns_clear_frame_area):
(ns_draw_fringe_bitmap):
(ns_draw_window_cursor):
(ns_draw_vertical_window_border):
(ns_draw_window_divider):
(ns_dumpglyphs_stretch):
(ns_draw_glyph_string): Only draw if ns_focus or ns_clip_to_row
return YES.
(ns_copy_bits): Remove superfluous calls to ns_(un)focus.
(ns_flush_display): New function.

### DIFF
--- a/src/nsterm.m
+++ b/src/nsterm.m
@@ -276,12 +276,7 @@ struct ns_display_info *x_display_list; /* Chain of existing displays */
 long context_menu_value = 0;
 
 /* display update */
-static struct frame *ns_updating_frame;
-static NSView *focus_view = NULL;
 static int ns_window_num = 0;
-#ifdef NS_IMPL_GNUSTEP
-static NSRect uRect;            // TODO: This is dead, remove it?
-#endif
 static BOOL gsaved = NO;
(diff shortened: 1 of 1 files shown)


cjohansson:
I'll make a new pull request with these commits from GNU Emacs trunk:

9ad0f1d15c Fix deprecation warning
7946445962 Make all NS drawing be done from drawRect
41fa88b99b ; Fix some doc typos

PR #913, opened 2018-10-10
WIP: Clippy returns!!

shaleh:

PR #914, opened 2018-10-10
More Clippy!!

nickdrozd:
Based on https://github.com/Wilfred/remacs/pull/913

As is evident, I take a pretty granular approach when it comes to commits for this kind of thing.

After this, there are just a few big classes of Clippy errors/warnings remaining. One of them has to do marking any function that dereferences a raw pointer as `unsafe`. Another one has to with changing `self` arguments to `&self`, or maybe the other way around. Those seem like bigger changes that should be discussed first, so I didn't do them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,8 @@ Thank you for contributing to Remacs!
   can be found in the section on [Remacs style](#remacs-style).
 * Check for Rust compiler warnings, as CI won't go green if there
   are warnings.
+* Try to check your code with Clippy. `cargo clippy` will run the linter.
+  You can install clippy with `rustup component add clippy-preview`.
 * Format your code with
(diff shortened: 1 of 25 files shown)


shaleh:
I am a fan of smaller commits usually. I was working quickly to get you started.

shaleh:
The moving of `&self` to `self` as Clippy suggests should be safe. If the `&` is needed the compiler will let you know.

shaleh:
Thanks for picking this up.

nickdrozd:
Added a few more changes. One of the big remaining Clippy warnings has to do with `v as EmacsInt`, something about lossy conversion. Someone else can take a look at that one, since I don't understand it. Also needed is finally adding a `make clippy` option to the makefile and hooking it up in CI, once the errors are taken care of.

nickdrozd:
Fixed merge conflicts

PR #915, opened 2018-10-11
Also .rustified_enum some more recently introduced enums from gnutls

jcreedcmu:
Hello! I'm a long time emacs user, familiar with C (though not yet expert at rust or emacs internals), and thought remacs was a very interesting idea, and thought perhaps I'd try to take a crack at porting some builtins.

While attempting to get it to build on my machine, I found I needed to add these `.rustified_enum` calls. Without them, the
```rust
                .default_enum_style(bindgen::EnumVariation::ModuleConsts)
```
in `build.rs` causes module wrappers for `gnutls_ctype_target_t` and `gnutls_fips_mode_t` and `gnutls_gost_paramset_t` to be created, e.g. the `bindgen`-generated `bindings.rs` contains
```rust
pub mod gnutls_ctype_target_t {
    pub type Type = u32;
    pub const GNUTLS_CTYPE_CLIENT: Type = 0;
    pub const GNUTLS_CTYPE_SERVER: Type = 1;
    pub const GNUTLS_CTYPE_OURS: Type = 2;
    pub const GNUTLS_CTYPE_PEERS: Type = 3;
}
```
and, strangely, references to these enum types later in `bindings.rs` erroneously refer to simply (for example) `gnutls_ctype_target_t` as opposed to `gnutls_ctype_target_t::Type`. I don't know whether this is a bug in `bindgen` or an unintended consequence of how it's configured and called from `build.rs`. For example,
```rust
extern "C" {
    pub fn gnutls_certificate_type_get2(
        session: gnutls_session_t,
        target: gnutls_ctype_target_t, // <- this should be gnutls_ctype_target_t::Type
    ) -> gnutls_certificate_type_t;
}
```
minus the comment I added, occurs in `bindings.rs`, and produces 

    error[E0573]: expected type, found module 'gnutls_ctype_target_t'

### DIFF
--- a/rust_src/build.rs
+++ b/rust_src/build.rs
@@ -1081,9 +1081,12 @@ fn run_bindgen() {
                 .rustified_enum("gnutls_compression_method_t")
                 .rustified_enum("gnutls_connection_end_t")
                 .rustified_enum("gnutls_credentials_type_t")
+                .rustified_enum("gnutls_ctype_target_t")
                 .rustified_enum("gnutls_digest_algorithm_t")
                 .rustified_enum("gnutls_ecc_curve_t")
                 .rustified_enum("gnutls_ext_parse_type_t")
+                .rustified_enum("gnutls_fips_mode_t")
+                .rustified_enum("gnutls_gost_paramset_t")
                 .rustified_enum("gnutls_group_t")
                 .rustified_enum("gnutls_handshake_description_t")
                 .rustified_enum("gnutls_initstage_t")
@@ -1210,7 +1213,7 @@ fn run_bindgen() {
                 .generate()
                 .expect("Unable to generate bindings");
 
-            // https://github.com/servo/rust-bindgen/issues/839
+            // https://github.com/rust-lang-nursery/rust-bindgen/issues/839
             let source = bindings.to_string();
             let re = regex::Regex::new(
                 r"pub use self\s*::\s*gnutls_cipher_algorithm_t as gnutls_cipher_algorithm\s*;",


jcreedcmu:
This is the version of gnutls I have installed:
```
$ yum info gnutls-devel
Last metadata expiration check: 28 days, 0:59:38 ago on Thu Sep 13 08:53:12 2018.
Installed Packages
Name         : gnutls-devel
Version      : 3.6.4
Release      : 1.fc28
Arch         : x86_64
Size         : 4.6 M
Source       : gnutls-3.6.4-1.fc28.src.rpm
Repo         : @System
From repo    : updates
Summary      : Development files for the gnutls package
URL          : http://www.gnutls.org/
License      : GPLv3+ and LGPLv2+
Description  : GnuTLS is a secure communications library implementing the SSL, TLS and DTLS
             : protocols and technologies around them. It provides a simple C language
             : application programming interface (API) to access the secure communications
             : protocols as well as APIs to parse and write X.509, PKCS #12, OpenPGP and
             : other required structures.
             : This package contains files needed for developing applications with
             : the GnuTLS library.
```

rust_src/build.rs
```diff
@@ -1210,7 +1213,7 @@ fn run_bindgen() {
                 .generate()
                 .expect("Unable to generate bindings");
 
-            // https://github.com/servo/rust-bindgen/issues/839
+            // https://github.com/rust-lang-nursery/rust-bindgen/issues/839
```
jcreedcmu: This repo seems to have moved over here

jcreedcmu:
These enums to appear to be new-ish. For example, `gnutls_ctype_target_t` came from https://github.com/gnutls/gnutls/commit/07180a41673174 (Aug 15 2018)

jcreedcmu:
I wasn't able to reproduce this weird "enum referred to by `module_name` name instead of `module_name::Type`" by running `bindgen` from the commandline.

If I make `blah.h` be
```c
typedef enum {
                         bar, baz, blap
} Foo;

void func(Foo x);
```
and do
    $ bindgen --constified-enum-module 'Foo' blah.h
(with `rust-toolchain` in the same directory containing `nightly-2018-09-09` to match `remacs`'s toolchain version)
I get the correct output
```rust
/* automatically generated by rust-bindgen */

pub mod Foo {
    pub type Type = u32;
    pub const bar: Type = 0;
    pub const baz: Type = 1;
    pub const blap: Type = 2;
}
extern "C" {
    pub fn func(x: Foo::Type);
}
```
as expected.

jcreedcmu:
Oho, but I *can* reproduce it using `https://github.com/rust-lang-nursery/rust-bindgen.git` at commit `7111bb`, as remacs does, using a minimal `build.rs` with the above `blah.h` of
```rust
extern crate bindgen;

use std::env;
use std::path::PathBuf;

fn main() {
    let bindings = bindgen::Builder::default()
        .header("blah.h")
        .default_enum_style(bindgen::EnumVariation::ModuleConsts)
        .generate()
        .expect("Unable to generate bindings");

    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
    bindings
        .write_to_file(out_path.join("bindings.rs"))
        .expect("Couldn't write bindings!");
}
```

jcreedcmu:
Filed issue https://github.com/rust-lang-nursery/rust-bindgen/issues/1418

Hope I'm not missing something obvious about how bindgen is supposed to be invoked.

jcreedcmu:
Ok, yeah, the `bindgen` folks confirm that this is a bindgen bug, this PR is probably not necessary --- it seems bad to have to track every new enum that some dependency introduces anyway.

jcreedcmu:
Oof, I don't know what the right resolution of this is, now, actually. Trying to bump the `bindgen` version all the way up to `b8cd8a161` (which in which @emilio just fixed rust-lang-nursery/rust-bindgen#1418) produces like 1200 other errors, so maybe some other version skew matters there...

In any event, this PR works for me as a temporary workaround to get remacs building. Feel free to merge or close depending on whether you think it's useful or a gross hack...

jcreedcmu:
Cool, thanks! I think the bindgen upgrade might not be so difficult after all. It might just be that https://github.com/rust-lang-nursery/rust-bindgen/pull/1364 or something related to it narrowed the effect of `blacklist_type` and all that's needed is replacing it in a few cases with `blacklist_function` or the like...

jcreedcmu:
A little more work needs to be done on the bindgen side, I think. Looks like prior tohttps://github.com/rust-lang-nursery/rust-bindgen/pull/1364, `blacklist_type` would blacklist anything regardless of `ItemKind`, and after it, there is no way to blacklist `Var`s and `Module`s at all.

jcreedcmu:
Cool, I'll give it a try if I find some time this weekend.

PR #917, opened 2018-10-12
Port memory-use-counts

nickdrozd:


rust_src/src/alloc.rs
```diff
@@ -0,0 +1,39 @@
...
+/// Garbage collection does not decrease them.
+/// The elements of the value are as follows:
+///   (CONSES FLOATS VECTOR-CELLS SYMBOLS STRING-CHARS MISCS INTERVALS STRINGS)
+/// All are in units of 1 = one object consed except for VECTOR-CELLS
+/// and STRING-CHARS, which count the total length of objects consed.
+/// MISCS include overlays, markers, and some internal types.
+/// Frames, windows, buffers, and subprocesses count as vectors
+///   (but the contents of a buffer's text do not count here).
+#[lisp_fn]
+pub fn memory_use_counts() -> LispObject {
```
shaleh: Do we not have a way to turn a Vector into the expected list so this is not as hard coded.

nickdrozd:
```rust
pub fn memory_use_counts() -> LispObject {
    unsafe {
        list(&[
            bounded_number(globals.cons_cells_consed),
            bounded_number(globals.floats_consed),
            bounded_number(globals.vector_cells_consed),
            bounded_number(globals.symbols_consed),
            bounded_number(globals.string_chars_consed),
            bounded_number(globals.misc_objects_consed),
            bounded_number(globals.intervals_consed),
            bounded_number(globals.strings_consed),
        ])
    }
}
```

That works, but there is probably a Rustier way.

rust_src/src/alloc.rs
nickdrozd:
```rust
pub fn memory_use_counts() -> LispObject {
    unsafe {
        list(&[
            bounded_number(globals.cons_cells_consed),
            bounded_number(globals.floats_consed),
            bounded_number(globals.vector_cells_consed),
            bounded_number(globals.symbols_consed),
            bounded_number(globals.string_chars_consed),
            bounded_number(globals.misc_objects_consed),
            bounded_number(globals.intervals_consed),
            bounded_number(globals.strings_consed),
        ])
    }
}
```

That works, but there is probably a Rustier way.

test/rust_src/src/alloc-tests.el
```diff
@@ -0,0 +1,12 @@
...
+
+(require 'ert)
+
+(ert-deftest memory-use-count ()
+  (let ((counts (memory-use-counts)))
+    (should (= 8 (length counts)))
+    (dolist (count counts)
+      (should (integerp count)))))
+
+(provide 'alloc-tests)
```
brotzeit: Is this necessary ?

nickdrozd: Do you mean the `provide` line? I don't know, `auto-insert` adds that and I always just leave it.

Wilfred: It's only necessary if users want to do `(require 'alloc-tests)`. It's harmless and other Emacs test files do it (e.g. abbrev-tests.el).

If a user really wanted to define their own alloc-tests.el, it'd still be fine I think, because none of the tests (for C or Rust) are put on `load-path` for users.

rust_src/src/alloc.rs
nickdrozd: Okay, I redid the commit to do it this way, except that I put the `From` in `lisp.rs`. I tried to do `vec![x,y,z,...].iter().map(bounded_number).collect()`, but there was some kind of error about something not being implemented for `FnMut` something.

PR #919, opened 2018-10-13
delete more non-gtk code in x related files.

harryfei:
After merging #885 and #910 , non-gtk logic isn't needed in x related files.


brotzeit:
Wow. How much more is coming ;)

harryfei:
> Wow. How much more is coming ;)

Just likes the chain reaction.

PR #930, opened 2018-10-17
LispBufferOrCurrent and LispBufferOrName types.

shaleh:
LispBufferOrName is either a LispBuffer or the name of one which we
can use to find the Buffer. nsberror is called for an invalid object.

LispBufferOrCurrent is a LispBufferOrName that uses the current
buffer when the object provided is nil.


shaleh:
I am interested in:

1. Do we care?
1. If so, are these the best names?

Currently, I think `lispbuffer_or_by_name` is wasted. As we port more it might become worthwhile.

The extra boilerplate for the handling of `if nil: current_buffer` is a little annoying and drives up the number of lines added.

rust_src/src/buffers.rs
```diff
@@ -488,6 +486,78 @@ fn assoc_ignore_text_properties(key: LispObject, list: LispObject) -> LispObject
     }
 }
 
+pub enum LispBufferOrName {
+    Buffer(LispBufferRef),
+}
```
db48x: Shouldn't this enum have a variant for the name?

shaleh: You mean repeat the enum name?

```
pub enum LispBufferOrName {
    LispBufferOrName(LispBufferRef)
}
```

The `Buffer` name is from my initial plan to have `Buffer` and `Name`. But I went with the current code instead.

brotzeit: Is this idiomatic rust ?

rust_src/src/buffers.rs
```diff
@@ -488,6 +486,78 @@ fn assoc_ignore_text_properties(key: LispObject, list: LispObject) -> LispObject
...
+            current_buffer()
+        } else {
+            buffer_or_name
+        };
+        let buffer = buffer_or_by_name(obj);
+        LispBufferOrCurrent::Buffer(buffer)
+    }
+}
+
+pub fn lispbuffer_or_by_name(buffer_or_name: LispObject) -> LispObject {
```
brotzeit: We should choose a name that suggests that it's similiar to `get_buffer` except it returns `nil` instead of throwing an error.  Maybe get_buffer and get_buffer_or_error. Would be consistent.

brotzeit: We should choose a name that suggests that it's similiar to get_buffer except it returns nil instead of throwing an error for nil. Maybe get_buffer and get_buffer_or_error. Would be consistent.

shaleh: Yeah. Or I might just drop the function for now.

rust_src/src/buffers.rs
shaleh: I thought about it some and came up with what I think is a better approach.

If we like this one, I will do something similar with `LispBufferOrCurrent`.

shaleh:
This should be neutral or even slightly negative number of lines. However, the verbosity of the new data type and rustfmt splitting out the lines pushed it positive. oh well.

rust_src/src/buffers.rs
db48x: I don't think I like this as much as an enum. I had envisoned an enum that held either a string or a buffer:

    enum LispBufferOrName {
        Buffer(LispObject),
        Name(LispObject),
    }

And then two From impls, like this:

    impl From<LispObject> for LispBufferOrName {
        fn from(v: LispObject) -> LispBufferOrName {
            if v.is_string() {
                LispBufferOrName::Name(v)
            } else {
                LispBufferOrName::Buffer(v.as_buffer_or_error())
            }
        }
    }

    impl From<LispBufferOrName> for Option<LispBufferRef> {
        fn from(v: LispBufferOrName) -> Option<LispBufferRef> {
            match v {
                LispBufferOrName::Buffer(b) => Ok(b),
                LispBufferOrName::Name(n) => {
                    cdr(assoc_ignore_text_properties(buffer_or_name, unsafe {
                        Vbuffer_alist
                    })).into()
                }
            }
        }
    }

The first would allow any `lisp_fn` to use `LispBufferOrName` as an argument type, and the second would allow them to trivially convert it into a buffer by calling `.into()`. Neither operation requires any extra allocation, or even calling methods with long names.

shaleh: I tried it. `nsberror` needs the name. So I did this instead. There are also slightly different handlings of nil so I could not easily put the type in LispFn. That was my goal. This was the result.

db48x: So have an `impl From<LispBufferOrName> for LispBufferRef` which calls `nsberror` instead of returning `None`.

shaleh: yeah, it got messy so I made this simplification. I can recreate the gnarly version so we can talk about it.

shaleh: I also want to work with a 'LispBufferOrCurrent' as well. Using structs this works well. Trying to make From and friends work it was a lot of replicated code and then again, different error handling. Sometimes. Plus, there are currently only a few calls.

brotzeit:
I think it's more important that it's easy to read.

brotzeit:
What are the advantages of using a struct over an enum here ? I don't know, but it feels like an enum fits better in this case.

shaleh:
Replaced with a newer PR

PR #933, opened 2018-10-17
delete lwlib.

harryfei:
* delete almost all the lwlib.
* move `lwlib/lwlib-widget.h` into `src/lwlib-widget.h`. Some codes depend on this header file.


harryfei:
rebased

PR #935, opened 2018-10-17
port eval-region

brotzeit:
What would 

```c
struct infile
{
  /* The input stream.  */
  FILE *stream;

  /* Lookahead byte count.  */
  signed char lookahead;

  /* Lookahead bytes, in reverse order.  Keep these here because it is
     not portable to ungetc more than one byte at a time.  */
  unsigned char buf[MAX_MULTIBYTE_LENGTH - 1];
} *infile;
```
look like in rust ?

### DIFF
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -2,13 +2,14 @@
 
 use libc;
 use std::ffi::CString;
+use std::ptr;
 
 use remacs_macros::lisp_fn;
 use remacs_sys;
-use remacs_sys::staticpro;
-use remacs_sys::EmacsInt;
-use remacs_sys::{build_string, read_internal_start, symbol_redirect};
-use remacs_sys::{globals, Qnil, Qread_char};
+use remacs_sys::{build_string, read_internal_start, readevalloop, specbind, staticpro,
+                 symbol_redirect, unbind_to, Fcons};
+use remacs_sys::{globals, EmacsInt};
+use remacs_sys::{Qeval_buffer_list, Qnil, Qread_char, Qstandard_output, Qsymbolp};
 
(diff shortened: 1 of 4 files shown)


db48x:
Stick it in a header file and let bindgen handle it.

rust_src/src/lread.rs
```diff
@@ -182,4 +184,56 @@ pub fn read(stream: LispObject) -> LispObject {
...
+/// which is the input stream for reading characters.
+///
+/// This function does not move point.
+#[lisp_fn(min = "2")]
+pub fn eval_region(
+    start: LispObject,
+    end: LispObject,
+    printflag: LispObject,
+    read_function: LispObject,
+) -> LispObject {
```
shaleh: Do we want to update this comment to reflect we have inherited it? Can we resolve it?

brotzeit: I don't understand this comment :)

shaleh: The FIXME that you carried over. What should we do with it?

brotzeit: I meant the FIXME.

brotzeit: I still don't get it. `eval-sexp-add-defvars` is a lisp function ?

db48x: Yes, it's in `elisp-mode.el`. `cconv.el` has some TODOs at the top:

    ;; TODO: (not just for cconv but also for the lexbind changes in general)
    ;; - let (e)debug find the value of lexical variables from the stack.
    ;; - make eval-region do the eval-sexp-add-defvars dance.

and so on. Since it's 8 years old, I guess it's not terribly important. I guess you can just leave the FIXME in place :)

brotzeit:
Should be good.

PR #937, opened 2018-10-18
Fix more clippy complaints

nickdrozd:


rust_src/src/marker.rs
```diff
@@ -559,7 +559,7 @@ impl LispBufferRef {
 
 /// Return the byte position corresponding to CHARPOS in B.
 #[no_mangle]
-pub extern "C" fn buf_charpos_to_bytepos(b: *mut Lisp_Buffer, charpos: isize) -> isize {
+pub unsafe extern "C" fn buf_charpos_to_bytepos(b: *mut Lisp_Buffer, charpos: isize) -> isize {
```
db48x: Why does clippy recommend making this function unsafe?

nickdrozd: I'm not sure exactly, but I think it requires that any function that directly manipulates a raw pointer be marked unsafe.

shaleh: @db48x clippy warns whenever there is `unsafe` within a function and the function signature does not reflect it.

rust_src/src/casefiddle.rs
```diff
@@ -215,32 +215,16 @@ pub extern "C" fn syms_of_casefiddle() {
...
-        initial_define_key(
-            meta_map,
-            'l' as i32,
-            CString::new("downcase-word").unwrap().as_ptr(),
-        );
-        initial_define_key(
-            meta_map,
-            'u' as i32,
-            CString::new("upcase-word").unwrap().as_ptr(),
-        );
```
shaleh: Maybe we could add a macro for these strings?

nickdrozd: I was thinking the same thing.

rust_src/src/marker.rs
db48x: That doesn't make any sense. A function that is marked as "unsafe" doesn't need to have any unsafe blocks inside of it; in fact it generally should not.

brotzeit: Is it not possible to configure clippy like rustfmt ?

nickdrozd: It is, but the safety recommendations seem reasonable to me. Or rather, since I don't completely understand the logic behind them, it seems reasonable to me to do whatever Clippy says with respect to safety unless there is a specific reason not to.

nickdrozd:
Can anyone make sense of these failures?

Wilfred:
macOS tests look like GnuPG version issues again. The Linux issue is stranger:

```
Test regex-tests-PTESTS backtrace:
  signal(ert-test-failed (((should-not (regex-tests-PTESTS)) :form (re
  ert-fail(((should-not (regex-tests-PTESTS)) :form (regex-tests-PTEST
  (if (not (unwind-protect (setq value-49 (apply fn-47 args-48)) (setq
  (let (form-description-51) (if (not (unwind-protect (setq value-49 (
  (let ((value-49 'ert-form-evaluation-aborted-50)) (let (form-descrip
  (let* ((fn-47 (function regex-tests-PTESTS)) (args-48 (condition-cas
  (closure (t) nil (let* ((fn-47 (function regex-tests-PTESTS)) (args-
  ert--run-test-internal(#s(ert--test-execution-info :test #s(ert-test
  ert-run-test(#s(ert-test :name regex-tests-PTESTS :documentation "Te
  ert-run-or-rerun-test(#s(ert--stats :selector (not (tag :expensive-t
  ert-run-tests((not (tag :expensive-test)) #f(compiled-function (even
  ert-run-tests-batch((not (tag :expensive-test)))
  ert-run-tests-batch-and-exit((not (tag :expensive-test)))
  eval((ert-run-tests-batch-and-exit '(not (tag :expensive-test))))
  command-line-1(("-L" ":." "-l" "ert" "-l" "src/regex-tests.el" "--ev
  command-line()
  normal-top-level()
Test regex-tests-PTESTS condition:
    (ert-test-failed
     ((should-not
       (regex-tests-PTESTS))
      :form
      (regex-tests-PTESTS)
      :value
      ("line number 341: Regex '^[a-z0-9]*$': Expected match; but no match")))
   FAILED   3/22  regex-tests-PTESTS
```

This is running some tests of the regex engine using the `PTESTS` file. I don't see how that's related at all. Does the test fail for you locally?

nickdrozd:
`epg-tests` always fails for me (is there a way we can skip it if the relevant requirements aren't installed?), but other than that everything passes locally.

nickdrozd:
Okay, I think I figured out the new Clippy interface, so I went through and updated the old plugin interface. While I was at it, I also turned up the Clippy level to `pedantic` and fixed a few of those complaints as well (but not all of them). I think this fixes https://github.com/Wilfred/remacs/issues/757? But I may have misunderstood it.

My method for running `clippy` is simple: go to `src/Makefile` (the generated one) and change `cargo build` to `cargo clippy`, then run `make`. That's not how it would work in CI, obviously, but running that locally now raises no errors.

All tests pass locally too, so hopefully that will happen in the build.

rust_src/src/cmds.rs
```diff
@@ -214,20 +214,18 @@ pub fn delete_char(n: EmacsInt, killflag: bool) -> () {
...
+    } else if n < 0 {
+        if pos < buffer.begv {
+            xsignal!(Qbeginning_of_buffer);
         } else {
-            unsafe { del_range(buffer.pt, pos) };
+            unsafe { del_range(pos, buffer.pt) };
         }
+    } else if pos > buffer.zv {
+        xsignal!(Qend_of_buffer);
     } else {
```
shaleh: Are you sure? `kill-forward-characters` might be subtly different.

shaleh: @nickdrozd my comment is on the code below this comment. Is `del_range` doing exactly what `kill-forward-chars` is doing?

nickdrozd: Before:
```rust
    if !killflag {
        if n < 0 {
            if pos < buffer.begv {
                xsignal!(Qbeginning_of_buffer);
            } else {
                unsafe { del_range(pos, buffer.pt) };
            }
        } else if pos > buffer.zv {
            xsignal!(Qend_of_buffer);
        } else {
            unsafe { del_range(buffer.pt, pos) };
        }
    } else {
        call_raw!(Qkill_forward_chars, LispObject::from(n));
    }
```

After:
```rust
    if killflag {
        call_raw!(Qkill_forward_chars, LispObject::from(n));
    } else if n < 0 {
        if pos < buffer.begv {
            xsignal!(Qbeginning_of_buffer);
        } else {
            unsafe { del_range(pos, buffer.pt) };
        }
    } else if pos > buffer.zv {
        xsignal!(Qend_of_buffer);
    } else {
        unsafe { del_range(buffer.pt, pos) };
    }
```

`kill-forward-chars` is called just when `killflag` is true, before and after. The major difference is that `if !killflag` is changed to `if killflag`, inverting the `if/else` blocks. But the logic should be the same.

The diff does a poor job of representing the changes in this case, because I guess it's trying to preserve continuity when there is none.

shaleh: !OH!

Thanks @nickdrozd. I did not look closely enough.

nickdrozd:
@brotzeit @db48x @harryfei Any questions / objections?

harryfei:
LGTM. :+1:

PR #939, opened 2018-10-19
port exit_recursive_edit and abort_recursive_edit

brotzeit:


rust_src/src/keyboard.rs
```diff
@@ -122,6 +123,36 @@ pub fn lucid_event_type_list_p(event: Option<LispCons>) -> bool {
     })
 }
 
+// _Noreturn will be added to prototype by make-docfile.
```
brotzeit: What does this mean and how do i add attributes to functions ?

nickdrozd: According to @db48x, there's no need to worry about it in Rust (see https://github.com/Wilfred/remacs/pull/904#issuecomment-427613854). Here, it's taken care of by the `!` return type. In C, the generated prototype becomes `_Noreturn EXFUN (Fabort_recursive_edit, 0);`, whatever that means.

shaleh: From the GCC docs:

```
noreturn
A few standard library functions, such as abort and exit, cannot return. GCC knows this automatically. Some programs define their own functions that never return. You can declare them noreturn to tell the compiler this fact. For example,
          void fatal () __attribute__ ((noreturn));
          
          void
          fatal (/* ... */)
          {
            /* ... */ /* Print error message. */ /* ... */
            exit (1);
          }
     
The noreturn keyword tells the compiler to assume that fatal cannot return. It can then optimize without regard to what would happen if fatal ever did return. This makes slightly better code. More importantly, it helps avoid spurious warnings of uninitialized variables.

The noreturn keyword does not affect the exceptional path when that applies: a noreturn-marked function may still return to the caller by throwing an exception or calling longjmp.

Do not assume that registers saved by the calling function are restored before calling the noreturn function.

It does not make sense for a noreturn function to have a return type other than void.

The attribute noreturn is not implemented in GCC versions earlier than 2.5. An alternative way to declare that a function does not return, which works in the current version and in some older versions, is as follows:

          typedef void voidfn ();
          
          volatile voidfn fatal;
```

shaleh: TL;DR, it allows for compilation optimization and checks in C code. It informs how we port using `-> !` syntax as @nickdrozd points out.

shaleh: We can drop the comment

rust_src/src/keyboard.rs
```diff
@@ -122,6 +123,36 @@ pub fn lucid_event_type_list_p(event: Option<LispCons>) -> bool {
...
+pub fn quit_recursive_edit(val: LispObject) -> ! {
```
shaleh: Maybe `val` should be bool? And we convert the bool in `Fthrow` call?

rust_src/src/keyboard.rs
```diff
@@ -122,6 +123,36 @@ pub fn lucid_event_type_list_p(event: Option<LispCons>) -> bool {
...
+        xsignal!(
+            Quser_error,
+            ::remacs_sys::make_string(
+                msg.as_ptr() as *const ::libc::c_char,
+                msg.len() as ::libc::ptrdiff_t,
+            )
+        );
+    }
+}
+
```
shaleh: Both of these could be `-> !` as well. Does the `lisp_fn` macro support that currently?

db48x: Yes, it does.

brotzeit: @db48x I have tried it but it didn't work.

db48x: What was the problem with it?

brotzeit: Some error that indicates it isn't supported. Sorry I forgot to copy it. We also don't have an occurrance in our code base of `lisp_fn` with `-> !`.

brotzeit:
```

error[E0277]: the trait bound `lisp::LispObject: std::convert::From<!>` is not satisfied
   --> src/keyboard.rs:144:1
    |
144 | #[lisp_fn]
    | ^^^^^^^^^^ the trait `std::convert::From<!>` is not implemented for `lisp::LispObject`
    |
    = help: the following implementations were found:
              <lisp::LispObject as std::convert::From<()>>
              <lisp::LispObject as std::convert::From<bool>>
              <lisp::LispObject as std::convert::From<std::option::Option<T>>>
              <lisp::LispObject as std::convert::From<std::vec::Vec<lisp::LispObject>>>
            and 23 others
    = note: required by `std::convert::From::from`
```

db48x: Heh, that's not an error message you see often. I did think it would work though.

db48x: Well, this actually seems to work:

```
impl From<!> for LispObject {
    fn from(_v: !) -> Self {
        // I'm surprized that this works
        Qnil
    }
}
```

db48x:
Yea, don't introduce any warnings :)

Add a `#[allow(unreachable_code)]` to the body of the function generated by the `lisp_fn` macro, on line 121 of `src/remacs-macros/lib.rs`.

db48x:
You could add some :)

brotzeit:
I would if I would understand the macro ;)

brotzeit:
But it's a lot more complicated(at least it seems to me) than other stuff that also isn't documented. And I wanted to start with the things I understand (more or less).

PR #945, opened 2018-10-21
port charsetp

brotzeit:


rust_src/src/charset.rs
```diff
@@ -0,0 +1,28 @@
...
+use remacs_sys::hash_lookup;
+use remacs_sys::Vcharset_hash_table;
+
+use lisp::defsubr;
+use lisp::LispObject;
+
+impl LispObject {
+    pub fn is_charset(self) -> bool {
+        unsafe {
+            let mut h_ref = Vcharset_hash_table.as_hash_table_or_error();
```
Wilfred: I like that this is explicitly showing use of `Vcharset_hash_table`. I've not seen this variable before, looks like it's not exposed to elisp.

Once all the charset functions are ported to rust, we might want to store this variable as a hash_table to save the typecheck on every call.

shaleh:
I updated the code to use the new style. My comment was stale from when GitHub was hiccuping.

PR #946, opened 2018-10-22
port more frame functions

brotzeit:


rust_src/src/frames.rs
```diff
@@ -305,4 +310,56 @@ pub fn frame_text_height(frame: LispObject) -> LispObject {
...
+
+/// Return width (in pixels) of vertical window dividers on FRAME.
+#[lisp_fn(min = "0", name = "right_divider_width")]
+pub fn frame_right_divider_width(frame: LispObject) -> LispObject {
+    let frame_ref = frame_or_selected(frame);
+    LispObject::from(frame_ref.right_divider_width)
+}
+
+/// Return width (in pixels) of horizontal window dividers on FRAME.
+#[lisp_fn(min = "0", name = "bottom_divider_width")]
```
Wilfred: The tests are failing in this PR, and I think it's because you've set the name. Remacs isn't getting a function called `frame-bottom-dviider-width`. I think this should be the other way round:

```
#lispfn(name="frame-bottom-divider-width")
pub fn bottom_divider_width(...)
```

shaleh: Why is `name` being used at all here? Is there a conflict somewhere?

Wilfred:
```
src/frame.c
3171 DEFUN ("frame-bottom-divider-width", Fbottom_divider_width, Sbottom_divider_width, 0, 1, 0,
3175   return make_number (FRAME_BOTTOM_DIVIDER_WIDTH (decode_any_frame (frame)));
3312   {"bottom-divider-width",	SYMBOL_INDEX (Qbottom_divider_width)},
3855   store_in_alist (alistptr, Qbottom_divider_width,
3856 		  make_number (FRAME_BOTTOM_DIVIDER_WIDTH (f)));
4235 x_set_bottom_divider_width (struct frame *f, Lisp_Object arg, Lisp_Object oldval)
4237   int old = FRAME_BOTTOM_DIVIDER_WIDTH (f);
4242       f->bottom_divider_width = new;
4243       adjust_frame_size (f, -1, -1, 4, 0, Qbottom_divider_width);
5444   DEFSYM (Qbottom_divider_width, "bottom-divider-width");
5847   defsubr (&Sbottom_divider_width);
```

It's consistent with the C code the Rust is replacing (the C code had different names), but I don't see any obvious conflict. We could probably remove it as far as I can see.

brotzeit: Oh I missed this. I take a look at it again.

rust_src/src/frames.rs
```diff
@@ -346,4 +351,56 @@ pub fn frame_text_height(frame: LispObject) -> LispObject {
...
 
+/// Return scroll bar width of FRAME in pixels.
+#[lisp_fn(min = "0")]
+pub fn frame_scroll_bar_width(frame: LispObject) -> LispObject {
+    let frame_ref = frame_or_selected(frame);
+    let width = match frame_ref.vertical_scroll_bar_type() {
+        vertical_scroll_bar_type::vertical_scroll_bar_left
+        | vertical_scroll_bar_type::vertical_scroll_bar_right => frame_ref.config_scroll_bar_width,
+        _ => 0,
+    };
```
shaleh: Why are you manually casting in each of these functions? Why not just return `EmacsInt`?

brotzeit:
```401 | pub fn frame_bottom_divider_width(frame: LispObject) -> EmacsInt {
    |                                                         -------- expected `i64` because of return type
402 |     let frame_ref = frame_or_selected(frame);
403 |     frame_ref.bottom_divider_width
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected i64, found i32
help: you can cast an `i32` to `i64`, which will sign-extend the source value
    |
403 |     frame_ref.bottom_divider_width.into()
```

`lisp_fn` should handle this right ?

shaleh: EmacsInt is a i64. The compiler warns about casts. So either `lisp_fn` is taught about i32 or you manually cast to an EmacsInt.

db48x: If you want `lisp_fn` to handle it for you, your function must return a specific integer type like i32: `pub fn frame_bottom_divider_width(frame: LispObject) -> i32`

brotzeit: IMO `lisp_fn` should be able to handle this.

brotzeit: I see you already told me here to return i32. I think that was too early in the morning for me :P

brotzeit:
Why is bindgen failing here ?

db48x:
Unknown. I just tried it myself and it works fine. It would be helpful to see what bindgen did generate, if it didn't generate those two methods.

brotzeit:
How can we see this in the travis build ?

db48x:
We've never used them, but Travis has the option to preserve files from a build. We really ought to configure it to preserve the build artifacts, including the log files from the lisp tests and the generated Rust files.You can set it to preserve bindings.rs at a minimum and commit that to this branch and when it rebuilds it will save it for you.

brotzeit:
It seems bindgen doesn't generate the methods when running without a window system.

brotzeit:
I need help with this...

shaleh:
Can you post what you have done? Where are you hung up?

brotzeit:
Do I have to add a variable in configure.ac or can I access `HAVE_WINDOW_SYSTEM` directly ? Or should I use `$window_system` ? How can I add a feature in rust based on this ?

db48x:
See https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section.

If you add this to the Cargo.toml file:

    [features]
    have-window-system = []

then in the code you'll be able conditionally compile something like this:

    #[cfg(feature = "have-window-system")]
    pub fn example() {
    }

It is then only a matter of turning this feature on and off at the right times. I thought at first that you would need to modify `configure.ac` so that it know how to turn the feature on and off inside the `Cargo.toml`, but actually you can do it in the `Makefile.in` instead. When it comes time to build the Rust code, have add the `--features "have-window-system"` command-line option to the `cargo build` command.

shaleh:
The Makefile.in can update the Cargo.toml:

```
[features]
default = ["use_gtk", "have_window_system"]
```

harryfei:
@brotzeit A reference for adding a cargo feature in remacs' rust code.

https://github.com/harryfei/remacs/commit/023be6a84db3cd4d379e4e25073df212b74e9623

shaleh:
I just cleared up the merge conflicts and updated the code to use `LispFrameOrSelected`.

rust_src/src/frames.rs
```diff
@@ -406,6 +408,28 @@ pub fn frame_bottom_divider_width(frame: LispFrameOrSelected) -> i32 {
...
+        | vertical_scroll_bar_type::vertical_scroll_bar_right => frame.config_scroll_bar_width,
+        _ => 0,
+    }
+}
+
+/// Return scroll bar height of FRAME in pixels.
+#[lisp_fn(min = "0")]
+pub fn frame_scroll_bar_height(frame: LispFrameOrSelected) -> i32 {
+    let frame: LispFrameRef = frame.into();
+    if frame.horizontal_scroll_bars() {
```
shaleh: The C code uses `FRAME_HAS_HORIZONTAL_SCROLL_BARS` to check if the window system is live or not. We should port that logic using the feature support then this function can be re-implemented simply.

shaleh: My recent change for LIBXML2 is a good guide for this. It is quite simple.

brotzeit:
This needs another review.

PR #948, opened 2018-10-23
Replace C style hash::lookup implementation with a more idiomatic Rust version.

shaleh:
Add a `HashLookupResult` which defines `Missing` and `Found`.

### DIFF
--- a/rust_src/src/hashtable.rs
+++ b/rust_src/src/hashtable.rs
@@ -21,6 +21,7 @@ pub enum HashLookupResult {
     Missing(EmacsUint),
     Found(isize),
 }
+use self::HashLookupResult::{Found, Missing};
 
 impl LispHashTableRef {
     pub fn allocate() -> LispHashTableRef {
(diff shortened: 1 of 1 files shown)


rust_src/src/hashtable.rs
```diff
@@ -207,12 +221,9 @@ pub fn copy_hash_table(mut table: LispHashTableRef) -> LispHashTableRef {
...
-    let idx = hash_table.lookup(key, ptr::null_mut());
-
-    if idx >= 0 {
-        hash_table.get_hash_value(idx)
-    } else {
-        dflt
+    match hash_table.lookup(key) {
+        Found(idx) => hash_table.get_hash_value(idx),
+        Missing(_) => dflt,
     }
```
db48x: I love this

shaleh: Yeah, I pointed @brotzeit at it and then realized how poor the UX was.

shaleh: I suppose we could get fancy and add `and_then` or `or_else` to it.

db48x:
Heh

PR #950, opened 2018-10-24
remove obsolete function eq_raw

brotzeit:


harryfei:
What about implementing the rust builtin `PartialEq` trait if we really need the `eq`?

brotzeit:
I asked this question a year ago, but there were objections because this way it's harder to distinguish between *ref types and LispObjects. I think this was the reason.

harryfei:
Ok, it make senses to me. Keep it now.

Issue #949, opened 2018-10-24
ld "multiple definition of" problems on Arch Linux


bb010g:
When attempting to `make` Remacs on Arch Linux, I get the following errors of the form ``/usr/bin/ld: $file.o:…/remacs/src/buffer.h:1425: multiple definition of `infile'; dispnew.o:…/remacs/src/buffer.h:1425: first defined here``:

<details><summary>Build log</summary>

```
bb010g ~/.../rust/remacs % make                                                                                                                                                                                                                                         (master)
make -C lib all
make[1]: Entering directory '/home/bb010g/Documents/rust/remacs/lib'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/home/bb010g/Documents/rust/remacs/lib'
make -C lib-src all
make[1]: Entering directory '/home/bb010g/Documents/rust/remacs/lib-src'
RUSTFLAGS= \
cargo build  --manifest-path=../rust_src/remacs-lib/Cargo.toml
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
make[1]: Leaving directory '/home/bb010g/Documents/rust/remacs/lib-src'
make -C src VCSWITNESS='$(srcdir)/../.git/logs/HEAD' all
make[1]: Entering directory '/home/bb010g/Documents/rust/remacs/src'
RUSTFLAGS="" \
[...]

PR #951, opened 2018-10-24
Fix issue #949.

harryfei:


nickdrozd:
I was running into that problem on Ubuntu 18. Tested locally, and this works.

PR #952, opened 2018-10-24
move methods from lisp.rs to related files

brotzeit:
lisp.rs is pretty bloated because of those methods. I would suggest to move them to the file where the type is defined. If there are no objections I will also move the rest of the methods.


brotzeit:
Do we really need numbers.rs AND floatfns.rs ? I just realized there is no numbers.c. I would make this one file ?

shaleh:
I like having the float support separate. It also has more macro magic.

shaleh:
@brotzeit check the rustformat version you have installed.

brotzeit:
flymake tests...

PR #955, opened 2018-10-25
More readme updating

brotzeit:


brotzeit:
You mean something similiar ? I would add a section about the progress and the midterm and longterm goals.

PR #965, opened 2018-10-27
[WIP] Port time-add & time-subtract

### LINKED ISSUE #932: Port time-add, time-subtract
From `editfns.c`

```c
DEFUN ("time-add", Ftime_add, Stime_add, 2, 2, 0,
       doc: /* Return the sum of two time values A and B, as a time value.
A nil value for either argument stands for the current time.
See `current-time-string' for the various forms of a time value.  */)
  (Lisp_Object a, Lisp_Object b)
{
  return time_arith (a, b, time_add);
}

DEFUN ("time-subtract", Ftime_subtract, Stime_subtract, 2, 2, 0,
       doc: /* Return the difference between two time values A and B, as a time value.
Use `float-time' to convert the difference into elapsed seconds.
A nil value for either argument stands for the current time.
See `current-time-string' for the various forms of a time value.  */)
  (Lisp_Object a, Lisp_Object b)
{
  return time_arith (a, b, time_subtract);
}
```

bb010g:
Currently causes a core dump during `make check`.

Fixes #932.

---
This change is [Reviewable](https://reviewable.io/reviews/wilfred/remacs/965)

### DIFF
--- a/rust_src/src/time.rs
+++ b/rust_src/src/time.rs
@@ -1,5 +1,6 @@
 //! Time support
 
+use std::cmp::max;
 use std::ptr;
 
 use libc::timespec as c_timespec;
@@ -7,12 +8,15 @@ use libc::{c_int, c_long, time_t};
 
 use remacs_lib::current_timespec;
 use remacs_macros::lisp_fn;
+use remacs_sys::Qnil;
 use remacs_sys::{lisp_time, EmacsInt};
 
-use lisp::defsubr;
-use lisp::LispObject;
+use lisp::{defsubr, ExternalPtr, LispObject};
 use lists::list;
 use numbers::MOST_NEGATIVE_FIXNUM;
+
+pub type LispTime = lisp_time;
+pub type LispTimeRef = ExternalPtr<LispTime>;
 
 const LO_TIME_BITS: i32 = 16;
 
@@ -131,7 +135,7 @@ pub unsafe extern "C" fn decode_time_components(
     low: LispObject,
     usec: LispObject,
     psec: LispObject,
-    result: *mut lisp_time,
+    mut result: LispTimeRef,
     dresult: *mut f64,
 ) -> c_int {
     let high = high;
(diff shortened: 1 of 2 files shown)


rust_src/src/time.rs
```diff
@@ -302,24 +313,26 @@ pub unsafe extern "C" fn lisp_time_struct(
...
+    error!("Invalid time specification")
 }
 
-/// Report that a time value is out of range for Emacs.
-fn time_overflow() -> ! {
-    error!("Specified time is not representable");
+/// Check a return value compatible with that of decode_time_components.
+#[no_mangle]
+pub extern "C" fn check_time_validity(validity: c_int) {
+    if validity <= 0 {
```
nickdrozd: This could be rewritten as a match on ordering.

rust_src/src/time.rs
```diff
@@ -1,19 +1,23 @@
...
 
-use lisp::defsubr;
-use lisp::LispObject;
+use lisp::{defsubr, ExternalPtr, LispObject};
 use lists::list;
 use numbers::MOST_NEGATIVE_FIXNUM;
 
+pub type LispTime = lisp_time;
+pub type LispTimeRef = ExternalPtr<LispTime>;
+
```
shaleh: :+1:

Wilfred:
@bb010g I've given you collaborator rights, so you can review other PRs if you wish :)

rust_src/src/time.rs
```diff
@@ -354,12 +367,90 @@ pub fn float_time(time: LispObject) -> LispObject {
...
+    let max_len = max(alen, blen);
+    let mut val = LispObject::cons(LispObject::from_C(t.ps as c_long), Qnil);
+    if max_len == 3 {
+        val = LispObject::cons(LispObject::from_C(t.us as c_long), val);
+    }
+    if max_len == 2 {
+        val = LispObject::cons(LispObject::from_C(t.lo as c_long), val);
+        val = LispObject::cons(LispObject::from_C(t.hi), val);
+    }
```
carlosmn: The `switch` originally is a bit confusing, but this is not what it would convert to. In the `default` case, where length is not 2 or 3 (which I take it to mean it's actually four or more since I don't think you can actually have just 1), you still *fall through* to the `case 3` and `case 2`. The `if`s should be reading `if max_len >= 3` etc and the "first" `ps` value here should be added in case where `max_len > 3`.

bb010g:
Closing in favor of #1157, which does this a lot more idiomatically.

PR #966, opened 2018-10-27
Port window-new-total to Rust.

### LINKED ISSUE #873: Port window-new-total
```c
DEFUN ("window-new-total", Fwindow_new_total, Swindow_new_total, 0, 1, 0,
       doc: /* Return the new total size of window WINDOW.

The new total size of WINDOW is the value set by the last call of
`set-window-new-total' for WINDOW.  If it is valid, it will be shortly
installed as WINDOW's total height (see `window-total-height') or total
width (see `window-total-width').  */)
  (Lisp_Object window)
{
  return decode_valid_window (window)->new_total;
}
```

AshtonKem:
Fixes #873.

### DIFF
--- a/test/rust_src/src/windows-tests.el
+++ b/test/rust_src/src/windows-tests.el
@@ -38,6 +38,10 @@
   (should (eq (window-old-point) 1))
   (should (eq (window-old-point (selected-window)) 1)))
 
+(ert-deftest window-new-total()
+  (should (eq (window-new-total) 9))
+  (should (eq (window-new-total (selected-window)) 9)))
+
 (ert-deftest window-use-time ()
   (let ((use-time (window-use-time)))
     (should (eq 'integer (type-of use-time)))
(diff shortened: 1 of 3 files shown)


test/rust_src/src/windows-tests.el
```diff
@@ -38,6 +38,10 @@
   (should (eq (window-old-point) 1))
   (should (eq (window-old-point (selected-window)) 1)))
```
shaleh: Awesome. Tests rock.

Why 9? Can we use a variable or at a minimum have a comment documenting the choice

AshtonKem:
I set it to 1 and ran it, and the test failed because the correct value was 9. I assume that there might be some minimum buffer height value somewhere, if I can find it I’ll reference it in the test.

test/rust_src/src/windows-tests.el
AshtonKem: That indeed does work, and is at least a little less magical than hard-coding a 9. It looks like the relationship between new-total and total-height is a bit more complicated than this call would imply, but that's probably well out of the scope of this MR.

shaleh: Agreed.

PR #977, opened 2018-10-28
update readme

brotzeit:


README.md
```diff
@@ -213,24 +215,11 @@ You can now run your shiny new Remacs build!
 $ RUST_BACKTRACE=1 src/remacs -q
 ```
 
-## Design Goals
+# Design Goals
 
 **Compatibility**: Remacs should not break existing elisp code, and
 ideally provide the same FFI too.
 
-**Similar naming conventions**: Code in Remacs should use the same
```
brotzeit: I moved this section to the wiki.

PR #979, opened 2018-10-30
Port internal-show-cursor{,-p}

### LINKED ISSUE #974: Port internal-show-cursor-p
```c
DEFUN ("internal-show-cursor-p", Finternal_show_cursor_p,
       Sinternal_show_cursor_p, 0, 1, 0,
       doc: /* Value is non-nil if next redisplay will display a cursor in WINDOW.
WINDOW nil or omitted means report on the selected window.  */)
  (Lisp_Object window)
{
  return decode_any_window (window)->cursor_off_p ? Qnil : Qt;
}
```

bb010g:
Closes #974.

---
This change is [Reviewable](https://reviewable.io/reviews/wilfred/remacs/979)

### DIFF
--- a/rust_src/src/dispnew.rs
+++ b/rust_src/src/dispnew.rs
@@ -7,13 +7,13 @@ use remacs_macros::lisp_fn;
 use remacs_sys::{clear_current_matrices, dtotimespec, fset_redisplay,
                  mark_window_display_accurate, timespec_add, timespec_sub,
                  wait_reading_process_output};
+use remacs_sys::{redisplaying_p, Qnil, Vframe_list, WAIT_READING_MAX};
 use remacs_sys::{EmacsDouble, EmacsInt, Lisp_Glyph};
-use remacs_sys::{Qnil, Vframe_list, WAIT_READING_MAX};
 
 use frames::{frame_live_or_selected, LispFrameRef};
 use lisp::{defsubr, ExternalPtr, LispObject};
 use terminal::{clear_frame, update_begin, update_end};
-use windows::LispWindowRef;
+use windows::{window_or_selected, LispWindowRef};
 
 pub type LispGlyphRef = ExternalPtr<Lisp_Glyph>;
 
@@ -103,6 +103,30 @@ pub extern "C" fn set_window_update_flags(w: LispWindowRef, on_p: bool) {
             Some(next.as_window_or_error())
         };
     }
+}
+
+/***********************************************************************
+			   Blinking cursor
+ ***********************************************************************/
+
+/// Set the cursor-visibility flag of WINDOW to SHOW.
+/// WINDOW nil means use the selected window.  SHOW non-nil means
+/// show a cursor in WINDOW in the next redisplay.  SHOW nil means
+/// don't show a cursor.
+#[lisp_fn]
+pub fn internal_show_cursor(window: LispObject, show: LispObject) {
+    // Don't change cursor state while redisplaying.  This could confuse
+    // output routines.
+    if !unsafe { redisplaying_p } {
+        window_or_selected(window).set_cursor_off_p(show.is_nil())
+    }
+}
+
+/// Value is non-nil if next redisplay will display a cursor in WINDOW.
+/// WINDOW nil or omitted means report on the selected window.
+#[lisp_fn]
+pub fn internal_show_cursor_p(window: LispObject) -> bool {
+    !window_or_selected(window).cursor_off_p()
 }
 
 include!(concat!(env!("OUT_DIR"), "/dispnew_exports.rs"));
--- a/src/dispnew.c
+++ b/src/dispnew.c
@@ -5957,35 +5957,6 @@ init_display (void)
 
 
 /***********************************************************************
-			   Blinking cursor
- ***********************************************************************/
-
-DEFUN ("internal-show-cursor", Finternal_show_cursor,
-       Sinternal_show_cursor, 2, 2, 0,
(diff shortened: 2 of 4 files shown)


src/dispnew.c
```diff
@@ -5956,35 +5956,6 @@ init_display (void)
...
-DEFUN ("internal-show-cursor", Finternal_show_cursor,
-       Sinternal_show_cursor, 2, 2, 0,
-       doc: /* Set the cursor-visibility flag of WINDOW to SHOW.
-WINDOW nil means use the selected window.  SHOW non-nil means
-show a cursor in WINDOW in the next redisplay.  SHOW nil means
-don't show a cursor.  */)
-  (Lisp_Object window, Lisp_Object show)
-{
-  /* Don't change cursor state while redisplaying.  This could confuse
-     output routines.  */
```
brotzeit: Why drop the comment ?

rust_src/src/dispnew.rs
```diff
@@ -105,4 +105,20 @@ pub extern "C" fn set_window_update_flags(w: LispWindowRef, on_p: bool) {
...
 
+/***********************************************************************
+			   Blinking cursor
+ ***********************************************************************/
+
+#[lisp_fn]
+pub fn internal_show_cursor(window: LispObject, show: LispObject) {
+    if !unsafe { redisplaying_p } {
+        window_or_selected(window).set_cursor_off_p(show.is_nil())
+    }
```
brotzeit: This function returns `Qnil`.

src/window.c
```diff
@@ -301,19 +301,6 @@ decode_live_window (register Lisp_Object window)
...
-{
-  struct window *w;
-
-  if (NILP (window))
-    return XWINDOW (selected_window);
-
-  CHECK_WINDOW (window);
-  w = XWINDOW (window);
-  return w;
-}
```
brotzeit: This function is still used I think.

src/window.c
```diff
@@ -301,19 +301,6 @@ decode_live_window (register Lisp_Object window)
   return XWINDOW (window);
 }
 
-struct window *
-decode_any_window (register Lisp_Object window)
```
brotzeit: This function is still used I think.

rust_src/src/dispnew.rs
shaleh: Making no return is the equivalent. I prefer the explicit return too.

src/window.c
shaleh: It is ported to Rust now.

src/window.c
```diff
@@ -301,19 +301,6 @@ decode_live_window (register Lisp_Object window)
...
-  struct window *w;
-
-  if (NILP (window))
-    return XWINDOW (selected_window);
-
-  CHECK_WINDOW (window);
-  w = XWINDOW (window);
-  return w;
-}
-
```
shaleh: Want to add `decode_valid_window` while here?

rust_src/src/dispnew.rs
brotzeit: Right, we discussed this recently.

src/dispnew.c
bb010g: Sorry, forgot to copy those over.

shaleh: I wrote https://gist.github.com/shaleh/c84d905b0bc755163beea8269be00f1d to help. `M-| rustize-emacs`. It needs some tweaks to catch all of the C code styles but it is a start.

rust_src/src/dispnew.rs
bb010g: What's the point of the explicit return if it's always returning `Qnil` so it can just return something? That reasoning is why it's elided with `()` in Rust—something has to be returned, but it's just noise (unless it's a sum arm, but you'd use `LispObject` for that).

shaleh: Clarity. I am not a fan of magic in my code personally. I'd be ok with the ()/Qnil case being dropped from `lisp_fn`.

If our mission is having a project open to new developers and maybe casual coders, the more obvious our code is the better off we will be.

bb010g: If we're trying to be anti-magic here, shouldn't we be using `-> ()` on all Rust functions, then? I don't see how a mental continuation of Rust functions that always return values into Lisp functions is broken here. If someone working on Remacs doesn't know that "void" Rust functions actually return unit, I think it's more valuable to tell them that in the porting guide while noting how "void" Lisp functions are declared and written than to clutter up the code with trailing `Qnil`s, like C has to do, and Rust almost had. This way, the existing pragmatic clarity of Rust's implicit unit return is kept while Rust code they may run into in the future that deals with the output of "void" functions is made less surprising. Additionally, this means that explicit `Qnil` returns are more likely to denote the termination of a list, and not just void, better conveying intent with Rust's type system.

brotzeit: @bb010g I think you're right. `lisp_fn` already contains a lot of magic and newcomers already have to take a look at the wiki. But we are using `-> !` though I don't know why it's not `-> ()`. Doesn't really matter I guess.

shaleh: `-> !` means "function never returns to caller" not "does not return a value". Anything that returns '!' is an executional dead end and the compiler can use this information.

`-> ()` means 'does not return a value'.

brotzeit: Not my day. `-> ()` seems good.

harryfei:
I think we could merge this. @shaleh

PR #980, opened 2018-10-30
Port internal_equal C function.

shaleh:


shaleh:
The bool vector check is failing. Not sure why, I have just started adding debugging to point the way. It looks like the conversion to A Vector pointer is off.

The Cons section is also not working correctly. This might be a flaw in the iterator?

I have not had the time to finish this properly but it is pretty close.

`internal_equal` is now in Rust. The original C is left behind as `reference_internal_equal` and is handling the broken sections above.

db48x:
I added some more printing; it fails to compare conses which aren't proper lists:

    cons: ('require . 'easy-mmode) vs ('require . 'pcase)
    Wrong type argument: listp, (require . easy-mmode)

shaleh:
I am in the process of pulling this PR apart.

PR #982, opened 2018-10-30
Port memory profiler defuns

nickdrozd:


rust_src/src/profiler.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// Start/restart the memory profiler.
+/// The memory profiler will take samples of the call-stack whenever a new allocation takes
+/// place. Note that most small allocations only trigger the profiler occasionally.
+/// See also `profiler-log-size' and `profiler-max-stack-depth'.
+#[lisp_fn]
+pub fn profiler_memory_start() -> bool {
+    unsafe {
+        if profiler_memory_running {
```
shaleh: Another inspiration for a `new_string!` macro.

rust_src/src/profiler.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    true
+}
+
+/// Stop the memory profiler.  The profiler log is not affected.
+/// Return non-nil if the profiler was running.
+#[lisp_fn]
+pub fn profiler_memory_stop() -> bool {
+    unsafe {
```
shaleh: Let's not alias `running` here. This way start and stop match.

rust_src/src/profiler.rs
```diff
@@ -0,0 +1,75 @@
...
+            error(error_str.as_ptr() as *const u8);
+        }
+
+        if memory_log.is_nil() {
+            memory_log = make_log(globals.profiler_log_size, globals.profiler_max_stack_depth)
+        }
+
+        profiler_memory_running = true;
+    }
+
```
shaleh: Maybe return `profiler_memory_running`? To match stop?

rust_src/src/profiler.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// Return the current memory profiler log.
+/// The log is a hash-table mapping backtraces to counters which represent the amount of memory
+/// allocated at those points. Every backtrace is a vector of functions, where the last few
+/// elements may be nil.
+/// Before returning, a new log is allocated for future samples.
+#[lisp_fn]
+pub fn profiler_memory_log() -> LispObject {
+    unsafe {
```
shaleh: The intent is to return the previous log so, maybe we should call this 'previous'?

rust_src/src/profiler.rs
shaleh: Sadly not.

shaleh:
@nickdrozd try the `new_string` macro next. That should be a good one.

PR #983, opened 2018-10-31
Use LispNumber for coerced markers.

shaleh:
This continues the idea from LispBufferOrName.


shaleh:
The `map_or` on LispObject is an experiment. Let's see how it works out. There are more places it would be useful in this PR.

PR #984, opened 2018-10-31
port message-box

brotzeit:


nickdrozd:
Ha, I had this one on my todo list for today!

Is it possible to run it in a test?

nickdrozd:
@shaleh Is there a reason we're using `CString::new` etc instead of `"OK".as_ptr() as *const ::libc::c_char`? That seems to work just as well.

brotzeit:
@nickdrozd I will add tests.

nickdrozd:
> @shaleh Is there a reason we're using `CString::new` etc instead of `"OK".as_ptr() as *const ::libc::c_char`? That seems to work just as well.

Actually it does not work just as well.

shaleh:
@nickdrozd keep in mind _all_ allocations that are visible to the C code are made in the C code. We are bound by the Lisp Garbage Collector which we have not ported yet.

brotzeit:
How can I access the message box contents ?

harryfei:
LGTM

shaleh:
I don't know that you can @brotzeit

PR #985, opened 2018-10-31
port copy-syntax-table

brotzeit:
I don't know how to handle `Vstandard_syntax_table`.

### DIFF
--- a/rust_src/src/syntax.rs
+++ b/rust_src/src/syntax.rs
@@ -1,7 +1,7 @@
 //! Functions related to syntax
 
 use remacs_macros::lisp_fn;
-use remacs_sys::{buffer_defaults, check_syntax_table, scan_lists, set_char_table_defalt};
+use remacs_sys::{buffer_defaults, scan_lists, set_char_table_defalt};
 use remacs_sys::{EmacsInt, Qnil, Qsyntax_table, Qsyntax_table_p};
 use remacs_sys::{Fcopy_sequence, Fset_char_table_parent};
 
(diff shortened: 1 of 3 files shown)


rust_src/src/remacs_sys.rs
```diff
@@ -75,6 +75,7 @@ pub enum BoolVectorOp {
 extern "C" {
     // these weren't declared in a header, for example
     pub static Vprocess_alist: Lisp_Object;
+    pub static Vstandard_syntax_table: Lisp_Object;
```
db48x: This is wrong. There's a `syntax_table` field on every `buffer`, and `Vstandard_syntax_table` is a macro that refers to the `syntax_table` field on a particular `buffer` struct which was set aside for the purpose of holding the default values for this and other fields. See https://github.com/Wilfred/remacs/blob/master/src/buffer.h#L1086-L1093

shaleh: https://github.com/Wilfred/remacs/blob/cdf7a08ef6a76a220661b9e6223d547780b825b6/src/syntax.h#L34

You could write a Rust macro to keep the idea.

rust_src/src/remacs_sys.rs
```diff
@@ -154,6 +155,7 @@ extern "C" {
         dest: Lisp_Object,
         op: BoolVectorOp,
     ) -> Lisp_Object;
```
shaleh: You should not need to manually add functions. Instead, make sure the function is exported in a C header.

brotzeit:
Should be ok.

rust_src/src/syntax.rs
```diff
@@ -65,4 +67,29 @@ pub fn standard_syntax_table() -> LispCharTableRef {
     unsafe { buffer_defaults.syntax_table_ }.into()
 }
 
+/// Construct a new syntax table and return it.
+/// It is a copy of the TABLE, which defaults to the standard syntax table.
+#[lisp_fn(min = "0")]
+pub fn copy_syntax_table(mut table: LispObject) -> LispObject {
+    let buffer_table = unsafe { buffer_defaults.syntax_table_ };
+    if table.is_not_nil() {
```
shaleh: Why not include the port of `check_syntax_table`?

shaleh: You can leave it off for now if you like.

PR #986, opened 2018-11-01
Port standard-syntax-table.

jaccarmac:
Address #969.

I'm watching #985 with great interest as well.

### DIFF
--- a/rust_src/src/syntax.rs
+++ b/rust_src/src/syntax.rs
@@ -1,6 +1,6 @@
 //! Functions related to syntax
 use remacs_macros::lisp_fn;
-use remacs_sys::scan_lists;
+use remacs_sys::{buffer_defaults, scan_lists};
 use remacs_sys::{EmacsInt, Qsyntax_table, Qsyntax_table_p};
 
 use chartable::LispCharTableRef;
@@ -57,6 +57,13 @@ fn check_syntax_table_p(table: LispCharTableRef) {
     if table.purpose != Qsyntax_table {
         wrong_type!(Qsyntax_table_p, LispObject::from(table))
     }
+}
+
+/// Return the standard syntax table.  This is the one used for new buffers.
+#[lisp_fn]
+pub fn standard_syntax_table() -> LispCharTableRef {
+    let buffer_table = unsafe { buffer_defaults.syntax_table_ };
+    buffer_table.as_char_table_or_error()
 }
 
 include!(concat!(env!("OUT_DIR"), "/syntax_exports.rs"));
--- a/rust_src/src/syntax.rs
+++ b/rust_src/src/syntax.rs
@@ -62,8 +62,7 @@ fn check_syntax_table_p(table: LispCharTableRef) {
 /// Return the standard syntax table.  This is the one used for new buffers.
 #[lisp_fn]
 pub fn standard_syntax_table() -> LispCharTableRef {
-    let buffer_table = unsafe { buffer_defaults.syntax_table_ };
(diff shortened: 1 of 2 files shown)


jaccarmac:
Trying to get back into the swing of things after not contributing in forever. Does the hubbub around Vstandard_syntax_table have anything to do with the new variable stuff?

rust_src/src/syntax.rs
```diff
@@ -59,4 +59,11 @@ fn check_syntax_table_p(table: LispCharTableRef) {
     }
 }
 
+/// Return the standard syntax table.  This is the one used for new buffers.
+#[lisp_fn]
+pub fn standard_syntax_table() -> LispCharTableRef {
+    let buffer_table = unsafe { buffer_defaults.syntax_table_ };
```
jaccarmac: I don't particularly like this but it seems to be the only good way to do this. Also prefer the type "conversion" over returning raw LispObject, but I see in other places with direct returns of variables of known types Rust just sees a LispObject. I suppose this is a problem that will slowly become less of one as more C turns into Rust?

shaleh:
> Trying to get back into the swing of things after not contributing in forever. Does the hubbub around Vstandard_syntax_table have anything to do with the new variable stuff?

No. In the C it is a macro to read a member of struct buffer. It was ported as an object incorrectly.

rust_src/src/syntax.rs
shaleh: As I mentioned on brotzeit's PR, perhaps representing `Vstandard_syntax_table` as `standard_syntax_table!` is a reasonable choice.

shaleh: Simplified:

```
/// Return the standard syntax table.  This is the one used for new buffers.
#[lisp_fn]
pub fn standard_syntax_table() -> LispCharTableRef {
    unsafe { buffer_defaults.syntax_table_ }.into()
}
```

jaccarmac: Ooh nice, I'll change that. Thanks!

Re: the macro. In that case we would need a macro for every member of buffer_defaults, as with C. Would it be worthwhile to create a wrapper for buffer_defaults without underscore tails? Creating a full-on equivalent to the C macro system is possible of course, just not sure exactly what it gets us.

(I'm not familiar with the Vx variables in Emacs, maybe there's some uniformity between them i am missing?)

shaleh: I have not researched the C code enough to have an opinion yet on this.

In general, I like having easily greppable names. But there is a cost v. benefit analysis for sure.
`buffer_defaults.something...` is not as obvious as `standard_something`.

jaccarmac: Fair enough, and easy enough to add to `remacs_sys` or somesuch. What do you think is the best way to deal with `unsafe` in that case? Leave the caller to mark `unsafe` as appropriate?

shaleh: Usually best.

Also, remas-sys twiddling should not be required for most things any more. If you need something from C make sure it is exported from a C header. bindgen will update the Rust side for you on the next build.

shaleh: There is no need for `BVAR` in the current code. Turn the LispObject into a LispBufferRef and then directly access the field. bindgen has setup the C structs for us.

shaleh: The advantage of `Vstandard_syntax_table` is it abstracts away the buffer and the C struct member name. So it is obviously "getting the standard syntax table" and not "um, accessing a field on a buffer. What does that field do....."

PR #987, opened 2018-11-01
[WIP] Port font-match-p.

jaccarmac:
Address #964.

### DIFF
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -3,7 +3,7 @@
 use remacs_macros::lisp_fn;
 use remacs_sys::font_match_p as c_font_match_p;
 use remacs_sys::{pvec_type, FONT_ENTITY_MAX, FONT_OBJECT_MAX, FONT_SPEC_MAX};
-use remacs_sys::{EmacsInt, Qfont_entity, Qfont_object, Qfont_spec, Qfont};
+use remacs_sys::{EmacsInt, Qfont, Qfont_entity, Qfont_object, Qfont_spec};
 
 use lisp::defsubr;
 use lisp::LispObject;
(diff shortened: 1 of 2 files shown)


rust_src/src/fonts.rs
```diff
@@ -112,4 +113,13 @@ pub fn fontp(object: LispObject, extra_type: LispObject) -> bool {
     })
 }
 
+/// Return t if and only if font-spec SPEC matches with FONT.  FONT is a font-spec, font-entity,
+/// or font-object.
+#[lisp_fn]
+pub fn font_match_p(spec: LispObject, font: LispObject) -> bool {
+    if !spec.is_font_spec() {wrong_type!(Qfont_spec, spec)}
```
jaccarmac: Is there a better pattern for here? I held off implementing anything bigger since I saw similar patterns elsewhere, but I seem to remember some kind of From impl for Option letting me automagically use types in my function signatures?

rust_src/src/fonts.rs
```diff
@@ -112,4 +113,17 @@ pub fn fontp(object: LispObject, extra_type: LispObject) -> bool {
     })
 }
 
+/// Return t if and only if font-spec SPEC matches with FONT.  FONT is a font-spec, font-entity,
+/// or font-object.
+#[lisp_fn]
+pub fn font_match_p(spec: LispObject, font: LispObject) -> bool {
```
shaleh: We could create `LispFontSpec` and `LispFontSomething`. I wrote 'something' since it is a font related object not necessarily a font.

shaleh: Might not be worth it. That depends on how many uses there are.

jaccarmac: Something similar already exists in that file. Can you clarify how your suggestion is different from what's there?

shaleh: if it exists, use it :-)

jaccarmac: Yeah, I'll work on the `From` impls there. The structs are a bit different than the ones elsewhere in Remacs which I didn't pick up on yesterday.

PR #999, opened 2018-11-03
upgrade rust toolchain into 2018-11-03

harryfei:


harryfei:
I reformat the code already. However, this CI is still failed. It's really odd.

harryfei:
I found the root cause. I forget to checkout from latest master when I do this change.

shaleh:
Too new for me on MacOS.

> toolchain 'nightly-2018-11-03-x86_64-apple-darwin' does not have the binary `rustfmt`

harryfei:
Just for confirming, did you install it manually by `rustup component add rustfmt-preview`?

shaleh:
No. I ran `rustup update`.

harryfei:
I found that we need to install the `rustfmt` manually everytime we upgrade the toolchain in `remacs`.

jaccarmac:
The rustfmt reinstall is a minor annoyance but perhaps worth moving the install into a bootstrap script? It's already in CI after all.

PR #1002, opened 2018-11-03
Port buffer string

krgn:
This PR addresses #875


harryfei:
You should use `make rustfmt` to format rust code.

krgn:
> You should use `make rustfmt` to format rust code.

done!

shaleh:
@krgn there is a rust-mode setting which runs rustfmt whenever you save the buffer.

krgn:
> @krgn there is a rust-mode setting which runs rustfmt whenever you save the buffer.

Yes, I have this set up actually, though in rustic-mode. I did some of the editing in vim though, where this didn't kick-in :laughing:

PR #1009, opened 2018-11-04
Merge f1204e

db48x:


db48x:
The error in the Linux build is `Symbol's function definition is void: gnutls-macs`. This didn't make much sense to me at first, but it occurred to me over lunch that because this function is defined in C, it could only turn out to be undefined if there were an `#ifdef` that removed it. As it happens, there is: https://github.com/Wilfred/remacs/blob/master/src/gnutls.c#L1849. Apparently we're not building with gnutls 3.x? It's plausible, but this check existed before this merge, and it built fine then.

shaleh:
I am fairly sure we use 2.x currently.

shaleh:
On my Mac, the gnupg test does not fail. flymake is the only consistent failure.
That one is weird because it passes when I recreate the elements of the test manually. It only fails when running in the test.

rust_src/src/process.rs
```diff
@@ -383,7 +383,7 @@ pub fn process_send_string(process: LispObject, mut string: LispStringRef) -> ()
     unsafe {
         send_process(
             cget_process(process),
-            string.data as *mut libc::c_char,
```
shaleh: In other places we have used `as_slice` for this kind of behavior. Should we also do that here?

db48x:
Thanks!

PR #1017, opened 2018-11-05
[#1016] port delete-frame

krgn:


shaleh:
It can feel a little silly to convert then from LispObject -> Rust type -> LispObject. Remember that this ugliness is temporary.

`delete_frame(frame.into(), force.into())`

brotzeit:
Btw you can [close issues automatically](https://help.github.com/articles/closing-issues-using-keywords/).

krgn:
I just pushed some changes and tested the build. Let me know what you think.

rust_src/src/frames.rs
```diff
@@ -348,4 +349,23 @@ pub fn frame_text_height(frame: LispObject) -> LispObject {
...
+/// FRAME must be a live frame and defaults to the selected one.
+///
+/// A frame may not be deleted if its minibuffer serves as surrogate
+/// minibuffer for another frame.  Normally, you may not delete a frame if
+/// all other frames are invisible, but if the second optional argument
+/// FORCE is non-nil, you may do so.
+///
+/// This function runs `delete-frame-functions' before actually
+/// deleting the frame, unless the frame is a tooltip.
+/// The functions are run with one argument, the frame to be deleted.
```
shaleh: I believe this is an issue because Rust will expose a `delete_frame` and C also exposes a `delete_frame`.
We usually name the `lisp_fn` `delete_frame_lisp` in this case.

If the C function `delete_frame` is simple enough, port it over and then the problem is gone.

krgn: Ah, I looked into that function, and its a very involved bit of code, hence I aliased the C function as `c_delete_frame` in Rust. It should pass tests now at least.

krgn: Oh, I forgot to add - since there may have been a misunderstanding on my part - I did test this works, manually. Should I still rename the function to `delete_frame_lisp`?

shaleh: Our goal is usually to have the Rust callable by C. While this does compile I am concerned that it might confused the reader. Let's use the `_lisp` name until we can port over the C code.

shaleh: Also, mind @db48x's request. The `into` calls are useless because you and converting LispObject to..... LispObject.

krgn: I just finished the changes @db48x requested

PR #1018, opened 2018-11-06
port select-window

ebpa:
@nickdrozd gave me an intro to porting functions to rust for remacs (thanks! ^_^)

Rust changes have been formatted with rustfmt.

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -1,14 +1,13 @@
 //! Generic frame functions.
 
 use remacs_macros::lisp_fn;
-use remacs_sys::{delete_frame as c_delete_frame, frame_dimension, output_method, Fcons,
-                 Fselect_window};
+use remacs_sys::{delete_frame as c_delete_frame, frame_dimension, output_method, Fcons};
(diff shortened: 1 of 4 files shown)


rust_src/src/frames.rs
```diff
@@ -181,7 +181,7 @@ pub fn set_frame_selected_window(
         error!("In `set-frame-selected-window', WINDOW is not on FRAME")
     }
     if frame_ref == selected_frame().as_frame().unwrap() {
-        unsafe { Fselect_window(window, norecord) }
+        select_window(window, norecord)
```
shaleh: This should be a call to `select_window_lisp`.

ebpa: @shaleh You prefer that stylistically over the `select_window_lisp as select_window` alias in the use declaration?  I can dig that.  I'll update the declaration.

shaleh: Code is meant to be read as much as it is executed. Imagine someone 30 years from now porting this from Rust to AmazeBalls3000 (the preferred language of the next generation). They will be scratching their head over the multiple `select_window`s.

PR #1021, opened 2018-11-06
Add LispWindowOrSelected, LispWindowLiveOrSelected, and LispWindowValidOrSelected.

shaleh:


shaleh:
If this approach is welcomed then applying the same logic to Frame is next.

PR #1023, opened 2018-11-07
port font-at

brotzeit:


rust_src/src/fonts.rs
```diff
@@ -129,4 +133,43 @@ pub fn font_match_p(spec: LispObject, font: LispObject) -> bool {
     unsafe { c_font_match_p(spec, font) }
 }
 
+/// Return a font-object for displaying a character at POSITION.
+/// Optional second arg WINDOW, if non-nil, is a window displaying
+/// the current buffer.  It defaults to the currently selected window.
+/// Optional third arg STRING, if non-nil, is a string containing the target
+/// character at index specified by POSITION.
+#[lisp_fn(min = "1", c_name = "font_at", name = "font-at")]
+pub fn font_at_lisp(
```
shaleh: `position` can be `LispNumber` no?

shaleh: `string` can be `Option<LispStringRef>` right?

brotzeit: position can't be `LispNumber` because of the coerce marker check. But I have to move the `position.as_fixnum_or_error()`. But you're right with string.

shaleh: See #983

shaleh: Because of the C `font_at` we can leave this be for now. Eventually, we can use the right types.

brotzeit:
I don't see the problem

```
Test test-font-at condition:
    (ert-test-failed
     ((should
       (eq
	(font-at ... win)
	(face-attribute ... :font)))
      :form
      (eq nil unspecified)
      :value nil))
   FAILED  2/2  test-font-at
```

brotzeit:
It works when I run this manually in emacs. I guess it has something to do with running without interactive terminal ?

shaleh:
Yeah the failing flymake test also works when I run it but fails when make does.

shaleh:
Merging this breaks CI. So you also need to disable the test.

brotzeit:
Should I comment it out with a note or delete it completely ?

rust_src/src/fonts.rs
```diff
@@ -143,4 +147,45 @@ pub fn find_font(spec: LispObject, frame: LispObject) -> LispObject {
...
+    position: LispObject,
+    window: LispWindowLiveOrSelected,
+    string: LispObject,
+) -> LispObject {
+    let mut w: LispWindowRef = window.into();
+    let cur_buf = ThreadState::current_buffer();
```
shaleh: This is likely wrong after the recent changes.

rust_src/src/fonts.rs
```diff
@@ -143,4 +147,45 @@ pub fn find_font(spec: LispObject, frame: LispObject) -> LispObject {
...
+
+    let pos = if let Some(s) = string.as_string() {
+        let pos = position.as_fixnum_or_error() as isize;
```
shaleh: @nickdrozd has been migrating us to `into()` and `from()` calls.
```suggestion
        let pos = EmacsInt::from(position) as isize;
```

rust_src/src/fonts.rs
```diff
@@ -143,4 +147,45 @@ pub fn find_font(spec: LispObject, frame: LispObject) -> LispObject {
...
+#[lisp_fn(min = "1", c_name = "font_at", name = "font-at")]
+pub fn font_at_lisp(
+    position: LispObject,
+    window: LispWindowLiveOrSelected,
+    string: LispObject,
+) -> LispObject {
+    let mut w: LispWindowRef = window.into();
+    let cur_buf = ThreadState::current_buffer();
+
+    let pos = if let Some(s) = string.as_string() {
```
shaleh: This should likely be a `match` not a `if let`.

rust_src/src/fonts.rs
```diff
@@ -143,4 +147,45 @@ pub fn find_font(spec: LispObject, frame: LispObject) -> LispObject {
...
+        if !(0 <= pos && pos < s.len_bytes()) {
+            args_out_of_range!(string, position);
+        }
+        pos
+    } else {
+        if w.contents != cur_buf.into() {
+            error!("Specified window is not displaying the current buffer");
+        }
+        position.as_number_coerce_marker_or_error();
+        let pos = position.as_fixnum_or_error() as isize;
```
shaleh:
```suggestion
        let pos = EmacsInt::from(position) as isize;
```

brotzeit:
Ready.

PR #1024, opened 2018-11-07
fix delete_frame_lisp

brotzeit:


rust_src/src/frames.rs
```diff
@@ -362,9 +362,9 @@ pub fn frame_text_height(frame: LispObject) -> LispObject {
 /// deleting the frame, unless the frame is a tooltip.
 /// The functions are run with one argument, the frame to be deleted.
 #[lisp_fn(min = "0", name = "delete-frame", c_name = "delete_frame")]
```
shaleh: The error was choosing Window instead of Frame. Why revert back to LispObject?

brotzeit: `delete-frame` takes 
```c
Lisp_Object
delete_frame (Lisp_Object frame, Lisp_Object force)
```

brotzeit:
```c
Lisp_Object
delete_frame (Lisp_Object frame, Lisp_Object force)
```

brotzeit: That's the bug you we're talking about in gitter.

shaleh: Because the `delete_frame` C function actually takes a `LispFrameLiveOrSelected`. Window was definitely wrong. But `LispFrameRef` is just as wrong. Even `Option<LispFrameRef>` is not fully sufficient.

PR #1025, opened 2018-11-07
Cut unnecessary nil returns

nickdrozd:
As I understand it, the lisp_fn macro converts Rust return values to
Lisp values. In particular, the unit () is converted to nil. Thus a
function that returns only Qnil might as well return nothing at all,
since the ultimate Lisp function will return nil anyway.

### DIFF
--- a/rust_src/src/casefiddle.rs
+++ b/rust_src/src/casefiddle.rs
@@ -16,7 +16,7 @@ use threads::ThreadState;
 use obarray::intern;
 use symbols::symbol_value;
 
-fn casify_word(flag: case_action, words: EmacsInt) -> LispObject {
+fn casify_word(flag: case_action, words: EmacsInt) {
     let buffer_ref = ThreadState::current_buffer();
 
     let far_end = match unsafe { scan_words(buffer_ref.pt, words) } {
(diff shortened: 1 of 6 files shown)


rust_src/src/casefiddle.rs
```diff
@@ -16,7 +16,7 @@ use threads::ThreadState;
 use obarray::intern;
 use symbols::symbol_value;
```
shaleh: `casify_word` is not a `lisp_fn`. Leave the useless return.

shaleh: Ok, nevermind. It looks like the return was only added to satisfy the DEFUN return requirement.

nickdrozd:
My general thinking that is that the Rust input and output types should be cinched down as tight as possible so as to enjoy the benefits of Rust compiler checks. Basically any `LispObject` should be considered not specific enough unless it has to be that. Is the object expected to be a cons? Or a symbol? Or what?

We've been doing this to some extent already. Is it worthwhile to go through and fix more instances of this? This PR takes care of the case with outputs of unit type `()`.

db48x:
Yes, I think it's worth it. At the moment we have a mixture primarily because the capabilities of the `lisp_fn` macro have been improved over time.

shaleh:
Agreed.

PR #1027, opened 2018-11-08
Port buffer, marker, and overlay bits.

shaleh:


shaleh:
@db48x this should be ready now.

brotzeit:
Can you split up your next PR that's like this. Reviewing small PRs is a lot easier.
Another question. Is it ok for you if we split big impls again. I think you can get a better overview of methods that belong together if they have their own impl. It's also really easy to find impls for the same type.

shaleh:
The `LispBufferRef` impl is kinda large, but not much we can do about it. There are many methods.

brotzeit:
We could use more impls than one.

shaleh:
What does breaking the `impl` block into pieces in the same file buy us? Pulling the `LispObject` impl apart and putting the pieces in the files near their match code was the right decision and closer matches Rust style. But the Buffer impl is still going to be right where it is.

brotzeit:
The impl is 300 lines long now. Even if we don't put it in different files it's easier to get an overview. This way I can look at the comment of an impl and ignore it if it doesn't contain what I'm looking for. I think it would at least make sense to group methods that only access struct fields without any logic. It's not important, but it certainly makes things easier for newcomers.

shaleh:
Is this common in other large Rust projects? I come from C and C++. Or functional languages. In C++ a class is as long as it is. TAGS helps. Similar for functional languages. A module has all of the functions its needs.

Personally, I find constantly hitting another `impl Blah` for the same type annoying. It feels like it is repeated boiler plate for no gain.

Grouping within the `impl` is a good idea and we can certainly do better. We can also put some block comments in.

shaleh:
@brotzeit I have minimal exposure to other big Rust projects. if this is a cultural idiom I am not familiar with, I can adapt.

brotzeit:
I don't know if that's the case.

shaleh:
Ok, bite sized pieces.

PR #1030, opened 2018-11-09
don't use bool for lisp_fn arguments

### LINKED ISSUE #1029: Cursor doesn't blink
When I start Remacs with `-Q`, the cursor blinks once and then stays dark, that is, it doesn't continue blinking.

brotzeit:
close #1029 

I think the problem with this is that elisp only knows nil and not nil.


shaleh:
This was the right fix the wrong way.

The problem was not that we used a `bool` the problem is we did not invert it. The code was looking for `!show` which is what the `is_nil` check is doing.

PR #1031, opened 2018-11-09
Revert "don't use bool for lisp_fn arguments"

shaleh:
Reverts Wilfred/remacs#1030

The problem was I did not invert the boolean. Tossing out the bool and going back to LispObject solves the problem but is heavy handed.

### DIFF
--- a/rust_src/src/dispnew.rs
+++ b/rust_src/src/dispnew.rs
@@ -114,12 +114,12 @@ pub extern "C" fn set_window_update_flags(w: LispWindowRef, on_p: bool) {
 /// show a cursor in WINDOW in the next redisplay.  SHOW nil means
 /// don't show a cursor.
 #[lisp_fn]
-pub fn internal_show_cursor(window: LispWindowOrSelected, show: LispObject) {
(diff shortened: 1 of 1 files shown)


brotzeit:
I see. And `lisp_fn` recognizes non nil as true ?

brotzeit:
Ok. I just wanted to be sure.

shaleh:
I understand. It is always good to check our assumptions.

nickdrozd:
Looks I spoke too soon before. Now the cursor blinks, but only for ~5 seconds, and then it stops. Moving the cursor or changing focus and then coming back will restart the blinking, but it will still only go for ~5 seconds before stopping. (Both with and without the change in this PR)

shaleh:
@nickdrozd can you identify a build of remacs where the cursor behaved correctly?

brotzeit:
Hmmm. It seems to work for me.

shaleh:
Even with `remacs -Q` it persists?

nickdrozd:
Yeah, I have no idea what it is.

PR #1036, opened 2018-11-11
Fix buffer-list showing duplicate entries

### LINKED ISSUE #994: buffers shown twice in helm buffer
I just cloned remacs and build the binary. 

I'm using Spacemacs, and found that in Helm buffer all buffers (except Messages) are shown twice.

nitishch:
Fixes #994.

The issue was a simple logical error. Do I need to have a test for this? This doesn't seem like an issue that might pop up again.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -20,6 +20,7 @@ use character::char_head_p;
 use chartable::LispCharTableRef;
 use data::Lisp_Fwd;
 use editfns::point;
+use frames::LispFrameRef;
 use lisp::defsubr;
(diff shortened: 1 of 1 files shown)


shaleh:
@Nitishch is if I apply this to my PR #1033 to refresh this function?

nitishch:
> @nitishch is it ok if I apply this to my PR #1033 to refresh this function?

Sure. I didn't see that PR.

shaleh:
I added a simple test that `(buffer-list (selected-frame))` is unique.

shaleh:
Closing in lieu of #1033

PR #1037, opened 2018-11-11
fixing

brotzeit:


rust_src/src/editfns.rs
```diff
@@ -393,7 +393,7 @@ pub fn char_after(mut pos: LispObject) -> Option<EmacsInt> {
 /// Remaining arguments form a sequence of PROPERTY VALUE pairs for text
 /// properties to add to the result.
 /// usage: (fn STRING &rest PROPERTIES)
```
shaleh: Why are you dropping the min=1? That is accurate.

brotzeit: Fixed.

rust_src/src/marker.rs
```diff
@@ -260,8 +254,8 @@ pub fn marker_insertion_type(marker: LispMarkerRef) -> bool {
 /// If ITYPE is non-nil, it means the marker advances when you insert text at it.
 /// If ITYPE is nil, it means the marker stays behind when you insert text at it.
 #[lisp_fn]
-pub fn set_marker_insertion_type(mut marker: LispMarkerRef, itype: LispObject) -> LispObject {
-    marker.set_insertion_type(itype.is_not_nil());
+pub fn set_marker_insertion_type(mut marker: LispMarkerRef, itype: bool) -> bool {
+    marker.set_insertion_type(itype);
     itype
```
brotzeit: I just realized that itype will be returned as lisp bool and not as the original value. I think we shouldn't do this ?

shaleh: Yeah, bad idea.

brotzeit:
Should be ok.

PR #1039, opened 2018-11-11
Update docs to mention rustize tool and preferring more specific types.

shaleh:


CONTRIBUTING.md
```diff
@@ -12,6 +12,13 @@ Thank you for contributing to Remacs!
 
 * Follow the existing style for new Rust code. Notes about style
   can be found in the section on [Remacs style](#remacs-style).
+* Use rust_src/admin/rustize-emacs to help start a port of a C function
+  or `DEFUN`. Mark a region around a `DEFUN` and run `M-| rustize-emacs`.
+  This will do the work to transform the definition into Rust. You can
+  then yank and paste it into the appropriate Rust file or start a new one.
+  Now you need to port the body of the function.
+* Try to use types where possible. `LispObject` is analogous to a void pointer
```
Wilfred: It's not directly analogous to a void pointer, is it? It's a tagged pointer. The `Object` example works but I find the `*void` example a little confusing. How about something like this?

```
* Try to use more exact types where possible. `LispObject` allows anything, and often we can do better.
```

shaleh: I used the `void*` analogy because in C we often start with a `void*` and cast it to something useful. But yeah, your suggestion works.

PR #1040, opened 2018-11-11
Port mapcar1.

shaleh:
Along the way refactor bool vector iterator.

This is a simpler PR than @birkenfeld's original. Whenever I switch
to `iter().collect()` I get failures and crashes. This code is a
reasonable starting point at least.

### DIFF
--- a/rust_src/src/lists.rs
+++ b/rust_src/src/lists.rs
@@ -722,4 +722,64 @@ pub fn circular_list(obj: LispObject) -> ! {
     xsignal!(Qcircular_list, obj);
 }
 
+fn mapcar_over_iterator<I: Iterator<Item = LispObject>>(
+    output: &mut [LispObject],
+    fun: LispObject,
+    it: I,
+) -> EmacsInt {
+    let mut mapped = 0;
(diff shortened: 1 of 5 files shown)


rust_src/src/lists.rs
```diff
@@ -722,4 +722,64 @@ pub fn circular_list(obj: LispObject) -> ! {
...
+) -> EmacsInt {
+    let mut safe_value = [Qnil];
+
+    let output = if vals.is_null() {
+        &mut safe_value
+    } else {
+        unsafe { std::slice::from_raw_parts_mut(vals, leni as usize) }
+    };
+
+    if let Some(v) = seq.as_vectorlike() {
```
nickdrozd: Is there a way to use a `match` instead of all these `if / else` blocks?

shaleh: Kind of, but I saw it as less efficient.

I can ask for the vector type and match on that. But then I still need to call `as_foo`. Which will check the type again.

shaleh:
Good question. I will look.

PR #1042, opened 2018-11-12
Port generate-new-buffer-name

nitishch:
There are a few tests for `generate-new-buffer-name` already. So I didn't add any.


rust_src/src/buffers.rs
```diff
@@ -1014,6 +1017,51 @@ pub fn erase_buffer() {
...
+/// Otherwise modify name by appending `<NUMBER>', incrementing NUMBER
+/// (starting at 2) until an unused name is found, and then return that name.
+/// Optional second argument IGNORE specifies a name that is okay to use (if
+/// it is in the sequence to be tried) even if a buffer with that name exists.
+///
+/// If NAME begins with a space (i.e., a buffer that is not normally
+/// visible to users), then if buffer NAME already exists a random number
+/// is first appended to NAME, to speed up finding a non-existent buffer.
+#[lisp_fn(min = "1")]
+pub fn generate_new_buffer_name(name: LispStringRef, ignore: LispObject) -> LispStringRef {
```
db48x: The ignore argument can be an `Option<LispStringRef>`

rust_src/src/buffers.rs
```diff
@@ -1014,6 +1017,51 @@ pub fn erase_buffer() {
...
+    {
+        return name;
+    }
+
+    let basename = if name.byte_at(0) == b' ' {
+        let range = Range::new(0, 1000000);
+        let mut rng = rand::thread_rng();
+        // Since random only contains ascii chars, we can use
+        // mock_unibyte_string
+        let suffix = mock_unibyte_string!(format!("-{}", range.ind_sample(&mut rng)));
```
db48x: Don't use `mock_unibyte_string` here. It's only intended to be used for testing, since the "string" that it creates doesn't participate in the Lisp garbage collector.

rust_src/src/buffers.rs
```diff
@@ -1014,6 +1017,51 @@ pub fn erase_buffer() {
...
+            return genname.into();
+        }
+        genname
+    } else {
+        name.into()
+    };
+
+    let mut suffix_count = 2;
+    loop {
+        let suffix = mock_unibyte_string!(format!("<{}>", suffix_count));
```
db48x: Same.

rust_src/src/buffers.rs
nitishch: It is intended to be that way. C code too allocates the Lisp_String on the stack if possible. mock_unibyte_string too does the same.

rust_src/src/buffers.rs
nitishch: `ignore` is used only as an argument to `string_equal`. If I make `ignore` an `Option<LispStringRef>`, I have to have `.into()` calls. I decided making it a `LispObject` would be simpler.

shaleh: Yeah, @brotzeit had a similar thing on one of their PRs recently. Sometimes letting the `LispObject` through is the right thing.

My take on it is whether the value is used in the Rust function or just passed on through. If Rust touches, make a Rust type and then use `.into()`. Otherwise, just leave it a `LispObject`.

rust_src/src/buffers.rs
shaleh: If that is the case for sure, we should come up with a generic "local allocator" which the mock system should call.

rust_src/src/buffers.rs
nitishch: Also, `ignore` here can be a symbol or a string. So it would be wrong to make it an `Option<LispStringRef>`.

rust_src/src/buffers.rs
nitishch: Currently the way to initialize the object is specific to the object's type. I'm not sure if we can generalize the allocator.

nitishch:
Fixed the build.

brotzeit:
Is this ready to merge ?

nitishch:
> Is this ready for merging ?

Yes.

nitishch:
Added `LispObject::local_unibyte_string` with the contents of `mock_unibyte_string!`.

rust_src/src/multibyte.rs
```diff
@@ -279,6 +279,25 @@ impl LispObject {
     pub fn empty_unibyte_string() -> LispStringRef {
         LispStringRef::from(unsafe { empty_unibyte_string })
     }
+
+    // Creates a LispObject on stack representing the unibyte string `s`
```
shaleh: How is this "on the stack"? https://doc.rust-lang.org/rust-by-example/std/box.html is clear that a Box value is a heap value. All heap must be C managed currently.

This code calls `Box<T>::into_raw` which passes ownership to the C code. However, the C code has not put this pointer into the garbage collector. Which means this is likely a memory leak.

The mock function was to emulate the creation of strings for small tests. If this really worked correctly we would be writing Rust unit tests.

To accept this code we need to be certain that the garbage collector will properly handle the cleanup. Is the call to `LispObject::tag_ptr` sufficient? I am not certain that it is.

nitishch: I didn't realize this. Will correct this.

nitishch: Even the memory used by `strcopy` was leaking.

nitishch: See the changes now. If these are fine, we can replace usage of `mock_unibyte_string` with the new `local_unibyte_string` too.

rust_src/src/buffers.rs
```diff
@@ -1057,6 +1060,51 @@ pub fn erase_buffer() {
...
+        name.into()
+    };
+
+    let mut suffix_count = 2;
+    loop {
+        let s = format!("<{}>", suffix_count);
+        local_unibyte_string!(suffix, s);
+        let candidate = unsafe { concat2(basename, suffix) };
+        let buf = get_buffer(LispBufferOrName::Name(candidate));
+        if buf.is_none() || string_equal(candidate, ignore) {
```
shaleh:
```suggestion
        if string_equal(candidate, ignore) || Fget_buffer(candidate).is_nil() {
```

Check for the string_equal first like in the C code. This will short circuit and avoid the call to `get_buffer`. Although not the creation of the `LispBufferOrName` object sadly. Which can be avoided if you use `Fget_buffer(candidate)` instead and check `is_nil()`.

rust_src/src/eval_macros.rs
```diff
@@ -283,3 +283,28 @@ macro_rules! per_buffer_var_idx {
         (unsafe { crate::remacs_sys::buffer_local_flags.$field }).as_natnum_or_error() as usize
     };
 }
+
+#[macro_export]
+#[allow(unused_macros)]
+macro_rules! local_unibyte_string {
+    ($name: ident, $string: expr) => {
+        let mut bytes: Vec<u8> = ($string).bytes().collect();
```
shaleh: Why is the comment below still here?

shaleh: Could this have been `$string.as_str().as_bytes_mut()`?

nitishch: That would mean making `$string` mutable.

shaleh: Is that a significant concern here?

rust_src/src/buffers.rs
```diff
@@ -1057,6 +1060,50 @@ pub fn erase_buffer() {
...
+/// (starting at 2) until an unused name is found, and then return that name.
+/// Optional second argument IGNORE specifies a name that is okay to use (if
+/// it is in the sequence to be tried) even if a buffer with that name exists.
+///
+/// If NAME begins with a space (i.e., a buffer that is not normally
+/// visible to users), then if buffer NAME already exists a random number
+/// is first appended to NAME, to speed up finding a non-existent buffer.
+#[lisp_fn(min = "1")]
+pub fn generate_new_buffer_name(name: LispStringRef, ignore: LispObject) -> LispStringRef {
+    if get_buffer(LispBufferOrName::Name(name.into())).is_none()
```
shaleh: Here too you switched the order and as below it should return to how it was.

rust_src/src/buffers.rs
```diff
@@ -1057,6 +1060,50 @@ pub fn erase_buffer() {
...
+        return name;
+    }
+
+    let basename = if name.byte_at(0) == b' ' {
+        let range = Range::new(0, 1000000);
+        let mut rng = rand::thread_rng();
+        let mut s = format!("-{}", range.ind_sample(&mut rng));
+        local_unibyte_string!(suffix, s);
+        let genname = unsafe { concat2(name.into(), suffix) };
+        if get_buffer(LispBufferOrName::Name(genname)).is_none() {
```
shaleh: Another way to write this would be:
```
match get_buffer(LispBufferOrName::Name(genname)) {
    None => return genname.into(),
    _ => genname
}
```

I am not sure which is better.

rust_src/src/eval_macros.rs
```diff
@@ -283,3 +283,26 @@ macro_rules! per_buffer_var_idx {
         (unsafe { crate::remacs_sys::buffer_local_flags.$field }).as_natnum_or_error() as usize
     };
 }
+
+// Creates a Lisp_String from $string on the stack and a LispObject
+// pointing the Lisp_String. Assigns the LispObject to $name
+#[macro_export]
+#[allow(unused_macros)]
+macro_rules! local_unibyte_string {
+    ($name: ident, $string: expr) => {
```
shaleh: Should the code be placed in a block to ensure the names do not conflict? For instance, what happens if you try to call this macro twice in the same namespace?

src/buffer.c
```diff
@@ -878,60 +878,6 @@ reset_buffer_local_variables (struct buffer *b, bool permanent_too)
     }
 }
```
shaleh: You omitted the comment below. It still sounds useful.

shaleh:
BTW, Clippy says this:

```
warning: long literal lacking separators
    --> src/buffers.rs:1082:35
     |
1082 |         let range = Range::new(0, 1000000);
     |                                   ^^^^^^^ help: consider: `1_000_000`
     |
     = note: #[warn(clippy::unreadable_literal)] on by default
     = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#unreadable_literal
```

shaleh:
@db48x this is looking more reasonable now. Mind giving it another look?

rust_src/src/eval_macros.rs
shaleh: Run a test where you call the macro twice in a row and see if the output is right.

rust_src/src/buffers.rs
nitishch: Done.

src/buffer.c
nitishch: Done.

nitishch:
> BTW, Clippy says this:

How are you running Clippy. When I run `cargo clippy`, I get the following error -

```
thread 'main' panicked at 'No EMACS_CFLAGS were specified, and we need them in order to run bindgen.', build.rs:648:17
```

If I run `EMACS_CFLAGS='' cargo clippy`, I get this error:

```
wrapper.h:1:10: fatal error: 'config.h' file not found
wrapper.h:1:10: fatal error: 'config.h' file not found, err: true
thread 'main' panicked at 'Unable to generate bindings: ()', libcore/result.rs:1009:5
```

shaleh:
1. `make clippy`
2. Travis runs it on all PRs, look at the build results.

rust_src/src/eval_macros.rs
nitishch: Added the test `test_local_unibyte_string`.

brotzeit:
Please merge this as a squash.

rust_src/src/buffers.rs
```diff
@@ -1057,6 +1060,52 @@ pub fn erase_buffer() {
...
+    } else {
+        name.into()
+    };
+
+    let mut suffix_count = 2;
+    loop {
+        let mut s = format!("<{}>", suffix_count);
+        local_unibyte_string!(suffix, s);
+        let candidate = unsafe { concat2(basename, suffix) };
+        if string_equal(candidate, ignore) || Fget_buffer(candidate).is_nil() {
```
brotzeit: I think you can use `get_buffer` here.

shaleh:
```suggestion
        if string_equal(candidate, ignore) || get_buffer(LispBufferOrName::Name(candidate)).is_none() {
```

shaleh: Since you know you have a string we can be more direct.

nitishch: Done

rust_src/src/buffers.rs
shaleh: I am ok with that. From reading other Rust code the `match` form with one side returning or erroring and the other being the "good" case is common. Especially in assignments.

```
let foo = match something {
    None => return THING, // or maybe error/die
    Some(value) => value,
};
```

nickdrozd:
I second @brotzeit 's motion to squash :)

shaleh:
Always squash.

shaleh:
@nitishch we all failed to look at the OSX build logs. This change segfaults on OSX. We can either revert it or fix it.

shaleh:
I opened #1132 to track this and let others know.

PR #1046, opened 2018-11-12
Pull from upstream to fix backup file behavior.

jaccarmac:


jaccarmac:
Now with the extent of changes of my second attempt without all the squashing. Hopefully tests will pass this time!

shaleh:
Great work sleuthing this!

jaccarmac:
I see that ChangeLog.2 has a lot of added lines. Which other things were added in error?

Wilfred:
Everything else looks fine to me. @brotzeit what were you referring to?

brotzeit:
I have directories in my local repo on master that were already removed and I don't know why. It seems we only have to delete the changelog.

jaccarmac:
#1063's up to remove the changelog. Phantom dirs are weird but don't seem to be caused by Git. Do they get marked for removal by `git clean -nxd`?

jaccarmac:
Git can get funky when the ignore list, staged files, tracked files, and files on disk interact. Don't have to do it often, but `git-clean` is a tool in my toolbox. FYI the `-n` flag does a dummy run, change it to `-f` to actually modify your file system.

PR #1050, opened 2018-11-13
remove ne

brotzeit:


brotzeit:
Ah now I remember. I got this idea from `PartialEq`.
```rust
fn ne(&self, other: &Rhs) -> bool
```

shaleh:
ah. But there is no need to specially implement it since the `!eq` works.

PR #1052, opened 2018-11-13
Port current-indentation

jcreedcmu:


rust_src/src/indent.rs
```diff
@@ -1,12 +1,43 @@
 //! Indentation functions
 
+use std::ptr;
+
+use editfns::{point, point_byte, point_min, point_min_byte};
+use lisp::defsubr;
 use remacs_macros::lisp_fn;
 
+
```
jcreedcmu: Whoops, introduced this accidentally during rebase.

jcreedcmu:
An attempt at https://github.com/Wilfred/remacs/issues/1005

Instead of actually porting the meat of the function, which lies in the (previously `static`) function `position_indentation`, I just exposed it in `indent.h` and called it from rust. Hope this is an okay incremental approach for a first attempt.

jcreedcmu:
Some questions I realized I don't know the answer to as I wrote this:

* Is there an easy way to do a `make clean` but *only* for the output of bindgen? When I edited the `.h` files, just doing `make` did not seem to realize that bindgen's output was invalidated.

* do I really want this pair of `_byte` functions that I introduced here, to simulate the C macros `PT_BYTE` and `BEGV_BYTE` in `buffer.h`? My simple tests still pass if I just put `-1`s in instead of calls to them, (imitating the other call to `find_newline` I found in `editfns.rs`) but I assume it's something to handle the case of e.g. multibyte unicode encodings or something?

* how am I supposed to run the clippy linter? I tried following the instructions in `CONTRIBUTING.md` but if I run
```$ cargo clippy```
from in the toplevel directory, it complains it can't find `Cargo.toml`, and if I run it from inside `rust_src`, I get

	 error: failed to run custom build command for `remacs v0.1.0 (.../remacs/rust_src)`
	 process didn't exit successfully: `.../remacs/rust_src/target/debug/build/remacs-62c2d27fb4cf7834/build-script-build` (exit code: 101)
	 --- stdout
	 cargo:rerun-if-env-changed=EMACS_CFLAGS
	 cargo:rerun-if-env-changed=SRC_HASH

	 --- stderr
	 thread 'main' panicked at 'No EMACS_CFLAGS were specified, and we need them in order to run bindgen.', build.rs:640:17
	 note: Run with `RUST_BACKTRACE=1` for a backtrace.

and I assume `EMACS_CFLAGS` normally comes out of the autogenrated Makefile somehow?

**Comment by shaleh** [comment]
Good question!!

> Is there an easy way to do a make clean but only for the output of bindgen? When I edited the .h files, just doing make did not seem to realize that bindgen's output was invalidated.

You can `touch src/lisp.h` and re-run make that should do what you want.

> do I really want this pair of _byte functions that I introduced here, to simulate the C macros PT_BYTE and BEGV_BYTE in buffer.h? My simple tests still pass if I just put -1s in instead of calls to them, (imitating the other call to find_newline I found in editfns.rs) but I assume it's something to handle the case of e.g. multibyte unicode encodings or something?

If we do want those function there is no need to make them `lisp_fn`. That exposes the function as a lisp accessible `(point-byte)`. For these function look at the top of buffers.rs and mimic what is already there for `BEGV`.

> how am I supposed to run the clippy linter?

Clippy had not worked for a while. Now the tool works again we need to update our support. What is needed is an addition to `src/Makefile.in` to call `cargo clippy`. Then the right environment will be set up.

**Comment by shaleh** [comment]
Oh, and don't feel bad about leaning on a C function to port another function.  It is expected.

**Comment by jcreedcmu** [comment]
I had the same experience regardless of whether I changed `indent.h` or `lisp.h`. Bindgen's output stayed the same either way. In any event, manually deleting everything in `target/` that had 'bindgen' in the filename was a sufficient workaround :)

I'll de-`#lisp_fun` those, then...

**Comment by shaleh** [comment]
That is odd. I never have an issue running make if `lisp.h` changes. Your brute force approach is not a bad one though.

File: rust_src/src/editfns.rs [hunk 2]
```diff
@@ -46,6 +46,13 @@ pub fn point() -> EmacsInt {
     buffer_ref.pt as EmacsInt
 }
```
shaleh: These belong in buffers.rs with the others.

rust_src/src/indent.rs
```diff
@@ -1,12 +1,36 @@
 //! Indentation functions
 
 use remacs_macros::lisp_fn;
+use std::ptr;
```
shaleh: Put this import above the remacs_macros line with an empty line between them.

jcreedcmu:
(I took off the `lisp_fn`, though, removing that makes perfect sense)

rust_src/src/editfns.rs
jcreedcmu: With `BEG` and `BEG_BYTE`, you mean? Can do.

db48x:
It's supposed to rerun the `rust_src/build.rs` whenever the value of the SRC_HASH variable changes, which is supposed to happen any time any `.c`, `.h`, or `.rs` file in any directory mentioned in `EMACS_CFLAGS` changes. If you can work out why it isn't working for you, then we can fix the bug.

shaleh:
The recent switch to Rust 2018 probably means you need to manually update rustfmt. Hopefully `rustup` will get updated so this is no longe needed.

PR #1054, opened 2018-11-13
add intspec to erase_buffer

brotzeit:
Only found this because I started to use remacs :P


shaleh:
I too have started using remacs as my day to day editor. So far, so good.

PR #1056, opened 2018-11-14
Revert "Pull from upstream to fix backup file behavior."

brotzeit:
I'm afraid we have to do this again.


shaleh:
Do we really have to revert the whole PR?

brotzeit:
I think it's not important, but I opened this as a reminder. However most of this PR gets removed.

PR #1065, opened 2018-11-16
LispFrameOrSelected

shaleh:


shaleh:
Frankly I need to better understand those hoops to document them.

shaleh:
I pulled next/prev frame out. They will get their own PR.

PR #1067, opened 2018-11-16
port editfns functions

brotzeit:


rust_src/src/editfns.rs
```diff
@@ -1177,4 +1319,39 @@ pub fn rust_find_field(
...
+    }
+}
+
+/// Return the contents of the field surrounding POS as a string.
+/// A field is a region of text with the same `field' property.
+/// If POS is nil, the value of point is used for POS.
+#[lisp_fn(min = "0")]
+pub fn field_string(pos: LispNumber) -> LispObject {
+    let mut isize: beg;
+    let mut isize: end;
```
shaleh: It looks like `rust_find_field` could return a tuple instead of using C style pointers to handle multiple returns.

shaleh: Yep, I made a PR based on this one. #1070.

I pulled the 3 users of find_field from this PR into it. Now there is no C version of the function and the pointers are gone.

rust_src/src/editfns.rs
```diff
@@ -1003,7 +1188,7 @@ pub fn save_current_buffer(args: LispObject) -> LispObject {
 
 /// Execute BODY, saving and restoring current buffer's restrictions.
 /// The buffer's restrictions make parts of the beginning and end invisible.
```
shaleh: The backslash is required.

PR #1081, opened 2018-11-19
port syntax.c functions

brotzeit:


rust_src/src/syntax.rs
```diff
@@ -104,4 +109,88 @@ pub fn copy_syntax_table(mut table: LispObject) -> LispObject {
...
+    // Avoid jumping out of an input field.
+    let tmp = constrain_to_field(
+        Some(LispNumber::Fixnum(val as EmacsInt)),
+        LispNumber::Fixnum(point as EmacsInt),
+        false,
+        false,
+        Qnil,
+    ) as isize;
+    unsafe { set_point(tmp) };
+    val == tmp
```
nitishch: You are comparing `tmp` to `val` while the C code compared it to output of `scan_words`.

rust_src/src/syntax.rs
```diff
@@ -104,4 +109,88 @@ pub fn copy_syntax_table(mut table: LispObject) -> LispObject {
...
+/// table, but `find-word-boundary-function-table', such as set up
+/// by `subword-mode', can change that.  If a Lisp program needs to
+/// move by words determined strictly by the syntax table, it should
+/// use `forward-word-strictly' instead.
+#[lisp_fn(min = "0", intspec = "^p")]
+pub fn forward_word(arg: Option<EmacsInt>) -> bool {
+    let arg = arg.unwrap_or(1);
+    let cur_buf = ThreadState::current_buffer();
+    let point = cur_buf.pt;
+
```
shaleh: While I like how this reads, it does not match the original C's intent.

Both orig_val and val are created by the call to `scan_words`. However, orig_val is not changed if it is 0 and val is.
The correct code written in this style would be (I used an assignment to simplify the return and avoid repetition.):

```
let (val, orig_val) = match unsafe { scan_words(point, arg) } {
        0 => {
           let val = if arg > 0 { cur_buf.zv } else { cur_buf.begv };
           (val, 0)
        }
        n => (n, n),
};
```

PR #1084, opened 2018-11-21
Port some file predicates

nickdrozd:


shaleh:
kick it...

rust_src/src/fileio.rs
```diff
@@ -49,4 +54,74 @@ pub fn recent_auto_save_p() -> bool {
     cur_buf.modifications_since_save() < cur_buf.auto_save_modified
 }
 
+/// Return t if file FILENAME is on a case-insensitive filesystem.
+/// The arg must be a string.
+#[lisp_fn]
+pub fn file_name_case_insensitive_p(filename: LispStringRef) -> bool {
+    let absname = unsafe { Fexpand_file_name(filename.into(), Qnil) };
+
+    /* If the file name has special constructs in it,
```
shaleh: We have been using `//` instead of `/*` for comments.

rust_src/src/fileio.rs
```diff
@@ -49,4 +54,74 @@ pub fn recent_auto_save_p() -> bool {
...
+
+/// Return t if file FILENAME exists (whether or not you can read it.)
+/// See also `file-readable-p' and `file-attributes'.
+/// This returns nil for a symlink to a nonexistent file.
+/// Use `file-symlink-p' to test for such links.
+#[lisp_fn]
+pub fn file_exists_p(filename: LispStringRef) -> bool {
+    let absname = unsafe { Fexpand_file_name(filename.into(), Qnil) };
+
+    /* If the file name has special constructs in it,
```
shaleh: Same here.

src/fileio.c
```diff
@@ -2193,27 +2193,6 @@ file_name_case_insensitive_p (const char *filename)
...
-  CHECK_STRING (filename);
-  filename = Fexpand_file_name (filename, Qnil);
-
-  /* If the file name has special constructs in it,
-     call the corresponding file handler.  */
-  handler = Ffind_file_name_handler (filename, Qfile_name_case_insensitive_p);
-  if (!NILP (handler))
-    return call2 (handler, Qfile_name_case_insensitive_p, filename);
-
-  filename = ENCODE_FILE (filename);
```
shaleh: This is your mistake in the Rust. You are recursively calling the Rust function instead of the C function.

This means `file-name-case-insensitive-p` needs a `file_name_case_insensitive_p_rust` version and name in the `lisp_fn`.

nickdrozd: Good catch! I had no idea what the problem was.

rust_src/Cargo.toml
```diff
@@ -14,6 +14,7 @@ remacs-lib = { version = "0.1.0", path = "remacs-lib" }
 remacs-macros = { version = "0.1.0", path = "remacs-macros" }
 base64 = "0.9"
 clippy = { version = "*", optional = true }
+errno = "0.2.3"
```
nitishch: `0.2.4` seems to be the latest version.

src/fileio.c
shaleh: I missed it the first time too.

I ran temacs in lldb. It would segfault so then I looked at the backtrace. There was a LONG list of the same function when it should not be recursing.

nickdrozd:
Yeah, I suppose so. `file-exists-p` and `file-executable-p` are adequately covered by existing tests, but not `file-name-absolute-p` or `file-name-case-insensitive-p`. The latter will require a little work with temp files, and I won't have time to do that for a few days. I'll pick it up on Monday.

shaleh:
:+1:

PR #1086, opened 2018-11-21
Port next-frame and previous-frame.

shaleh:
Introduce `for_each_frame` macro.


brotzeit:
I couldn't find any tests for these functions. I think this should be easy to verify this works ?

rust_src/src/frames.rs
```diff
@@ -414,4 +422,86 @@ pub fn delete_frame_lisp(frame: LispObject, force: bool) {
...
+
+    // Go through the list at most twice. This treats the frame list
+    // as a circular list allowing for 'next' to occur before FRAME.
+    // Once the desired frame is found in the list, the next frame that is
+    // a valid candidate will be returned regardless of its position.
+    while passed < 2 {
+        for_each_frame!(f => {
+	    if passed > 0 {
+	        let tmp = unsafe { candidate_frame(f.into(), frame_obj, miniframe) };
+	        if !tmp.is_nil() {
```
brotzeit: I don't want to be pedantic and I would also prefer this way. But the team decided differently and I think we should try to use `is_not_nil` consistently.

shaleh:
I will have to see if writing a test is feasible. It is a good idea if it can be done.

shaleh:
I wrote this simple set of tests:

```
(require 'ert)

(ert-deftest test-next-frame ()
  "next-frame will skip the current frame. In this test there is the base frame
   and the new frame."
  (let ((frame (make-frame)))
    (should (equal 2 (length (frame-list))))
    (should-not (equal frame (next-frame)))
    (select-frame (previous-frame))
    (delete-frame frame t)))

(ert-deftest test-previous-frame ()
  "previous-frame will skip the current frame. In this test there is the base frame
   and the new frame."
  (let ((frame (make-frame)))
    (should (equal 2 (length (frame-list))))
    (should (not (equal frame (previous-frame))))
    (select-frame (previous-frame))
    (delete-frame frame t)))
```

However, it fails at the call to `make-frame` because of the terminal being unknown.

PR #1087, opened 2018-11-22
Port copy-keymap

shaleh:
Thanks @accidentalrebel for your work.

### DIFF
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -13,14 +13,18 @@ use crate::{
     eval::{autoload_do_load, unbind_to},
     keyboard::lucid_event_type_list_p,
     lisp::{defsubr, LispObject},
-    lists::nth,
+    lists::{nth, setcdr},
     obarray::intern,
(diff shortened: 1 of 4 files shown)


rust_src/src/keymap.rs
```diff
@@ -614,4 +618,67 @@ pub fn describe_vector_lisp(vector: LispObject, mut describer: LispObject) {
...
+                elt = LispObject::cons(front, unsafe { copy_keymap_item(cons_1.cdr()) });
+            }
+        }
+
+        setcdr(tail.as_cons().unwrap(), list!(elt));
+        tail = tail.as_cons().unwrap().cdr();
+        keymap = cons.cdr();
+    }
+
+    setcdr(tail.as_cons().unwrap(), keymap);
```
brotzeit: Don't we use `as_cons_or_error` instead of `unwrap` ?

shaleh: Usually. But here we know we have cons. We need an equivalent to XCAR and friends.

PR #1090, opened 2018-11-22
Port string-width.

agraven:
This should solve issue #991


agraven:
Porting lisp_string_width seems like the most proper solution, and might also save someone work down the line, so it's the one I'll try to implement

shaleh:
Sounds good. We also have a Gitter channel (https://gitter.im/remacs-discuss/Lobby) if you have questions.

agraven:
I've implemented it using a call to `lisp_string_width` for now, since that function is a much deeper rabbit hole of additional functions that would need to be ported than I'd anticipated

rust_src/src/multibyte.rs
```diff
@@ -87,6 +87,15 @@ impl LispStringRef {
         s.size
     }
 
+    /// Return width of STRING when displayed in the current buffer. Width is
+    /// measured by how many columns it occupies on the screen. When calculating
+    /// width of a multibyte character in STRING, only the base leading-code is
+    /// considered; the validity of the following bytes is not checked.  Tabs in
+    /// STRING are always taken to occupy `tab-width' columns.
+    pub fn string_width(self) -> ptrdiff_t {
```
shaleh:
```suggestion
    pub fn width(self) -> usize {
```

We can avoid "stuttering". The lisp name is `<type>-<field>` and directly analogous to object oriented `<type>::<field>`.

Also, we are using `usize` instead of `ptrdiff_t`. Same thing, different name.

rust_src/src/strings.rs
```diff
@@ -167,6 +167,16 @@ pub fn clear_string(mut string: LispStringRef) {
     string.mark_as_unibyte();
 }
 
+/// Return width of STRING when displayed in the current buffer. Width is
+/// measured by how many columns it occupies on the screen. When calculating
+/// width of a multibyte character in STRING, only the base leading-code is
+/// considered; the validity of the following bytes is not checked.  Tabs in
+/// STRING are always taken to occupy `tab-width' columns.
+#[lisp_fn]
+pub fn string_width(string: LispStringRef) -> EmacsInt {
```
shaleh: If you make the switch to `usize` above you can return `usize` here as well and skip the cast.

rust_src/src/multibyte.rs
agraven: Oh right, I hadn't realized it was following that convention. I remember GTK's `gtk_text_tag_table_new()` style names, really makes you appreciate a proper module system.

rust_src/src/strings.rs
agraven: Both the suggested tweaks have been added

shaleh:
Functionally this is good to go once the CI checks pass.

Could you add some simple tests?

    (should (eq (string-width "\t") 8))) ;; Tabs are expanded.
    (should (eq (string-with "abc" 3)))

A couple of utf8/multibyte tests would also be nice.

shaleh:
All looks good. Thanks @agraven !

Issue #1085, opened 2018-11-21
flymake fails tests on MacOS


shaleh:
I spent a few minutes debugging this today.

flymake on header files wants to check the C file that includes it instead of the header itself. There is a C file in the directory that includes `some-problems.h` but the code does not find it. Instead it looks at `no-problems.h` and then quits.

I can reproduce the flymake behavior by simply opening some-problems.h. The mode-line says `[0 0]` and no lines are highlighted as problematic.

I did not spend enough time to see why the C file was not properly inspected.

PR #1093, opened 2018-11-25
Port `window-new-normal` to Rust

### LINKED ISSUE #1048: Port window-new-normal
```c
DEFUN ("window-new-normal", Fwindow_new_normal, Swindow_new_normal, 0, 1, 0,
       doc: /* Return new normal size of window WINDOW.

The new normal size of WINDOW is the value set by the last call of
`set-window-new-normal' for WINDOW.  If valid, it will be shortly
installed as WINDOW's normal size (see `window-normal-size').  */)
  (Lisp_Object window)
{
  return decode_valid_window (window)->new_normal;
}
```

Gastove:
Commit includes a Rust implementation of `window-new-normal`, removal
of the old C function and defsubr for same, and a test to assure
`windows-new-normal` correctly returns the value set by
`set-window-new-normal`.

Closes #1048

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1058,6 +1058,18 @@ pub fn scroll_down(arg: LispObject) {
     unsafe { scroll_command(arg, -1) };
 }
 
+/// Return new normal size of window WINDOW.
+/// WINDOW must be a valid window and defaults to the selected one.
+///
+/// The new normal size of WINDOW is the value set by the last call of
+/// `set-window-new-normal' for WINDOW.  If valid, it will be shortly
+/// installed as WINDOW's normal size (see `window-normal-size').
+#[lisp_fn(min = "0")]
+pub fn window_new_normal(window: LispWindowValidOrSelected) -> LispObject {
+    let win: LispWindowRef = window.into();
+    win.new_normal
+}
+
 /// Return the new total size of window WINDOW.
 /// WINDOW must be a valid window and defaults to the selected one.
 ///
--- a/test/rust_src/src/windows-tests.el
+++ b/test/rust_src/src/windows-tests.el
@@ -43,10 +43,25 @@
   (should (eq (window-new-total (selected-window)) (window-total-height))))
 
 (ert-deftest window-new-normal ()
+  "Effectively tests both `window-new-normal' (the getter) and
+`set-window-new-normal' (the setter)."
+  ;; Can we change normal on this window?
   (set-window-new-normal nil 1.0)
   (should (= (window-new-normal) 1.0))
   (set-window-new-normal nil 0.23)
-  (should (= (window-new-normal) 0.23)))
+  (should (= (window-new-normal) 0.23))
+
+  ;; Can we correctly get the value of a different window?
+  (let ((current-window-expected-normal 1.0)
+        (other-window-expected-normal 0.5)
+        (current-window (selected-window))
+        (other-window (split-window)))
+    (set-window-new-normal current-window current-window-expected-normal)
+    (set-window-new-normal other-window other-window-expected-normal)
+    ;; Normal for current-window should be the same
+    (should (= (window-new-normal) current-window-expected-normal))
+    (select-window other-window)
+    (should (= (window-new-normal) other-window-expected-normal))))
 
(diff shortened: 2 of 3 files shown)


rust_src/src/windows.rs
```diff
@@ -1058,6 +1058,18 @@ pub fn scroll_down(arg: LispObject) {
...
+/// Return new normal size of window WINDOW.
+/// WINDOW must be a valid window and defaults to the selected one.
+///
+/// The new normal size of WINDOW is the value set by the last call of
+/// `set-window-new-normal' for WINDOW.  If valid, it will be shortly
+/// installed as WINDOW's normal size (see `window-normal-size').
+#[lisp_fn(min = "0")]
+pub fn window_new_normal(window: LispWindowValidOrSelected) -> LispObject {
+    let win: LispWindowRef = window.into();
+    win.new_normal
```
shaleh: When we port to Rust, we try to return as specific a type as possible. From my reading of the code, it looks like `new_normal` is a floating type. Perhaps this should return `EmacsDouble` and the return should be `win.new_normal.as_float_or_error()`? Or are there cases where `nil` is valid? If so, the return can be `Option<EmacsDouble>` and the code can return `win.new_normal.as_float()`.

The reason for the specific return is it allows for the Rust code to act on native types instead of marshalling to and from Lisp primitives.

Gastove: Ahh, a very good point. 

As best as I can tell, there are no cases where nil is valid -- but to be honest, I'm new here, and I'd welcome advice on this. To my read, returning `EmacsDouble` sounds ideal.

shaleh: Let's start out cautious and return `Option<EmacsDouble>`. We can switch to plain `EmacsDouble` as the port continues.

test/rust_src/src/windows-tests.el
```diff
@@ -42,6 +42,12 @@
   (should (eq (window-new-total) (window-total-height)))
   (should (eq (window-new-total (selected-window)) (window-total-height))))
 
+(ert-deftest window-new-normal ()
+  (set-window-new-normal nil 1.0)
+  (should (= (window-new-normal) 1.0))
+  (set-window-new-normal nil 0.23)
+  (should (= (window-new-normal) 0.23)))
+
```
nitishch: Since this is the only test for `window-new-normal`, can you add tests calling it with a window argument.

rust_src/src/windows.rs
nitishch: It appears `window-new-normal` can return any `LispObject`. `(set-window-new-normal nil "abc") (window-new-normal)` returns `"abc"`.

I thought this is a mistake in restricting the value but see `lisp/window.el` line 5351 -

`      (unless (eq (window-new-normal sub) 'ignore)`

This suggests that `window-new-normal` returning a symbol is an expected behaviour.

shaleh: I suspect the assignment of a string is lazy checking. However, it does look like a symbol is checked. Which means the return should remain LispObject.

The tests should reflect the checking against a symbol. The string is a mistake we should perhaps prevent at some point.

Gastove: Okay. I'll move this back to `LispObject`, perhaps with a `TODO` comment to investigate and refine the type as more of `window.c` is ported.

test/rust_src/src/windows-tests.el
Gastove: Can do :+1:

Gastove:
Ah good, the tests are failing on CI. These tests pass for me locally, _and_ they're about a different part of the code than I changed. I guess I'll dig in to that next.

nickdrozd:
@Gastove My guess is that your test is introducing side-effects that aren't getting undone before the next test. I don't know enough about windows or ERT to add anything more specific than that, but that's what I would look into.

shaleh:
Agreed @nickdrozd. Perhaps the right thing to do is open a new window and mod it.

Gastove:
Yeah -- I just wish I could replicate the behavior locally :/

Gastove:
OK. The build now passes for Linux. It fails for OSX on flymake, but I believe that's brought up in #1085

PR #1095, opened 2018-11-25
cosmetic changes

brotzeit:


rust_src/src/marker.rs
```diff
@@ -30,39 +30,39 @@ impl LispMarkerRef {
         unsafe { mem::transmute(self.as_ptr()) }
     }
```
shaleh: I am not opposed to moving to `isize`. However, we should be consistent. Does this mean we also drop the `c_int` usage as well?

rust_src/src/symbols.rs
```diff
@@ -23,6 +23,10 @@ use crate::{
 pub type LispSymbolRef = ExternalPtr<Lisp_Symbol>;
 
 impl LispSymbolRef {
```
shaleh: Does it make more sense to drop this style and only use `From` traits?

brotzeit: You mean `as_lisp_obj` ?

rust_src/src/marker.rs
brotzeit: I thought we only use C types when rust functions are used by C functions.

rust_src/src/symbols.rs
shaleh: Yes.

brotzeit: Makes sense to me. This method has always looked a little strange.

PR #1107, opened 2018-11-27
Port call-process

nickdrozd:


rust_src/src/buffers.rs
```diff
@@ -163,6 +163,10 @@ impl LispBufferRef {
         self.filename_
     }
```
shaleh: These accessors are a relic from before we had bindgen. Does having this method help in a significant way?

rust_src/src/callproc.rs
```diff
@@ -0,0 +1,82 @@
...
+/// determined by `unhandled-file-name-directory'), or "~" otherwise.  If
+/// you want to run a process in a remote directory use `process-file'.
+///
+/// usage: (call-process PROGRAM &optional INFILE DESTINATION DISPLAY &rest ARGS)
+#[lisp_fn(min = "1", name = "call-process", c_name = "call_process")]
+pub fn call_process_lisp(args: &mut [LispObject]) -> LispObject {
+    let count = c_specpdl_index();
+
+    let infile = if args.len() >= 2 && args[1].is_not_nil() {
+        unsafe { Fexpand_file_name(args[1], ThreadState::current_buffer().directory()) }
```
shaleh: There is a `CHECK_STRING` call here in the C code. Do we need the equivalent?

rust_src/src/buffers.rs
nickdrozd: No, I just copied the existing style. Those `field_` names are pretty ugly though, so I think this looks better. Have we settled on a standard way of doing this?

rust_src/src/callproc.rs
nickdrozd: There is `encoded_file.as_string_or_error()`, so a non-string error will get caught one way or another. Does it matter if it happens before or after `encode_file_name`?

nickdrozd:
`call-process` doesn't return the command output, just an int, so I don't know how to test it aside from just a basic smoke test. But it's already used pretty extensively in the existing Lisp tests, so another smoke test doesn't seem necessary.

rust_src/src/buffers.rs
nickdrozd: Okay, I got rid of it.

shaleh:
Merge at your leisure.

PR #1109, opened 2018-11-28
Fix 'usage' in docstrings.

shaleh:
Emacs stores the function usage without the function name like this:

    (fn foo &rest args)

When we find a function with an explicit `usage:` section we keep
the part after the function name.

If the function does not have usage we generate it.

### DIFF
--- a/rust_src/remacs-lib/docfile.rs
+++ b/rust_src/remacs-lib/docfile.rs
@@ -59,7 +59,15 @@ pub unsafe extern "C" fn scan_rust_file(
             }
             in_docstring = true;
             if line.starts_with("/// usage: (") {
-                docstring_usage.push_str(&line[11..]);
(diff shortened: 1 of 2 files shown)


nickdrozd:
Nice! Could you add some tests? I think these assertions should do it:

```
(should
 (equal
  (car (help-split-fundoc (documentation 'funcall-interactively) 'funcall-interactively))
  "(funcall-interactively FUNCTION &rest ARGUMENTS)"))

(should
 (equal
  (car (help-split-fundoc (documentation 'message) 'message))
  "(message FORMAT-STRING &rest ARGS)"))
```

(They could be made more elegant, but whatever.)

shaleh:
Tests are definitely a good idea. It was 3am so I called it good enough :-)

PR #1116, opened 2018-12-02
Correct deprecated AppKit Color API

gganley:
Ever since the release macOS 10.14 Mojave there have been issues running the GUI of Emacs. This is taken from GNU Emacs to replace the deprecated APIs

### DIFF
--- a/src/nsterm.m
+++ b/src/nsterm.m
@@ -8068,8 +8068,11 @@ not_in_argv (NSString *arg)
 
 - (void)drawRect: (NSRect)rect
 {
-  int x = NSMinX (rect), y = NSMinY (rect);
-  int width = NSWidth (rect), height = NSHeight (rect);
+  int x, y, width, height;
+  const NSRect *rects;
+  NSInteger numRects;
+  NSColor *bgCol = ns_lookup_indexed_color
+    (NS_FACE_BACKGROUND (FRAME_DEFAULT_FACE (emacsframe)), emacsframe);
 
   NSTRACE ("[EmacsView drawRect:" NSTRACE_FMT_RECT "]",
            NSTRACE_ARG_RECT(rect));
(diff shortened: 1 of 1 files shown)


shaleh:
Bah. One more time.

gganley:
These were cherry picked from emacs/scratch/ns-drawing. Let me see if they were merged into master.

shaleh:
Does emacs master work? If so, we can continue to port back from it. If not, then we need this cherry pick

shaleh:
Does emacs master work? If so, we can continue to port back from it. If not, then we need this cherry pick

gganley:
Just to clarify, are you asking if the emacs master has the depreciation issue?

shaleh:
Yes, sorry if I was not clear.

gganley:
I compiled emacs HEAD from brew and it works just fine, would you like the commands I ran and pictures of it running or is this sufficient.

shaleh:
Just knowing that emacs HEAD works is enough. That means there are commits in HEAD that we can backport to our tree. Thanks for doing the legwork @gganley.

A potentially quick way to identify the fix is to find the .m files here https://github.com/emacs-mirror/emacs/tree/master/src and then look at their commit history. Hopefully a commit message is clear enough and we can cherry pick those commits.

gganley:
So you'd rather me cherry-pick from master rather than this branch?

shaleh:
Yes we would. It makes the merges later easier.

shaleh:
@gganley @daichimukai what do we do with this PR?

daichimukai:
@ggnaley Now necessary commits seems to be merged in GNU Emacs's master commit. Are you interested in cherry-picking?

gganley:
Dupe of #1338, they cherry picked it already.

daichimukai:
Closing due to that #1338 has merged.

PR #1118, opened 2018-12-02
Port support for cons in internal_equal.

shaleh:
However, there was a snag. It turns out `TailsIter` is over eager when
it comes to checking the end condition. In C there is a function called `CHECK_LIST_END`
used after `FOR_EACH_TAIL` for example `CHECK_LIST_END (sequence, sequence);`. It ensures
that the list was fully walked or at least exited early due to a decision and not because the last item checked was not a CONS cell. If the last item was not a proper cons cell like say `(0 . 0)` then `CHECK_LIST_END` will complain. But we do not always want that.

This is the beginning of a refactor/reorganization of the LispCons iterators. `TailsIter::new` now takes an extra parameter which contains the error to raise when the check fails or None. For this PR the code has been left as it was and the lack of check must be requested with `iter_tails_noendchecked`. In the near future this function will replace `iter_tails` and a new `iter_tails_endchecked` will be added. However, this change requires going through all of our ported code and finding the places that need to be updated. To prepare for that there is a `iter_tails_v2()` which exposes two enums to controls whether end checking is done and whether circular list checking is done.

As a further step towards clarity and simplification the extra `unchecked` iterators have been removed. Their functionality is now available using the enum controls.

To reiterate -- the existing behavior has not been changed. If the code calls `iter_tails()` then the behavior before this PR continues. However, this PR provides all of the tools needed to update the tree.


rust_src/src/lists.rs
```diff
@@ -313,6 +323,58 @@ impl LispCons {
...
         }
     }
+
+    pub fn equal(
+        self,
+        other: LispCons,
+        kind: equal_kind::Type,
+        depth: i32,
+        ht: LispObject,
+    ) -> bool {
```
shaleh: I dislike this code duplication. However, the iterators are of different type which prevents the logic from being collapsed. The type of iterator and the check for equality for each item is the only difference between the two stanzas.

One option is a macro. But it felt wrong.

It would look like this:

```
if kind == equal_kind::EQUAL_NO_QUIT {
            let mut it1 = LispObject::from(self).iter_tails_unchecked();
            let mut it2 = LispObject::from(other).iter_tails_unchecked();

             cons_equal!(it1, it2,
                         0, Qnil,   // these two are for the check on each item
                         depth + 1, ht) // and these are for when none matched.
} else {
            let mut it1 = LispObject::from(self).iter_tails_noendchecked();
            let mut it2 = LispObject::from(other).iter_tails_noendchecked();

             cons_equal!(it1, it2,
                         depth + 1, ht,   // these two are for the check on each item
                         depth + 1, ht) // and these are for when none matched.
}
```

rust_src/src/lists.rs
```diff
@@ -142,8 +152,8 @@ impl Iterator for TailsIter {
     fn next(&mut self) -> Option<Self::Item> {
         match self.tail.as_cons() {
             None => {
```
shaleh: This is where the existing code is wrong. No error should be raised unless requested.

rust_src/src/lists.rs
```diff
@@ -55,13 +55,17 @@ impl LispObject {
     /// of cons cells ending in nil.  Otherwise a wrong-type-argument error
     /// will be signaled.
     pub fn iter_tails(self) -> TailsIter {
-        TailsIter::new(self, Some(Qlistp))
+        TailsIter::new(self, Some(Qlistp), Some(Qlistp))
+    }
+
+    pub fn iter_tails_noendchecked(self) -> TailsIter {
```
shaleh: This is the short term solution.

rust_src/src/lists.rs
shaleh: To resolve this, the trait `LispConsIterator` has been added which has `Iterator` has a supertrait.
Also, while cleaning up the Iterators I added a macro to create the Car based Iterators.

The diff is messy. Avoid it and look at the actual Rust code.

shaleh: Nevermind, code is happy now.

rust_src/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 #![feature(never_type)]
 #![feature(const_fn_union)]
 #![feature(ptr_offset_from)]
+#![feature(self_struct_ctor)]
```
shaleh: this allows a constructor to use `Self(foo)` instead of `SomeType(foo)`.

rust_src/src/lisp.rs
```diff
@@ -450,23 +450,25 @@ impl LispObject {
 /// `$data` should be an `alist` and `$iter_item` type should implement `From<LispObject>`
 macro_rules! impl_alistval_iter {
     ($iter_name:ident, $iter_item:ty, $data: expr) => {
-        pub struct $iter_name {
-            tails: CarIter,
-        }
+        pub struct $iter_name(CarIter);
```
shaleh: Make a `newtype`. Simpler.

rust_src/src/lisp.rs
```diff
@@ -450,23 +450,25 @@ impl LispObject {
...
 
         impl $iter_name {
             pub fn new() -> Self {
-                Self {
-                    tails: CarIter::new($data, Some(Qlistp)),
-                }
```
shaleh: Same behavior. Nothing has changed.

rust_src/src/lists.rs
```diff
@@ -51,70 +51,141 @@ impl LispObject {
         self.is_cons() || self.is_nil()
     }
 
+    pub fn iter_tails_v2(
```
shaleh: Nothing calls this function right now.

shaleh:
OK. This has a much nicer to read diff. Yes, there is a little more code here than is strictly needed. However, no existing functionality is affected. Plus, this opens the door for smaller, directed PRs to use the new functionality.

shaleh:
This is ready to merge.

PR #1126, opened 2018-12-05
Port ding, redisplay

nickdrozd:


rust_src/src/dispnew.rs
```diff
@@ -136,4 +143,65 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
     !win.cursor_off_p()
 }
 
+/***********************************************************************
```
shaleh: We have left these comments behind in the past.

rust_src/src/dispnew.rs
```diff
@@ -136,4 +143,65 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
...
+	   Terminal Related Lisp Functions
+***********************************************************************/
+
+/// Beep, or flash the screen.
+/// Also, unless an argument is given,
+/// terminate any keyboard macro currently executing.
+#[lisp_fn(min = "0")]
+pub fn ding(arg: Option<LispObject>) {
+    if let Some(_) = arg {
```
shaleh: This would read better as a `match`.

```
match arg {
    Some(_) => {},
    None => {},
}
```

rust_src/src/dispnew.rs
```diff
@@ -136,4 +143,65 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
...
+/// preempted by arriving input, even if `redisplay-dont-pause' is nil.
+/// If `redisplay-dont-pause' is non-nil (the default), redisplay is never
+/// preempted by arriving input, so FORCE does nothing.
+///
+/// Return t if redisplay was performed, nil if redisplay was preempted
+/// immediately by pending input.
+#[lisp_fn(min = "0")]
+pub fn redisplay(force: Option<LispObject>) -> bool {
+    let force: bool = if let Some(_) = force { true } else { false };
+
```
shaleh: `let force = force.is_some();`

rust_src/src/dispnew.rs
```diff
@@ -136,4 +143,65 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
...
 
+/***********************************************************************
+	   Terminal Related Lisp Functions
+***********************************************************************/
+
+/// Beep, or flash the screen.
+/// Also, unless an argument is given,
+/// terminate any keyboard macro currently executing.
+#[lisp_fn(min = "0")]
+pub fn ding(arg: Option<LispObject>) {
```
shaleh: I do not see the gain in Option<LispObject>. If the user provides no arguments then arg will be Qnil.

nickdrozd:
```
 GEN      rust_src/src/dispnew-tests.log
/bin/bash: line 2: 10241 Illegal instruction     (core dumped) HOME=/nonexistent EMACSLOADPATH= LC_ALL=C EMACS_TEST_DIRECTORY=/home/travis/build/Wilfred/remacs/test "../src/remacs" -batch --no-site-file --no-site-lisp -L ":." -l ert -l rust_src/src/dispnew-tests.el --eval "(ert-run-tests-batch-and-exit (quote (not (or (tag :expensive-test) (tag :unstable)))))" > rust_src/src/dispnew-tests.log 2>&1
Running 2 tests (2018-12-06 05:45:12+0000)
make[3]: *** [rust_src/src/dispnew-tests.log] Error 132
```

rust_src/src/dispnew.rs
nickdrozd: My general thinking is that it's good to give the type checker as much logic as possible. In this case we take one argument, and it's a `LispObject`. But we know at compile time that its value doesn't matter at all provided that it's non-nil. That is, all that matters is whether it's nil or not. Thus we can tell the compiler to look for just two possible (sets of) values. We can't use `bool`, because any non-nil value is a valid argument, eg `(ding "ding")`, and that can't restricted. But `Option` is another way of representing two possibilities, so `Option<LispObject>` works.

Actually, even that's too broad. It also doesn't matter if the thing is a `LispObject` or not, since its value is never used. Really what it should be is something like `Option<T> where LispObject: From<T>`. Unfortunately it doesn't seem to be possible to use `<T>` with `lisp_fn` at the moment.

This isn't important for `ding` specifically of course, but it's something to think about generally, since this pattern comes up a lot.

rust_src/src/dispnew.rs
nickdrozd: There are a few more of these in `dispnew.rs`. Should they be removed?

rust_src/src/dispnew.rs
shaleh: I see that view. I largely share it. However, we are implementing the runtime for a typeless system. And at some point we will care about efficiency. `LispObject` is an `Option` type already. It is nil or a value. I hope as we become independent of the C that `LispObject` can be a newtype for `Option<LispValue>` and nil becomes `None`. Until then, we should be wise with what we have.

rust_src/src/dispnew.rs
shaleh: I personally like the grouping comments. We have been copying code without worrying about how easy it is to ready. At some point we are going to need to take some time and make the project readable, discoverable, etc.

I am not certain bringing the block descriptors now matters.

shaleh:
Are you using Linux or can you launch a Linux VM easily? The rr debugger makes catching these failures much nicer than lldb or gdb.

shaleh:
Are you using Linux or can you launch a Linux VM easily? The rr debugger makes catching these failures much nicer than lldb or gdb.

nickdrozd:
```
(gdb) disassemble ding
Dump of assembler code for function remacs::dispnew::ding:
   0x00000000006f27e0 <+0>:	sub    $0x28,%rsp
   0x00000000006f27e4 <+4>:	mov    %rdi,0x18(%rsp)
   0x00000000006f27e9 <+9>:	mov    0x18(%rsp),%rdi
   0x00000000006f27ee <+14>:	callq  0xc83990 <remacs::lisp::LispObject::is_nil>
   0x00000000006f27f3 <+19>:	mov    %al,0x17(%rsp)
   0x00000000006f27f7 <+23>:	mov    0x17(%rsp),%al
   0x00000000006f27fb <+27>:	test   $0x1,%al
   0x00000000006f27fd <+29>:	jne    0x6f2801 <remacs::dispnew::ding+33>
   0x00000000006f27ff <+31>:	jmp    0x6f280c <remacs::dispnew::ding+44>
   0x00000000006f2801 <+33>:	lea    -0x2cb592(%rip),%rax        # 0x427276 <bitch_at_user>
   0x00000000006f2808 <+40>:	callq  *%rax
=> 0x00000000006f280a <+42>:	ud2    
   0x00000000006f280c <+44>:	lea    0xd71531(%rip),%rax        # 0x1463d44 <noninteractive>
   0x00000000006f2813 <+51>:	testb  $0x1,(%rax)
   0x00000000006f2816 <+54>:	je     0x6f2829 <remacs::dispnew::ding+73>
   0x00000000006f2818 <+56>:	mov    $0x7,%edi
   0x00000000006f281d <+61>:	callq  *0xa3f6ad(%rip)        # 0x1131ed0
   0x00000000006f2823 <+67>:	mov    %eax,0x10(%rsp)
   0x00000000006f2827 <+71>:	jmp    0x6f284f <remacs::dispnew::ding+111>
   0x00000000006f2829 <+73>:	callq  0x6af8a0 <remacs::frames::selected_frame>
   0x00000000006f282e <+78>:	mov    %rax,0x20(%rsp)
   0x00000000006f2833 <+83>:	lea    0x20(%rsp),%rdi
   0x00000000006f2838 <+88>:	callq  0x6b45b0 <<remacs::lisp::ExternalPtr<T>>::as_mut>
   0x00000000006f283d <+93>:	mov    %rax,0x8(%rsp)
   0x00000000006f2842 <+98>:	mov    0x8(%rsp),%rdi
   0x00000000006f2847 <+103>:	callq  *0xa414a3(%rip)        # 0x1133cf0
   0x00000000006f284d <+109>:	jmp    0x6f284f <remacs::dispnew::ding+111>
   0x00000000006f284f <+111>:	add    $0x28,%rsp
   0x00000000006f2853 <+115>:	retq   
End of assembler dump.
```
One of those instructions does not belong.

nickdrozd:
The failure comes from executing `(ding)`. It takes the `arg.is_nil()` branch and then AFAICT the `bitch_at_user()` call completes successfully, after which that `ud2` instruction raises an error and it crashes.

shaleh:
I have a guess.

`bitch_at_user` is marked in Rust as never returning. However, if you look at the function in C sometimes it does return after dinging the bell while others it raise a signal. You could try removing both the blacklist in build.rs and the potentially incorrect entry in remacs_sys.rs and see if it just magically works. If that fails, perhaps we can fix what is broken there.

It is also a relatively simple function. We could port it to Rust and see if that helps.

nickdrozd:
DING DING DING

nickdrozd:
`bitch_at_user` uses a C macro called `INTERACTIVE`:

```c
/* Nonzero if input is coming from the keyboard.  */

#define INTERACTIVE (NILP (Vexecuting_kbd_macro) && !noninteractive)
```

Is there a good way to deal with macros like that? Just write a function?

rust_src/src/dispnew.rs
```diff
@@ -136,4 +143,57 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
...
+        unsafe {
+            if noninteractive {
+                putchar_unlocked(7);
+            } else {
+                ring_bell(selected_frame().as_mut())
+            }
+        }
+    }
+}
+
```
shaleh: bitch_at_user is a version of `ding` that sometimes signals.....

```
void
bitch_at_user (void)
{
  if (noninteractive)
    putchar_unlocked (07);
  else if (!INTERACTIVE)  /* Stop executing a keyboard macro.  */
    {
      const char *msg
	= "Keyboard macro terminated by a command ringing the bell";
      Fsignal (Quser_error, list1 (build_string (msg)));
    }
  else
    ring_bell (XFRAME (selected_frame));
}
```

shaleh: It seems to me that `ding` would be written:

```
  if (noninteractive)
    putchar_unlocked (07);
  else if (NILP(arg) && !INTERACTIVE)  /* Stop executing a keyboard macro.  */
    {
      const char *msg
	= "Keyboard macro terminated by a command ringing the bell";
      Fsignal (Quser_error, list1 (build_string (msg)));
    }
  else
    ring_bell (XFRAME (selected_frame));
}
```

shaleh: I'd port `bitch_at_user` as above and call it `ding_internal`. Have `ding` and `bitch_at_user` call it and `bitch_at_user` always passes in NIL.

shaleh: As an aside, that is an archaic name for a function that we should consider retiring.

rust_src/src/dispnew.rs
```diff
@@ -143,13 +143,34 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
     !win.cursor_off_p()
 }
 
+/// Return whether input is coming from the keyboard.
+// Corresponds to the INTERACTIVE macro in commands.h.
+pub fn is_interactive() -> bool {
+    unsafe { globals.Vexecuting_kbd_macro.is_nil() && !noninteractive }
+}
+
```
shaleh: If we gave `ding_internal` an argument we could collapse the Lisp ding and ding_internal code. The only difference between them is whether to call when interactive but currently in a non-interactive mode.

rust_src/src/dispnew.rs
```diff
@@ -143,13 +143,34 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
...
+#[no_mangle]
+pub extern "C" fn ding_internal() {
+    unsafe {
+        if noninteractive {
+            putchar_unlocked(7);
```
shaleh: Beware the Emacs source number formats. The `7` here was `07` which in C is octal. https://www.geeksforgeeks.org/octal-numbers-c/

In this case `07` is decimal 7. So it works out.

rust_src/src/dispnew.rs
```diff
@@ -143,13 +143,34 @@ pub fn internal_show_cursor_p(window: LispWindowOrSelected) -> bool {
...
+        } else if !is_interactive() {
+            // Stop executing a keyboard macro.
+            let msg = "Keyboard macro terminated by a command ringing the bell";
+            Fsignal(Quser_error, list!(msg.into()));
```
shaleh: We usually use `xsignal!` in the Rust code.

nickdrozd:
@shaleh Anything else?

Issue #886, opened 2018-10-01
Can't build GNU Emacs or Remacs on macOS 10.14 Mojave - fatal error: 'libxml/tree.h' file not found


cjohansson:
Might be off-topic but I get this error on macOS 10.14 Mojave, after running `make`:

`xml.c:23:10: fatal error: 'libxml/tree.h' file not found`

Both `xcode-select --install` and `brew install libxml2` is installed. I get the same error on GNU Emacs as well. Do you get this error or do you know about a solution to it?

I didn't get this error before updating macOS to Mojave.

shaleh:
I have bit upgraded yet.

socketwiz:
I also upgraded to Mojave and I ran into this problem.  I solved it by using [homebrew](https://brew.sh/) to install `libxml2`:

```shell
$ brew install libxml2
```
If you have already installed `libxml2` then you can see what environment variables it wants you to set by doing `brew info libxml2`.  Here is what it said on my machine.

```shell
$ export LDFLAGS="-L/usr/local/opt/libxml2/lib"
$ export CPPFLAGS="-I/usr/local/opt/libxml2/include"
$ export PKG_CONFIG_PATH="/usr/local/opt/libxml2/lib/pkgconfig"
```

Once you have your environment updated with `libxml2` then you need to re-run `configure` so it can pickup `libxml2` from the environment you've setup.

```shell
$ ./configure --enable-rust-debug
```

Now you should be able to re-run make.  It built to completion for me when I did this.  It took just over 2 hours on my machine.

```shell
$ make
```

db48x:
> $ export PKG_CONFIG_PATH="/usr/local/opt/libxml2/lib/pkgconfig"

This is all you need. The `configure` script will run `pkg-config` to find out what to add to `LDFLAGS` and `CPPFLAGS`.

cjohansson:
Yes it's working @db48x , I have this in my `~/.bash_profile` now:

```bash
export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/local/opt/libxml2/lib/pkgconfig:/opt/X11/li
    b/pkgconfig:$PKG_CONFIG_PATH
```

Running `brew link libxml2` gives that as an advice as well but I only tried the first two advices (LDFLAGS and CPPFLAGS) which didn't work.

[...]

PR #1129, opened 2018-12-07
Updated Mojave build instructions for libxml2

cjohansson:
[#886](https://github.com/Wilfred/remacs/issues/886)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ more Emacs-y.
     To use the installed version of `makeinfo` instead of the built-in 
     (`/usr/bin/makeinfo`) one, you'll need to make sure `/usr/local/opt/texinfo/bin` 
     is before `/usr/bin` in `PATH`.
-    Mojave: `export PKG_CONFIG_PATH="/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg"`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -149,12 +149,12 @@ more Emacs-y.
...
    MacOS:
    
-        brew install libxml2 gnutls texinfo
+        brew install gnutls texinfo
         
     To use the installed version of `makeinfo` instead of the built-in 
     (`/usr/bin/makeinfo`) one, you'll need to make sure `/usr/local/opt/texinfo/bin` 
     is before `/usr/bin` in `PATH`.
-    Mojave: `export PKG_CONFIG_PATH="/usr/local/opt/libxml2/lib/pkgconfig"`
+    Mojave: `export PKG_CONFIG_PATH="/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg"`
```
shaleh: Does the user need to install this first? Or is only pointing at the `.pkg` sufficient.

cjohansson: No I think this should be available once Xcode Command Line tools are installed and they are a requirement for compiling stuff on macOS, you install them with `xcode-select --install`

cjohansson:
I added instructions for installing the headers, it seems the export command doesn't work, it must have been something else in my environment that made it appear to work last time

shaleh:
Thank you for your diligence @cjohansson.

PR #1130, opened 2018-12-08
[WIP] add libvterm support

brotzeit:
Thanks to [emacs-libvterm](https://github.com/akermu/emacs-libvterm) vterm seems to work in remacs.
I just pulled in the code from the module and I will now start the cleanup. 
Any kind of help is appreciated, but the PR isn't ready for a review. If you want to improve something just push to this branch.

TODO:
- [ ] use pseudovector instead of Lisp_Misc_Save_Value
- [ ] use libvterm functions where possible
- [ ] update make stuff

### DIFF
--- a/rust_src/src/vterm.rs
+++ b/rust_src/src/vterm.rs
@@ -2,7 +2,7 @@
 
 use remacs_macros::lisp_fn;
 
-use libc::{c_char, c_int, c_uchar, c_void};
+use libc::{c_char, c_int, c_uchar, c_void, size_t};
 
 use std::{cmp, mem};
 
@@ -16,8 +16,8 @@ use crate::{
     multibyte::LispStringRef,
     obarray::intern,
     remacs_sys::{
-        allocate_vterm, color_to_rgb_string, get_col_offset, mysave_value, refresh_lines,
-        rgb_string_to_color, row_to_linenr, term_process_key, vterm_output_read,
+        allocate_vterm, color_to_rgb_string, get_col_offset, is_key, mysave_value, refresh_lines,
+        rgb_string_to_color, row_to_linenr, term_process_key, utf8_to_codepoint, vterm_output_read,
         vterm_screen_callbacks, vterm_screen_set_callbacks, VtermScrollbackLine,
     },
 
@@ -29,12 +29,13 @@ use crate::{
 
     // vterm
     remacs_sys::{
-        vterm_get_size, vterm_input_write, vterm_new, vterm_obtain_screen, vterm_obtain_state,
-        vterm_output_get_buffer_current, vterm_screen_enable_altscreen, vterm_screen_flush_damage,
-        vterm_screen_reset, vterm_screen_set_damage_merge, vterm_set_size, vterm_set_utf8,
-        vterm_state_get_cursorpos, vterm_state_set_default_colors, vterm_state_set_palette_color,
-        VTermColor, VTermDamageSize, VTermModifier, VTermPos, VTermProp, VTermRect,
-        VTermScreenCell, VTermState, VTermValue,
+        vterm_get_size, vterm_input_write, vterm_keyboard_key, vterm_keyboard_unichar, vterm_new,
+        vterm_obtain_screen, vterm_obtain_state, vterm_output_get_buffer_current,
+        vterm_screen_enable_altscreen, vterm_screen_flush_damage, vterm_screen_reset,
+        vterm_screen_set_damage_merge, vterm_set_size, vterm_set_utf8, vterm_state_get_cursorpos,
+        vterm_state_set_default_colors, vterm_state_set_palette_color, VTermColor, VTermDamageSize,
+        VTermKey, VTermModifier, VTermPos, VTermProp, VTermRect, VTermScreenCell, VTermState,
+        VTermValue,
     },
     threads::ThreadState,
     vectors::length,
@@ -253,7 +254,7 @@ pub fn vterminal_update(
     unsafe {
         if string.is_not_nil() {
             let mut utf8 = code_convert_string_norecord(string, Qutf_8, true).as_string_or_error();
-            let mut len = STRING_BYTES(utf8.as_mut()) as isize + 1;
+            let mut len = STRING_BYTES(utf8.as_mut()) as usize;
 
             let mut v: Vec<c_uchar> = Vec::with_capacity(len as usize);
 
@@ -261,7 +262,7 @@ pub fn vterminal_update(
                 v.as_mut_ptr() as *mut c_void,
                 utf8.data_ptr() as *mut c_void,
                 len as libc::size_t,
-            ) as *mut c_uchar;
+            ) as *const c_char;
 
             let mut modifier = VTermModifier::VTERM_MOD_NONE;
             if shift {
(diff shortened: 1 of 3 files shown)


rust_src/src/vterm.rs
```diff
@@ -0,0 +1,418 @@
...
+    let buffer_lnum =
+        call!(LispObject::from(intern("line-number-at-pos"))).as_fixnum_or_error() as c_int;
+
+    let state: *mut VTermState = vterm_obtain_state((*term).vt);
+    let mut pos: VTermPos = std::mem::zeroed();
+    vterm_state_get_cursorpos(state, &mut pos);
+
+    let cursor_lnum = row_to_linenr(term, pos.row);
+    call!(
+        LispObject::from(intern("goto-line")),
```
shaleh: Same question about `goto-line` here.

rust_src/src/vterm.rs
```diff
@@ -0,0 +1,418 @@
...
+        buffer_lnum = ThreadState::current_buffer().zv as c_int;
+        let del_cnt = buffer_lnum - height - (*term).sb_size as c_int + (*term).sb_pending;
+
+        if del_cnt > 0 {
+            buffer_lnum = ThreadState::current_buffer().zv as c_int;
+        }
+
+        let buf_index = buffer_lnum - height + 1;
+        call!(
+            LispObject::from(intern("goto-line")),
```
shaleh: and here

rust_src/src/vterm.rs
```diff
@@ -0,0 +1,418 @@
...
+
+        // is_not_nil shouldn't be necessary
+        if output.is_not_nil() {
+            call!(LispObject::from(intern("vterm-flush-output")), output);
+        }
+    }
+}
+
+#[lisp_fn]
+pub fn vterm_write_input(input: LispObject) -> LispObject {
```
shaleh:
```suggestion
pub fn vterm_write_input(input: LispStringRef) -> LispObject {
```

rust_src/src/vterm.rs
```diff
@@ -0,0 +1,418 @@
...
+        let mut s = input.as_string_or_error();
+        vterm_input_write(
+            (*term).vt,
+            s.sdata_ptr(),
+            STRING_BYTES(s.as_mut()) as usize + 1,
+        );
+        vterm_screen_flush_damage((*term).vts);
+    }
+
+    LispObject::from(0)
```
shaleh: Why not just return int or EmacsInt?

src/vterm.c
```diff
@@ -0,0 +1,474 @@
...
+    length++;
+  }
+  Lisp_Object text = vterm_render_text(buffer, length, &lastCell);
+  call1(intern ("insert"), text);
+
+  return text; 
+}
+
+VTermScreenCallbacks vterm_screen_callbacks = {
+    .damage = vterm_damage,
```
leonerd: Curious. Is this attempting to make a set of callbacks that reuse the original `libvterm` functions in most of the slots, but just adds some scrollback handling? This feels a slightly awkward way to go about it - perhaps we can find a better way around that?

src/vterm.c
```diff
@@ -0,0 +1,474 @@
...
+    .movecursor = vterm_movecursor,
+    /* .settermprop = term_settermprop, */
+    .resize = vterm_resize,
+    .sb_pushline = term_sb_push,
+    .sb_popline = term_sb_pop,
+    
+};
+
+void
+vterm_sb_buffer_size () {
```
leonerd: I'd suggest not making new non-static functions in the `vterm_...` namespace; in case `libvterm` itself starts adding more.

src/vterm.c
```diff
@@ -0,0 +1,474 @@
...
+    vterm_keyboard_key(term->vt, VTERM_KEY_INS, modifier);
+  } else if (is_key(key, len, "<delete>")) {
+    vterm_keyboard_key(term->vt, VTERM_KEY_DEL, modifier);
+  } else if (is_key(key, len, "<home>")) {
+    vterm_keyboard_key(term->vt, VTERM_KEY_HOME, modifier);
+  } else if (is_key(key, len, "<end>")) {
+    vterm_keyboard_key(term->vt, VTERM_KEY_END, modifier);
+  } else if (is_key(key, len, "<prior>")) {
+    vterm_keyboard_key(term->vt, VTERM_KEY_PAGEUP, modifier);
+  } else if (is_key(key, len, "<f0>")) {
```
leonerd: Could maybe these be done parametrically by looking for prefix `<f` and suffix `>` then parsing the middle as being a decimal integer? That would extend beyond 12 then if required.

src/vterm.c
```diff
@@ -0,0 +1,474 @@
...
+    cells[col].width = 1;
+  }
+
+  free(sbrow);
+
+  return 1;
+}
+
+static size_t
+codepoint_to_utf8_c(const uint32_t codepoint, unsigned char buffer[4]) {
```
leonerd: This (and the subsequent `utf8_to_codepoint`) are very similar to functions that already exist in `libvterm`. It might be easier just to expose those publicly then you can use them directly.

src/vterm.c
```diff
@@ -0,0 +1,474 @@
...
+      if (cell.width > 1) {
+        offset += cell.width - 1;
+      }
+    }
+    col += cell.width;
+  }
+  return offset;
+}
+
+static bool compare_cells(VTermScreenCell *a, VTermScreenCell *b) {
```
leonerd: This too is almost identical to an internal `libvterm` function - might be best to expose and use that.

src/vterm.c
brotzeit: I just copied the code from emacs-libvterm and modified it so it works in remacs and didn't really take a look at libvterm, so far.

> perhaps we can find a better way around that

Sounds good =)
I don't know much, but libvterm seems to work really well. Thanks for your work!

src/vterm.c
brotzeit: True. I wanted to avoid confusion regarding term and vterm, but I will change it to something different.

src/vterm.c
brotzeit: Makes sense.

src/vterm.c
brotzeit: Good to know. I will make a TODO list.

brotzeit:
I tried to allocate the struct as a pseudovector similiar to terminal.
```c
Term *
allocate_vterm (void) {
  global_term =  ALLOCATE_ZEROED_PSEUDOVECTOR
    (struct Term, next_vterm, PVEC_VTERM);
}
```
I added the field
```c
struct Term *next_vterm;
```
to the term struct.

It compiles and runs, but after a few insertions remacs segfaults. Any idea ?

rust_src/src/vterm.rs
```diff
@@ -0,0 +1,418 @@
...
+    if (*term).invalid_end >= (*term).invalid_start {
+        let mut height: i32 = 0;
+        let mut width: i32 = 0;
+        vterm_get_size((*term).vt, &mut height, &mut width);
+
+        // Term height may have decreased before `invalid_end` reflects it.
+        let line_start = row_to_linenr(term, (*term).invalid_start);
+
+        call!(
+            LispObject::from(intern("goto-line")),
```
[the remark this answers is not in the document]
shaleh: Calling lisp is ok. Calling `interactive` lisp is frowned on.

brotzeit:
> Are you using malloc directly or are you using lisp_malloc?

I thought the macro is doing the allocation for pseudovectors.

brotzeit:
@shaleh I removed `next` for now, but it still doesn't work.

```c
struct Term *
allocate_vterm (void) {
  return ALLOCATE_ZEROED_PSEUDOVECTOR
    (struct Term, vt, PVEC_VTERM);
}

struct Term {
  union vectorlike_header header;

  Lisp_Object dummy;

  VTerm *vt;
  VTermScreen *vts;
  // buffer used to:
  //  - convert VTermScreen cell arrays into utf8 strings
  //  - receive data from libvterm as a result of key presses.
  VtermScrollbackLine **sb_buffer; // Scrollback buffer storage for libvterm  
  VtermCursor cursor;
  
  bool is_invalidated;
  
  size_t sb_current;          // number of rows pushed to sb_buffer
  size_t sb_size;             // sb_buffer size
  // "virtual index" that points to the first sb_buffer row that we need to
  // push to the terminal buffer when refreshing the scrollback. When negative,
  // it actually points to entries that are no longer in sb_buffer (because the
  // window height has increased) and must be deleted from the terminal buffer
  int sb_pending;

  int invalid_start, invalid_end; // invalid rows in libvterm screen
};
```

I've read the docs a little more and when I understand correctly non LispObjects has to be after the last LispObject and you tell `ALLOCATE_ZEROED_PSEUDOVECTOR` with the second argument where it has to start allocating different types.

brotzeit:
@db48x @Wilfred any idea ?

brotzeit:
It seems bindgen has problems handling the libvterm enum `VTermProp`.

```
error[E0573]: expected type, found module `VTermProp`
   --> src/vterm.rs:537:11
    |
537 |     prop: VTermProp,
    |           ^^^^^^^^^ did you mean `VTermPos`?
```

I've never seen this before. Any idea ?

PR #1139, opened 2018-12-14
Port find-font function to Rust

Compro-Prasad:


rust_src/src/fonts.rs
```diff
@@ -131,4 +133,15 @@ pub fn font_match_p(spec: LispObject, font: LispObject) -> bool {
     unsafe { c_font_match_p(spec, font) }
 }
 
+/// Return a font-entity matching with FONT-SPEC on the current frame.
+/// Optional 2nd argument FRAME, if non-nil, specifies the target frame.
+#[lisp_fn]
+pub fn rust_find_font(spec: LispObject, frame: LispObject) -> LispObject {
+    let mut val = unsafe { Flist_fonts(spec, frame, LispObject::from(1), Qnil) };
+    if consp(val) {
```
shaleh: instead of `consp` you can use `if let` style.

```
if let Some(cons) = val.as_cons() {
    cons.car()
} else {
    val
}
```

rust_src/src/fonts.rs
```diff
@@ -131,4 +133,15 @@ pub fn font_match_p(spec: LispObject, font: LispObject) -> bool {
     unsafe { c_font_match_p(spec, font) }
 }
 
+/// Return a font-entity matching with FONT-SPEC on the current frame.
+/// Optional 2nd argument FRAME, if non-nil, specifies the target frame.
+#[lisp_fn]
+pub fn rust_find_font(spec: LispObject, frame: LispObject) -> LispObject {
+    let mut val = unsafe { Flist_fonts(spec, frame, LispObject::from(1), Qnil) };
```
shaleh: no need for `mut` if you use the style below.

rust_src/src/fonts.rs
```diff
@@ -131,4 +133,15 @@ pub fn font_match_p(spec: LispObject, font: LispObject) -> bool {
     unsafe { c_font_match_p(spec, font) }
 }
 
+/// Return a font-entity matching with FONT-SPEC on the current frame.
+/// Optional 2nd argument FRAME, if non-nil, specifies the target frame.
+#[lisp_fn]
```
shaleh: There is a tool in admin/ called `rustize-emacs`. Give it a go. If the second argument can be left out then this needs a `min`.
```suggestion
#[lisp_fn(min = "1")]
```

rust_src/src/fonts.rs
shaleh: or `match`

```
match val.as_cons() {
    Some(cons) => cons.car(),
    None => val,
}
```

shaleh: or `val.as_cons().map_or(val, |cons| cons.car())`

shaleh: I would probably choose the `match`.

rust_src/src/fonts.rs
Compro-Prasad: Yep. Forgot that.

Compro-Prasad:
I think its ready to be merged.

PR #1146, opened 2018-12-17
XML2 optional compilation support

### LINKED ISSUE #1075: Build fails when libxml2 library is absent
In case `libxml2` library is not installed, `./configure` step figures the library to be absent and as far as I understand, the corresponding C code is not built using `#define`s. But Rust code uses methods `init_libxml2_functions` in `xml.rs` file anyway and this leads to the compilation error -

```
error[E0432]: unresolved imports `crate::remacs_sys::init_libxml2_functions`, `crate::remacs_sys::parse_region`
 --> src/xml.rs:9:18
  |
9 |     remacs_sys::{init_libxml2_functions, parse_region},
  |                  ^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^ no `parse_region` in `remacs_sys`. Did you mean to use `write_region`?
  |                  |
  |                  no `init_libxml2_functions` in `remacs_sys`

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `remacs`.
```

Conditional compilation has to be propagated to Rust code too.

shaleh:
Set the feature "use-xml2" when the compilation support LIBXML2.

Closes #1075


harryfei:
@shaleh 

I use `cargo build --features xxx` to do the same thing in my `webrender` branch.

I think it's cleaner than this way. It doesn't need any `Cargo.toml.in` template file.

https://github.com/harryfei/remacs/blob/webrender/configure.ac#L4930

shaleh:
That also works with what I added.

I added the template because we will have a few of these. As a developer, I don't want to remember to pass gobs of command line options. As I have it set up a simple `cargo check` or `cargo build` will work.

harryfei:
Okay. It makes sense.

shaleh:
Once we are self hosting on Rust the `configure` script style may not work for us. At that point we will need to look into how we want to handle this. Asking people to edit the `default` line in `[features]` might be how we do that. But today we use configure and it makes sense to have the Rust build set up with the same options as the C build. Otherwise there is bound to be confusion.

Thanks for asking @harryfei . I should have been more explicit.

Issue #1140, opened 2018-12-14
build.rs fails with "no such file or directory" error


agraven:
Building fails with a panic in build.rs with a panic, giving the message
```
thread 'main' panicked at 'called 'Result::unwrap()' on an 'Err' value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', libcore/result.rs:1009:5
```

The relevant line in the stack trace is `9: build_script_build::main`. It for some reason does not print which line number the error occurs in.

I am running Ubuntu Comsic (18.10), and gcc version 8.2.0. I did not give configure any arguments, but I did supply make with the argument `EMACS_CFLAGS:="-Isrc -isystem /usr/include/c++/8/tr1"`, since otherwise it failed with the error of not being able to find `stdbool.h`

agraven:
I added the details to the original post.

shaleh:
There is also this noise which needs to be cleaned up when using the new gcc.

```
cc1: warning: -Wabi won't warn about anything [-Wabi]
cc1: note: -Wabi warns about differences from the most up-to-date ABI, which is also used by default
cc1: note: use e.g. -Wabi=11 to warn about changes from GCC 7
```

shaleh:
I cannot reproduce the failure in a Docker container. #1141 has the updates needed to make a Cosmic based container.

shaleh:
I just saw this failure on a Debian machine that I upgraded from Stretch to unstable aka not quite Buster. Have not had a chance to debug it yet.

I also saw that we still fail to always detect clang during configure. I needed to install libclang-dev.

shaleh:
I installed the `clang-7` package and poof, it worked.

agraven:
This fixed it for me as well. I suppose `clang-7` should be added to the list of required packages in README.md then?

shaleh:
Should we close this @agraven or is there something else we should fix?

agraven:
No, the solution has been found, so I'm closing it.

PR #1147, opened 2018-12-17
Add clang-7 to list of dependencies

agraven:
This should prevent encountering any issues with a missing `stdbool.h`, as seen in issue #1140.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ more Emacs-y.
 
         apt install texinfo libjpeg-dev libtiff-dev \
           libgif-dev libxpm-dev libgtk-3-dev libgnutls28-dev \
-          libncurses5-dev libxml2-dev libxt-dev
+          libncurses5-dev libxml2-dev libxt-dev clang-7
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ more Emacs-y.
 
         apt install texinfo libjpeg-dev libtiff-dev \
           libgif-dev libxpm-dev libgtk-3-dev libgnutls28-dev \
-          libncurses5-dev libxml2-dev libxt-dev
+          libncurses5-dev libxml2-dev libxt-dev clang-7
```
shaleh: Instead of putting it here, let's update the section right above it where it suggests installing `build-essential`. It also suggests installed `clang` which in my case at least should have brought in `clang-7`. Let's break out that apt section to resemble the section below so it is clear that they should run this. Also, let's add `clang-dev` since people usually miss it.

agraven: By `clang-dev` I assume you're referring to `libclang-dev`?

PR #1148, opened 2018-12-17
Port load-average

carlosmn:


rust_src/src/fns.rs
```diff
@@ -306,4 +308,39 @@ pub extern "C" fn internal_equal_misc(
...
+#[lisp_fn(min = "0")]
+pub fn load_average(use_floats: LispObject) -> LispObject {
+    let mut load_avg: [libc::c_double; 3] = [0.0, 0.0, 0.0];
+    let loads = unsafe { libc::getloadavg(load_avg.as_mut_ptr(), 3) };
+
+    if loads < 0 {
+        error!("load-average not implemented for this operating system");
+    }
+
+    (0..loads as usize).rev().fold(Qnil, |acc, i| {
```
nickdrozd: `From<Vec<LispObject>> for LispObject` is implemented, so I think you can return `loads` as-is (or map whatever float converter) rather than manually constructing a Lisp list.

carlosmn: So it is. I've updated it so we generate a `Vec<LispObject>` and we generate the list out of that with `From`.

shaleh:
rust-mode has an option to call `rustfmt` on every save. I highly recommend it.

rust_src/src/fns.rs
```diff
@@ -306,4 +308,39 @@ pub extern "C" fn internal_equal_misc(
...
+///
+/// If the 5-minute or 15-minute load averages are not available, return a
+/// shortened list, containing only those averages which are available.
+///
+/// An error is thrown if the load average can't be obtained.  In some
+/// cases making it work would require Emacs being installed setuid or
+/// setgid so that it can read kernel information, and that usually isn't
+/// advisable.
+#[lisp_fn(min = "0")]
+pub fn load_average(use_floats: LispObject) -> LispObject {
```
```suggestion
pub fn load_average(use_floats: bool) -> LispObject {
```

Can `use_floats` be type 'bool`?

carlosmn: It looks like it can be. I just tested it out and we do get `true` for anything non-`nil` which is the behaviour I was trying to keep by accepting the `LispObject`.

nickdrozd: That alters the API. `(load-average t)` and `(load-average "asdf")` should return the same thing. It's not a pattern that fits in very well with the Rust compiler, which is what motivated me to try to use `Option` in https://github.com/Wilfred/remacs/pull/1126.

rust_src/src/fns.rs
```diff
@@ -306,4 +308,41 @@ pub extern "C" fn internal_equal_misc(
...
+///
+/// If the 5-minute or 15-minute load averages are not available, return a
+/// shortened list, containing only those averages which are available.
+///
+/// An error is thrown if the load average can't be obtained.  In some
+/// cases making it work would require Emacs being installed setuid or
+/// setgid so that it can read kernel information, and that usually isn't
+/// advisable.
+#[lisp_fn(min = "0")]
+pub fn load_average(use_floats: bool) -> LispObject {
```
nickdrozd: You can change the return type to `Vec<LispObject>` and drop the final `LispObject::from`. The `lisp_fn` macro uses magic to convert everything to `LispObject`. `memory_use_counts` in `alloc.rs` is another example of this.

nickdrozd: In this case, we know that the result will be a list of ints or a list of floats, so even `Vec<LispObject>` is too general. But I don't know how to make it more specific. Ideally converting a `Vec` to `LispObject` would recursively convert its items as well. Is there a way to do that? A generic type or something like that?

(That would be outside the scope of this PR of course.)

rust_src/src/fns.rs
shaleh: @nickdrozd anything that is truthy will become true and nil is false. It should work as expected no?

rust_src/src/fns.rs
shaleh: I think we would need to write a bunch of `impl From<Vec<EmacsInt>> for LispObject` style code. Should not be too complicated. Might even be able to semi-automate it with a macro.

rust_src/src/fns.rs
shaleh: @nickdrozd in this case `use_floats` is not consumed by any of the functions inside `load_average`. In your PR it was.

rust_src/build.rs
```diff
@@ -341,7 +341,7 @@ fn handle_file(mod_path: &PathBuf) -> Result<Option<ModuleData>, BuildError> {
                     e.kind(),
                     format!("Failed to open {}: {}", mod_info.path.to_string_lossy(), e),
                 )
-                .into())
+                .into());
```
shaleh: ? Where did this come from? Is your rustfmt not matching ours?

rust_src/src/dispnew.rs
```diff
@@ -55,7 +55,7 @@ pub fn sleep_for(seconds: EmacsDouble, milliseconds: Option<EmacsInt>) {
 }
 
 /**********************************************************************
-		    Redrawing Frames
```
shaleh: Also a new, unexpected change.

rust_src/src/frames.rs
```diff
@@ -446,13 +446,13 @@ pub fn next_frame(frame: LispFrameOrSelected, miniframe: LispObject) -> LispFram
     // a valid candidate will be returned regardless of its position.
     while passed < 2 {
         for_each_frame!(f => {
-	    if passed > 0 {
```
shaleh: And more unexpected changes...

rust_src/build.rs
carlosmn: it did come from rustfmt. I just installed `rustfmt-nightly` seems I seem to need that with the toolchain configured in the repo. Unfortunately I did make sure to only add the file I touched, but then I reflexibly did `git commit -am` so I still sent everything over. I'll force-push with the correct commit.

shaleh: You want `rustup component add rustfmt-preview`.

rust_src/src/fns.rs
shaleh: @carlosmn the thing to watch is where a function uses a value as a bool in itself but passes it on elsewhere and the original type is needed.

Something like this for instance:

    pub fn foo(value: LispObject) -> LispObject {
        if value.is_nil() {
            other_function(1)
        else {
            other_function(value)
        }
    }

In this case `other_function` is actually looking for a number and if we just converted the bool to a number it would be '1' which is not ok.

rust_src/src/fns.rs
carlosmn: The returned values are either floats or integers, so a type that describes that would probably have to be some wrapping type like `enum EmacsNumber { Double(EmacsDouble), Int(EmacsInt) }` which should then be able to again use `From` on the inner type.

shaleh: `LispNumber` is our wrapper for a thing that can be a float or int.

carlosmn: Ah there it is. I have some changes locally that do let us return `Vec<LispNumber>` with what seems like a reasonable `impl From<Vec<LispNumber>> for LispObject` but that's getting out of scope for this change.

shaleh: yeah. If it is worth a follow on PR, then cool.

rust_src/src/fns.rs
nickdrozd: `lisp_fn` is full of surprises. Just to assuage my lingering doubts though, could you add a test case that verifies that `(load-average)` returns a list of ints and `(load-average "asdf")` returns a list of floats? Or any other weird inputs (`(load-average '(list of floats))`, for instance).

shaleh: @nickdrozd perhaps we should dedicate some code to validating that `lisp_fn` behaves as expected. That way we 1) have less doubt and 2) have something to test any changes against.

rust_src/src/fns.rs
```diff
@@ -306,4 +308,39 @@ pub extern "C" fn internal_equal_misc(
...
+/// shortened list, containing only those averages which are available.
+///
+/// An error is thrown if the load average can't be obtained.  In some
+/// cases making it work would require Emacs being installed setuid or
+/// setgid so that it can read kernel information, and that usually isn't
+/// advisable.
+#[lisp_fn(min = "0")]
+pub fn load_average(use_floats: bool) -> Vec<LispObject> {
+    let mut load_avg: [libc::c_double; 3] = [0.0, 0.0, 0.0];
+    let loads = unsafe { libc::getloadavg(load_avg.as_mut_ptr(), 3) };
```
shaleh: This is a Unix call. It works on Mac, Linux, BSD, etc. But it will bomb on Windows.

shaleh: There is code in w32.c to emulate it. You could wrap this in a `cfg!()` check.

rust_src/src/fns.rs
```diff
@@ -306,4 +309,43 @@ pub extern "C" fn internal_equal_misc(
...
+/// shortened list, containing only those averages which are available.
+///
+/// An error is thrown if the load average can't be obtained.  In some
+/// cases making it work would require Emacs being installed setuid or
+/// setgid so that it can read kernel information, and that usually isn't
+/// advisable.
+#[lisp_fn(min = "0")]
+pub fn load_average(use_floats: bool) -> Vec<LispObject> {
+    let mut load_avg: [libc::c_double; 3] = [0.0, 0.0, 0.0];
+    let loads = if cfg!(windows) {
```
shaleh: This style does not pass clippy.

What you need to do is make a `getloadavg` function which compiles to either the remacs_sys version or the libc version and call that function here.

carlosmn: With all the same rustup, cargo and rustc versions I see at https://travis-ci.org/Wilfred/remacs/jobs/469420967 I `rustup component add clippy-preview` but running `make clippy` does not show me the same error. I've pushed up something that should be OK and I'll see if the Travis output agrees.

rust_src/src/remacs_sys.rs
```diff
@@ -105,6 +105,8 @@ extern "C" {
     pub fn allocate_misc(t: Lisp_Misc_Type) -> LispObject;
     #[cfg(windows)]
     pub fn file_attributes_c(filename: LispObject, id_format: LispObject) -> LispObject;
+    #[cfg(windows)]
```
shaleh: Did you have to add this for it to compile?

rust_src/src/fns.rs
shaleh: Success!

rust_src/src/fns.rs
shaleh: The tests look reasonable.

rust_src/src/remacs_sys.rs
carlosmn: I didn't try without it as it looked like other W32 functions were there as well so I followed the pattern. I can try removing it but as neither I nor CI build on Windows, I don't know if we'll actually notice if it doesn't work.

carlosmn: Just realised you probably just mean the `cfg` bit. It compiles without it, so it should be fine, I've added another commit to do that.

shaleh: Yeah, we used to have a working Windows CI but it was too flaky. I want to get one working again. Just not sure where to host it. Most people want to build things with Windows tools and we want to build with Rust and friends. Which is what broke the previous CI.

carlosmn:
It did turn out to be a lot more involved than expected. I guess  I should have known better than to start with a POSIX call :)

PR #1149, opened 2018-12-18
Port set-window-new-total

### LINKED ISSUE #973: Port set-window-new-total
```c
DEFUN ("set-window-new-total", Fset_window_new_total, Sset_window_new_total, 2, 3, 0,
       doc: /* Set new total size of WINDOW to SIZE.
Return SIZE.

Optional argument ADD non-nil means add SIZE to the new total size of
WINDOW and return the sum.

The new total size of WINDOW, if valid, will be shortly installed as
WINDOW's total height (see `window-total-height') or total width (see
`window-total-width').

Note: This function does not operate on any child windows of WINDOW.  */)
     (Lisp_Object window, Lisp_Object size, Lisp_Object add)
{
  struct window *w = decode_valid_window (window);

  CHECK_NUMBER (size);
  if (NILP (add))
    wset_new_total (w, size);
  else
    wset_new_total (w, make_number (XINT (w->new_total) + XINT (size)));

  return w->new_total;
}
```

mukuact:
This should resolve #973

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1074,6 +1074,41 @@ pub fn window_new_total(window: LispWindowValidOrSelected) -> LispObject {
     win.new_total
 }
 
+/// Setter for new total of Window
+pub extern "C" fn wset_new_total(win: &mut LispWindowRef, new_total: EmacsInt) {
+    win.new_total = new_total.into();
+}
+
+/// Set new total size of WINDOW to SIZE.
+/// WINDOW must be a valid window and defaults to the selected one.
+/// Return SIZE.
+///
+/// Optional argument ADD non-nil means add SIZE to the new total size of
+/// WINDOW and return the sum.
+///
+/// The new total size of WINDOW, if valid, will be shortly installed as
+/// WINDOW's total height (see `window-total-height') or total width (see
+/// `window-total-width').
+///
+/// Note: This function does not operate on any child windows of WINDOW.
+#[lisp_fn(min = "2")]
+pub fn set_window_new_total(
+    window: LispWindowValidOrSelected,
+    size: EmacsInt,
+    add: LispObject,
+) -> LispObject {
+    let win: LispWindowRef = window.into();
+
+    let new_total: EmacsInt = if add.is_nil() {
+        size
+    } else {
+        EmacsInt::from(win.new_total) + size
+    };
+    wset_new_total(&mut win, new_total);
+
+    win.new_total
+}
+
 #[no_mangle]
 pub extern "C" fn wset_update_mode_line(mut w: LispWindowRef) {
     // If this window is the selected window on its frame, set the
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1075,6 +1075,7 @@ pub fn window_new_total(window: LispWindowValidOrSelected) -> LispObject {
 }
 
 /// Setter for new total of Window
+#[no_mangle]
 pub extern "C" fn wset_new_total(win: &mut LispWindowRef, new_total: EmacsInt) {
     win.new_total = new_total.into();
 }
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1074,12 +1074,6 @@ pub fn window_new_total(window: LispWindowValidOrSelected) -> LispObject {
     win.new_total
 }
 
-/// Setter for new total of Window
-#[no_mangle]
-pub extern "C" fn wset_new_total(win: &mut LispWindowRef, new_total: EmacsInt) {
-    win.new_total = new_total.into();
-}
-
 /// Set new total size of WINDOW to SIZE.
 /// WINDOW must be a valid window and defaults to the selected one.
 /// Return SIZE.
@@ -1105,8 +1099,7 @@ pub fn set_window_new_total(
     } else {
         EmacsInt::from(win.new_total) + size
     };
-    wset_new_total(&mut win, new_total);
-
+    win.new_total = new_total.into();
     win.new_total
 }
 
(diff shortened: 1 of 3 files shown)


rust_src/src/windows.rs
```diff
@@ -1074,6 +1074,42 @@ pub fn window_new_total(window: LispWindowValidOrSelected) -> LispObject {
...
+    add: LispObject,
+) -> LispObject {
+    let win: LispWindowRef = window.into();
+
+    let new_total: EmacsInt = if add.is_nil() {
+        size
+    } else {
+        EmacsInt::from(win.new_total) + size
+    };
+    wset_new_total(&mut win, new_total);
```
shanavas786: You can directly assign `win.new_total` here.
I don't see any advantage in porting setter method `wset_new_total` as it will be removed from `C` code base once all of its callers are ported.

shanavas786: You can directly assign `win.new_total` here.
I don't see any advantage in porting setter method `wset_new_total` as it will be removed from `C` code base once all of its callers are ported.

mukuact: OK. that's right. 
However, I don't understand why the original C code uses setter instead of direct assignment. 
Anyway, I will directly assign win.new_total.

rust_src/src/windows.rs
```diff
@@ -1074,6 +1074,35 @@ pub fn window_new_total(window: LispWindowValidOrSelected) -> LispObject {
...
+/// The new total size of WINDOW, if valid, will be shortly installed as
+/// WINDOW's total height (see `window-total-height') or total width (see
+/// `window-total-width').
+///
+/// Note: This function does not operate on any child windows of WINDOW.
+#[lisp_fn(min = "2")]
+pub fn set_window_new_total(
+    window: LispWindowValidOrSelected,
+    size: EmacsInt,
+    add: LispObject,
```
agraven: It would be more in line with conventions to make `add` an `Option<LispObject>`  and then `match` on it, similarly to how the example in `README.md` does.

rust_src/src/windows.rs
```diff
@@ -1074,6 +1074,35 @@ pub fn window_new_total(window: LispWindowValidOrSelected) -> LispObject {
...
+) -> LispObject {
+    let mut win: LispWindowRef = window.into();
+
+    let new_total: EmacsInt = if add.is_nil() {
```
agraven: This type annotation is superfluous.

rust_src/src/windows.rs
mukuact: 👍

mukuact: I have a question!
Which the ADD would be if LispObject is nil, Some(Qnil) or None?

agraven: That'd be None

rust_src/src/windows.rs
shaleh: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=12215

rust_src/src/windows.rs
shaleh: We tend to skip `Option<LispObject>`. It does not add much _and_ you still need to check if the object is nil.

rust_src/src/windows.rs
```diff
@@ -1074,6 +1074,34 @@ pub fn window_new_total(window: LispWindowValidOrSelected) -> LispObject {
...
+/// The new total size of WINDOW, if valid, will be shortly installed as
+/// WINDOW's total height (see `window-total-height') or total width (see
+/// `window-total-width').
+///
+/// Note: This function does not operate on any child windows of WINDOW.
+#[lisp_fn(min = "2")]
+pub fn set_window_new_total(
+    window: LispWindowValidOrSelected,
+    size: EmacsInt,
+    add: Option<LispObject>,
```
shaleh: It looks like `add` can simply be a `bool`

shaleh:
Install `rustup component add rustfmt-preview` and then set your rust-mode to call rustfmt whenever a buffer is saved. `(setq rust-format-on-save t)`

rust_src/src/windows.rs
mukuact: It is the best that `add` is `LispObject`, I think.

The rust's `bool` cannot be put here because `add` is an optional arg. 
There's no implement for `From(Lispobject) for Some(bool)`, so `Some(bool)` cannot be here.
`Option<LispObject>` need checking `None` and checking `add.is_nil()`.
If this is `LispObject`, `add.is_nil()` is `nil` either there is no optional args or args is nil.

shaleh: I am ok with it staying `LispObject`. But you are perhaps confused about how the optional arguments are handled.

Qnil is passed as the argument if one is not provided, otherwise some other LispObject value is used. So the `From<bool>` in lisp.rs will work just find. Qnil is False and anything else is True.

For many of our functions we use `Option<T>`. Qnil becomes `None` and anything else becomes Some(T). The LispObject having a value of Qnil because the user did not provide or they sent `nil` on purpose both end up being the same `None`. Typically, the `as_foo()` method of `LispObject` is called to create a `Foo`. For instance if the function takes an optional LispWindow then `LispObject.as_window()` is called by the `lisp_fn` macro via the `Option<LispWindow>` From implementation.

rust_src/src/windows.rs
db48x: You want `add` to just be a `bool` in this case :)

mukuact: 👍

rust_src/src/windows.rs
shaleh: Perhaps this is a place where our documentation could use an improvement? If so, would you like to take a try at updating it?

PR #1153, opened 2018-12-19
Implement From<Vec<T>> for LispObject and convert one use

carlosmn:
This came out of the discussion in #1148 where the return value for `load-average`/`load_average` ended up being too generic.

This lets us return `Vec<LispNumber>` from a function we want to expose to lisp.

### DIFF
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -258,6 +258,19 @@ impl From<Vec<LispObject>> for LispObject {
     }
 }
 
+impl<T> From<Vec<T>> for LispObject
+where
+    LispObject: From<T>,
+{
+    default fn from(v: Vec<T>) -> LispObject {
+        list(
+            &v.into_iter()
+                .map(LispObject::from)
+                .collect::<Vec<LispObject>>(),
+        )
+    }
+}
+
 impl From<LispObject> for bool {
     fn from(o: LispObject) -> Self {
         o.is_not_nil()
--- a/rust_src/src/fns.rs
+++ b/rust_src/src/fns.rs
@@ -9,6 +9,7 @@ use crate::{
     lisp::defsubr,
     lisp::LispObject,
     lists::{assq, car, get, member, memq, put, LispCons},
+    numbers::LispNumber,
     obarray::loadhist_attach,
     objects::equal,
     remacs_sys::Vautoload_queue,
@@ -334,7 +335,7 @@ unsafe fn getloadaverage(loadavg: *mut libc::c_double, nelem: libc::c_int) -> li
 /// setgid so that it can read kernel information, and that usually isn't
 /// advisable.
 #[lisp_fn(min = "0")]
-pub fn load_average(use_floats: bool) -> Vec<LispObject> {
+pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
     let mut load_avg: [libc::c_double; 3] = [0.0, 0.0, 0.0];
     let loads = unsafe { getloadaverage(load_avg.as_mut_ptr(), 3) };
 
@@ -345,9 +346,9 @@ pub fn load_average(use_floats: bool) -> Vec<LispObject> {
     (0..loads as usize)
         .map(|i| {
             if use_floats {
-                LispObject::from(load_avg[i])
+                LispNumber::Float(load_avg[i])
             } else {
-                LispObject::from((100.0 * load_avg[i]) as i64)
+                LispNumber::Fixnum((100.0 * load_avg[i]) as i64)
             }
         })
         .collect()
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -18,6 +18,7 @@
 #![feature(const_fn_union)]
 #![feature(ptr_offset_from)]
 #![feature(self_struct_ctor)]
+#![feature(specialization)]
 
 extern crate errno;
 #[macro_use]
--- a/rust_src/src/numbers.rs
+++ b/rust_src/src/numbers.rs
@@ -187,16 +187,6 @@ impl From<LispNumber> for LispObject {
     }
 }
 
-impl From<Vec<LispNumber>> for LispObject {
-    fn from(v: Vec<LispNumber>) -> LispObject {
-        LispObject::from(
-            v.into_iter()
-                .map(LispObject::from)
-                .collect::<Vec<LispObject>>(),
-        )
-    }
-}
-
 impl LispObject {
     pub fn is_number(self) -> bool {
         self.is_fixnum() || self.is_float()
--- a/rust_src/src/numbers.rs
+++ b/rust_src/src/numbers.rs
@@ -187,6 +187,16 @@ impl From<LispNumber> for LispObject {
     }
 }
 
+impl From<Vec<LispNumber>> for LispObject {
+    fn from(v: Vec<LispNumber>) -> LispObject {
+        LispObject::from(
+            v.into_iter()
+                .map(LispObject::from)
+                .collect::<Vec<LispObject>>(),
+        )
+    }
+}
+
 impl LispObject {
     pub fn is_number(self) -> bool {
         self.is_fixnum() || self.is_float()


nickdrozd:
:+1: 

I still feel like I don't understand Rust at all. Why is it `into_iter` here rather than `iter`?

carlosmn:
`into_iter` consumes the object, whereas `iter` just borrows it. One big difference in this case is that we get `LispNumber` instead of `&LispNumber` that we'd get with `iter`. As we're consuming the input to the function anyway, it makes sense to tell the compiler we're not interested in the values after running the conversion.

In this case I would expect the compiler to realise it's all the same as the function is pretty small, but it does let us do `.map(LispObject::from)` whereas with `iter` we'd have to do `.map(|x| LispObject::from(*x))`. Because of the borrow we'd have to dereference (which I think copies).

shaleh:
I wanted to suggest we could be more generic:

```
@@ -187,12 +188,12 @@ impl From<LispNumber> for LispObject {
     }
 }
 
-impl From<Vec<LispNumber>> for LispObject {
-    fn from(v: Vec<LispNumber>) -> LispObject {
-        LispObject::from(
-            v.into_iter()
+impl<T: Into<LispObject>> From<Vec<T>> for LispObject {
+    fn from(v: Vec<T>) -> LispObject {
+        list(
+            &(v.into_iter()
                 .map(LispObject::from)
-                .collect::<Vec<LispObject>>(),
+                .collect::<Vec<LispObject>>()),
         )
     }
 }
```

But the compilers says this conflicts with the `From<Vec<LispObject>>` implementation. Bah!

@carlosmn since there is no rush on this PR, would you see if you can convince Rust to use this more generic version?

carlosmn:
This part of the language is a bit black magic, but yeah I can give it a go.

shaleh:
Agreed. If we can't, we can't. But this would remove some boiler plate.
If the language does not directly support it, maybe a macro would allow us to cut down on the repetition.

shaleh:
What I do not understand is why `Into<LispObject>` trait matches a `LispObject`. I would not think a type implements `Into` on itself. Maybe we have a sloppy `From` definition gumming up the works.

shaleh:
BTW, sitting in `rust_src` and running `cargo check` is a fast way to experiment once you have run the build at least once.

carlosmn:
The documentation for into says we shouldn't be implementing it

>Library authors should not directly implement this trait, but should prefer implementing the From trait, which offers greater flexibility and provides an equivalent Into implementation for free, thanks to a blanket implementation in the standard library.

https://doc.rust-lang.org/std/convert/trait.Into.html

shaleh:
Right, we never implement `Into` directly. Only `From`.

Also, we are an app not a lib so we have more lee way.

I suspect this line is our issue

> From and Into are reflexive, which means that all types can into themselves and from themselves

shaleh:
I used `Into<LispObject>` because I could not see a way to say `From<T> for LispObject`.

nickdrozd:
```rust
impl<T> From<Vec<T>> for LispObject
where
    LispObject: From<T>,
{
    fn from(v: Vec<T>) -> Self {
        let lisp_obj_vec = v
            .into_iter()
            .map(LispObject::from)
            .collect::<Vec<LispObject>>();

        list(&lisp_obj_vec)
    }
}
```

shaleh:
It _might_ be possible to use my version but specify the LispObject version as the `default`.

shaleh:
I like it @nickdrozd . I would not have thought to use the `where` syntax in that way. But it still does not compile happily.

nickdrozd:
What are you getting? The following changes compile fine for me:

```
modified   rust_src/src/fns.rs
@@ -9,6 +9,7 @@ use crate::{
     lisp::defsubr,
     lisp::LispObject,
     lists::{assq, car, get, member, memq, put, LispCons},
+    numbers::LispNumber,
     obarray::loadhist_attach,
     objects::equal,
     remacs_sys::Vautoload_queue,
@@ -334,7 +335,7 @@ unsafe fn getloadaverage(loadavg: *mut libc::c_double, nelem: libc::c_int) -> li
 /// setgid so that it can read kernel information, and that usually isn't
 /// advisable.
 #[lisp_fn(min = "0")]
-pub fn load_average(use_floats: bool) -> Vec<LispObject> {
+pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
     let mut load_avg: [libc::c_double; 3] = [0.0, 0.0, 0.0];
     let loads = unsafe { getloadaverage(load_avg.as_mut_ptr(), 3) };
 
@@ -345,9 +346,9 @@ pub fn load_average(use_floats: bool) -> Vec<LispObject> {
     (0..loads as usize)
         .map(|i| {
             if use_floats {
-                LispObject::from(load_avg[i])
+                LispNumber::Float(load_avg[i])
             } else {
-                LispObject::from((100.0 * load_avg[i]) as i64)
+                LispNumber::Fixnum((100.0 * load_avg[i]) as i64)
             }
         })
         .collect()
modified   rust_src/src/lisp.rs
@@ -252,9 +252,17 @@ impl From<()> for LispObject {
     }
 }
 
-impl From<Vec<LispObject>> for LispObject {
-    fn from(v: Vec<LispObject>) -> Self {
-        list(&v)
+impl<T> From<Vec<T>> for LispObject
+where
+    LispObject: From<T>,
+{
+    fn from(v: Vec<T>) -> Self {
+        let lisp_obj_vec = v
+            .into_iter()
+            .map(LispObject::from)
+            .collect::<Vec<LispObject>>();
+
+        list(&lisp_obj_vec)
     }
 }
```

shaleh:
You dropped the original `Vec<LispObject>`. I have been trying to keep it since it is more efficient for that case.

carlosmn:
In order to support both `Vec<T>` and `Vec<LispObject>` we'd have to opt-in to specialization, which is an unstable feature.

It looks like we're opting into a bunch of other features, so maybe that's not a big deal. I'll push up a version that uses that.

shaleh:
Yeah we lean HEAVILY into unstable.

shaleh:
New code looks nice. :+1:

nickdrozd:
Supposing we couldn't use specialization and had to choose between the generic version and the `Vec<LispObject>` version, do we know for sure that the generic version would be less efficient? Perhaps the compiler could say "Oh, we're mapping `LispObject` into `LispObject`, just leave it."

carlosmn:
Since we monomophosize the function calls, it would know at least that the `LispObject::from` call happens on a `LispObject` already so it should be fairly efficient either way. I would expect that maybe the `collect` would cause some extra temporary allocation, but I guess we'd have to decompile it to know.

rust_src/src/lisp.rs
```diff
@@ -258,6 +258,19 @@ impl From<Vec<LispObject>> for LispObject {
     }
 }
 
+impl<T> From<Vec<T>> for LispObject
+where
+    LispObject: From<T>,
+{
+    default fn from(v: Vec<T>) -> LispObject {
+        list(
```
nickdrozd: If we're keeping the generic version and the specialized `Vec<LispObject>` version, then should the generic one be `LispObject::from(v.into_iter()...)` (that is, convert the `Vec<T>` into `Vec<LispObject>` and then convert that into `LispObject`)? Otherwise the `list(&...)` logic is being duplicated.

shaleh: I explicitly duplicated it to ensure we did not recurse and cause issues.

shaleh: Which may be an unfounded concern. But I saw recursive explosion in some of my attempts to get this working this morning.

shaleh: My thinking is since we are saying the `Vec<LispObject>` case is a specialization it seems odd to have the default implementation depend on it.

shaleh: If there are better arguments then I can be swayed.

nickdrozd: My mental model of what's going on here is mostly based on the behavior of Python objects plus whatever I feel like the compiler should be able to figure out.

carlosmn:
I disassembled a version without the `impl From<Vec<LispObject>> for LispObject` version and it looks like there are enough smarts in rustc/llvm to realise what's going on.

I couldn't even find a call to `list` itself, but there are calls to `Fcons` to it looks like it realises that it's building a vector just to iterate over it in `list` and it it does it inside the `from` function.

There is one call to allocating a vector, but that seems to be as a result of the call to `fold` in `list` rather than the `collect`.

It's not a big deal wither way I think, but it does look like the "zero-cost abstractions" promise from Rust does hold here.

Issue #1134, opened 2018-12-11
delete-frame is not an interactive command


2mc:
Contrary to Emacs, `delete-frame` appears not to be an interactive command.

Because of this, when triggering the standard key-binding `C-x 5 0,` REmacs throws the error

    wrong-type-argument commandp delete-frame

and does not delete the frame.

brotzeit:
Thanks for the bug report. The intspec in the `lisp_fn` macro is missing. Want to give it a try ?

brotzeit:
You just have to add `intspec = ""`.

PR #1154, opened 2018-12-19
Mark delete-frame as interactive

agraven:
Fixes users not being able to close frames, solving issue #1134

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -418,7 +418,12 @@ pub fn frame_bottom_divider_width(frame: LispFrameOrSelected) -> i32 {
 /// This function runs `delete-frame-functions' before actually
 /// deleting the frame, unless the frame is a tooltip.
 /// The functions are run with one argument, the frame to be deleted.
-#[lisp_fn(min = "0", name = "delete-frame", c_name = "delete_frame")]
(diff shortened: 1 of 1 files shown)


rust_src/src/frames.rs
```diff
@@ -418,7 +418,12 @@ pub fn frame_bottom_divider_width(frame: LispFrameOrSelected) -> i32 {
 /// This function runs `delete-frame-functions' before actually
 /// deleting the frame, unless the frame is a tooltip.
 /// The functions are run with one argument, the frame to be deleted.
-#[lisp_fn(min = "0", name = "delete-frame", c_name = "delete_frame")]
+#[lisp_fn(
+    intspec = "",
```
shaleh: We usually add the `intspec` to the end like with `DEFUN` in C. But the order is not really important.

Wilfred:
@agraven I've added you as a collaborator, so you can also review/merge other people's PRs if you wish :)

cc @brotzeit

brotzeit:
@Wilfred Thanks.

@agraven Unfortunately it's not possible to review changes to the wiki, so just push you're changes.

shaleh:
You can make a branch. There is no formal support but we could read it and give you feedback.

PR #1157, opened 2018-12-20
Port time-add, time-subtract and time-less-p

carlosmn:
This doesn't go as far as #965 and leaves the types untouched. It also makes use of the new `From` magic and returns `Vec<LispNumber>` for the time. :man_dancing: 

The carries from one unit to the next in `time_add` and `time_subtract` have been ported to a more explicit and rusty form, but this does mean we've introduced conditionals which are not there in the C version. I don't know how big a deal this actually is, but we could have a version that casts e.g. `(us > 1_000_000 as u32)` so we can add/multiply 0 and 1.

### DIFF
--- a/rust_src/src/editfns.rs
+++ b/rust_src/src/editfns.rs
@@ -1,6 +1,7 @@
 //! Lisp functions pertaining to editing.
 
 use std;
+use std::cmp::max;
 use std::ptr;
 
 use libc;
@@ -28,12 +29,12 @@ use crate::{
     remacs_sys::{
         buffer_overflow, build_string, current_message, del_range, del_range_1, downcase,
         find_before_next_newline, find_newline, get_char_property_and_overlay, globals, insert,
-        insert_and_inherit, insert_from_buffer, make_buffer_string, make_buffer_string_both,
-        make_save_obj_obj_obj_obj, make_string_from_bytes, maybe_quit, message1, message3,
-        record_unwind_current_buffer, record_unwind_protect, save_excursion_restore,
-        save_restriction_restore, save_restriction_save, scan_newline_from_point,
-        set_buffer_internal_1, set_point, set_point_both, styled_format, update_buffer_properties,
-        STRING_BYTES,
+        insert_and_inherit, insert_from_buffer, lisp_time, make_buffer_string,
+        make_buffer_string_both, make_save_obj_obj_obj_obj, make_string_from_bytes, maybe_quit,
+        message1, message3, record_unwind_current_buffer, record_unwind_protect,
+        save_excursion_restore, save_restriction_restore, save_restriction_save,
+        scan_newline_from_point, set_buffer_internal_1, set_point, set_point_both, styled_format,
+        update_buffer_properties, STRING_BYTES,
     },
     remacs_sys::{
         Fadd_text_properties, Fcopy_sequence, Fget_pos_property, Fnext_single_char_property_change,
@@ -42,6 +43,7 @@ use crate::{
     remacs_sys::{Qboundary, Qfield, Qinteger_or_marker_p, Qmark_inactive, Qnil, Qt},
     textprop::get_char_property,
     threads::{c_specpdl_index, ThreadState},
+    time::{lisp_time_struct, time_overflow, LO_TIME_BITS},
     util::clip_to_bounds,
     windows::selected_window,
 };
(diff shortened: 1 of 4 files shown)


rust_src/src/editfns.rs
```diff
@@ -1393,4 +1395,120 @@ pub fn delete_and_extract_region(
...
 }
 
+fn time_arith<F: FnOnce(lisp_time, lisp_time) -> lisp_time>(
+    a: LispObject,
+    b: LispObject,
+    op: F,
+) -> Vec<LispNumber> {
+    let mut alen: c_int = 0;
+    let mut blen: c_int = 0;
+    let ta = unsafe { lisp_time_struct(a, &mut alen) };
```
shaleh: In theory, we could have `LispTime` in Rust which would implement the conversion logic right? Then the parameters could be `LispTime` instances. This would end up being:

    fn time_arith<F: FnOnce(LispTime, LispTime) -> LispTime>(
        a: LispTime,
        b: LispTime,
        op: F,
    ) -> Vec<EmacsInt> {
        let new_t = op(a, b);  // either check for overflow here or in the `to_vector` call below.
        new_t.to_vector()
    }

The `maxlen` would need to be computed and stored as part of the `op`.

Worth it? Over kill?

rust_src/src/editfns.rs
```diff
@@ -1393,4 +1395,120 @@ pub fn delete_and_extract_region(
...
+        v.push(LispNumber::Fixnum(t.us.into()));
+    }
+    if maxlen > 3 {
+        v.push(LispNumber::Fixnum(t.ps.into()));
+    }
+
+    v
+}
+
+fn time_add(ta: lisp_time, tb: lisp_time) -> lisp_time {
```
shaleh: This code is nicer to read than the previous version.

rust_src/src/editfns.rs
```diff
@@ -1393,4 +1395,120 @@ pub fn delete_and_extract_region(
...
+
+/// Return the difference between two time values A and B, as a time value. Use `float-time' to
+/// convert the difference into elapsed seconds.  A nil value for either argument stands for the
+/// current time.  See `current-time-string' for the various forms of a time value.
+#[lisp_fn(name = "time-subtract", c_name = "time_subtract")]
+pub fn time_subtract_lisp(a: LispObject, b: LispObject) -> Vec<LispNumber> {
+    time_arith(a, b, time_subtract)
+}
+
+macro_rules! return_if_different {
```
shaleh: Cute. I like it.

rust_src/src/editfns.rs
```diff
@@ -1393,4 +1395,120 @@ pub fn delete_and_extract_region(
...
+/// argument stands for the current time.  See `current-time-string' for the various forms of a
+/// time value.
+#[lisp_fn]
+pub fn time_less_p(t1: LispObject, t2: LispObject) -> bool {
+    let mut t1len: c_int = 0;
+    let mut t2len: c_int = 0;
+    let a = unsafe { lisp_time_struct(t1, &mut t1len) };
+    let b = unsafe { lisp_time_struct(t2, &mut t2len) };
+
+    return_if_different!(a.hi, b.hi);
```
nickdrozd: Could this whole thing be simplified to `a.hi < b.hi || ... || a.ps < b.ps`?

shaleh: The original C says "a != b && a < b". Maybe to optimize the == being a faster check? 🤷‍♂️

rust_src/src/editfns.rs
carlosmn: I went with `into_vec` which accepts a length as we need to keep track of the length of the inputs. Whenever we finally move `LispTime` wholly into Rust, then we can start tracking that in the struct itself through conversions, but for now I think keeping it separate still makes the most sense.

carlosmn:
I probably went overboard with `LispTime` `impl`s but we have few things left that use `struct lisp_time` from C so we should be able to have it be a normal Rust struct soon.

rust_src/src/editfns.rs
shaleh:
```
fn time_arith<F>(a: LispObject, b: LispObject, op: F) -> Vec<EmacsInt>
where
    F: FnOnce(LispTime, LispTime) -> LispTime
```

Puts the function definition first for the reader and the where clause explains what an 'op' is.

rust_src/src/editfns.rs
```diff
@@ -1393,4 +1395,95 @@ pub fn delete_and_extract_region(
...
+    let t = op(ta, tb);
+    if LispObject::fixnum_overflow(t.hi) {
+        time_overflow();
+    }
+
+    let maxlen = max(alen, blen) as usize;
+
+    t.into_vec(maxlen)
+}
+
```
shaleh:
```
impl Sub for LispTime {
    type Output = LispTime;

    fn sub(self, other: Self) -> Self::Output {
    }
}
```

Along with a match `Add` round out the LispTime. Now you can say `time1 - time2`. If you copy the `time_add` logic into the `add` trait method poof. Done. Then the calls below become `time_arith(a, b, |ta, tb| ta + tb)`.

rust_src/src/time.rs
```diff
@@ -16,7 +17,62 @@ use crate::{
...
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for LispTime {
+    fn cmp(&self, other: &LispTime) -> Ordering {
+        return_if_different!(self.hi, other.hi);
+        return_if_different!(self.lo, other.lo);
+        return_if_different!(self.us, other.us);
+        return_if_different!(self.ps, other.ps);
```
bb010g: This should be a [`then_with`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then_with) chain, i.e. `self.hi.cmp(other.hi).then_with(|| self.lo.cmp(other.lo)).then_with(…).then_with(…)`.

carlosmn: Huh, TIL. Is there some reason to use `then_with` rather than `then`?

shaleh: `then` computes its argument even if the check is unneeded. By passing in closures, `then_with` will not evaluate the next comparison until it is actually needed.

Equal.then(expensive())

Equal.then_with(expensive)

rust_src/src/time.rs
```diff
@@ -18,6 +20,105 @@ use crate::{
...
+
+#[allow(clippy::suspicious_arithmetic_impl)]
+impl Add for LispTime {
+    type Output = LispTime;
+
+    fn add(self, other: LispTime) -> LispTime {
+        let mut hi = self.hi + other.hi;
+        let mut lo = self.lo + other.lo;
+        let mut us = self.us + other.us;
+        let mut ps = self.ps + other.ps;
```
nickdrozd: These mutating, procedural calculations could be rewritten to be pure. My guess is that the compiler would be able to do more with it that way. I also think the logic would be clearer, but that's a matter of opinion.

shaleh: I am not so certain there is a pay off there.

```
let (us,ps) = if ps > 1_000_000 {
    (us - 1_000_000, ps + 1)
else {
    (us, ps)
};
let (us, lo) = if us > 1_000_000 {
    (us - 1_000_000, lo + 1)
} else {
    (us, lo)
};
let (lo, hi) = if lo > 1 << LO_TIME_BITS {
   (lo - (1 << LO_TIME_BITS), hi + 1)
} else {
    (lo, hi)
};
```

Is that really better?

nickdrozd: I would calculate auxiliary variables `ps_diff`, `us_diff` etc first, and then calculate `ps`, `us` etc just once, since I don't like variable shadowing (it's no better than mutation from the reader's perspective).

I also tend to think that using a different implementation is per se a good thing (as long as the different implementation is at least as fast, clear, etc).

rust_src/src/time.rs
```diff
@@ -18,6 +20,105 @@ use crate::{
...
+
+        if ps > 1_000_000 {
+            ps += 1;
```
shaleh: You flipped `ps` and `us` here. Here is the C:

```
  us += (1000000 <= ps);
  ps -= (1000000 <= ps) * 1000000;
```

shaleh: Also, the checks should be '>=' right? In the original code, if `ps` is 1_000_000 then `us` is incremented and `ps` is reduced.

carlosmn: Yeah... I'll go over both versions again and double check.

PR #1158, opened 2018-12-21
Break up implementation of Debug trait.

shaleh:
To facilitate this, introduce `force_foo` methods which are analogous to `XFOO` in the C code. These can replace some of our usage of `as_foo_or_error()` when we are certain of the type of the object.

What I'd like to see is a bit of bikeshedding over the naming I used. is `force_foo` clear enough? I wanted the name to be short so it did not use up too much space in code and detract from the real work. The C `XFOO` is fairly perfect in that regard.


rust_src/src/lisp.rs
```diff
@@ -576,78 +589,30 @@ impl LispObject {
...
-            Lisp_Type::Lisp_Int0 | Lisp_Type::Lisp_Int1 => {
-                write!(f, "{}", self.as_fixnum().unwrap())?;
-            }
-            Lisp_Type::Lisp_Misc => {
-                write!(f, "#<MISC @ {:#X}: VAL({:#X})>", self_ptr, self.to_C())?;
-            }
-            Lisp_Type::Lisp_String => {
-                write!(f, "{:?}", display_string(*self))?;
-            }
+            Lisp_Type::Lisp_Symbol => write!(f, "{:?}", self.force_symbol()),
```
nickdrozd: Why are there all these separate `force_whatever` functions? It seems like it would be easier to name them all `force` or `force_type` or something, and some of these branches could be consolidated.

shaleh: They are all members of LispObject so they need different names.

The intent here is to provide the equivalent of XINT or XCONS. These functions do no checks and act like cast operators.

shaleh: They are also meant to be fast / least effort. Having a if/else or match logic would defeat that purpose. These would replace some of the `as_foo_or_error()` calls when we know for certain the type of object we have. Like, we know we just made a CONS cell no need to wrap it an Option and unwrap it using `as_cons_or_error()`.

shaleh: I could have done this as two PRs. But I thought the existing type `match` was the perfect reason to add the `force_` functions. The alternative is like I did in the `internal_equal` PR:

```
if let (Some(a), Some(b)) = (thing1.as_foo(), thing2.as_foo()) {
    ...
} else if let (Some(a), Some(b)) = (thing1.as_bar(), thing2.as_bar()) {
    ...
}
```

But the direct type and force is way more efficient.

krobelus: I'd be fine with `force_foo` but we could also go with `as_foo_unchecked` which is more discoverable since we already have several `as_foo` and `as_foo_or_error`. Given that the `XFOO` macros are used a lot, same might hold in Rust code (?) which would make a short name more appealing.

shaleh: I chose `force_foo` because it was shorter than `as_foo_unchecked`. I expect these to be used frequently. Another option would be macros like `CONS!()` or `FIXNUM!()` which evaluate to something like `obj.as_foo().unwrap()` or `Option<LispFoo>::from(obj).unwrap()` internally.

rust_src/src/vectors.rs
```diff
@@ -94,6 +98,31 @@ impl LispObject {
     }
 }
 
+impl Debug for LispVectorlikeRef {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+        if let Some(v) = self.as_vector() {
+            let mut first = true;
+            write!(f, "[")?;
+            for el in v.as_slice() {
```
nickdrozd: Is there a better way to do this? Maybe `v.as_slice()[1:]` or something like that? Toggles are ugly.

brotzeit:
I think we can still rename the methods if we find a better solution.

rust_src/src/vectors.rs
shaleh: Hmm, I think I have a solution. Thanks for the prodding @nickdrozd.

rust_src/src/vectors.rs
```diff
@@ -96,6 +100,32 @@ impl LispObject {
     }
 }
 
+impl Debug for LispVectorlikeRef {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
```
shaleh: @nickdrozd I needed to add `slice_patterns` support. This is much cleaner.

nickdrozd: Nice!

nickdrozd:
@shaleh I ventured out into the wasteland of Github's online editor to try and fix the merge conflict. I think the change is correct? But maybe not.

shaleh:
I know what you mean. I use it fairly often but it is always questionable. Thanks for doing it.

shaleh:
The diff looks reasonable. You replaced some code with my `force_string` and in another place you replaced some code with `.into()`.

PR #1161, opened 2018-12-21
Port close-font

agraven:
Port the lisp function close-font. To achieve this, the `LispFrameLiveOrSelected` and `LispFontObjectRef` types were added. This solves issue #1013.


rust_src/src/fonts.rs
```diff
@@ -97,6 +103,70 @@ impl FontExtraType {
...
 
+pub type LispFontObjectRef = ExternalPtr<Lisp_Font_Object>;
+
+impl LispFontObjectRef {
+    pub fn as_lisp_obj(self) -> LispObject {
+        LispObject::tag_ptr(self, Lisp_Type::Lisp_Vectorlike)
+    }
+
+    pub fn add_log(self, action: &str, result: LispObject) {
+        let c_str = CString::new(action).unwrap();
```
shaleh: Is this safe?

rust_src/src/fonts.rs
```diff
@@ -97,6 +103,70 @@ impl FontExtraType {
...
+        unsafe { font_add_log(c_str.as_ptr(), self.as_lisp_obj(), result) }
+    }
+
+    pub fn close(mut self, mut frame: LispFrameRef) {
+        if data::aref(self.as_lisp_obj(), FONT_TYPE_INDEX.into()).is_nil() {
+            // Already closed
```
shaleh: Instead of the aref another option is to convert to vectorlike and use get. I am not sire which is best.

rust_src/src/fonts.rs
```diff
@@ -97,6 +103,70 @@ impl FontExtraType {
...
+}
+
+impl From<LispFontObjectRef> for LispObject {
+    fn from(f: LispFontObjectRef) -> Self {
+        f.as_lisp_obj()
+    }
+}
+
+impl From<LispObject> for LispFontObjectRef {
+    fn from(o: LispObject) -> Self {
```
shaleh: I think the `as_lisp_obj` idiom is tech debt and we should only use into/from instead.

shaleh: Thoughts?

rust_src/src/fonts.rs
```diff
@@ -143,4 +213,12 @@ pub fn find_font(spec: LispObject, frame: LispObject) -> LispObject {
     }
 }
 
+/// Close FONT-OBJECT
```
shaleh: @nickdrozd this is a place where the comment does not match the variable. Should we name the arg `font_object`?

rust_src/src/frames.rs
```diff
@@ -133,6 +133,21 @@ impl LispFrameOrSelected {
...
+pub struct LispFrameLiveOrSelected(LispFrameRef);
+
+impl From<LispObject> for LispFrameLiveOrSelected {
+    fn from(obj: LispObject) -> LispFrameLiveOrSelected {
+        LispFrameLiveOrSelected(obj.map_or_else(selected_frame, |f| f.as_live_frame_or_error()))
+    }
+}
+
+impl From<LispFrameLiveOrSelected> for LispFrameRef {
+    fn from(f: LispFrameLiveOrSelected) -> LispFrameRef {
```
brotzeit: LispFrameRef -> Self

rust_src/src/frames.rs
```diff
@@ -133,6 +133,21 @@ impl LispFrameOrSelected {
     }
 }
 
+#[derive(Clone, Copy)]
+pub struct LispFrameLiveOrSelected(LispFrameRef);
+
+impl From<LispObject> for LispFrameLiveOrSelected {
+    fn from(obj: LispObject) -> LispFrameLiveOrSelected {
```
brotzeit: LispFrameLiveOrSelected -> Self

rust_src/src/fonts.rs
```diff
@@ -143,4 +213,12 @@ pub fn find_font(spec: LispObject, frame: LispObject) -> LispObject {
...
+#[lisp_fn]
```
brotzeit: min="1"

rust_src/src/fonts.rs
nickdrozd: Yeah, they should always match, except in cases where the argument name is a Rust keyword (eg `type`). It would be nice to avoid those cases too.

rust_src/src/fonts.rs
agraven: I think so, since the function is only called using hardcoded strings by other functions and not with any user-inputted strings, so unless a function calls it with a broken string there shouldn't be any issues here.

rust_src/src/fonts.rs
```diff
@@ -97,6 +103,73 @@ impl FontExtraType {
...
+        if data::aref(self.as_lisp_obj(), FONT_TYPE_INDEX.into()).is_nil() {
+            // Already closed
+            return;
+        }
+        self.add_log("close", LispObject::from(false));
+        unsafe {
+            if let Some(f) = (*self.driver).close {
+                f(self.as_mut())
+            }
+            #[cfg(feature = "window-system")]
```
shaleh: If the C compilation does not add an entry to a struct this will still fail to compile. Which can be frustrating. You may end up needing two internal functions that are compile time picked.

rust_src/src/fonts.rs
agraven: I just copied the behavior form the c function `font_close_object`, I'll see if the other way is nicer

rust_src/src/fonts.rs
agraven: That's actually quite reasonable, it would be more slightly more idiomatic and safer in the long term to stick to from/into

rust_src/src/fonts.rs
shaleh: Yes, but `intern` is called on it in `font_add_log`. Will the interpreter try to garbage collect that at some point?

rust_src/src/fonts.rs
agraven: Yeah, that was a mistake on my part.

rust_src/src/fonts.rs
shaleh: No allocations made in Rust are safe to the C side. The C code may try to free it.

agraven: Hm, I hadn't considered that. Is there an established safer way to send strings to c functions?

agraven:
I just realized, the `font_close_object` function I based the implementation of `LispFontObjectRef::close` on is no longer used. Should I remove it?

rust_src/src/fonts.rs
brotzeit: I think we already had this. I would say open an issue ?

rust_src/src/fonts.rs
```diff
@@ -143,4 +227,12 @@ pub fn find_font(spec: LispObject, frame: LispObject) -> LispObject {
     }
 }
 
+/// Close FONT-OBJECT
+#[lisp_fn(min = "1")]
+pub fn close_font(font_object: LispFontObjectRef, frame: LispFrameLiveOrSelected) {
+    let frame: LispFrameRef = frame.into();
+    //unsafe { font_close_object(frame.as_mut(), object) }
```
shaleh: Don't leave commented out code.

shaleh:
The Mac full build fails because the window system is set but the member is missing. Looks odd.

rust_src/src/fonts.rs
agraven: Oops, that was an accident

shaleh:
I will try to get my work Mac out later to see if I can quickly identify the issue.

rust_src/src/fonts.rs
shaleh: Ok, I just took a few minutes and tracked this to the bottom. The `CString` is not used directly. The contents are `memcpy`ed out. There is no issue here.

rust_src/src/fonts.rs
```diff
@@ -97,6 +103,69 @@ impl FontExtraType {
...
+        let c_str = CString::new(action).unwrap();
+        unsafe { font_add_log(c_str.as_ptr(), self.into(), result) }
+    }
+
+    pub fn close(mut self, mut _frame: LispFrameRef) {
+        if data::aref(self.into(), FONT_TYPE_INDEX.into()).is_nil() {
+            // Already closed
+            return;
+        }
+        self.add_log("close", LispObject::from(false));
```
shaleh: In the C code this is a macro call.

```
#define FONT_ADD_LOG(ACTION, ARG, RESULT)       \
  do {                                          \
    if (! EQ (Vfont_log, Qt))                   \
      font_add_log ((ACTION), (ARG), (RESULT)); \
  } while (false)
```

Which means we either a) need a macro or b) need the `add_log` method to check the value before doing anything.

agraven: I think checking the value is the best solution. However, something that strikes me as odd is that `font_add_log` itself already does the exact same check in the form of
```c
if (EQ (Vfont_log, Qt))
  return;
```
so implementing the check the macro does as well is doing just doing the check twice, which seems kind of pointless.

shaleh: Indeed. I missed that.

shaleh:
I still owe a Mac debug session.

shaleh:
This PR just became a bit more complicated.

Your code is failing because it is trying to access the X Window specific implementation on MacOS which uses a different window system. That is what the `.x.` is referring to. On MacOS the member is called `ns`. You can see this by looking at xterm.h and nsterm.h for the definition of the `FRAME_DISPLAY_INFO` macro.

1) `close-font` is only defined when `FONT_DEBUG` is defined. It is wrapped in an IFDEF in the C code. So, you need another Rust feature.
2) There is a different `display_info` type for each window system: X, NS (Apple), Win, etc. In C the code works because each struct defines `n_fonts`. Rust will not accept this. You will either end up with feature gated functions or implementing a trait for `display_info` that supports asking for the number of fonts and decrementing or incrementing them.

shaleh:
The new types and feature gating work would help other porters and is worthwhile. You can either complete this PR or grab them and make a simpler PR for another function and return to this one afterwards.

agraven:
I'm going to complete the PR. As you said, it can be useful to others in the future

brotzeit:
@agraven is still something missing ? I need this merged for #946.

agraven:
@brotzeit i still need to add compatability for window systems other than x, I'll have it done tomorrow probably

agraven:
I've implemented the accesses to the different versions of `display_info` in the simplest way possible. I think long term as more functions get ported it will probably smarter to make a more proper interface for doing so, but for now I think this PR needs to be finished.

brotzeit:
@shaleh good ?

brotzeit:
I would merge this ?

configure.ac
```diff
@@ -5442,7 +5442,20 @@ if test "$CANNOT_DUMP" != "yes"; then
...
+    CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"window-system\", "
+fi
+case "$window_system" in
+    x11)
+        CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"x11\", "
+    ;;
+    nextstep)
+	CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"nextstep\", "
+    ;;
+    w32)
```
shaleh: The `w32` is likely too generic. Perhaps `window_system_x`, `window_system_nextstep`, and `window_system_ms_win`?

rust_src/src/fonts.rs
```diff
@@ -97,6 +103,74 @@ impl FontExtraType {
...
+
+impl From<LispFontObjectRef> for LispObject {
+    fn from(f: LispFontObjectRef) -> Self {
+        LispObject::tag_ptr(f, Lisp_Type::Lisp_Vectorlike)
+    }
+}
+
+impl From<LispObject> for LispFontObjectRef {
+    fn from(o: LispObject) -> Self {
+        o.as_font_object_or_error()
```
shaleh: @nickdrozd has been working to remove these methods and replace them with From conversions. Fold the method code here and drop it.

rust_src/src/fonts.rs
```diff
@@ -97,6 +103,74 @@ impl FontExtraType {
...
+    }
+}
+
+impl From<LispObject> for Option<LispFontObjectRef> {
+    fn from(o: LispObject) -> Self {
+        o.as_font_object()
```
shaleh: Same here, look at the recent PRs if you need guidance.

rust_src/src/frames.rs
```diff
@@ -129,6 +129,21 @@ impl LispFrameOrSelected {
     }
 }
 
+#[derive(Clone, Copy)]
+pub struct LispFrameLiveOrSelected(LispFrameRef);
+
+impl From<LispObject> for LispFrameLiveOrSelected {
+    fn from(obj: LispObject) -> Self {
+        LispFrameLiveOrSelected(obj.map_or_else(selected_frame, |f| f.as_live_frame_or_error()))
```
shaleh: This should also be refactored but can wait for another PR.

rust_src/src/fonts.rs
agraven: Yeah, I noticed the pr's about that. done.

configure.ac
agraven: I was worried that would be an issue. Done, though I used dashes as that's the convention with feature names.

rust_src/src/fonts.rs
```diff
@@ -144,31 +144,20 @@ impl From<LispFontObjectRef> for LispObject {
 
 impl From<LispObject> for LispFontObjectRef {
     fn from(o: LispObject) -> Self {
-        o.as_font_object_or_error()
+        Option::<LispFontObjectRef>::from(o).unwrap_or_else(|| wrong_type!(Qfont_object, o))
```
shaleh: If that felt ugly you could write:

    match o.into() {
        Some(font) => font,
        None => wrong_type!(Qfont_object, o),
    }

agraven: Ah yes, I missed that. That's a lot less ugly.

PR #1166, opened 2018-12-22
Cut as_lisp_obj

nickdrozd:
Prefer into where possible, or else LispObject::from

### DIFF
--- a/rust_src/src/chartable.rs
+++ b/rust_src/src/chartable.rs
@@ -12,7 +12,7 @@ use crate::{
         pvec_type, Lisp_Char_Table, Lisp_Sub_Char_Table, Lisp_Type, More_Lisp_Bits,
         CHARTAB_SIZE_BITS,
     },
-    remacs_sys::{Qchar_code_property_table, Qchar_table_p, Qnil},
+    remacs_sys::{Qchar_code_property_table, Qchar_table_p},
 };
 
 pub type LispCharTableRef = ExternalPtr<Lisp_Char_Table>;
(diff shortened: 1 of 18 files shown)


nickdrozd:
This takes care of all of them except for `LispObarrayRef`, which does some borrowing stuff that I haven't been able to figure out.

rust_src/src/chartable.rs
```diff
@@ -235,11 +235,7 @@ pub fn set_char_table_parent(mut chartable: LispCharTableRef, parent: Option<Lis
         }
     }
 
-    chartable.parent = if let Some(p) = parent {
-        p.as_lisp_obj()
-    } else {
-        Qnil
-    };
+    chartable.parent = if let Some(p) = parent { p.into() } else { Qnil };
```
shaleh:
```suggestion
    chartable.parent = parent.map_or(Qnil, |p| p.into());
```

I think the `map_or` makes more sense. But if not, wouldn't `match` be a better fit? Not on you, this was existing code. I think we over used the `if let Some()` idiom.

shaleh:
I do not see the issue with `LispObarrayRef`. It is a new type wrapper around `LispObject` like `LispCons`.

shaleh:
Can you also remove the `LispCons` `as_obj`?

nickdrozd:
I think that should take care of all of them.

rust_src/src/chartable.rs
nickdrozd: We can do even better than that: `parent: Option<LispCharTableRef>`, so all we need is `chartable.parent = parent.into()`.

Issue #499, opened 2017-11-22
Emacs internal string encoding


db48x:
Currently some of the ported functions (like the error! macro) are using Rust's strings to handle data from an Emacs string or buffer. Rust's strings are quite nice, but Emacs has a subtly different requirements that make them the wrong choice in almost all cases.

If we look in the Emacs manual, section 22.2 says:

>    Internally, Emacs uses its own multibyte character encoding, which is
> a superset of the “Unicode” standard.  This internal encoding allows
> characters from almost every known script to be intermixed in a single
> buffer or string.  Emacs translates between the multibyte character
> encoding and various other coding systems when reading and writing
> files, and when exchanging data with subprocesses.

The Elisp manual, in section 32.1, elaborates:

> Emacs buffers and strings support a large repertoire of characters from
> many different scripts, allowing users to type and display text in
> almost any known written language.
> 
>    To support this multitude of characters and scripts, Emacs closely
> follows the “Unicode Standard”.  The Unicode Standard assigns a unique
> number, called a “codepoint”, to each and every character.  The range of
> codepoints defined by Unicode, or the Unicode “codespace”, is
> ‘0..#x10FFFF’ (in hexadecimal notation), inclusive.  Emacs extends this
> range with codepoints in the range ‘#x110000..#x3FFFFF’, which it uses
> for representing characters that are not unified with Unicode and “raw
> 8-bit bytes” that cannot be interpreted as characters.  Thus, a
> character codepoint in Emacs is a 22-bit integer.
> 
>    To conserve memory, Emacs does not hold fixed-length 22-bit numbers
> that are codepoints of text characters within buffers and strings.
> Rather, Emacs uses a variable-length internal representation of
> characters, that stores each character as a sequence of 1 to 5 8-bit
> bytes, depending on the magnitude of its codepoint(1).  For example, any
> ASCII character takes up only 1 byte, a Latin-1 character takes up 2
> bytes, etc.  We call this representation of text “multibyte”.

To summarize, Emacs internally stores strings in a format which is exactly like UTF-8 except that the allowed range of codepoints is larger. Thus while the largest codepoints in UTF-8 are encoded in 4 bytes, in Emacs the largest codepoints need 5 bytes. It uses these extra codepoints to encode characters which don't have any unicode equivalent. This allows an Emacs buffer to hold text from multiple character sets during editing (they all get reencoded to a common character set when the buffer is written to disk). We need to implement a string type which follows these rules, and use it instead of the built-in str and String types.

[...]

PR #1168, opened 2018-12-22
Fix base64 decoding into multibyte string

carlosmn:
When we're told decode into a multibyte string, we don't want to
restore the raw bytes and neither do we want to stuff the bytes into a
utf-8 string. Emacs comes with its own encoding which was lost when
converting the code to Rust.

Perform this encoding step when dealing with multibyte output and set
the tests against outputs taken from running the upstream
implementation.

---

This is somewhat related to #499 as we can't return the bare decoded data nor can we let the Rust stdlib deal with it as a string

### DIFF
--- a/rust_src/src/base64.rs
+++ b/rust_src/src/base64.rs
@@ -85,19 +85,33 @@ pub unsafe extern "C" fn base64_decode_1(
             }
             if multibyte {
                 // Decode non-ASCII bytes into UTF-8 pairs.
-                let s: String = decoded[..decoded_length]
-                    .iter()
-                    .map(|&byte| byte as char)
-                    .collect();
+                let s = encode_multibyte_string(&decoded[..decoded_length]);
                 let s_len = s.len();
                 ptr::copy_nonoverlapping(s.as_ptr(), to as *mut u8, s_len);
-                s.len() as isize
+                s_len as isize
             } else {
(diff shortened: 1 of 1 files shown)


rust_src/src/base64.rs
```diff
@@ -268,6 +285,37 @@ fn test_multibyte_base64_decode_1() {
...
+    let decoded_multibyte = vec![
+        193, 131, 192, 177, 193, 131, 192, 177, 193, 131, 192, 177, 193, 131, 192, 177, 193, 131,
+        192, 161, 193, 131, 192, 161, 193, 131, 192, 161, 193, 131, 192, 161, 193, 131, 192, 161,
+        193, 131, 192, 161, 193, 132, 192, 145, 193, 132, 192, 145, 193, 132, 192, 145, 193, 132,
+        192, 145, 193, 132, 192, 145, 193, 132, 192, 145, 193, 132, 192, 145, 193, 131, 192, 176,
+        10,
+    ];
+    let mut n = 0;
+    let mut decoded: Vec<u8> = Vec::new();
+    decoded.resize(compute_decode_size(input.len()), 0);
```
shaleh: Why not use `with_capacity`?

shaleh: Makes sense. Thanks.

PR #1169, opened 2018-12-23
Port base64-decode-region and base64-encode-region and rustify support code

carlosmn:
This build on top of #1168 so the rustification part doesn't conflict. We start by fixing the logic in `byte_pos_addr` which did not take the gap/gpt into account and thus would return wrong values when we tried to use it after `move_gap_both`.

After moving the implementation for `base64-{decode,encode}-region` there were no more base64 functions in C, so we no longer need the `extern "C"` or unsafe buffer operations. We can move those support functions to use Rust types. We can also remove functions meant to figure out how much to preallocate.

### DIFF
--- a/rust_src/src/base64.rs
+++ b/rust_src/src/base64.rs
@@ -16,7 +16,6 @@ use crate::{
         del_range_both, del_range_byte, insert, insert_1_both, make_unibyte_string, move_gap_both,
         set_point, set_point_both, signal_after_change, temp_set_point_both,
     },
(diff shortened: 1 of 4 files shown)


carlosmn:
There were some issues with the meaning of unibyte vs multibyte and unfortuately the exists tests confused what they meant so that testing with the Czech string was buggy.

I've pushed up a commit to fix that handling, which makes this PR not as clean as I would have liked, but oh well.

PR #1176, opened 2018-12-24
Port insert, insert-before-markers, insert-and-inherit, insert-and-inherit-before-markers

carlosmn:
Port this family of functions as well as the functions they use to insert their arguments as only two of them have some C calling them so we can make a lot of these functions be safe.

This also introduced a wrapper `insert_slice` to be a Rusty version of `insert` which we can call with a slice as we usually want to do instead of a pointer+length pair; analogous to the `insert_string` that C can use for a NUL-terminated string.


rust_src/src/remacs_sys.rs
```diff
@@ -127,6 +127,15 @@ extern "C" {
         dest: Lisp_Object,
         op: BoolVectorOp,
     ) -> Lisp_Object;
+    pub fn insert_from_string_1(
```
shaleh: Don't manually add things to remacs_sys. Instead, copy it to lisp.h or another header. Bindgen will find it and export it.

carlosmn: Makes sense.

carlosmn:
It doens't look like the failing tests have much to do with the changes I've made, and I can repro the sort order error on remacs master and going back some.

Upstream has [f4b5ff20a663db](http://git.savannah.gnu.org/cgit/emacs.git/commit/?id=f4b5ff20a663db483ee4bb3e60c34ce4cd2d6d70) which seems to explain why it might fail though I don't know why other builds are succeeding.

shaleh:
What is the plan for finishing this PR?

carlosmn:
Oh I completely forgot this was still outstanding, I'll rebase and see if the tests are still failing and if so try to figure out what on earth is making them fail at a distance.

rust_src/src/editfns.rs
```diff
@@ -1447,4 +1450,284 @@ pub fn widen() {
...
+    pos: ptrdiff_t,
+    pos_byte: ptrdiff_t,
+    length: ptrdiff_t,
+    length_byte: ptrdiff_t,
+    inherit: bool,
+) {
+    let current_buffer = ThreadState::current_buffer_unchecked();
+    let opoint = current_buffer.pt;
+
+    let s = string.as_string_or_error();
```
shaleh: this can be `let s: LispStringRef = string.into();`. See @nickdrozd's recent PRs.

rust_src/src/editfns.rs
```diff
@@ -1447,4 +1450,284 @@ pub fn widen() {
...
+    if s.len_chars() == 0 {
```
shaleh: Does `LispStringRef` have an `empty` predicate? Should it?

carlosmn: It doesn't seem to. `slice` and `str` and friends all seem to have `is_empty` so I suppose it makes sense that this would too.

shaleh: No need to add it in this PR. I will open a tech debt issue to investigate and add it.

shaleh: #1230

rust_src/src/editfns.rs
carlosmn: Changed and fixed up.

brotzeit:
Can you add tests for insert-before-markers and insert-and-inherit. Insert is widely used in tests and I couldn't find a place where insert-and-inherit-before-markers is used so I guess we can ignore them.

rust_src/src/editfns.rs
shaleh: 1230 landed, you can use it now.

brotzeit:
I will try to add some tests.

brotzeit:
Hmm I took a look at it and it seems the tests have to cover many cases with markers, overlays, multibyte, unibyte...

I suggest we merge this and open an issue for both functions insert-before-markers and insert-and-inherit.

shaleh:
Let's try that again.

Issue #1167, opened 2018-12-22
Something is up with LispMarkerRef / marker functions


nickdrozd:
One build of https://github.com/Wilfred/remacs/pull/1163 failed with a crash resembling https://github.com/Wilfred/remacs/issues/820 and https://github.com/Wilfred/remacs/issues/1145:

```
 thread '<unnamed>' panicked at 'assertion failed: charpos <= bytepos', src/marker.rs:351:13
 stack backtrace:
    0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
              at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
    1: std::sys_common::backtrace::_print
              at libstd/sys_common/backtrace.rs:71
    2: std::panicking::default_hook::{{closure}}
              at libstd/sys_common/backtrace.rs:59
              at libstd/panicking.rs:211
    3: std::panicking::default_hook
              at libstd/panicking.rs:227
    4: std::panicking::rust_panic_with_hook
              at libstd/panicking.rs:476
    5: std::panicking::begin_panic
    6: remacs::marker::set_marker_internal
    7: Fmatch_data
              at /home/travis/build/Wilfred/remacs/src/search.c:2653
    8: eval_sub
              at /home/travis/build/Wilfred/remacs/src/eval.c:1549
    9: Flet
   10: eval_sub
              at /home/travis/build/Wilfred/remacs/src/eval.c:1495
   11: eval_sub
              at /home/travis/build/Wilfred/remacs/src/eval.c:1614
   12: Fprogn
   13: funcall_lambda
              at /home/travis/build/Wilfred/remacs/src/eval.c:2106
   14: apply_lambda
              at /home/travis/build/Wilfred/remacs/src/eval.c:1967
   15: eval_sub
              at /home/travis/build/Wilfred/remacs/src/eval.c:1618
   16: Fsetq
   17: eval_sub
              at /home/travis/build/Wilfred/remacs/src/eval.c:1495
   18: Fprogn
   19: funcall_lambda
              at /home/travis/build/Wilfred/remacs/src/eval.c:2106
   20: apply_lambda
              at /home/travis/build/Wilfred/remacs/src/eval.c:1967
   21: eval_sub
              at /home/travis/build/Wilfred/remacs/src/eval.c:1618
   22: FletX
   23: eval_sub
              at /home/travis/build/Wilfred/remacs/src/eval.c:1495
   24: Fprogn
   25: funcall_lambda
[...]

Issue #1145, opened 2018-12-16
'assertion failed: buffer_ref.beg() <= charpos && charpos <= buffer_ref.z()', src/marker.rs:618:5


gaak99:
Twas doing some ```helm``` I think...

not running with ```RUST_BACKTRACE=1``` so no backtrace of note and no core file.

```
git log --oneline HEAD|head -1
de43398 Updated Mojave build instructions for libxml2 (#1129)
```

gaak99:
Am now running with ```RUST_BACKTRACE=1``` and will try to reproduce (or at least remember the sequence of ops to trigger assertion).

shaleh:
This has been something of a ghost lately. Recompiling or just running again has resolved it. If you can find a reproducible case that would be super helpful.

brotzeit:
I'm optimistic that the fix works.

Issue #820, opened 2018-07-23
Crash: 'assertion failed: charpos == bytepos', src/marker.rs:247:13


nickdrozd:
I was messing with a package that deals with hooks and local variables (see https://github.com/Malabarba/Nameless/pull/25) and Remacs crashed abruptly. The error message:

```
thread '<unnamed>' panicked at 'assertion failed: charpos == bytepos', src/marker.rs:247:13
```

I haven't been able to reproduce it, and I don't remember exactly what I was doing when it happened. Wish I could be of more help :cat2: 

`Remacs 26.0.60 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30) of 2018-07-19`

db48x:
That's going to be fun. Could you run Remacs with RUST_BACKTRACE=1, just in case this happens again? It'll cause it to print a stack trace when it asserts like this, which can be very helpful for tracking down the problem.

brotzeit:
This looks familiar :/

brotzeit:
So what can we do about this ? Is there a way to run it automatically and repeatedly in a debugger until the error occurs ? Or what are other ways to catch non deterministic bugs ?

brotzeit:
Let's reopen an issue if somebody sees it again. I didn't see this error for a long time.

PR #1177, opened 2018-12-25
Fix LispMarkerRef tagging when converting to LispObject and make some use of it

carlosmn:
The implmentation of `From<LispMakerRef> for LispObject` transmutes the memory which means we fail to tag the pointer. As the receiver expects it to be tagged it thinks it's a `Lisp_Symbol` which confuses the code quite a bit and can cause assertions to fail.

Fix that and tag the pointer as "misc". Since this is fixed, make use of the conversions by accepting and returning `LispMarkerRef` in some lisp function implementations which would crash `temacs` before this change.

---

This tagging error is the cause of #1167 and it's probably related to  #820 and #1145 though I can't be positive as those are assertions on the contents.


nickdrozd:
It's a Christmas miracle.

PR #1179, opened 2018-12-26
Port window-redisplay-end-trigger and set-window-redisplay-end-trigger

agraven:
This solves issue #1165

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1182,7 +1182,7 @@ pub fn window_parameters(window: LispWindowValidOrSelected) -> LispObject {
 /// Return WINDOW's redisplay end trigger value.
 /// WINDOW must be a live window and defaults to the selected one.
 /// See `set-window-redisplay-end-trigger' for more information.
-#[lisp_fn]
+#[lisp_fn(min = "0")]
 pub fn window_redisplay_end_trigger(window: LispWindowLiveOrSelected) -> LispObject {
     let win: LispWindowRef = window.into();
     win.redisplay_end_trigger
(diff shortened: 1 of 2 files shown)


rust_src/src/windows.rs
```diff
@@ -1179,4 +1179,30 @@ pub fn window_parameters(window: LispWindowValidOrSelected) -> LispObject {
...
+}
+
+/// Set WINDOW's redisplay end trigger value to VALUE.
+/// WINDOW must be a live window and defaults to the selected one.  VALUE
+/// should be a buffer position (typically a marker) or nil.  If it is a
+/// buffer position, then if redisplay in WINDOW reaches a position beyond
+/// VALUE, the functions in `redisplay-end-trigger-functions' are called
+/// with two arguments: WINDOW, and the end trigger value.  Afterwards the
+/// end-trigger value is reset to nil.
+#[lisp_fn(min = "2")]
```
shaleh: This one does not. `2, 2, 0` means "min=2, max=2, interactive=NULL". The previous functions "0, 1, 0" means min-0, max=1.

shaleh:
BTW, there is a tool in rust_src/admin called 'rustize-emacs'. If you mark a region and then C-uM-| rustize-emacs it will convert the function definition to Rust. It should work on DEFUNs or on standard C functions. I say should because the parsing is just a regex so it fails sometime. Tweaks welcomed.

One of the advantages of the tool is it handles the min, max, interactive bits.

rust_src/src/windows.rs
shaleh: company-mode was correct.

(window-redisplay-end-trigger) is defined like this:

```
DEFUN ("window-redisplay-end-trigger", Fwindow_redisplay_end_trigger,
       Swindow_redisplay_end_trigger, 0, 1, 0, /* ...ellided doc.... */)
```

The first 0 is "min"
The 1 is "max"
and the final 0 is the NULL pointer for interactive.

If you look at `(set-window-redisplay-end-trigger)` it has 2,2,0. Since the first two numbers match there is no "min" for us to set. The min and the max are equal and match the number of arguments.

agraven: Oh, right. Thank you for the clarification.

PR #1180, opened 2018-12-27
Skip running bindgen if C source hasn't changed

carlosmn:
The cargo build script performs a few different tasks so we run it
whenever any of the source code has changed.

A particularly expensive task is running bindgen to generate Rust code
from the C code automatically. This can take minutes and we run it
even if only Rust code has changed, which would not produce a change.

Teach the Makefile to keep track of a hash for the C source code and
tell the cargo build script. Teach the build script to skip running
bindgen unless the Makefile has determined that it needs to be run.

### DIFF
--- a/rust_src/build.rs
+++ b/rust_src/build.rs
@@ -639,6 +639,13 @@ fn run_bindgen() {
             .expect("Could not create bindings.rs");
         return;
     }
+    // Set by the Makefile to non-empty value if we need to generate the bindings
+    let do_build = std::env::var_os("BUILD_BINDINGS");
+    match do_build.as_ref() {
+        Some(s) if s == "" => return,
+        None => return,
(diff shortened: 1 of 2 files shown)


carlosmn:
There's bound to be some edge cases here.... like it considers the task done based on the hash file so if you kill the bindgen run it will still consider it actually done, but that's probably a minor issue in practice.

Since we're already relying on a Makefile rule to say when we explicitly need it, the next logical step is to separate a executable that does bindgen and whatever else we do that depends on the C code (that can then be made to actually let make know whether it finishe). But this should provide some relief until that or something similar comes about.

shaleh:
Which means we should close this PR correct?

carlosmn:
yeah let's close this in favour of #1185

PR #1181, opened 2018-12-27
Port frame-after-make-frame (fix #1106)

### LINKED ISSUE #1106: Port frame-after-make-frame
```c
DEFUN ("frame-after-make-frame",
       Fframe_after_make_frame,
       Sframe_after_make_frame, 2, 2, 0,
       doc: /* Mark FRAME as made.
FRAME nil means use the selected frame.  Second argument MADE non-nil
means functions on `window-configuration-change-hook' are called
whenever the window configuration of FRAME changes.  MADE nil means
these functions are not called.

This function is currently called by `make-frame' only and should be
otherwise used with utter care to avoid that running functions on
`window-configuration-change-hook' is impeded forever.  */)
  (Lisp_Object frame, Lisp_Object made)
{
  f->after_make_frame = !NILP (made);
  f->inhibit_horizontal_resize = false;
  f->inhibit_vertical_resize = false;
  return made;
}
```

r-darwish:
This is my first attempt with the remacs project.

I think I got the implementation straight but when I launch remacs it says `Symbol’s function definition is void: frame-after-make-frame`.

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -508,4 +508,22 @@ pub fn previous_frame(frame: LispFrameOrSelected, miniframe: LispObject) -> Lisp
     }
 }
 
+/// Mark FRAME as made.
+/// FRAME nil means use the selected frame.  Second argument MADE non-nil
+/// means functions on `window-configuration-change-hook' are called
+/// whenever the window configuration of FRAME changes.  MADE nil means
+/// these functions are not called.
+///
+/// This function is currently called by `make-frame' only and should be
+/// otherwise used with utter care to avoid that running functions on
+/// `window-configuration-change-hook' is impeded forever.
+#[lisp_fn]
+pub fn frame_after_make_frame(frame: LispFrameOrSelected, made: LispObject) -> LispObject{
+    let mut frame_ref = frame.live_or_error();
+    frame_ref.set_after_make_frame(made.is_not_nil());
+    frame_ref.set_inhibit_horizontal_resize(false);
+    frame_ref.set_inhibit_vertical_resize(false);
+    made
+}
+
 include!(concat!(env!("OUT_DIR"), "/frames_exports.rs"));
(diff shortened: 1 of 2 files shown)


rust_src/src/frames.rs
```diff
@@ -508,4 +508,22 @@ pub fn previous_frame(frame: LispFrameOrSelected, miniframe: LispObject) -> Lisp
...
+/// FRAME nil means use the selected frame.  Second argument MADE non-nil
+/// means functions on `window-configuration-change-hook' are called
+/// whenever the window configuration of FRAME changes.  MADE nil means
+/// these functions are not called.
+///
+/// This function is currently called by `make-frame' only and should be
+/// otherwise used with utter care to avoid that running functions on
+/// `window-configuration-change-hook' is impeded forever.
+#[lisp_fn]
+pub fn frame_after_make_frame(frame: LispFrameOrSelected, made: LispObject) -> LispObject {
```
nickdrozd: `made` could be a `bool` instead (with the function returning `bool`).

shaleh: But it is passed through to further code. So it can't be.

carlosmn: It's used as as boolean and then returned, so `made: bool` and returning `bool` should work just fine.

brotzeit: The problem is that non-nil is t in elisp and non-nil would be returned. So if `made` is "foo" it would return t instead of "foo".

shaleh:
@r-darwish this looks ready to go. Switch `made` to be a `bool` as @nickdrozd and @carlosmn suggest and this can be merged.

We suspect you need to run `make clean` and re-run the build. This should resolve your problem testing the change. It works for 2 of us.

r-darwish:
Thanks for the feedback. I should be able to make the necessary changes at Sunday

rust_src/src/frames.rs
carlosmn: Fair point, it's returning the same object which could be anything. The documentation doesn't mention what it returns so it's all rather underspecified.

brotzeit:
@r-darwish Thanks!

shaleh:
@r-darwish on further review, it is safer to leave `made` as a LispObject. So no need for changes. Good work here.

PR #1185, opened 2018-12-28
Extract bindings generation and tighten dependency rules

carlosmn:
Instead of generating the bindings code to be inserted into
`remacs_sys.rs` from the build script, extract that code into its own
command that can be called from the Makefile.

These new rules for make can now depend on the header files which is
the source of our data, avoiding regeneration when the Rust or C
source files change.

These generated files are now stored under the source directory as we
do not have access to cargo's `OUT_DIR` variable. They go into their
own directory to break dependency cycles between header files and Rust
source code.


carlosmn:
I've kept the `SRC_HASH` stuff as it looks like maybe the exports generation might actually need it. Either way it runs pretty quickly.

As a further step in extracting bindings generation, I wonder if it might be faster overall to generate one `bindings_X.rs` file per header file we want to examine. Any O(n^2) behaviour in bindgen should get much better, plus we get to parallelize the work. But that's for another day.

carlosmn:
This was working locally but it looks like I might have messed up cleaning up my assortment of changes.

There's a circular dependency when I make the ato-generated files depend on `$(srcdir)/*.h` which make then breaks in such a way that we don't actually build the code. I'm not sure what the circle is, but I suppose it'd be related to `globals.h` depending on the Rust files.

carlosmn:
It turns out that is was due to `macuvs.h` being a generated file that depends on `emacs-bootstrap` to get created. Filtering it out looks to make this work though we won't regenerate if it changes. It is included in `wrapper.h` so presumably we might be interested in its data at some point, but it doesn't look like we use its contents now. It does come with the source code so we won't be missing it and it looks like it doesn't actually change when buliding it either way.

carlosmn:
Running the bindings generator in release mode brings down my local times from a bit over 3m down to 11s (once the binary is built). I'm not sure if we'll notice the difference too much in Travis as compilation will likely take a bit longer, but at least for local development this should help quite a bit.

rust_src/remacs-bindings/src/main.rs
```diff
@@ -0,0 +1,336 @@
+extern crate bindgen;
```
harryfei: We can remove this `extern crate` line now because of `2018 edition`.

rust_src/.gitignore
```diff
@@ -1 +1,2 @@
 target
+generated/
```
harryfei: I think `generated` is better than `generated/`.

carlosmn: Better how? `generated` is less specific and would also match files that end up named `generated`.

harryfei: It makes sense.

Wilfred:
@carlosmn I've added you as a collaborator, so you can also review other people's PRs if you wish :)

PR #1186, opened 2018-12-28
use Self

brotzeit:


brotzeit:
Ok now ?

brotzeit:
Let me think about it. Maybe I can come up with something better when I'm back home ;)

shaleh:
When I squash merge I copy the PR's description and use that as the commit message.

brotzeit:
@shaleh it's about the contents of the commit message.

shaleh:
Which are all lost when you squash.

brotzeit:
Is this about this PR or another one where I didn't squash correctly =)

shaleh:
:-)

My comments are general. If PR is made with a good description it is best to use it as the commit message. At work, we actually use a template for the PR (which GitHub supports and automates). The process to merge is:

 1. Click edit on the PR message.
 1. Select all, copy.
 1. Click Squash and merge
 1. Paste into the space, overriding the original contents.

shaleh:
I have not implemented the template because we have 2 PR styles at the moment. Ports are usually obvious. The other PR is like this one. Once we move to mostly Rust work I think a template is a great idea.

shaleh:
@brotzeit if you edit the PR description we can merge this with a Squash using the method I describe above.

brotzeit:
I'm still confused =)
Is this good to go or not ? This is only one commit I don't understand why we are talking about squashing.

shaleh:
@db48x is asking for a more helpful commit message. When we are porting functions a simple "Porting foo" is enough. For other work we should document what, why, etc. A year from now we may be crawling through commit messages looking to fix a bug.

You can either: `git commit --amend` and push again or write it here on the PR in the description and ensure that is pasted in when "Squash" is chosen which lets you re-write the git commit message. Squash works even for one commit.

brotzeit:
Yeah, I know. But I can't think of a better commit message ;)

PR #1191, opened 2018-12-30
port copy-alist

brotzeit:


rust_src/src/fns.rs
```diff
@@ -362,4 +362,32 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+/// Elements of ALIST that are not conses are also shared.
+#[lisp_fn]
+pub fn copy_alist(mut alist: LispObject) -> LispObject {
+    if alist.is_nil() {
+        return alist;
+    }
+
+    let new_alist = unsafe { lisp_concat(1, &mut alist, Lisp_Type::Lisp_Cons, false) };
+
+    let mut tem = new_alist;
```
shanavas786: what about

```rust
new_alist.iter_tails().map(|tail| set_car_here);
```

brotzeit: I will give it a try.

brotzeit: Is this a case where I could use `fold` ?

rust_src/src/fns.rs
```diff
@@ -362,4 +362,32 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+    while tem.is_not_nil() {
```
shaleh: This looks like it could be written:

    for elt in tem.iter_tails(/* no checks */) {
        let front = elt.car();
        if let Some((car, cdr)) = front.into() {
            elt.set_car((car, cdr).into());
        }
    }

You are correct in asking if this could be a fold. If we had better control of allocations the pseudo code would look more like:

    let new_alist_vec = alist.iter_cars(/* no checks */).fold(Vec<LispObject>::new(), |accum, elt| {
        let new_cons: (LispObject, LispObject) = elt.into();
        accum.append(new_cons);
        accum
    });
    new_alist_vec.into()

shaleh: @nickdrozd I think this example shows a place where the casting style is less effective. How would you write this?

shaleh: What we want is a way to say:

    if elt.is_cons() {
        accum.append(copy of elt);
    }

nickdrozd: Would a `while let` simplify it?

rust_src/src/fns.rs
```diff
@@ -362,4 +362,32 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+        debug_assert!(tem.is_cons());
+        let t = tem.as_cons().unwrap();
+        let car = t.car();
+
+        if let Some(c) = car.as_cons() {
+            setcar(t, LispObject::cons(c.car(), c.cdr()));
```
nickdrozd: So `t` is the memory location pointed at by the `LispObject` `tem`? This is getting into confusing object identity territory.

nickdrozd: So `t` is a cons cell composed of `car` and `cdr`, and `car is a cons cell composed of `c.car()` and `c.cdr()`. Then we take those last two items, and recompose them into a new cons cell, then make that new cons cell the new car of `t`. Is that right?

nickdrozd: https://github.com/Wilfred/remacs/pull/1262 should make that call to `LispObject::cons` unnecessary.

rust_src/src/fns.rs
shaleh: @nickdrozd I was referring to my hypothetical `fold` version. For now my first version or something close to it is probably the best.

```
for elt in tem.iter_tails(/* no checks */) {
    let front = elt.car();
    if let Some((car, cdr)) = front.into() {
        elt.set_car((car, cdr).into());
    }
}
```

rust_src/src/fns.rs
shaleh: The version I recommend above is much simpler.

brotzeit:
@shaleh Thanks.

brotzeit:
Somebody please review.

PR #1192, opened 2018-12-31
Port list-system-processes (fix #1187)

### LINKED ISSUE #1187: Port list-system-processes
```c
DEFUN ("list-system-processes", Flist_system_processes, Slist_system_processes,
       0, 0, 0,
       doc: /* Return a list of numerical process IDs of all running processes.
If this functionality is unsupported, return nil.

See `process-attributes' for getting attributes of a process given its ID.  */)
  (void)
{
  return list_system_processes ();
}
```

r-darwish:


rust_src/src/process.rs
```diff
@@ -528,4 +528,13 @@ pub fn process_running_child_p(mut process: LispObject) -> LispObject {
     }
 }
 
+/// Return a list of numerical process IDs of all running processes.
+/// If this functionality is unsupported, return nil.
+///
+/// See `process-attributes' for getting attributes of a process given its ID.
+#[lisp_fn]
+pub fn list_system_processes() -> LispObject {
```
brotzeit: Can you use `list_system_processes_rust` and `#[lisp_fn(name="list-system-processes"]`. That's what we usually do in this case.

PR #1196, opened 2019-01-02
Port a couple of functions to rust

shanavas786:


rust_src/src/buffers.rs
```diff
@@ -48,6 +48,7 @@ use crate::{
 
 pub const BEG: ptrdiff_t = 1;
 pub const BEG_BYTE: ptrdiff_t = 1;
+pub const UNKNOWN_MODTIME_NSECS: EmacsInt = -2;
```
shaleh: Should this be imported via remacs_sys?

rust_src/src/buffers.rs
```diff
@@ -383,6 +384,45 @@ impl LispBufferRef {
...
+        self.directory_ = if null_curr_buff {
+            Qnil
+        } else {
+            curr_buff.directory_
+        };
+        self.modtime = unsafe { make_timespec(0, UNKNOWN_MODTIME_NSECS) };
+        self.modtime_size = -1;
+        self.save_length_ = 0.into();
+        self.last_window_start = 1;
+        /* It is more conservative to start out "changed" than "unchanged".  */
```
shaleh: C style comment.
```suggestion
        // It is more conservative to start out "changed" than "unchanged".
```

rust_src/src/buffers.rs
```diff
@@ -1122,6 +1161,36 @@ pub unsafe fn per_buffer_idx(offset: isize) -> isize {
...
+        if overlays.is_null() {
+            return;
+        }
+
+        let mut overlays_iter = LispOverlayRef::from_ptr(list as *mut c_void)
+            .unwrap_or_else(|| panic!("Invalid overlay reference."))
+            .iter();
+
+        let mut overlay = overlays_iter.next();
+        while let Some(mut ov) = overlay {
```
shaleh: Why is this a `while` and not a `for`? I see `while` as "until we are done, not sure how long that is" and `for` as "for each item in this constrained container".

rust_src/src/buffers.rs
```diff
@@ -383,6 +384,45 @@ impl LispBufferRef {
         let pos = buffer_bytes.add(offset) as *mut LispObject;
         *pos = value;
     }
+
+    pub fn reset(&mut self) {
+        let curr_buff = ThreadState::current_buffer();
+        let null_curr_buff = ThreadState::is_current_buffer_null();
```
shaleh: This feels like a C style solution. Perhaps the `current_buffer()` needs to return an `Option`?

shaleh:
The CI failure is a segfault or other crash.

rust_src/src/buffers.rs
shanavas786: it needs to change `next` field. If we change it in a `for loop`, iterator will terminate immediately.
`while let` allows me to change the item without disturbing the iterator.

shaleh: ah. Makes sense.

PR #1197, opened 2019-01-02
Make rustdoc

r-darwish:
This feature wasn't requested here. However, I find it very useful when developing, and I think it might be of use to other developers as well.


shaleh:
Should we include this in the CI testing.

Issue #1175, opened 2018-12-24
Port move-to-column


brotzeit:
```c
DEFUN ("move-to-column", Fmove_to_column, Smove_to_column, 1, 2,
       "NMove to column: ",
       doc: /* Move point to column COLUMN in the current line.
Interactively, COLUMN is the value of prefix numeric argument.
The column of a character is calculated by adding together the widths
as displayed of the previous characters in the line.
This function ignores line-continuation;
there is no upper limit on the column number a character can have
and horizontal scrolling has no effect.

If specified column is within a character, point goes after that character.
If it's past end of line, point goes to end of line.

Optional second argument FORCE non-nil means if COLUMN is in the
middle of a tab character, change it to spaces.
In addition, if FORCE is t, and the line is too short to reach
COLUMN, add spaces/tabs to get there.

The return value is the current column.  */)
  (Lisp_Object column, Lisp_Object force)
{
  ptrdiff_t pos, prev_col;
  EMACS_INT col;
  EMACS_INT goal;

  CHECK_NATNUM (column);
  goal = XINT (column);

  col = goal;
  pos = ZV;
  scan_for_column (&pos, &col, &prev_col);

  SET_PT (pos);

  /* If a tab char made us overshoot, change it to spaces
     and scan through it again.  */
  if (!NILP (force) && col > goal)
    {
      int c;
      ptrdiff_t pos_byte = PT_BYTE;

      DEC_POS (pos_byte);
      c = FETCH_CHAR (pos_byte);
      if (c == '\t' && prev_col < goal)
	{
	  ptrdiff_t goal_pt, goal_pt_byte;

	  /* Insert spaces in front of the tab to reach GOAL.  Do this
	     first so that a marker at the end of the tab gets
	     adjusted.  */
	  SET_PT_BOTH (PT - 1, PT_BYTE - 1);
	  Finsert_char (make_number (' '), make_number (goal - prev_col), Qt);

	  /* Now delete the tab, and indent to COL.  */
	  del_range (PT, PT + 1);
	  goal_pt = PT;
	  goal_pt_byte = PT_BYTE;
	  Findent_to (make_number (col), Qnil);
	  SET_PT_BOTH (goal_pt, goal_pt_byte);

	  /* Set the last_known... vars consistently.  */
	  col = goal;
	}
    }

  /* If line ends prematurely, add space to the end.  */
  if (col < goal && EQ (force, Qt))
    Findent_to (make_number (col = goal), Qnil);

[...]

PR #1198, opened 2019-01-02
Port move-to-column

agraven:
This solves issue #1175.


brotzeit:
Do you think it makes sense to add some tests ?

agraven:
I know there are some tests that fail if this function misbehaves, but I suppose it would still make sense write tests for the function itself. I'm unsure if I have the elisp skills to write said tests, but I'll give it a go.

rust_src/src/indent.rs
```diff
@@ -74,9 +74,9 @@ pub fn current_column() -> EmacsInt {
 ///
 /// The return value is the current column.
 #[lisp_fn(min = "1", intspec = "NMove to column")]
-pub fn move_to_column(column: EmacsInt, force: bool) -> EmacsInt {
+pub fn move_to_column(column: EmacsInt, force: LispObject) -> EmacsInt {
     column.check_natnum();
-    let buffer = &mut ThreadState::current_buffer();
+    let buffer = &mut ThreadState::current_buffer().unwrap();
```
shaleh: Use the `unchecked` variant instead of manually unwrapping.

rust_src/src/indent.rs
```diff
@@ -103,8 +103,10 @@ pub fn move_to_column(column: EmacsInt, force: bool) -> EmacsInt {
 
                 // Delete the tab and indent to COL
                 del_range(buffer.pt, buffer.pt + 1);
+                let goal_pt = buffer.pt;
```
shaleh: Why add the temporary variables?

agraven: Because `Findent_to` changes the value of `buffer.pt{,_byte}`

rust_src/src/indent.rs
agraven: Fixed

rust_src/src/indent.rs
shaleh: Ah, ok.

rust_src/src/indent.rs
shaleh: Did it revert?

shaleh:
Think of it like this. "Load a buffer with known content. Move to a location in the buffer. Is (point) where you expect?"

rust_src/src/indent.rs
```diff
@@ -54,6 +56,78 @@ pub fn current_column() -> EmacsInt {
...
+///
+/// Optional second argument FORCE non-nil means if COLUMN is in the
+/// middle of a tab character, change it to spaces.
+/// In addition, if FORCE is t, and the line is too short to reach
+/// COLUMN, add spaces/tabs to get there.
+///
+/// The return value is the current column.
+#[lisp_fn(min = "1", intspec = "NMove to column")]
+pub fn move_to_column(column: EmacsInt, force: LispObject) -> EmacsInt {
+    column.check_natnum();
```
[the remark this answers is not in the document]
agraven: Thanks for the pointer, was not aware that type existed.

agraven:
I just realized I made a mistake, this implementation only turns the part of the tab the left of `column` to spaces, instead of the entire tab.

test/rust_src/src/indent-tests.el
```diff
@@ -4,6 +4,31 @@
 
 (require 'ert)
 
+(ert-deftest move-to-column ()
```
shaleh: Great tests. I think it would be better if they were broken up into smaller functions so it was easier to debug a failure.

agraven: Done. I got by by using the backtrace to see where things went wrong, but I can see how breaking it up makes it a lot easier to quickly see what misbehaved.

PR #1204, opened 2019-01-03
Port clear-face-cache (fix #1164)

### LINKED ISSUE #1164: Port clear-face-cache
```c
DEFUN ("clear-face-cache", Fclear_face_cache, Sclear_face_cache, 0, 1, 0,
       doc: /* Clear face caches on all frames.
Optional THOROUGHLY non-nil means try to free unused fonts, too.  */)
  (Lisp_Object thoroughly)
{
  clear_face_cache (!NILP (thoroughly));
  face_change = true;
  windows_or_buffers_changed = 53;
  return Qnil;
}
```

r-darwish:
1. I'd like to give that `53` a meaningful constant name, but I don't know what this value means
2. I can't get this to compile, even when cleaning the build. This is the first time I'm making a new Rust module in Remacs so I probably missed something:

```
/usr/bin/ld: composite.o: in function `Fclear_composition_cache':
/home/roeyd/src/remacs/src/composite.c:700: undefined reference to `Fclear_face_cache'
/usr/bin/ld: fontset.o: in function `free_realized_fontsets':
/home/roeyd/src/remacs/src/fontset.c:1327: undefined reference to `Fclear_face_cache'
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/rust_src/src/xfaces.rs
+++ b/rust_src/src/xfaces.rs
@@ -1,0 +1,21 @@
+//! "Face" primitives.
+
+use remacs_macros::lisp_fn;
+
+use crate::{
+    lisp::defsubr,
+    remacs_sys::{clear_face_cache, set_face_change, windows_or_buffers_changed},
+};
+
+/// Clear face caches on all frames.
+/// Optional THOROUGHLY non-nil means try to free unused fonts, too.
+#[lisp_fn(min = "0", name = "clear-face-cache", c_name = "clear_face_cache")]
+pub fn clear_face_cache_lisp(thoroughly: bool) {
+    unsafe {
+        clear_face_cache(thoroughly);
+        set_face_change(true);
+        windows_or_buffers_changed = 53;
+    }
+}
+
+include!(concat!(env!("OUT_DIR"), "/xfaces_exports.rs"));
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -122,6 +122,7 @@ mod util;
 mod vectors;
 mod window_configuration;
 mod windows;
+mod xfaces;
 mod xml;
 
 #[cfg(all(not(test), target_os = "macos", feature = "unexecmacosx"))]
--- a/src/dispextern.h
+++ b/src/dispextern.h
@@ -1812,6 +1812,8 @@ GLYPH_CODE_P (Lisp_Object gc)
 
 extern bool face_change;
 
+void set_face_change(bool value);
+
 /* For reordering of bidirectional text.  */
 
 /* UAX#9's max_depth value.  */
--- a/src/xfaces.c
+++ b/src/xfaces.c
@@ -356,6 +356,11 @@ static struct face *realize_non_ascii_face (struct frame *, Lisp_Object,
 			      Utilities
  ***********************************************************************/
 
+void set_face_change(bool value)
+{
+    face_change = value;
+}
+
 #ifdef HAVE_X_WINDOWS
 
 #ifdef DEBUG_X_COLORS
@@ -644,17 +649,6 @@ clear_face_cache (bool clear_fonts_p)
       clear_image_caches (Qnil);
     }
 #endif /* HAVE_WINDOW_SYSTEM */
-}
-
-DEFUN ("clear-face-cache", Fclear_face_cache, Sclear_face_cache, 0, 1, 0,
-       doc: /* Clear face caches on all frames.
-Optional THOROUGHLY non-nil means try to free unused fonts, too.  */)
-  (Lisp_Object thoroughly)
-{
-  clear_face_cache (!NILP (thoroughly));
-  face_change = true;
-  windows_or_buffers_changed = 53;
-  return Qnil;
 }
 
 
@@ -6303,7 +6297,6 @@ syms_of_xfaces (void)
   defsubr (&Sdump_face);
   defsubr (&Sshow_face_resources);
 #endif /* GLYPH_DEBUG */
-  defsubr (&Sclear_face_cache);
   defsubr (&Stty_suppress_bold_inverse_default_colors);
 
 #if defined DEBUG_X_COLORS && defined HAVE_X_WINDOWS


brotzeit:
Did you run `cargo clean` in `rust_src/` ?

r-darwish:
Yeah. I ran `make clean`. The build also seems to fail in [Travis](https://travis-ci.org/Wilfred/remacs/jobs/474846263#L1601). 

There's a different error in Travis from what I described, but it's probably due to the fact that I didn't run `./configure` with `WERROR_CFLAGS='-Werror -Wno-error=deprecated-declarations'` like Travis did. The two errors probably caused by the same problem. 

The function in `composite.c` tries to call `Fclear_face_cache`, which is now in Rust. Can a C function call a Lisp function implemented in Rust?

rust_src/src/xfaces.rs
```diff
@@ -0,0 +1,25 @@
+use crate::{
```
shaleh: 1) The file should begin with `//! Description`. Grab the one from the top of the C file and update as appropriate.
2) Our style is to list std, then external crates, then our crates, then the src/ modules.

rust_src/src/xfaces.rs
```diff
@@ -0,0 +1,25 @@
...
+    lisp::defsubr,
+    remacs_sys::{clear_face_cache, windows_or_buffers_changed},
+};
+use remacs_macros::lisp_fn;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+lazy_static! {
+    static ref FACE_CHANGE: AtomicBool = AtomicBool::new(false);
```
shaleh: Do not replicate the variables from C when porting. This leads to bugs where C sets a value and Rust does not see it.
Remove the original `FACE_CHANGE` and see how it goes.

rust_src/src/xfaces.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/// Clear face caches on all frames.
+/// Optional THOROUGHLY non-nil means try to free unused fonts, too.
+#[lisp_fn(name = "clear-face-cache")]
```
shaleh: There are two issues here. The first is why the compilation breaks. As written, this exports a function called `Frust_clear_face_cache`. To fix this `cname` also needs to be set. The other issue is `thoroughly` is meant to be optional. This means settings `min = "0"`.

rust_src/src/xfaces.rs
```diff
@@ -0,0 +1,25 @@
...
+pub fn rust_clear_face_cache(thoroughly: bool) {
+    unsafe {
+        clear_face_cache(thoroughly);
+    }
+    FACE_CHANGE.store(true, Ordering::SeqCst);
+    unsafe {
+        windows_or_buffers_changed = 53;
```
shaleh: `53` is an arbitrary value. If you look at the code `windows_or_buffers_changed` is set to a different value each time it is assigned. This allows one to debug where the value came from. The value is only checked for zero or non-zero.

shaleh: See this comment in xdisp.c.

```
  Mostly for historical reasons, windows_or_buffers_changed can also take
   other non-zero values.  In that case, the precise value doesn't matter (it
   encodes the cause of the setting but is only used for debugging purposes),
   and what it means is that we shouldn't pay attention to any `redisplay' bits
   and we should simply try and redisplay every window out there.
```

shaleh:
I opened #1205 after reviewing this PR. We were not consistently using `c_name` which could be confusing.

rust_src/src/xfaces.rs
```diff
@@ -0,0 +1,25 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+
+lazy_static! {
+    static ref FACE_CHANGE: AtomicBool = AtomicBool::new(false);
+}
+
+/// Clear face caches on all frames.
+/// Optional THOROUGHLY non-nil means try to free unused fonts, too.
+#[lisp_fn(name = "clear-face-cache")]
+pub fn rust_clear_face_cache(thoroughly: bool) {
```
shaleh: The proper name is `clear_face_cache_lisp`.

rust_src/src/xfaces.rs
r-darwish: If I remove the variable from C then I'll need to export a getter and a setter from Rust and use it in the C code until everything is ported. Do I understand correctly?

rust_src/src/xfaces.rs
r-darwish: Do we want to keep the `53` floating here or do we want to convert it to an enum that looks something like:

```rust
enum ChangeCode {
    ClearFaceCache = 53,
    ...
}
```

?

rust_src/src/xfaces.rs
r-darwish: https://github.com/r-darwish/remacs/blob/clear-face-cache/src/xfaces.c#L22

Should I just write the first line, which is "Faces.", or should I convert the whole paragraph?

shaleh: `"Face" primitives.` is what you want.

rust_src/src/xfaces.rs
shaleh: Sounds right. Which is why we usually use the C variables until all of the code which needs them is ported then port the variable last.

rust_src/src/xfaces.rs
shaleh: The use case is a developer in a debugger who sees '53' in `windows_or_buffers_changed` so they know the most recent thing to set the value was `Fclear_face_cache`. The enum would not help here since the C code is still using an int.

rust_src/src/xfaces.rs
r-darwish: Maybe I'm missing something, but I don't see the original `face_change` listed in the doc of `remacs_sys`. How can I access the original C variable from Rust?

shaleh: If the variable is only used in the one C file then you need to expose it in a header.

shaleh: or right a C function to set/get and expose them.

r-darwish: It appears [here](https://github.com/r-darwish/remacs/blob/clear-face-cache/src/dispextern.h#L1813).  Is it unavailable in `emacs_sys` due to [this](https://github.com/r-darwish/remacs/blob/clear-face-cache/rust_src/remacs-bindings/src/main.rs#L258) blacklist?

r-darwish: I added up an setter function because I don't want to alter existing C code. We can add the getter one it's needed.

rust_src/src/xfaces.rs
```diff
@@ -0,0 +1,21 @@
...
+use remacs_macros::lisp_fn;
+
+use crate::{
+    lisp::defsubr,
+    remacs_sys::{clear_face_cache, set_face_change, windows_or_buffers_changed},
+};
+
+/// Clear face caches on all frames.
+/// Optional THOROUGHLY non-nil means try to free unused fonts, too.
+#[lisp_fn(name = "clear-face-cache", min = "0", c_name = "clear_face_cache")]
```
shaleh:
```suggestion
#[lisp_fn(min = "0", name = "clear-face-cache", c_name = "clear_face_cache")]
```

r-darwish:
@shaleh I don't understand why only some of the Travis builds failed. Could you please take a look?

shaleh:
The MacOS CI has been failing for a while, we have an open issue to resolve that. If you see only flymake and epg failures then the tests passed on Mac.

As for the Linux failure, hmmm. Not sure. The 4 CI tests are Linux and Mac, then Linux and Mac again but this time without the Graphical UI enabled.

r-darwish:
CI passes now. Is there anything else that needs to be done?

PR #1205, opened 2019-01-03
If the Rust name is not the expected C name we have to set it.

shaleh:
The function name exported to C is the Rust name with a capital F
prepended. If this is not the intended name then we have to set the
correct name with `c_name`. The value of `name` is _NOT_ used because
it could be an invalid C (and Rust) name like for instance `+`.

Also, this PR begins standardizing our naming. Rust versions of a DEFUN
which cannot use the orignal name are named `foo_lisp`.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -1038,7 +1038,11 @@ pub fn delete_overlay(overlay: LispOverlayRef) {
 
 /// Delete all overlays of BUFFER.
 /// BUFFER omitted or nil means delete all overlays of the current buffer.
-#[lisp_fn(min = "0", name = "delete-all-overlays")]
+#[lisp_fn(
+    min = "0",
+    name = "delete-all-overlays",
+    c_name = "delete_all_overlays"
+)]
 pub fn delete_all_overlays_lisp(buffer: LispBufferOrCurrent) {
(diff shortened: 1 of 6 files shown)


brotzeit:
I thought we only set c_name when the lisp function is called from C ?

shaleh:
Which can lead to confusion for new porters. Why not just be consistent? The C code was.

brotzeit:
Less noise and you can see if a lisp function is used in the rust/C code base. But it doesn't really matter since we eventually change the names so there are no conflicts ?

shaleh:
Some sources of conflict are due to there being an internal function with the same name. Having the `lisp_fn` version be named `foo_lisp` is a nice indicator that the two functions belong together. In other places we have been able to merge the two functions once all of the C side uses are gone.

When there is no more C, the `c_name` support and the `F` prepended functions all go away.

brotzeit:
I would open an issue to derive `c_name` from `name` ?

shaleh:
As mentioned in the description. We can _NOT_ use `name`. The name might be '+' for instance. Not a valid Rust or C identifier. `c_name` cannot be derived from anything but the Rust function name. Which is what happens currently.

brotzeit:
Ah yeah I forgot. It's a little silliy we have to do that for a hand full of functions.

PR #1206, opened 2019-01-03
respect our naming conventions

brotzeit:
Thanks @shaleh

### DIFF
--- a/rust_src/src/process.rs
+++ b/rust_src/src/process.rs
@@ -12,8 +12,8 @@ use crate::{
     multibyte::LispStringRef,
     remacs_sys::{
         add_process_read_fd, current_thread, delete_read_fd, emacs_get_tty_pgrp,
-        get_process as cget_process, list_system_processes, send_process,
-        setup_process_coding_systems, update_status, Fmapcar, STRING_BYTES,
+        list_system_processes, send_process, setup_process_coding_systems, update_status, Fmapcar,
+        STRING_BYTES,
     },
     remacs_sys::{pvec_type, EmacsInt, Lisp_Process, Lisp_Type, Vprocess_alist},
     remacs_sys::{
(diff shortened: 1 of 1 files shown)


brotzeit:
Hmmm clippy. I'm not sure about that...

shaleh:
If you look at it, there is no reason for the `if` as you have the code currently. `cget_process` was a different function. If there is now only `get_process` then get rid of the `if`.

rust_src/src/process.rs
```diff
@@ -265,15 +265,11 @@ pub fn set_process_plist(mut process: LispProcessRef, plist: LispObject) -> Lisp
 /// nil, indicating the current buffer's process.
 #[lisp_fn]
 pub fn process_status(process: LispObject) -> LispObject {
-    let p = if process.is_string() {
-        get_process(process)
-    } else {
-        unsafe { cget_process(process) }
+    let mut process = match get_process(process) {
+        Qnil => return process,
+        p => p.as_process_or_error(),
```
shaleh: @nickdrozd has been working to remove these `or_error` calls and replacing them with `.into()` instead.

brotzeit: Ah, right.

brotzeit: We need to update the wiki pages...

shaleh: I opened #1253

brotzeit: I never liked this _or_error naming.

shaleh: pros and cons. `as_foo_or_error()` lets me know it will die if it fails. `foo.into()` is now sometimes safe and sometimes not depending on whether it returns an Option or if there is a fall back. But yeah, the code is moving towards a little less verbosity which is probably good.

PR #1207, opened 2019-01-03
Intern should only accept strings

nickdrozd:
This is the GNU Emacs behavior.

### DIFF
--- a/rust_src/src/minibuf.rs
+++ b/rust_src/src/minibuf.rs
@@ -367,7 +367,7 @@ pub fn read_command_or_variable(
     if name.is_nil() {
         name
     } else {
-        lisp_intern(name, Qnil)
+        lisp_intern(name.into(), None)
     }
 }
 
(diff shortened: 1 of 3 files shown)


shaleh:
`fn lookup(self, name)` wants a Symbol or a String. If that was represented by a type you could avoid converting the object from a String to a Object to a String. The intern code would look like:

`self.lookup(LispSymbolOrString::from(string));  // string is a LispStringRef still`

nickdrozd:
@shaleh I don't know, that seems like a lot of ceremony for a one-off mixed type. It would be nice if we could somehow specify it in place, like `<LispSymbolRef | LispStringRef>`, or something like that.

shaleh:
Perhaps. But it is common in strong type systems. If we could name it we could use a Union type like Option. Have a Symbol wrapper and a String wrapper.

nickdrozd:
There are a few other places where that type would be useful, like `string-equal`, so I am convinced of its usefulness. But implementing that is outside the scope of this change. I'll open an issue.

PR #1210, opened 2019-01-04
Add from_lispobject_for_option macro

nickdrozd:
Ideally, we would be able to write this:

```rust
  impl<T: From<LispObject>> From<LispObject> for Option<T> {
      fn from(o: LispObject) -> Option<T> {
          if o.is_nil() {
              None
          } else {
              Some(T::from(o))
          }
      }
  }
```

Due to the so-called "orphan rule", this is not possible; external
traits (like `From`) can't be implemented for external types (like
`Option`). This may change in the future, but for now a bunch of
boilerplate is required. To get around this, we can use a macro.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -417,11 +417,7 @@ impl From<LispBufferRef> for LispObject {
     }
 }
 
-impl From<LispObject> for Option<LispBufferRef> {
-    fn from(o: LispObject) -> Self {
-        o.as_buffer()
-    }
-}
+from_lispobject_for_option!(LispBufferRef);
 
 impl LispObject {
     pub fn is_overlay(self) -> bool {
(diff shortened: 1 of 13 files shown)


rust_src/src/buffers.rs
```diff
@@ -553,19 +545,7 @@ impl From<LispObject> for LispBufferOrName {
...
 }
 
-impl From<LispObject> for Option<LispBufferOrName> {
-    fn from(v: LispObject) -> Option<LispBufferOrName> {
-        if v.is_nil() {
-            None
-        } else if v.is_string() {
-            Some(LispBufferOrName::Name(v))
-        } else if v.is_buffer() {
-            Some(LispBufferOrName::Buffer(v))
```
shaleh: This is a change in behavior. Now `or_error` is called before, this would return None.

rust_src/src/lists.rs
```diff
@@ -240,15 +240,7 @@ impl From<LispObject> for LispCons {
     }
 }
 
-impl From<LispObject> for Option<LispCons> {
-    fn from(o: LispObject) -> Self {
-        if o.is_list() {
```
shaleh: Also a change in behavior. Here if `as_cons()` was called on a fixnum the result would be None. Now it is an error.

nickdrozd:
Serves me right for seeing that it compiles and then immediately submitting :)

nickdrozd:
I get it, the test function is different for each one. It can be `is_not_nil`, `is_cons`, etc to see when `Some` should be returned.

PR #1212, opened 2019-01-05
add libvterm support

brotzeit:
I have a TODO list of the review from the old PR, but I would like to get this merged as I still hope I get some help with this =)

However I still have problems with configure.ac. I can't get this to work.
`.rustified_enum("VTermProp")` doesn't work since #1185. It worked before...

Apart from that, it seems to be usable. But resizing needs to get fixed.


shaleh:
Perhaps the right thing to do is leave this in a feature branch while the kinks are worked through.

brotzeit:
@shaleh ok I will work some more on this. Can you please take a look at configure.ac. I don't know why `HAVE_LIBVTERM` doesn't work.

shaleh:
@brotzeit I pushed a fix to the configure script and used it to make vterm.c become empty if libvterm is not found. That may need some tweaking, but the detection is now solid.

`EMACS_CHECK_MODULES` is all you needed. It uses `pkgconfig` to detect the library in question. An extra gotcha is the library defines itself as `vterm` in its .pc file, not `libvterm`.

When `EMACS_CHECK_MODULES` finds a library it sets `HAVE_$LIBRARY=yes` for you. Otherwise it is set to no. The extra bit which looks weird is you then have to call `AC_DEFINE` which exports the `HAVE_FOO` so it is in config.h and usable in C land.

shaleh:
Ok, I just pushed some changes that should enable you to allocate the `vterminal` as a pseudovector. `vterm_new` is not defined so I could not easily test it. Details are in the commit message.

brotzeit:
@shaleh Thanks. You're changes regarding the pseudovector were the same as mine, but it didn't work when I tried this because we were didn't tag correctly.

With `LispObject::tag_ptr(v, Lisp_Type::Lisp_Vectorlike)` it seems to work fine.

shaleh:
@brotzeit you also missed adding the header to the top of `struct vterminal`.

brotzeit:
No, I had that one. First post after your last post in https://github.com/Wilfred/remacs/pull/1130 ;)

I guess without that it wouldn't have ran in the first place. It was just the segfault after a few insertions when `vterm` was already running. 

Tagging solved the problem. I wonder if 
```rust
    pub fn from_ptr(ptr: *mut c_void) -> Option<Self> {
        unsafe { ptr.as_ref().map(|p| mem::transmute(p)) }
    }
```
is also a source of potential problems ?

shaleh:
Yeah, can be potentially dangerous if the wrong type is used. However, it is taking a pointer after it has been untagged. And the object returned should be retagged. If that is true it is safe.

rust_src/build.rs
```diff
@@ -349,10 +349,20 @@ fn env_var(name: &str) -> String {
 }
 
 // What to ignore when walking the list of files
+#[cfg(feature = "libvterm")]
 fn ignore(path: &str) -> bool {
     path == "" || path.starts_with('.') || path == "lib.rs" || path == "functions.rs"
 }
 
+#[cfg(not(feature = "libvterm"))]
```
shaleh: Instead of repeating the base `ignore` maybe you should make a `base_ignore` which is all that the vterm version uses and the non-vterm version is it plus vterm.rs.

rust_src/src/vterm.rs
```diff
@@ -53,9 +53,9 @@ impl LispVterminalRef {
         (height, width)
     }
 
-    pub fn set_size(self, rows: EmacsInt, cols: EmacsInt) {
+    pub fn set_size(self, rows: c_int, cols: c_int) {
```
shaleh: Do we want to continue using `c_int` or should we transition to `i32`?

brotzeit:
I have another problem that confuses me. When I use `describe-variable` with the buffer local variable `vterm--term` in `vterm-mode`, I get a segfault.

brotzeit:
And resizing horizontally after window size change also doesn't work correctly.

brotzeit:
`vterm_set_size` only sets the height and not the width. @leonerd any idea what I'm doing wrong here ? It seems to work for neovim...

rust_src/src/vterm.rs
```diff
@@ -0,0 +1,584 @@
...
+    pub fn get_size(self) -> (i32, i32) {
+        let mut height: i32 = 0;
+        let mut width: i32 = 0;
+        unsafe { vterm_get_size((*self).vt, &mut height, &mut width) };
+        (height, width)
+    }
+
+    pub fn set_size(self, rows: i32, cols: i32) {
+        unsafe {
+            vterm_set_size((*self).vt, rows, cols);
```
brotzeit: @leonerd Here. When I resize the window while htop is running it only updates the height. But when I exit htop and run it again it also gets the correct width. Maybe the reason is how emacs handles resizing...

brotzeit:
@leonerd I have read the term in neovim and emacs-libvterm so I thought this describes what we both mean ;)

But that's what I thought. Thanks.

shaleh:
@brotzeit once @Wilfred sets up the remacs project you could bring libvterm there instead of adding more code to this quite large repo.

brotzeit:
@shaleh Can you give me some source on how to add this to the build process. I was hoping I can take this from [emacs-libvterm](https://github.com/akermu/emacs-libvterm/commit/3929a9d573d89365fc68053824045ef654ef1146) but I guess I can't use this cmake stuff ?

rust_src/src/vterm.rs
akermu: The issue is caused by the `COLUMNS` and `LINES` environment variable. Unset these before invoking htop and resizing will work.

shaleh:
My expectation is the user first builds the vterm library and puts it somewhere linkable. Like says /usr/local/lib. Then they build remacs and the remacs configure script will find and set up the linking for vterm. Does this match your expectation @brotzeit?

brotzeit:
The only problem I see is that users get the wrong version for our remacs implementation. That's the reason why you told me to add libvterm to the remacs organization right ?

shaleh:
Having the crate in the remacs org makes keeping the versions in sync easier. The user installs the vterm lib and then builds remacs.

brotzeit:
I gave up with libvterm as a submodule. It seemed to me like a nice idea, but I don't get how this works.

I think with @akermu 's fixes it works pretty well. It would be great if we could start a final review.

brotzeit:
Scrollback still needs fixing, but since we make this optional this should be ok. We just mention in the readme libvterm support is not stable. Maybe some users get motivated to help with it when they see it works.

leonerd:
It might be of interest to you folks that I've now made a v0.1 tarball release. Though bits of API are still verymuch subject to change, so expect that v0.2 will probably have some incompatible bits in it.

http://www.leonerd.org.uk/code/libvterm/

PR #1213, opened 2019-01-05
impl From<(S, T)> for LispObject

nickdrozd:
Calls to LispObject::cons would look cleaner if they were just
tuples. We can have the compiler do the conversions for us. This makes
the code a little Lispier.

Note that the conversion works recursively, so nested pairs don't need
to be handled as a special case.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -827,14 +827,13 @@ pub extern "C" fn nsberror(spec: LispObject) -> ! {
 /// However, the overlays you get are the real objects that the buffer uses.
 #[lisp_fn]
 pub fn overlay_lists() -> LispObject {
-    let list_overlays = |ol: LispOverlayRef| -> LispObject {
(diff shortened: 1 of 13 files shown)


nickdrozd:
Whenever I learn a cool new programming technique, I try to use it as much as possible. Currently that technique for me is generics. (After years of being on the side of dynamic types, I'm trying to give static typing a shot.)

shaleh:
Just remember code is read as much as it is executed. If we make it too hard to understand it can make it more difficult to bring in new coders. Plus, if we put the code down for a bit it can become foreign even to ourself.

nickdrozd:
That's a good point. I think as far as legibility goes, there are two groups of people who are interested in Remacs, people who want to learn Emacs internals and people who want to mess with Rust. This change pushes some of the complexity out of the Lispier parts of the code and into the Rustier parts. Hopefully this will satisfy both groups: the Emacs internals people don't want to deal with the more draconian parts of Rust, while the Rust people might enjoy seeing more powerful parts of the language used.

PR #1214, opened 2019-01-05
Port interrupt-process

SanchayanMaity:
Provide a Rust implementation for the interrupt-process
function.

This implements #1195.

Signed-off-by: Sanchayan Maity <maitysanchayan@gmail.com>

### DIFF
--- a/rust_src/src/process.rs
+++ b/rust_src/src/process.rs
@@ -5,6 +5,7 @@ use remacs_macros::lisp_fn;
 
 use crate::{
     buffers::{current_buffer, get_buffer, LispBufferOrName, LispBufferRef},
+    eval::run_hook_with_args_until_success,
     lisp::defsubr,
     lisp::{ExternalPtr, LispObject, ProcessIter},
     lists::{assoc, car, cdr, plist_put},
@@ -17,8 +18,8 @@ use crate::{
     remacs_sys::{pvec_type, EmacsInt, Lisp_Process, Lisp_Type, Vprocess_alist},
     remacs_sys::{
         QCbuffer, QCfilter, QCsentinel, Qcdr, Qclosed, Qexit, Qinternal_default_process_filter,
-        Qinternal_default_process_sentinel, Qlisten, Qlistp, Qnetwork, Qnil, Qopen, Qpipe,
-        Qprocessp, Qreal, Qrun, Qserial, Qstop, Qt,
+        Qinternal_default_process_sentinel, Qinterrupt_process_functions, Qlisten, Qlistp,
+        Qnetwork, Qnil, Qopen, Qpipe, Qprocessp, Qreal, Qrun, Qserial, Qstop, Qt,
     },
 };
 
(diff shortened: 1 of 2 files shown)


shaleh:
Start here https://github.com/Wilfred/remacs/blob/master/CONTRIBUTING.md then read on the wiki too https://github.com/Wilfred/remacs/wiki/Porting-Cookbook

rust_src/src/process.rs
```diff
@@ -537,4 +538,24 @@ pub fn list_system_processes_rust() -> LispObject {
...
 
+/// Interrupt process PROCESS
+/// PROCESS may be a process, a buffer, or the name of a process or buffer.
+/// No arg or nil means current buffer's process.
+/// Second arg CURRENT-GROUP non-nil means send signal to
+/// the current process-group of the process's controlling terminal
+/// rather than to the process's own process group.
+/// If the process is a shell, this means interrupt current subjob
+/// rather than the shell.
+
```
shaleh: No gaps should be here. Add `///` for the empty lines.

rust_src/src/process.rs
```diff
@@ -537,4 +538,24 @@ pub fn list_system_processes_rust() -> LispObject {
...
+/// If CURRENT-GROUP is `lambda', and if the shell owns the terminal,
+/// don't send the signal.
+
+/// This function calls the functions of `interrupt-process-functions' in
+/// the order of the list, until one of them returns non-`nil'.  */)
+#[lisp_fn(name="interrupt-process")]
```
shaleh: The only reason to use `name` is if the Lisp DEFUN has the same name as an existing function. In the C code name in the `DEFUN` macro is changed from `foo-bar` to `Ffoo_bar`. In Rust we define `foo_bar` and a macro creates a wrapper function called `Ffoo_bar` for us.

I do not see another `interrupt_process` so there is no need to name the function `_rust`. Also, we are trying to transition to using `_lisp` for the name of functions that are interfaces to the DEFUNs and `_rust` for internal functions that need to be differentiated from C or DEFUN code.

rust_src/src/process.rs
```diff
@@ -537,4 +538,24 @@ pub fn list_system_processes_rust() -> LispObject {
...
+pub fn interrupt_process_rust(process: LispObject,
```
shaleh:
```suggestion
pub fn interrupt_process(process: LispObject,
```

rust_src/src/process.rs
shaleh:
```suggestion
#[lisp_fn(min = "0")]
```

After `Sinterrupt_process` there are 3 numbers. They represent the minimum number of required arguments, the maximum allowed, and finally whether the DEFUN is interactive. Zero here works like a NULL. Otherwise there will be a string.

SanchayanMaity:
@shaleh Thank you very much for the feedback. I will go through it and do the necessary changes.

rust_src/src/process.rs
```diff
@@ -537,4 +538,23 @@ pub fn list_system_processes_rust() -> LispObject {
...
+/// the current process-group of the process's controlling terminal
+/// rather than to the process's own process group.
+/// If the process is a shell, this means interrupt current subjob
+/// rather than the shell.
+///
+/// If CURRENT-GROUP is `lambda', and if the shell owns the terminal,
+/// don't send the signal.
+///
+/// This function calls the functions of `interrupt-process-functions' in
+/// the order of the list, until one of them returns non-`nil'.  */)
```
brotzeit: The docstring needs fixing `*/)`

SanchayanMaity: Fixed

rust_src/src/process.rs
```diff
@@ -537,4 +538,23 @@ pub fn list_system_processes_rust() -> LispObject {
...
+/// rather than the shell.
+///
+/// If CURRENT-GROUP is `lambda', and if the shell owns the terminal,
+/// don't send the signal.
+///
+/// This function calls the functions of `interrupt-process-functions' in
+/// the order of the list, until one of them returns non-`nil'.
+#[lisp_fn(min = "0")]
+pub fn interrupt_process(process: LispObject, current_group: LispObject) -> LispObject {
+    run_hook_with_args_until_success(&mut [process, current_group])
```
shaleh: What happened to `Qinterrupt_process_functions`?

shaleh: You can port this from C to rust. Remove the `DEFSYM` in the C code and add `def_lisp_sym!()` in the Rust code. There are examples in eval.rs.

PR #1215, opened 2019-01-05
Add tuple conversions, cut as_tuple and as_cons_or_error

nickdrozd:
This change complements 6f4ac71065. I wanted to make it possible to
destructure arbitrary conses, as in

```rust
  let (a, (b, c)) = obj.into();
```

but I don't think that's possible. A little destructuring is possible
though, and this replaces `as_tuple` with it. `as_cons_or_error` is
also cut along the way, in favor of `From`/`into`.

A few style changes have also been made wrt destructuring, like

```
  -        self.valcell.as_cons_or_error().cdr()
  +        let (_, d) = self.valcell.into();
  +        d
```

I think this is a little more in keeping with the Rust mindset of
exhaustive matching -- the caller explicitly acknowledges that there
is another value in the cons cell, and that that value is being thrown
away.

### DIFF
--- a/rust_src/src/crypto/mod.rs
+++ b/rust_src/src/crypto/mod.rs
@@ -123,8 +123,10 @@ fn get_coding_system_for_buffer(
         // Check file-coding-system-alist.
         let mut args = [Qwrite_region, start, end, file_name];
         let val = unsafe { Ffind_operation_coding_system(4, args.as_mut_ptr()) };
-        if val.is_cons() && val.as_cons_or_error().cdr().is_not_nil() {
-            return val.as_cons_or_error().cdr();
+        if let Some((_, d)) = val.into() {
+            if d.is_not_nil() {
+                return d;
+            }
         }
     }
     if buffer.buffer_file_coding_system_.is_not_nil() {
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -79,15 +79,15 @@ where
 /// If COND yields nil, and there are no ELSE's, the value is nil.
 /// usage: (if COND THEN ELSE...)
 #[lisp_fn(name = "if", c_name = "if", min = "2", unevalled = "true")]
-pub fn lisp_if(args: LispObject) -> LispObject {
-    let cell = args.as_cons_or_error();
-    let cond = unsafe { eval_sub(cell.car()) };
-    let cnsq = cell.cdr().as_cons_or_error();
+pub fn lisp_if(args: LispCons) -> LispObject {
+    let (cond, consq) = args.into();
+    let (then, else_) = consq.into();
+    let result = unsafe { eval_sub(cond) };
 
-    if cond.is_not_nil() {
-        unsafe { eval_sub(cnsq.car()) }
+    if result.is_not_nil() {
+        unsafe { eval_sub(then) }
     } else {
-        progn(cnsq.cdr())
+        progn(else_)
     }
 }
 
@@ -105,7 +105,7 @@ pub fn cond(args: LispObject) -> LispObject {
     let mut val = Qnil;
 
     for clause in args.iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off) {
-        let (head, tail) = clause.as_cons_or_error().as_tuple();
+        let (head, tail) = clause.into();
         val = unsafe { eval_sub(head) };
         if val != Qnil {
             if tail.is_not_nil() {
@@ -140,7 +140,7 @@ pub extern "C" fn prog_ignore(body: LispObject) {
 /// usage: (prog1 FIRST BODY...)
 #[lisp_fn(min = "1", unevalled = "true")]
 pub fn prog1(args: LispCons) -> LispObject {
-    let (first, body) = args.as_tuple();
+    let (first, body) = args.into();
 
     let val = unsafe { eval_sub(first) };
     progn(body);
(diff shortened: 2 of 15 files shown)


rust_src/src/crypto/mod.rs
```diff
@@ -123,8 +123,11 @@ fn get_coding_system_for_buffer(
         // Check file-coding-system-alist.
         let mut args = [Qwrite_region, start, end, file_name];
         let val = unsafe { Ffind_operation_coding_system(4, args.as_mut_ptr()) };
-        if val.is_cons() && val.as_cons_or_error().cdr().is_not_nil() {
-            return val.as_cons_or_error().cdr();
+        if val.is_cons() {
```
shaleh: I feel this reads better as `if let Some(cons) = val.as_cons()`. What do you think?

rust_src/src/eval.rs
```diff
@@ -79,15 +79,15 @@ where
 /// If COND yields nil, and there are no ELSE's, the value is nil.
 /// usage: (if COND THEN ELSE...)
 #[lisp_fn(name = "if", c_name = "if", min = "2", unevalled = "true")]
-pub fn lisp_if(args: LispObject) -> LispObject {
-    let cell = args.as_cons_or_error();
-    let cond = unsafe { eval_sub(cell.car()) };
-    let cnsq = cell.cdr().as_cons_or_error();
```
shaleh: This reads nicely.

rust_src/src/crypto/mod.rs
nickdrozd:
```rust
        if let Some((_, d)) = val.into() {
            if d.is_not_nil() {
                return d;
            }
        }
```

nickdrozd:
@shaleh Could you give these new changes a look? Nothing crazy in there, but I don't want to merge without approval.

rust_src/src/keymap.rs
```diff
@@ -643,14 +639,14 @@ pub extern "C" fn copy_keymap_1(chartable: LispObject, idx: LispObject, elt: Lis
...
     let mut tail = list!(Qkeymap);
     let copy = tail;
 
-    keymap = keymap.as_cons_or_error().cdr(); // Skip the `keymap' symbol.
+    let (_, mut keymap) = keymap.into(); // Skip the `keymap' symbol.
 
-    while let Some(cons) = keymap.as_cons() {
-        let mut elt = cons.car();
+    while let Some((kma, kmd)) = keymap.into() {
+        let mut elt = kma;
```
shaleh: Why make `kma` when you really wanted `mut elt`?

rust_src/src/lists.rs
```diff
@@ -236,14 +232,14 @@ impl Iterator for CarIter {
 
 impl From<LispObject> for LispCons {
     fn from(o: LispObject) -> Self {
-        o.as_cons_or_error()
+        o.as_cons().unwrap_or_else(|| wrong_type!(Qconsp, o))
```
shaleh: Would this work as:

    match o.into() {
        Some(cell) => cell,
        None => wrong_type!(Qconsp, o),
    }

shaleh: That would remove one of the final `as_cons` calls.

rust_src/src/lists.rs
```diff
@@ -262,6 +258,28 @@ impl<S: Into<LispObject>, T: Into<LispObject>> From<(S, T)> for LispObject {
...
+}
+
+impl From<LispObject> for (LispObject, LispObject) {
+    fn from(o: LispObject) -> Self {
+        LispCons::from(o).into()
+    }
+}
+
+impl From<LispObject> for Option<(LispObject, LispObject)> {
+    fn from(o: LispObject) -> Self {
```
shaleh: This would avoid checking if `o` is a LispCons twice. Is it better?

```
match Option<LispCons>::from(o) {
     Some(cell) => Some(cell.into()),
     None => None,
}
```

rust_src/src/eval.rs
```diff
@@ -513,21 +511,21 @@ pub fn macroexpand(mut form: LispObject, environment: LispObject) -> LispObject
             // SYM is not mentioned in ENVIRONMENT.
             // Look at its function definition.
             def = autoload_do_load(def, sym, Qmacro);
-            match def.as_cons() {
-                Some(cell) => {
-                    let func = cell.car();
+            match def.into() {
+                Some((car, cdr)) => {
```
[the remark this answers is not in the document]
nickdrozd: fixed

rust_src/src/eval.rs
```diff
@@ -724,9 +721,9 @@ pub extern "C" fn FUNCTIONP(object: LispObject) -> bool {
                         }
                     }
 
-                    return match it.rest().as_cons() {
+                    return match it.rest().into() {
                         None => true,
-                        Some(c) => c.car().is_nil(),
```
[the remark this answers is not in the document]
nickdrozd: I think explicit destructuring is cleaner than calling a method on an opaque object. Arguably it is Rustier too.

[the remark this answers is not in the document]
shaleh: Potentially there could be cases where it is too wasteful to call both car and cdr, make a tuple, and unpack it. But otherwise I have no issues here.

rust_src/src/lists.rs
nickdrozd: This change and the next one cause a segfault (not sure which one). I'll investigate it later.

PR #1216, opened 2019-01-06
Port mapc

### LINKED ISSUE #916: Port mapc
``` c

DEFUN ("mapc", Fmapc, Smapc, 2, 2, 0,
       doc: /* Apply FUNCTION to each element of SEQUENCE for side effects only.
Unlike `mapcar', don't accumulate the results.  Return SEQUENCE.
SEQUENCE may be a list, a vector, a bool-vector, or a string.  */)
  (Lisp_Object function, Lisp_Object sequence)
{
  register EMACS_INT leni;

  leni = XFASTINT (Flength (sequence));
  if (CHAR_TABLE_P (sequence))
    wrong_type_argument (Qlistp, sequence);
  mapcar1 (leni, 0, function, sequence);

  return sequence;
}
```

krobelus:
Fixes #916

### DIFF
--- a/rust_src/src/fns.rs
+++ b/rust_src/src/fns.rs
@@ -1,5 +1,7 @@
 //* Random utility Lisp functions.
 
+use std::ptr;
+
 use libc;
 
 use remacs_macros::lisp_fn;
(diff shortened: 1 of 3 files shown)


rust_src/src/fns.rs
```diff
@@ -90,6 +94,16 @@ pub fn quote(args: LispCons) -> LispObject {
     args.car()
 }
 
+#[lisp_fn]
+pub fn mapc(function: LispObject, sequence: LispObject) -> LispObject {
+    let leni = length(sequence) as EmacsInt;
+    if unsafe { CHAR_TABLE_P(sequence) } {
```
```suggestion
    if sequence.is_char_table() {
```

shaleh: We prefer to use methods where possible instead of porting the C macros.

krobelus: thanks, I missed that one!

rust_src/src/fns.rs
```diff
@@ -90,6 +93,16 @@ pub fn quote(args: LispCons) -> LispObject {
     args.car()
 }
 
+#[lisp_fn]
+pub fn mapc(function: LispObject, sequence: LispObject) -> LispObject {
+    let leni = length(sequence) as EmacsInt;
```
krobelus: not sure if `leni` as name is ok
same for the cast but that's a simple solution for now

shaleh: Both are fine.

PR #1218, opened 2019-01-07
Sync from Emacs 2018-03-15

shaleh:
The conflicts were in the docs mostly.
The C code is preparing for bignum support but it is mostly in our ported
code and we already call it `fixnum`.


nickdrozd:
Looks like this was squashed in rather than merged.

shaleh:
Yeah, I wish I had a way to stop me from doing that. GitHub remembers what I did last. It is almost always what I want. Grrrrr.

PR #1220, opened 2019-01-08
Port signal

### LINKED ISSUE #1152: Port signal
```c
DEFUN ("signal", Fsignal, Ssignal, 2, 2, 0,
       doc: /* Signal an error.  Args are ERROR-SYMBOL and associated DATA.
This function does not return.

An error symbol is a symbol with an `error-conditions' property
that is a list of condition names.
A handler for any of those names will get to handle this signal.
The symbol `error' should normally be one of them.

DATA should be a list.  Its elements are printed as part of the error message.
See Info anchor `(elisp)Definition of signal' for some details on how this
error message is constructed.
If the signal is handled, DATA is made available to the handler.
See also the function `condition-case'.  */
       attributes: noreturn)
  (Lisp_Object error_symbol, Lisp_Object data)
{
  signal_or_quit (error_symbol, data, false);
  eassume (false);
}
```

krobelus:
Fixes #1152

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -1,6 +1,6 @@
 //! Generic Lisp eval functions
 
-use std::ptr;
+use std::{ptr, unreachable};
 
 use libc::c_void;
 
@@ -22,7 +22,7 @@ use crate::{
         backtrace_debug_on_exit, build_string, call_debugger, check_cons_list, do_debug_on_call,
         do_one_unbind, eval_sub, find_symbol_value, funcall_lambda, funcall_subr, globals,
         grow_specpdl, internal_catch, list2, maybe_gc, maybe_quit, record_in_backtrace,
-        record_unwind_save_match_data, specbind, COMPILEDP, MODULE_FUNCTIONP,
+        record_unwind_save_match_data, signal_or_quit, specbind, COMPILEDP, MODULE_FUNCTIONP,
     },
     remacs_sys::{pvec_type, EmacsInt, Lisp_Compiled, Set_Internal_Bind},
     remacs_sys::{Fapply, Fdefault_value, Fload, Fpurecopy},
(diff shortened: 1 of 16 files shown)


krobelus:
this is missing some things, i had a stale build locally
working on it..

rust_src/src/eval.rs
```diff
@@ -1225,4 +1225,29 @@ pub fn catch(args: LispCons) -> LispObject {
...
+/// normally be one of them.
+///
+/// DATA should be a list.  Its elements are printed as part of the
+/// error message.  See Info anchor `(elisp)Definition of signal' for
+/// some details on how this error message is constructed.
+/// If the signal is handled, DATA is made available to the handler.
+/// See also the function `condition-case'.
+#[lisp_fn]
+pub fn signal(error_symbol: LispObject, data: LispObject) -> ! {
+    xsignal(error_symbol, data)
```
shaleh: I like the idea of standardizing on one implementation. However, `xsignal` is not it. In Rust we already have a `xisgnal` macro which handles arbitrary arguments. Now the developer needs to switch between a macro and a function of the same name. That is not good practice.

By porting `Fsignal` you also created `signal`. Which means there is no need for an internal function. Let `signal` be the implementation and `xsignal!` be the macro we prefer everyone uses to access it. When we port C code we keep the `xsignal` calls and adjust them to be `xsignal!` macro calls. So leave the `xsignal` function in place for C.

krobelus: Thanks, that makes sense!
I had some issues because the Rust-generated `Fsignal` is not marked noreturn in the type system.
This is not trivial to change because the signatures in `struct Lisp_Subr` have to return a `Lisp_Object` (It doesn't seem possible to make the members noreturn).

Anyway, I have cleaned up the mess now, and I just use `std::unreachable()!` after calling `Fsignal`.
This way there are far fewer changes and new code will use `xsignal!` anyway (so people won't need to add `std::unreachable!()`)

rust_src/src/eval_macros.rs
```diff
@@ -17,13 +17,15 @@ macro_rules! xsignal {
     ($symbol:expr) => {
         #[allow(unused_unsafe)]
         unsafe {
-            crate::remacs_sys::Fsignal($symbol, crate::remacs_sys::Qnil);
+            crate::eval::Fsignal($symbol, crate::remacs_sys::Qnil);
```
shaleh: Call `signal` not `Fsignal`.

rust_src/src/eval.rs
```diff
@@ -1274,4 +1274,25 @@ pub fn catch(args: LispCons) -> LispObject {
...
+/// normally be one of them.
+///
+/// DATA should be a list.  Its elements are printed as part of the
+/// error message.  See Info anchor `(elisp)Definition of signal' for
+/// some details on how this error message is constructed.
+/// If the signal is handled, DATA is made available to the handler.
+/// See also the function `condition-case'.
+#[lisp_fn]
+pub fn signal(error_symbol: LispObject, data: LispObject) -> ! {
+    #[cfg(test)]
```
krobelus: Moved this here from `functions.rs`

rust_src/src/eval_macros.rs
krobelus: Done. Also I moved the panic if it's called in `cfg(test)` to the signal function.
make check still errors with "undefined reference to `signal_or_quit`", I'm trying to find out why

rust_src/src/eval.rs
shaleh: Try:

    #[cfg(test)]
    {
        panic!("Fsignal called during tests.");
    }
    #[cfg(not(test))]
    {
         unsafe { signal_or_quit(error_symbol, data, false) };
         unreachable!();
    }

krobelus:
Sadly, there are still issues:
`Fsignal` is called several times from C code, resulting in errors like this

`coding.c:10566:1: error: control may reach end of non-void function [-Werror,-Wreturn-type]`

Because C doesn't know that `Fsignal` doesn't return. I think we shouldn't ignore this, but either use `xsignal` for those cases in C source, or we manage to communicate the noreturn attribute across the language boundary. What do you think?

rust_src/src/eval.rs
krobelus: yeah, much better

krobelus:
Should be about right now with the second commit. There are test failures but it seems they have been introduced before.

brotzeit:
Good ?

PR #1221, opened 2019-01-08
Rewrite as_type methods in terms of From

nickdrozd:
Currently a lot of type conversions are written as methods on
LispObject, and the From trait implementations are written in terms of
those. It should be the other way around, so move the as_type logic
into the From implementations and make the as_type functions just do
casts. This will make it easier to remove them entirely in the future,
allowing the type system to do more work.

A few as_type functions, like as_window_configuration_or_error, were
only used in a few places, so those were cut entirely.

### DIFF
--- a/rust_src/src/chartable.rs
+++ b/rust_src/src/chartable.rs
@@ -27,27 +27,23 @@ impl LispObject {
     }
 
     pub fn as_char_table(self) -> Option<LispCharTableRef> {
-        self.as_vectorlike().and_then(|v| v.as_char_table())
-    }
-
-    pub fn as_char_table_or_error(self) -> LispCharTableRef {
-        if let Some(chartable) = self.as_char_table() {
-            chartable
-        } else {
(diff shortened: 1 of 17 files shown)


rust_src/src/chartable.rs
```diff
@@ -27,27 +27,23 @@ impl LispObject {
...
-            wrong_type!(Qchar_table_p, self)
-        }
+        self.into()
     }
 }
 
 impl From<LispObject> for LispCharTableRef {
     fn from(o: LispObject) -> Self {
-        o.as_char_table_or_error()
+        if let Some(chartable) = o.as_char_table() {
```
shaleh: Should this be a `match`?

    match o.as_char_table() {
        Some(chartable) => chartable,
        None => wrong_type!(Qchar_table_p, o),
    }

nickdrozd: Probably, but this was basically just a cut/paste job. That change can be made in another pass.

PR #1231, opened 2019-01-12
port error-message-string

### LINKED ISSUE #1079: Port error-message-string
```c
DEFUN ("error-message-string", Ferror_message_string, Serror_message_string,
       1, 1, 0,
       doc: /* Convert an error value (ERROR-SYMBOL . DATA) to an error message.
See Info anchor `(elisp)Definition of signal' for some details on how this
error message is constructed.  */)
  (Lisp_Object obj)
{
  struct buffer *old = current_buffer;
  Lisp_Object value;

  /* If OBJ is (error STRING), just return STRING.
     That is not only faster, it also avoids the need to allocate
     space here when the error is due to memory full.  */
  if (CONSP (obj) && EQ (XCAR (obj), Qerror)
      && CONSP (XCDR (obj))
      && STRINGP (XCAR (XCDR (obj)))
      && NILP (XCDR (XCDR (obj))))
    return XCAR (XCDR (obj));

  print_error_message (obj, Vprin1_to_string_buffer, 0, Qnil);

  set_buffer_internal (XBUFFER (Vprin1_to_string_buffer));
  value = Fbuffer_string ();

  Ferase_buffer ();
  set_buffer_internal (old);

  return value;
}
```

mukuact:
I dont find print.rs, so make it.  ~~But, it cannot compile for some missing boilerplate codes ...~~

It will close #1079

### DIFF
--- a/rust_src/src/print.rs
+++ b/rust_src/src/print.rs
@@ -1,0 +1,46 @@
+use remacs_macros::lisp_fn;
+
+use crate::{
+    lisp::LispObject,
+    lists::{car, cdr, car_safe, cdr_safe},
+    threads::ThreadState,
+    remacs_sys::{
+        print_error_message, set_buffer_internal, Fbuffer_string, Ferase_buffer, set_buffer_internal,
+    },
+}
+
+/* A buffer which is used to hold output being built by prin1-to-string.  */
+Lisp_Object Vprin1_to_string_buffer;
+
+/// Convert an error value (ERROR-SYMBOL . DATA) to an error message.
+/// See Info anchor `(elisp)Definition of signal' for some details on how this
+/// error message is constructed.
+#[lisp_fn]
+pub fn error_message_string(obj: LispObject) -> LispObject {
+    let old = ThreadState::current_buffer_unchecked();
+    let value: LispObject;
+
+    /* If OBJ is (error STRING), just return STRING.
+    That is not only faster, it also avoids the need to allocate
+    space here when the error is due to memory full.  */
+    match obj.into() {
+        Some((car, cdr)) if car.eq(Qerror) => {
+            if cdr.car_safe().is_string() && cdr.cdr_safe().is_nil() {
+                return cdr.car();
+            }
+        }
+        _ => {}
+    }
+    unsafe {
+        print_error_message(obj, Vprin1_to_string_buffer, 0, Qnil);
+
+        set_buffer_internal(Vprin1_to_string_buffer.as_buffer());
+        value = Fbuffer_string();
+
+        Ferase_buffer();
+        set_buffer_internal(old);
+    }
+    return value;
+}
+
+include!(concat!(env!("OUT_DIR"), "/print_exports.rs"));
--- a/rust_src/src/print.rs
+++ b/rust_src/src/print.rs
@@ -3,7 +3,7 @@
 use remacs_macros::lisp_fn;
 
 use crate::{
-    buffers::{as_buffer_or_error, LispBufferRef},
+    buffers::LispBufferRef,
     lisp::LispObject,
     lists::LispCons,
     remacs_sys::Vprin1_to_string_buffer,
--- a/rust_src/src/print.rs
+++ b/rust_src/src/print.rs
@@ -17,7 +17,7 @@ use crate::{
 /// error message is constructed.
 #[lisp_fn]
 pub fn error_message_string(obj: LispCons) -> LispObject {
-    let old = ThreadState::current_buffer_unchecked();
+    let mut old = ThreadState::current_buffer_unchecked();
     let value: LispObject;
 
     // If OBJ is (error STRING), just return STRING.
--- a/rust_src/src/print.rs
+++ b/rust_src/src/print.rs
@@ -22,7 +22,7 @@ pub fn error_message_string(obj: LispCons) -> LispObject {
 
     // If OBJ is (error STRING), just return STRING.
     // That is not only faster, it also avoids the need to allocate
-    // space here when the error is due to memory full.  
+    // space here when the error is due to memory full.
     match obj.into() {
         (error, data) if error.eq(Qerror) => {
             if let Some((string, nil)) = data.into() {
--- a/rust_src/src/print.rs
+++ b/rust_src/src/print.rs
@@ -3,7 +3,7 @@
 use remacs_macros::lisp_fn;
 
 use crate::{
-    lisp::defsubr,
+    buffers::{as_buffer_or_error, LispBufferRef},
     lisp::LispObject,
     lists::LispCons,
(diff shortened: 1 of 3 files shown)


rust_src/src/print.rs
```diff
@@ -0,0 +1,46 @@
...
+    lisp::LispObject,
+    lists::{car, cdr, car_safe, cdr_safe},
+    threads::ThreadState,
+    remacs_sys::{
+        print_error_message, set_buffer_internal, Fbuffer_string, Ferase_buffer, set_buffer_internal,
+    },
+}
+
+/* A buffer which is used to hold output being built by prin1-to-string.  */
+Lisp_Object Vprin1_to_string_buffer;
```
db48x: No need to declare this variable; the C code already does, and we should reuse it. You can import it from remacs_sys.

rust_src/src/print.rs
```diff
@@ -0,0 +1,46 @@
...
+    let old = ThreadState::current_buffer_unchecked();
+    let value: LispObject;
+
+    /* If OBJ is (error STRING), just return STRING.
+    That is not only faster, it also avoids the need to allocate
+    space here when the error is due to memory full.  */
+    match obj.into() {
+        Some((car, cdr)) if car.eq(Qerror) => {
+            if cdr.car_safe().is_string() && cdr.cdr_safe().is_nil() {
+                return cdr.car();
```
db48x: This whole match is pretty nice, but I think it can be improved slightly by better naming. Instead of calling the variables `car` and `cdr`, why not `error_symbol` and `error_data`?

rust_src/src/print.rs
```diff
@@ -0,0 +1,46 @@
+use remacs_macros::lisp_fn;
```
shaleh: Minor style nit.

We start each file with a comment details what it contains. You can re-use the C code's comment here:

`//! Lisp object printing and output streams.`

rust_src/src/print.rs
shaleh: @nickdrozd your tuple addition really shines here.

shaleh:
Do resolve the rustfmt failure be sure to install `rustup component add rustfmt-preview`. Then read the docs for `rust-mode` and set the variable so rustfmt is called whenever you save a buffer.

rust_src/src/print.rs
nickdrozd: It might be possible to destructure this even more, like

```rust
if let Some((error, (string, rest)) = obj.into() {/* ... */}
```

rust_src/src/print.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/* A buffer which is used to hold output being built by prin1-to-string.  */
+Lisp_Object Vprin1_to_string_buffer;
+
+/// Convert an error value (ERROR-SYMBOL . DATA) to an error message.
+/// See Info anchor `(elisp)Definition of signal' for some details on how this
+/// error message is constructed.
+#[lisp_fn]
+pub fn error_message_string(obj: LispObject) -> LispObject {
```
nickdrozd:
```suggestion
pub fn error_message_string(obj: LispCons) -> LispObject {
```

nickdrozd: `LispCons` would need to be imported.

rust_src/src/print.rs
```diff
@@ -0,0 +1,49 @@
...
+                if string.is_string() && nil.is_nil() {
+                    return string;
+                }
+            }
+        }
+        _ => {}
+    }
+
+    unsafe {
+        print_error_message(obj.into(), Vprin1_to_string_buffer, None, Qnil);
```
mukuact: How to give a nullptr for an argument in rust?

rust_src/src/print.rs
mukuact: Thanks!

rust_src/src/print.rs
mukuact: It's nice to naming like that!

rust_src/src/print.rs
mukuact: And I’ｍ wondering about an ownership issue.
at [l.26](https://github.com/Wilfred/remacs/pull/1231/files#diff-78ac0a6803c74006beb627ba7442cdcfR26), 
I use `obj.into()` , which takes the ownership of `obj` and bind it to match arm's expressions.
, then `if let Some(string, nil) = value.into()` also takes that ownership.
Here, if it matches to the `_` arm, or don't get into if-let clause, where `obj`'s ownership go?
and,
Why this reuse obj makes the code complies without error?

agraven: To pass a null pointer you want to use `std::ptr::null()` or `std::ptr::null_mut()` depending on if you want it to be const or mut.

agraven: And you should note, as lisp objects are `Copy`, since they're more or less a fancier pointer (that's not the full story but good enough for our purposes), so you generally don't have to worry about ownership issues with them.

rust_src/src/print.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// Convert an error value (ERROR-SYMBOL . DATA) to an error message.
+/// See Info anchor `(elisp)Definition of signal' for some details on how this
+/// error message is constructed.
+#[lisp_fn]
+pub fn error_message_string(obj: LispCons) -> LispObject {
+    let old = ThreadState::current_buffer_unchecked();
+    let value: LispObject;
+
+    /* If OBJ is (error STRING), just return STRING.
```
shaleh: remacs style uses `//` instead of `/* */` style comments.

src/print.c
```diff
@@ -855,37 +855,6 @@ safe_debug_print (Lisp_Object arg)
     }
 }
```
shaleh: At the end of the file you also need to remove the `defsubr` line which refers to this function.

rust_src/src/print.rs
```diff
@@ -0,0 +1,49 @@
...
+                }
+            }
+        }
+        _ => {}
+    }
+
+    unsafe {
+        print_error_message(obj.into(), Vprin1_to_string_buffer, None, Qnil);
+
+        set_buffer_internal(Vprin1_to_string_buffer.as_buffer_or_error().as_mut());
```
shaleh: Is Rust able to compile: `Vprin1_to_string_buffer.into().as_mut()`? I suspect not. `LispBufferRef::from(Vprin1_to_string_buffer).as_mut()` will work though.

We are moving away from the `as_foo_or_error` methods. The `From` method will error if it cannot convert. Safe conversions are done using `Option<Foo>::from()`.

src/print.c
mukuact: ok

brotzeit:
Good ?

shaleh:
@db48x your opinion?

PR #1232, opened 2019-01-12
Port yes-or-no-p

carlosmn:
The first commit changes some `LispObjects` to `bool` which I noticed as I was porting many `Qnil`s. Then we port `yes-or-no-p`. The C version checks the length of the input but it just seems like a microoptimisation we probably don't care about.

We compare the strings as byte slices and it seems like it'd be a good follow-up to allow comparison of `LispStringRef` and strings or byte slice via `PartialEq`.

Annoyingly the parameter is allowed to be `nil` upstream which is treated as an empty string but we don't do that here.


rust_src/src/fns.rs
```diff
@@ -375,4 +383,61 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+pub fn yes_or_no_p(prompt: LispStringRef) -> bool {
+    let use_popup = unsafe {
+        (globals.last_nonmenu_event.is_nil() || globals.last_nonmenu_event.is_cons())
+            && globals.use_dialog_box
+            && !globals.last_input_event.is_nil()
+    };
+
+    if use_popup {
+        unsafe { redisplay_preserve_echo_area(4) };
+        let pane = list![LispObject::cons("Yes", Qt), LispObject::cons("No", Qnil)];
```
shaleh: @nickdrozd added support for a 2-tuple of LispObjects to be converted to a Cons cell.

```suggestion
        let pane = list![("Yes", Qt).into(), ("No", Qnil).into()];
```

rust_src/src/fns.rs
```diff
@@ -375,4 +383,61 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+        let menu = LispObject::cons(prompt, pane);
```
```suggestion
        let menu: LispObject = (prompt, pane).into();
```

rust_src/src/fns.rs
carlosmn: It turns out this won't compile,

    error[E0283]: type annotations required: cannot resolve `_: std::convert::Into<lisp::LispObject>`

and I'm not finding a way to do this in one expression. It does work if I extract each into its own

    let a: LispObject = ("Yes", Qt).into();

but that seems less readable than before.

carlosmn:
the tuple-into-object only worked with one level of indirection so `list![("Yes", Qt).into(), ("No", Qnil).into()]` would fail to compile with not knowing what to convert.

However we also taught the compiler that you can convert a `Vec<LispObject>` into a `LispObject` by building a list so I tried that and that was enough type hints for the compiler to be happy. I don't know if we could add annotations in the `list!` variant but this should end up compiling to about the same (probably?)

rust_src/src/fns.rs
```diff
@@ -375,4 +383,61 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+            Qnil,
+            Qnil,
+            false,
+            Qyes_or_no_p_history,
+            Qnil,
+            false,
+        ))
+        .into();
+
+        if ans.as_slice() == "yes".as_bytes() {
```
shaleh: I added `slice_patterns` to the list of features recently. Can this be rewritten as a match?

rust_src/src/fns.rs
```diff
@@ -375,4 +383,61 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+
+    if use_popup {
+        unsafe { redisplay_preserve_echo_area(4) };
+        let pane: LispObject = vec![("Yes", Qt).into(), ("No", Qnil).into()].into();
+        let menu: LispObject = (prompt, pane).into();
+        return unsafe { Fx_popup_dialog(Qt, menu, Qnil) }.into();
+    }
+
+    let yes_or_no: LispObject = "(yes or no) ".into();
+    let prompt = concat(&mut vec![prompt.into(), yes_or_no]).into();
```
nickdrozd: I would think there's a cleaner way to do this one, but I'm not sure what it is.

rust_src/src/fns.rs
```diff
@@ -375,4 +383,61 @@ pub fn load_average(use_floats: bool) -> Vec<LispNumber> {
...
+pub fn yes_or_no_p(prompt: LispStringRef) -> bool {
+    let use_popup = unsafe {
+        (globals.last_nonmenu_event.is_nil() || globals.last_nonmenu_event.is_cons())
+            && globals.use_dialog_box
+            && globals.last_input_event.is_not_nil()
+    };
+
+    if use_popup {
+        unsafe { redisplay_preserve_echo_area(4) };
+        let pane: LispObject = vec![("Yes", Qt).into(), ("No", Qnil).into()].into();
```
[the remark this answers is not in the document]
shaleh: Am I missing something or is `pane` unused?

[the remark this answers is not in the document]
shaleh: Yes, this works just fine.

rust_src/src/fns.rs
shaleh:
```
macro_rules! lisp_concat {
    ($($tt:tt),+) => {
        crate::fns::concat(&mut vec![$(($tt).into(), )*]);
    };
}
```

shaleh:
@carlosmn @nickdrozd I updated this PR by first merging it and then second applying our review suggestions.

rust_src/src/fns.rs
shaleh: I think this is a good idea but outside the scope of this PR.

shaleh:
Thanks again @carlosmn.

Issue #1209, opened 2019-01-04
Implement LispSymbolOrString type


nickdrozd:
There are a few functions that accept either strings or symbols, like `string-lessp`, `string-equal`, and `intern-soft`. Currently these all take `LispObject` as argument and then cast them with `LispObject::symbol_or_string_as_string`. Example:

```rust
#[lisp_fn]
pub fn string_lessp(string1: LispObject, string2: LispObject) -> bool {
    let s1 = LispObject::symbol_or_string_as_string(string1);
    let s2 = LispObject::symbol_or_string_as_string(string2);

    s1.as_slice() < s2.as_slice()
}
```

This would be better handled with a dedicated `LispSymbolOrString` type, giving something like:

```rust
#[lisp_fn]
pub fn string_lessp(string1: LispSymbolOrString, string2: LispSymbolOrString) -> bool {
    string1.as_slice() < string2.as_slice()
}
```

Implement this type, along with whatever functions are required (`From<LispSymbolOrString> for LispObject`, etc).

shaleh:
We should also define `From<LispSymbolOrString> for LispStringRef` and also for `LispSymbolRef`. This will simplify the call sites. When a dev calls `string_lessp` they can call it like `string_lessp(string1.into(), string2.into())`.

nickdrozd:
`process.rs` also has some places that could use a union type of some kind.

```
/// Return the process named NAME, or nil if there is none.
#[lisp_fn(name = "get-process", c_name = "get_process")]
pub fn get_process_lisp(name: LispObject) -> LispObject {
    if name.is_process() {
        name
    } else {
        name.as_string_or_error();
        cdr(assoc(name, unsafe { Vprocess_alist }, Qnil))
    }
}
```

PR #1237, opened 2019-01-13
impl Display for LispObject, cut as_string_or_error

nickdrozd:


rust_src/src/editfns.rs
```diff
@@ -420,7 +420,7 @@ pub fn propertize(args: &[LispObject]) -> LispObject {
 
     // the unwrap call is safe, the number of args has already been checked
     let first = it.next().unwrap();
-    let orig_string = first.as_string_or_error();
+    let orig_string = LispStringRef::from(*first);
```
shaleh: Why use `From` here and `Into` in the previous change?

rust_src/src/eval.rs
```diff
@@ -232,7 +232,7 @@ pub fn function(args: LispCons) -> LispObject {
                         // dynamically.
 
                         let docstring = unsafe { eval_sub(car(tail)) };
-                        docstring.as_string_or_error();
```
shaleh: This only verifying that `docstring` is a string right? The case hides the purpose.

rust_src/src/eval.rs
```diff
@@ -837,8 +837,8 @@ pub fn autoload_do_load(
         if equal(fun, fundef) {
             error!(
                 "Autoloading file {} failed to define function {}",
-                car(car(unsafe { globals.Vload_history })).as_string_or_error(),
-                sym.symbol_name().as_string_or_error()
+                car(car(unsafe { globals.Vload_history })),
```
shaleh: This is nice.

rust_src/src/process.rs
```diff
@@ -134,7 +134,7 @@ pub fn get_process_lisp(name: LispObject) -> LispObject {
     if name.is_process() {
         name
     } else {
-        name.as_string_or_error();
```
shaleh: This is another spot where the intent is to verify the type.

rust_src/src/editfns.rs
nickdrozd: I tried it a few different ways and this is the only one that worked. Something about the dereference I guess.

rust_src/src/eval.rs
nickdrozd: Changed this one to be clearer.

rust_src/src/process.rs
nickdrozd: Yeah, that's an indication that the function signature needs to be tightened up. I added a comment about this in https://github.com/Wilfred/remacs/issues/1209#issuecomment-452903395

shaleh: The C code has things like `CHECK_FIXNUM`. Perhaps something like 
that is helpful here? `CHECK_STRING`.

I like the cast operators for simplifying the code. However, we will need to be sure they are efficient at some point.

nickdrozd: For now I'm trusting in the Rust propaganda that says the compiler is able to do lots of fancy type deductions while still producing fast code. We'll have to profile to know for sure. Do we have any profiling machinery set up?

nickdrozd: I think those `CHECK_TYPE` operators are only needed because the C code can't get any finer-grained type-wise than `Lisp_Object`. In this particular case the type should be `ProcessOrString` or whatever.

shaleh: I mostly trust the Rust compiler. It is the marshalling and demarshalling in and out of C that concerns me. It may not be a valid concern, but I know remacs feels slower in places.

shaleh: As for profiling and the list since we are C hosted and Rust is only a library I am not sure where to start. I have not had the time to research it.

Issue #1235, opened 2019-01-13
Port compare-strings


brotzeit:
```c
DEFUN ("compare-strings", Fcompare_strings, Scompare_strings, 6, 7, 0,
       doc: /* Compare the contents of two strings, converting to multibyte if needed.
The arguments START1, END1, START2, and END2, if non-nil, are
positions specifying which parts of STR1 or STR2 to compare.  In
string STR1, compare the part between START1 (inclusive) and END1
\(exclusive).  If START1 is nil, it defaults to 0, the beginning of
the string; if END1 is nil, it defaults to the length of the string.
Likewise, in string STR2, compare the part between START2 and END2.
Like in `substring', negative values are counted from the end.

The strings are compared by the numeric values of their characters.
For instance, STR1 is "less than" STR2 if its first differing
character has a smaller numeric value.  If IGNORE-CASE is non-nil,
characters are converted to upper-case before comparing them.  Unibyte
strings are converted to multibyte for comparison.

The value is t if the strings (or specified portions) match.
If string STR1 is less, the value is a negative number N;
  - 1 - N is the number of characters that match at the beginning.
If string STR1 is greater, the value is a positive number N;
  N - 1 is the number of characters that match at the beginning.  */)
  (Lisp_Object str1, Lisp_Object start1, Lisp_Object end1, Lisp_Object str2,
   Lisp_Object start2, Lisp_Object end2, Lisp_Object ignore_case)
{
  ptrdiff_t from1, to1, from2, to2, i1, i1_byte, i2, i2_byte;

  CHECK_STRING (str1);
  CHECK_STRING (str2);

  /* For backward compatibility, silently bring too-large positive end
     values into range.  */
  if (INTEGERP (end1) && SCHARS (str1) < XINT (end1))
    end1 = make_number (SCHARS (str1));
  if (INTEGERP (end2) && SCHARS (str2) < XINT (end2))
    end2 = make_number (SCHARS (str2));

  validate_subarray (str1, start1, end1, SCHARS (str1), &from1, &to1);
  validate_subarray (str2, start2, end2, SCHARS (str2), &from2, &to2);

  i1 = from1;
  i2 = from2;

  i1_byte = string_char_to_byte (str1, i1);
  i2_byte = string_char_to_byte (str2, i2);

  while (i1 < to1 && i2 < to2)
    {
      /* When we find a mismatch, we must compare the
	 characters, not just the bytes.  */
      int c1, c2;

[...]

PR #1239, opened 2019-01-14
[WIP] Port compare-strings

agraven:
Solves #1235.

While I think I have the right general approach here, I'm unsure whether this implementation is correct, as `LispStringRefIterator` states it replaces `FETCH_STRING_CHAR_ADVANCE`, and the c implementation uses `FETCH_STRING_CHAR_AS_MULTIBYTE_ADVANCE`.

There are also errors that I have not been able to debug. Defining `validate_debug` causes bootstrapping to fail with the error
```
Using load-path (/home/amanda/Get/remacs/lisp /home/amanda/Get/remacs/lisp/emacs-lisp /home/amanda/Get/remacs/lisp/progmodes /home/amanda/Get/remacs/lisp/language /home/amanda/Get/remacs/lisp/international /home/amanda/Get/remacs/lisp/textmodes /home/amanda/Get/remacs/lisp/vc)
Loading emacs-lisp/byte-run...
Loading emacs-lisp/backquote...
Loading subr (source)...
End of file during parsing: /home/amanda/Get/remacs/lisp/subr.el
make[1]: *** [Makefile:807: bootstrap-emacs] Error 255
```
and building without including the `validate_debug` implementation causes [a failure to build the elisp files](https://gist.github.com/agraven/33a42c2b116e2bab92f4b9d0e80526a7)

### DIFF
--- a/rust_src/src/fns.rs
+++ b/rust_src/src/fns.rs
@@ -1,6 +1,7 @@
 //* Random utility Lisp functions.
 
-use std::ptr;
+use std::convert::TryFrom;
+use std::{cmp, ptr};
 
 use libc;
 
@@ -16,6 +17,7 @@ use crate::{
     numbers::LispNumber,
     obarray::loadhist_attach,
     objects::equal,
+    remacs_sys::args_out_of_range_3,
     remacs_sys::Fload,
     remacs_sys::Vautoload_queue,
     remacs_sys::{concat as lisp_concat, globals, record_unwind_protect},
@@ -25,6 +27,44 @@ use crate::{
     threads::c_specpdl_index,
     vectors::length,
 };
+
+#[no_mangle]
+pub unsafe extern "C" fn validate_subarray(
+    array: LispObject,
+    from: LispObject,
+    to: LispObject,
+    size: isize,
+    ifrom: *mut isize,
+    ito: *mut isize,
+) {
+    let (f, t) = validate_subarray_rust(array, from.into(), to.into(), size);
+
+    *ifrom = f;
+    *ito = t;
+}
+
+pub fn validate_subarray_rust(
+    array: LispObject,
+    from: Option<EmacsInt>,
+    to: Option<EmacsInt>,
+    size: isize,
+) -> (isize, isize) {
+    let f = match from {
+        None => 0,
+        Some(f) if f < 0 => f as isize + size,
+        Some(f) => f as isize,
+    };
+    let t = match to {
+        None => 0,
+        Some(t) if t < 0 => t as isize + size,
+        Some(t) => t as isize,
+    };
+
+    if !(0 <= f && f <= t && t <= size) {
+        unsafe { args_out_of_range_3(array, from.into(), to.into()) };
+    }
+    (f, t)
+}
 
 /// Return t if FEATURE is present in this Emacs.
 ///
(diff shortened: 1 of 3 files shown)


rust_src/src/fns.rs
```diff
@@ -26,6 +28,44 @@ use crate::{
...
+    to: Option<EmacsInt>,
+    size: isize,
+) -> (isize, isize) {
+    let f = match from {
+        None => 0,
+        Some(f) if f < 0 => f as isize + size,
+        Some(f) => f as isize,
+    };
+    let t = match to {
+        None => 0,
```
shaleh: This is your bug. `t` should default to `size` not zero.

rust_src/src/fns.rs
```diff
@@ -26,6 +28,44 @@ use crate::{
...
+        Some(t) if t < 0 => t as isize + size,
+        Some(t) => t as isize,
+    };
+
+    if !(0 <= f && f <= t && t <= size) {
+        unsafe { args_out_of_range_3(array, from.into(), to.into()) };
```
shaleh: There is no need to call the C directly. We have a macro `args_out_of_range!`.

shaleh: However, `into()` doesn't work. You need `LispObject::from()`.

agraven:
I added an iterator on `LispStringRef` that replicates `FETCH_STRING_CHAR_AS_MULTIBYTE_ADVANCE` using `make_char_multibyte`. Should make porting other invocations of the macro easier.

The error I'm getting now is:
```
Loading language/english...
Language environment not defined: "English"
```
I'll try to debug it further.

agraven:
That was a subtle one, the error turned out to be a missing + 1, which meant the function thought it hadn't compared the full range it was supposed to.

agraven:
Pretty close to having properly fixed this now. It now gives the right results for almost all strings, but there's an args out of range error with the assamese string "গুপ্তশব্দ" that I'm still determining the exact source of.

shaleh:
Could we be hitting a subtle issue in moving from the C cast to the Rust cast?

`# define lisp_h_XINT(a) (XLI (a) >> INTTYPEBITS)`

That is what `XINT` is defined as. Notice the bit shifting.

```
 /* Number of bits in a Lisp fixnum tag.  */
    INTTYPEBITS = GCTYPEBITS - 1,
```

`GCTYPEBITS` is 3.

shaleh:
@agraven are you still stuck here?

agraven:
Haven't given it a look yet. Will do soon

agraven:
Ok, I have an idea of what's going wrong here. The issue was `make_char_multibyte` was being called on unconditionally in the `char_indices_multibyte` iterator, when it should only be called on string that aren't multibyte.

agraven:
Hm, it's not quite there yet, there's still some comparisons of chinese characters that throw an args out of range error, but we're getting closer.

shaleh:
@agraven can this be resurrected?

agraven:
I think so, the issues with this are definitely minor, I just need to do enough debugging to find out where the issue stems from.

agraven:
The problem was that did case insensitive comparisons by transforming to lowercase instead of uppercase, which doesn't work because of technicalities with unicode.

PR #1240, opened 2019-01-15
Update to new 2019-01-19 nightly.

shaleh:
Update external crate versions where possible.
Uses official bindgen 0.46 instead of git repo hash.
Use full path in call to `defsubr` in the macro so it does not need to be imported everywhere.
Fix clippy and rustfmt complaints.
Remove some dead code.


rust_src/src/crypto/mod.rs
```diff
@@ -63,205 +52,6 @@ fn hash_alg(algorithm: LispSymbolRef) -> HashAlg {
     }
 }
```
shaleh: This was all waiting for `extract_data_from_object` to be ported. No sense keeping this zombie alive.

shaleh:
This is a noisy patch with little actual change.

daichimukai:
Well, nightly-2019-01-21 and nightly-2019-01-22 lack clippy component at least for x86_64 linux system?

[Reference](https://mexus.github.io/rustup-components-history/x86_64-unknown-linux-gnu.html)

brotzeit:
@daichimukai Interesting.

agraven:
@brozeit it's pretty inconsistent which nightly versions have what tools. There's a page listing the latest nightly each component supports, but I can't remember exactly where to find it

shaleh:
brotzeit's rebase did not appear to be happy. I ran it again from my original branch. Let's see how this one goes.

shaleh:
Stuck on cerrors timing out....

shaleh:
Ok, so something is up with the `cerror` checks apparently. Looks like it is time to revisit.

I have what I hope is a quick solution in mind. We can add a new Makefile target in `src/` which only does the C build but none of the Rust. The core build will depend on this target.

shaleh:
Nevermind that about the Makefile. I misinterpreted the 'cerrors' name. The actual intent is to verify that `lisp_fn` macros show proper details. Now to see if we can simplify this.

shaleh:
This will be rebased on #1280 when it is approved.

PR #1233, opened 2019-01-13
Port call-process-region

agraven:
This solves #1229.

Some of the c functions used were ported as well. Additionally, since I ported `record_unwind_protect_int` I ported all the other `record_unwind_protect` function while I was at it since they are almost identical.

PR #1248, opened 2019-01-16
Port record_unwind_protect functions

agraven:
Port the `record_unwind_protect` set of functions.

These changes are split from #1233.

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -1,8 +1,8 @@
 //! Generic Lisp eval functions
 
-use libc::c_void;
-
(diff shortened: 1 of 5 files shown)


rust_src/src/eval.rs
```diff
@@ -1,5 +1,7 @@
 //! Generic Lisp eval functions
 
+use libc::c_void;
```
shaleh: std is first, then other crates.

agraven: Ah shoot, should have double checked. Fixed.

PR #1249, opened 2019-01-16
Update to latest base64 crate (0.10.0)

### LINKED ISSUE #1241: Update to latest base64 crate
Today we are able to use a `Config` work around to set the line end characters and expected line length. However the current version of base64 no longer supports this.

Can we just use base64 as is?
Do we try to send a PR?
Do we need to investigate adding a vendor/ area so we can keep old versions of crates we use around?

daichimukai:
From version 0.10.0, base64 crate does not provide the function to
wrap encoded string in the expected line length and delete whitespaces
from the string to be decoded. The former has pulled into the
line-wrap crate and the latter can be achived by a short code.

Closes #1241.

### DIFF
--- a/rust_src/src/base64.rs
+++ b/rust_src/src/base64.rs
@@ -2,6 +2,8 @@
 use std::{cmp::max, slice};
 
 use libc::{c_char, c_uchar};
+use line_wrap::LineEnding;
+
 use remacs_macros::lisp_fn;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


rust_src/src/base64.rs
```diff
@@ -2,6 +2,7 @@
 use std::{cmp::max, slice};
 
 use libc::{c_char, c_uchar};
+use line_wrap::LineEnding;
```
shaleh: Minor style nit. Please leave the blank line so `use remac_macros` is on a line by itself.

rust_src/src/base64.rs
```diff
@@ -20,19 +21,9 @@ use crate::{
...
-        base64_crate::Config::new(
-            base64_crate::CharacterSet::Standard,
-            true, // pad
-            true, // strip whitespace
-            base64_crate::LineWrap::Wrap(76, base64_crate::LineEnding::LF),
-        )
-    } else {
-        base64_crate::STANDARD
-    };
+    let config = base64_crate::STANDARD;
```
shaleh: `pad` is still true for `STANDARD`?

rust_src/src/base64.rs
daichimukai: Roger that :+1:

rust_src/src/base64.rs
daichimukai: Yes for this version `base64 (v0.10.0)` (No padding version is provided as `STANDARD_NO_PAD`).

PR #1256, opened 2019-01-18
Cut is_nil, impl Not for LispObject

nickdrozd:


shaleh:
I want to like this. But `is_not_nil` stops me. This leads to confusion where the positive is written one way and the negative completely differently.

I'd rather write == Qnil and == Qt than have this as long the positive is more complicated.

nickdrozd:
I have another commit that more or less converts every instance of `is_not_nil` to `into`. I wish there was a way to have it automatically convert in boolean contexts. The asymmetry is very annoying.

nickdrozd:
We can leave it up for a while and see what we can come up with.

shaleh:
I supposed we could be like the Javascript community and use `!!foo`. Ick.

agraven:
I mean, it's not *that* much worse that rusts `&*` :p

nickdrozd:
I posted this question to Rust Reddit: https://www.reddit.com/r/rust/comments/ah7b2d/automatic_boolean_casting/

brotzeit:
I think the type indication is worth the verbosity. Our goal is to eventually remove nil from rust as far as possible right ?

nickdrozd:
PR is out of date, so closing in favor of https://github.com/remacs/remacs/issues/1450

PR #1257, opened 2019-01-18
Cut more type_or_error methods

nickdrozd:


rust_src/src/buffers.rs
```diff
@@ -590,7 +582,7 @@ impl From<LispObject> for LispBufferOrName {
         if v.is_string() {
             LispBufferOrName::Name(v)
         } else {
-            v.as_buffer_or_error();
+            LispBufferRef::from(v);
```
shaleh: This is another instance of spurious casting. `check_buffer` would be cheaper.

rust_src/src/editfns.rs
```diff
@@ -1113,7 +1113,7 @@ pub fn buffer_substring_no_properties(mut beg: LispObject, mut end: LispObject)
 // offload some work from GC.
 #[no_mangle]
 pub extern "C" fn save_excursion_save() -> LispObject {
-    let window = selected_window().as_window_or_error();
+    let window: LispWindowRef = selected_window().into();
```
shaleh: Are we approaching the time for `selected_window` to return a `LispWindowRef`?

rust_src/src/buffers.rs
nickdrozd: Maybe it would be better to make that an `else if` and make the final `else` a `wrong_type!` or something like that.

rust_src/src/editfns.rs
nickdrozd: Maybe. There are a few uses of `selected_window().as_window()`, and I don't understand what guarantees are made about the selected window.

shaleh: `Option<LispWindowRef>` is likely the right thing for it to return. This allows for a nil selection initially.

shaleh: But I think it is almost always `Some`.

PR #1258, opened 2019-01-18
Port record and make-record

### LINKED ISSUE #1172: Port make-record
```c
DEFUN ("make-record", Fmake_record, Smake_record, 3, 3, 0,
       doc: /* Create a new record.
TYPE is its type as returned by `type-of'; it should be either a
symbol or a type descriptor.  SLOTS is the number of non-type slots,
each initialized to INIT.  */)
  (Lisp_Object type, Lisp_Object slots, Lisp_Object init)
{
  CHECK_NATNUM (slots);
  EMACS_INT size = XFASTINT (slots) + 1;
  struct Lisp_Vector *p = allocate_record (size);
  p->contents[0] = type;
  for (ptrdiff_t i = 1; i < size; i++)
    p->contents[i] = init;
  return make_lisp_ptr (p, Lisp_Vectorlike);
}
```

agraven:
This closes #1172.

### DIFF
--- a/rust_src/src/alloc.rs
+++ b/rust_src/src/alloc.rs
@@ -65,6 +65,10 @@ pub fn bool_vector(args: &mut [LispObject]) -> LispObject {
     vector
 }
 
+/// Create a new record.
+/// TYPE is its type as returned by `type-of'; it should be either a
+/// symbol or a type descriptor.  SLOTS is the number of non-type slots,
+/// each initialized to INIT.
 #[lisp_fn]
 pub fn make_record(r#type: LispObject, slots: EmacsUint, init: LispObject) -> LispObject {
     let size = slots + 1;
@@ -79,6 +83,11 @@ pub fn make_record(r#type: LispObject, slots: EmacsUint, init: LispObject) -> Li
     }
 }
 
+/// Create a new record.
+/// TYPE is its type as returned by `type-of'; it should be either a
+/// symbol or a type descriptor.  SLOTS is used to initialize the record
+/// slots with shallow copies of the arguments.
+/// usage: (record TYPE &rest SLOTS)
 #[lisp_fn]
 pub fn record(args: &mut [LispObject]) -> LispObject {
     unsafe {
--- a/rust_src/src/alloc.rs
+++ b/rust_src/src/alloc.rs
@@ -1,16 +1,13 @@
 //! Storage allocation and gc
 
-use libc::c_void;
-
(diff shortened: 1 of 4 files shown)


rust_src/src/alloc.rs
```diff
@@ -57,4 +65,39 @@ pub fn bool_vector(args: &mut [LispObject]) -> LispObject {
...
+pub fn make_record(r#type: LispObject, slots: EmacsUint, init: LispObject) -> LispObject {
+    let size = slots + 1;
+    unsafe {
+        let ptr = allocate_record(size as i64);
+        let contents = (*ptr).contents.as_mut_slice(size as usize);
+        contents[0] = r#type;
+        for rec in contents.iter_mut().skip(1) {
+            *rec = init;
+        }
+        make_lisp_ptr(ptr as *mut c_void, Lisp_Vectorlike)
```
brotzeit: AFAIK we use tag_ptr in rust

rust_src/src/alloc.rs
```diff
@@ -57,4 +65,39 @@ pub fn bool_vector(args: &mut [LispObject]) -> LispObject {
     vector
 }
 
+/// Create a new record.
+/// TYPE is its type as returned by `type-of'; it should be either a
+/// symbol or a type descriptor.  SLOTS is the number of non-type slots,
+/// each initialized to INIT.
+#[lisp_fn]
+pub fn make_record(r#type: LispObject, slots: EmacsUint, init: LispObject) -> LispObject {
```
shaleh: That '#' is a typo right?

agraven: No, it's a raw identifier to preserve the original argument name, but if that's not necessary then I'll change it.

nickdrozd: There are a few other instances where a `type` arg was renamed as `itype` or something like that. We should fix those with this technique.

shaleh: Remember to update the docstring when you rename an argument.

And kudos, I learned something about Rust today.

agraven: It's a relatively recent addition if I remember correctly, I think it got added to stable within the last 6 months or so.

rust_src/src/alloc.rs
agraven: This only works on `ExternalPtr`s though, it seems unnecessary to create an `ExternalPtr` only to immediately extract its contained pointer.

PR #1262, opened 2019-01-18
Make LispCons::set_c<ad>r generic

nickdrozd:


rust_src/src/lists.rs
```diff
@@ -324,16 +324,16 @@ impl LispCons {
     }
 
     /// Set the car of the cons cell.
-    pub fn set_car(self, n: LispObject) {
+    pub fn set_car<T: Into<LispObject>>(self, n: T) {
```
shaleh: :+1:

shaleh:
Related to this, Rust code should try to use `foo.set_car(value)` instead of `setcar(foo, value)`

shaleh:
In theory, we could do this to every function that isn't shared with C and isn't an exported lisp_fn.

What are the boundaries of our use of generics? Where should we do this and where should we not?

nickdrozd:
I think type conversion noise should be avoided whenever possible, so if implementing a function generically can do that, then it should be done. `LispObject::cons` was a good example of that: forcing callers to convert arguments to `LispObject` was annoying. I don't know if there are currently any other cases like that.

Issue #1253, opened 2019-01-17
Update wiki to document recent casting changes


shaleh:
@nickdrozd has begun migrating us to lean more heavily into Rust style casts instead of using `as_foo_or_error`. We should document this on the wiki and elsewhere.

krobelus:
I've done a write-up of my understanding of how we want to do casts.
Let me know what you think. We could use this as base for a wiki entry.

Every value in Emacs Lisp can be converted to `LispObject`.
There are no implicit conversions in Rust but explicit conversion is expected to
be implemenented via the `From` trait.

```rust
impl From<Foo> for LispObject {
    fn from(f: Foo) -> Self {
        ...
    }
}
```

Conversely, to convert a `LispObject` to some more specialized type:

```rust
impl From<LispObject> for Option<Foo> {
    fn from(o: LispObject) -> Self {
        ...
    }
}
```

Unpacking the `Option` from above and signalling an error if the conversion
failed is a very common operation. Therefore every type should also provide
this trivial implementation:

```rust
impl From<LispObject> for Foo {
    fn from(o: LispObject) -> Self {
        Option::from(o).unwrap_or_else(|| wrong_type!(Qfoop))
    }
}
```

By implementing the `From` trait, we can use `.into()` to perform 
type conversion when the target type is unambiguous - otherwise you can always
use `Foo::from()`.

Consider the following example where `divide` is a function that accepts two
values of type`LispNumber`. We can easily convert the arguments without
specifiying the target type.

```rust
fn reciprocal(x: LispObject) -> LispNumber {
    divide(1.into(), x.into())
}
```

If above conversion fails, then `divide` is never called.  Note that such
a function should probably just accept `LispNumber` instead of `LispObject`
but we have to match legacy interfaces before we can improve them
incrementally.

These methods have the advantage of not requiring us to remember new function
names for every type, it's just `into` or `from` plus the name of the type.
Of course you have to be aware that a conversion might cause a `wrong-type-argument`
error but that's what we want most of the time.

shaleh:
Casting is not always cheap. It depends on the type.

I suspect a better approach is to use `Lisp_type` and a macro.

PR #1264, opened 2019-01-19
Port compute-motion

krobelus:


rust_src/src/lists.rs
```diff
@@ -260,7 +263,7 @@ impl Iterator for CarIter {
 
 impl From<LispObject> for LispCons {
     fn from(o: LispObject) -> Self {
-        o.as_cons().unwrap_or_else(|| wrong_type!(Qconsp, o))
+        o.as_cons_or_error()
```
daichimukai: As far as I understand (I'm not fully catching up :disappointed_relieved:), it seems to be better to use a former implementation as currently @nickdrozd and others working on improving the type conversion mechanism for Rust codes (e.g. `.as_<type>()` and `.as_<type>_or_error()` should be `into()`, inherited from `From`. We should make it as documents (#1253)). #1221 and related issues might be helpful reference.

daichimukai:
Nice work :sparkles:.  Please wait for someone's review as I commented.

rust_src/src/lists.rs
krobelus: Good point, I wasn't aware of this!
I changed some conversions to use `into()` (or `Foo::from()` if type annotations are necessary)

rust_src/src/indent.rs
```diff
@@ -134,4 +139,128 @@ pub extern "C" fn invalidate_current_column() {
...
+
+    let [to_hpos, to_vpos] = if topos.is_nil() {
+        [
+            window_width,
+            unsafe { window_internal_height(win.as_mut()) } as EmacsInt,
+        ]
+    } else {
+        number_pair(topos.into())
+    };
+    let [hscroll, tab_offset] = if offsets.is_nil() {
```
agraven: You can use a match statement here if you change the type of offsets.

agraven:
Good work. The only changes to make here are minor ones.

rust_src/src/frames.rs
```diff
@@ -69,6 +69,18 @@ impl LispFrameRef {
...
     }
+    /// Replaces FRAME_WINDOW_P
+    pub fn is_gui_window(self) -> bool {
+        #[allow(unused_variables)]
+        let method = self.output_method();
+        #[cfg(feature = "HAVE_X_WINDOWS")]
+        return method == output_method::output_x_window;
+        #[cfg(feature = "HAVE_NTGUI")]
+        return method == output_method::output_w32;
+        #[cfg(feature = "HAVE_NS")]
```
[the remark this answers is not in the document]
krobelus: oops :S

rust_src/src/frames.rs
```diff
@@ -69,6 +69,15 @@ impl LispFrameRef {
             return 0;
         }
     }
+
+    /// Replaces FRAME_WINDOW_P
+    pub fn is_gui_window(self) -> bool {
+        #[allow(unused_variables)]
+        let method = self.output_method();
+        (cfg!(feature = "window-system-x11") && method == output_method::output_x_window)
```
shaleh: https://github.com/alexcrichton/cfg-if might be a little cleaner. Worth a look.

krobelus: yeah, that definitely looks like the way to go, especially with code that should't be compiled. Here there's not much difference but I've already had an instance of `#[cfg(foo)]` + `#[cfg(not(foo))]`

rust_src/src/indent.rs
```diff
@@ -178,4 +180,126 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
...
+pub fn compute_motion(
+    from: LispNumber,
+    frompos: LispCons,
+    to: LispNumber,
+    topos: Option<LispCons>,
+    width: Option<EmacsInt>,
+    offsets: Option<LispCons>,
+    window: LispWindowLiveOrSelected,
+) -> LispObject {
+    fn number_pair(cons: LispCons) -> [EmacsInt; 2] {
```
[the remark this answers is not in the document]
shaleh: You should be able to use the 2-tuple as a function argument. `Option<(LispObject, LispObject)>`.

rust_src/src/indent.rs
```diff
@@ -178,4 +180,126 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
...
+            win.as_mut(),
+        )
+    };
+
+    list!(
+        pos.bufpos,
+        pos.hpos,
+        pos.vpos,
+        pos.prevhpos,
+        if pos.contin != 0 { Qt } else { Qnil }
```
shaleh: Or `(pos.contin != 0).into()`. Both forms are valid.

krobelus: turns out it even works without `into()`, because `cons` does that :)

rust_src/src/indent.rs
```diff
@@ -178,4 +179,123 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
...
+/// LINE of a certain window, pass the window's starting location as
+/// FROM and the window's upper-left coordinates as FROMPOS.  Pass the
+/// buffer's (point-max) as TO, to limit the scan to the end of the
+/// visible section of the buffer, and pass LINE and COL as TOPOS.
+#[lisp_fn]
+pub fn compute_motion(
+    from: LispNumber,
+    frompos: (LispObject, LispObject),
+    to: LispNumber,
+    topos: Option<(LispObject, LispObject)>,
```
shaleh: For now, let's leave off the tuple argument types. We can revisit this once we update docfile's parser.

krobelus: certainly. should be done now

rust_src/src/frames.rs
```diff
@@ -87,6 +88,19 @@ impl LispFrameRef {
     }
 }
 
+cfg_if! {
+    if #[cfg(feature = "window-system-x11")] {
```
shaleh: Instead of an `Option` here you can use the rest of the enum values. `termcap` is the non-gui interface. `initial` is the first value. See frame.h and termhooks.h.

rust_src/src/frames.rs
```diff
@@ -87,6 +88,19 @@ impl LispFrameRef {
     }
 }
```
shaleh: As I ponder this a bit, I think all of this could be a `const fn`: https://doc.rust-lang.org/unstable-book/language-features/const-fn.html instead of defining a global variable.

rust_src/src/frames.rs
krobelus: Sorry for the long delay.
I have simplified it now to do a match on the frame's output method.
`initial` and `termcap` are not considered a GUI window, anything else is.
This way I don't even need this helper function. It should still be equivalent to the `FRAME_WINDOW_P` macro, assuming that builds for, say X  don't create `output_w32` frames.

rust_src/src/indent.rs
```diff
@@ -179,4 +180,124 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
...
+/// LINE of a certain window, pass the window's starting location as
+/// FROM and the window's upper-left coordinates as FROMPOS.  Pass the
+/// buffer's (point-max) as TO, to limit the scan to the end of the
+/// visible section of the buffer, and pass LINE and COL as TOPOS.
+#[lisp_fn]
+pub fn compute_motion(
+    from: LispNumber,
+    frompos: LispObject,
+    to: LispNumber,
+    topos: Option<LispObject>,
```
shaleh: LispObject is already essentially a `Option` type. No need to wrap it in one here. Check if `topos` is nil below and all should be fine.

rust_src/src/indent.rs
```diff
@@ -179,4 +180,124 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
...
+    width: Option<EmacsInt>,
+    offsets: Option<LispObject>,
```
shaleh: Same with `offsets`. No need for Option wrapper.

rust_src/src/indent.rs
```diff
@@ -8,18 +8,19 @@ use crate::{
     buffers::{point_byte, point_min_byte},
     editfns::{insert_char, point, point_min},
     lisp::LispObject,
+    marker::buf_charpos_to_bytepos,
     multibyte::Codepoint,
+    numbers::LispNumber,
     remacs_sys::globals,
-    remacs_sys::EmacsUint,
```
shaleh: We have been separating types from functions in an attempt to reduce merge conflicts. Which means the imports should look something like:
```
remacs_sys::{EmacsInt, EmacsUint},
remacs_sys::{
    self, del_range, find_newline, position_indentation, sanitize_tab_width, scan_for_column, set_point,
    set_point_both, window_internal_height,
},
remacs_sys::Qt,
```

shaleh: BTW, window_internal_height is a good candidate for an easy PR.

krobelus: done, i also put `globals` with the other global variables (that are not symbols)

rust_src/src/indent.rs
```diff
@@ -179,4 +179,128 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
...
+    to: LispNumber,
+    topos: LispObject,
+    width: Option<EmacsInt>,
+    offsets: LispObject,
+    window: LispWindowLiveOrSelected,
+) -> LispObject {
+    let frompos_tuple: (LispObject, LispObject) = frompos.into();
+    let from_hpos = frompos_tuple.0.into();
+    let from_vpos = frompos_tuple.1.into();
+
```
shaleh: Here is an alternative approach:
```
diff --git a/rust_src/src/indent.rs b/rust_src/src/indent.rs
index 9bd8301066..1ce3d8cb8d 100644
--- a/rust_src/src/indent.rs
+++ b/rust_src/src/indent.rs
@@ -8,6 +8,7 @@ use crate::{
     buffers::{point_byte, point_min_byte},
     editfns::{insert_char, point, point_min},
     lisp::LispObject,
+    lists::LispCons,
     marker::buf_charpos_to_bytepos,
     multibyte::Codepoint,
     numbers::LispNumber,
@@ -224,16 +225,15 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
 #[lisp_fn]
 pub fn compute_motion(
     from: LispNumber,
-    frompos: LispObject,
+    frompos: LispCons,
     to: LispNumber,
     topos: LispObject,
     width: Option<EmacsInt>,
     offsets: LispObject,
     window: LispWindowLiveOrSelected,
 ) -> LispObject {
-    let frompos_tuple: (LispObject, LispObject) = frompos.into();
-    let from_hpos = frompos_tuple.0.into();
-    let from_vpos = frompos_tuple.1.into();
+    let from_hpos: EmacsInt = frompos.car().into();
+    let from_vpos: EmacsInt = frompos.cdr().into();

     let mut win: LispWindowRef = window.into();
     let window_width = EmacsInt::from(win.body_width(false))
```

krobelus: certainly better, done

rust_src/src/frames.rs
```diff
@@ -30,9 +30,12 @@ impl LispFrameRef {
         !self.terminal.is_null()
     }
 
-    // Awaiting Wilfred#1264
+    /// Replaces FRAME_WINDOW_P
     pub fn is_gui_window(self) -> bool {
-        cfg!(feature = "window_system")
```
shaleh: This appears to be a reasonable approach.

rust_src/src/indent.rs
```diff
@@ -179,4 +180,127 @@ pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
...
+        - if cfg!(feature = "window-system") && win.get_frame().is_gui_window() {
+            0
+        } else {
+            1
+        };
+
+    let (to_hpos, to_vpos) = if topos.is_nil() {
+        (
+            window_width,
+            EmacsInt::from(unsafe { window_internal_height(win.as_mut()) }),
```
shaleh: `window_internal_height` should be a member of `LispWindowRef`, something like `internal_height()`. The needed C function can be a light wrapper around it. here we already have a LispWindowRef so we can use the member directly and get rid of both the unsafe and the EmacsInt conversion.

agraven: I ported it. How is the integer conversion skippable though? All window size related functions use `i32`.

agraven:
@shaleh are there any other needed changes to this or is it good to go?

brotzeit:
Is it possible to add tests for this function ?

PR #1269, opened 2019-01-21
Port x_focus_frame

jubnzv:


rust_src/src/frames.rs
```diff
@@ -593,4 +594,22 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
...
+/// Set the input focus to FRAME.
+/// FRAME nil means use the selected frame.  Optional argument NOACTIVATE
+/// means do not activate FRAME.
+/// If there is no window system support, this function does nothing.
+#[lisp_fn(min = "1")]
+pub fn x_focus_frame(frame: LispFrameLiveOrSelected, noactivate: LispObject) -> LispObject {
+    #[cfg(feature = "window-system")]
+    {
+        let frame_ref: LispFrameRef = frame.into();
+        let frame_ref_ptr = frame_ref.as_ptr() as *mut Lisp_Frame;
```
shaleh: You can simply this by using `as_mut()`.

rust_src/src/frames.rs
```diff
@@ -593,4 +594,22 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
     frame_ref.focus_frame
 }
 
+/// Set the input focus to FRAME.
+/// FRAME nil means use the selected frame.  Optional argument NOACTIVATE
+/// means do not activate FRAME.
+/// If there is no window system support, this function does nothing.
```
shaleh: There should be an empty line with just `///` above the line 'If there is no....'.

rust_src/src/frames.rs
```diff
@@ -8,6 +8,7 @@ use crate::{
     lisp::defsubr,
     lisp::{ExternalPtr, LispObject},
     lists::{LispConsCircularChecks, LispConsEndChecks},
+    remacs_sys::x_focus_frame as c_x_focus_frame,
```
shaleh: No need to import this one separately. Group it with the other functions.

agraven:
If you look at the travis results, you can see that it fails when it tries to build without a window system. The specific error it gives is `error[E0432]: unresolved import 'crate::remacs_sys::x_focus_frame'  --> src/frames.rs:11:5`.

That means you need to check at compile time whether there is a window system with `#[cfg(feature = "window-system")]`. This is the case wherever `#ifdef HAVE_WINDOW_SYSTEM` is used in C code.

jubnzv:
Oh, it seems to work. Thank you for the thoughtful review. I just beginning in Rust and it takes some time to figure out with basics.

shaleh:
@jubnzv this project exists in part to encourage people to learn Rust. Welcome aboard.

rust_src/src/frames.rs
```diff
@@ -593,4 +593,23 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
     frame_ref.focus_frame
 }
 
+/// Set the input focus to FRAME.
+/// FRAME nil means use the selected frame.  Optional argument NOACTIVATE
+/// means do not activate FRAME.
+///
+/// If there is no window system support, this function does nothing.
+#[lisp_fn(min = "1")]
+pub fn x_focus_frame(_frame: LispFrameLiveOrSelected, _noactivate: LispObject) -> LispObject {
```
agraven: You can set the type of `noactivate` to bool, that way you can pass it directly to `c_x_focus_frame`.

rust_src/src/frames.rs
```diff
@@ -44,7 +44,7 @@ impl LispFrameRef {
             match self.vertical_scroll_bar_type() {
                 vertical_scroll_bar_type::vertical_scroll_bar_left
                 | vertical_scroll_bar_type::vertical_scroll_bar_right => {
-                    return self.config_scroll_bar_width
+                    return self.config_scroll_bar_width;
                 }
                 _ => return 0,
```
agraven: The return statements here are redundant, you can remove them.

rust_src/src/frames.rs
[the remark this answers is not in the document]
shaleh: I am sorry @agraven, I thought I hit review earlier.

rust_src/src/frames.rs
agraven: Disregard what I just wrote, I'd actually suggest removing the added semicolon and leaving this be for now, as there's another open PR that fixes this.

rust_src/src/frames.rs
```diff
@@ -507,13 +507,13 @@ pub fn next_frame(frame: LispFrameOrSelected, miniframe: LispObject) -> LispFram
     // a valid candidate will be returned regardless of its position.
     while passed < 2 {
         for_each_frame!(f => {
```
shaleh: This change and the one above it do not belong and will conflict with other active PRs that are resolving the situation. Please remove them from this PR.

rust_src/src/frames.rs
```diff
@@ -593,4 +593,22 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
...
+pub fn x_focus_frame(_frame: LispFrameLiveOrSelected, _noactivate: bool) -> LispObject {
+    #[cfg(feature = "window-system")]
+    {
+        let mut frame_ref: LispFrameRef = _frame.into();
+        unsafe {
+            c_x_focus_frame(frame_ref.as_mut(), _noactivate);
+        }
+    }
+
+    Qnil
```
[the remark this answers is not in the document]
shaleh: If the function only and always returns `Qnil` you can define the Rust function like this:

    #[lisp_fn]
    pub fn some_function(/* args.. */) {
        // code but no returns.
    }

and the right thing will happen.

PR #1272, opened 2019-01-21
Port window-body-height and window-body-width

### LINKED ISSUE #1188: Port window-body-height and window-body-width
```c

DEFUN ("window-body-height", Fwindow_body_height, Swindow_body_height, 0, 2, 0,
       doc: /* Return the height of WINDOW's text area.
WINDOW must be a live window and defaults to the selected one.  Optional
argument PIXELWISE non-nil means return the height of WINDOW's text area
in pixels.  The return value does not include the mode line or header
line or any horizontal divider.

If PIXELWISE is nil, return the largest integer smaller than WINDOW's
pixel height divided by the character height of WINDOW's frame.  This
means that if a line at the bottom of the text area is only partially
visible, that line is not counted.  */)
  (Lisp_Object window, Lisp_Object pixelwise)
{
  return make_number (window_body_height (decode_live_window (window),
					  !NILP (pixelwise)));
}

DEFUN ("window-body-width", Fwindow_body_width, Swindow_body_width, 0, 2, 0,
       doc: /* Return the width of WINDOW's text area.
WINDOW must be a live window and defaults to the selected one.  Optional
argument PIXELWISE non-nil means return the width in pixels.  The return
value does not include any vertical dividers, fringes or marginal areas,
or scroll bars.

If PIXELWISE is nil, return the largest integer smaller than WINDOW's
pixel width divided by the character width of WINDOW's frame.  This
means that if a column at the right of the text area is only partially
visible, that column is not counted.

Note that the returned value includes the column reserved for the
continuation glyph.  */)
  (Lisp_Object window, Lisp_Object pixelwise)
{
  return make_number (window_body_width (decode_live_window (window),
					 !NILP (pixelwise)));
```

agraven:
This closes #1188.

Because the vast majority of functions used in `window_body_{height,width}` are macros, I had to add quite a few functions to `LispWindowRef`.

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -22,16 +22,21 @@ use crate::{
         estimate_mode_line_height, minibuf_level,
         minibuf_selected_window as current_minibuf_window, scroll_command, select_window,
         selected_window as current_window, set_buffer_internal, set_window_hscroll,
-        update_mode_lines, window_body_width, window_list_1, window_menu_bar_p, window_tool_bar_p,
-        wset_redisplay,
+        update_mode_lines, window_list_1, window_menu_bar_p, window_tool_bar_p, wset_redisplay,
     },
-    remacs_sys::{face_id, glyph_matrix, pvec_type, EmacsInt, Lisp_Type, Lisp_Window},
     remacs_sys::{
-        Qbottom, Qceiling, Qfloor, Qheader_line_format, Qmode_line_format, Qnil, Qnone, Qt,
+        face_id, glyph_matrix, pvec_type, vertical_scroll_bar_type, EmacsInt, Lisp_Type,
+        Lisp_Window,
+    },
+    remacs_sys::{
+        Qceiling, Qfloor, Qheader_line_format, Qleft, Qmode_line_format, Qnil, Qnone, Qright, Qt,
         Qwindow_live_p, Qwindow_valid_p, Qwindowp,
     },
     threads::ThreadState,
 };
+
+#[allow(unused_imports)]
+use crate::remacs_sys::Qbottom;
 
 pub type LispWindowRef = ExternalPtr<Lisp_Window>;
 
@@ -50,6 +55,12 @@ impl LispWindowRef {
     /// contents slot is non-nil.
     pub fn is_valid(self) -> bool {
         self.contents.is_not_nil()
+    }
+
+    // Equivalent to WINDOW_RIGHTMOST_P
+    /// True if window W has no other windows to its right on its frame.
+    pub fn is_rightmost(self) -> bool {
+        self.right_pixel_edge() == self.get_frame().root_window().right_pixel_edge()
     }
 
     // Equivalent to WINDOW_BOTTOMMOST_P
@@ -180,7 +191,6 @@ impl LispWindowRef {
         }
     }
 
-    // Equivalent to window_body_height
     /// Return the number of lines/pixels of W's body. Don't count any mode
     /// or header line or horizontal divider of W. Rounds down to nearest
     /// integer when not working pixelwise.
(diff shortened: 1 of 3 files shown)


rust_src/src/windows.rs
```diff
@@ -51,10 +52,30 @@ impl LispWindowRef {
         self.contents.is_not_nil()
     }
 
+    // Equivalent to WINDOW_BOTTOMMOST_P
```
krobelus: Should we expose ports of macros like this to C under their original name if they are still used?

agraven: This is actually done with a few macros in `windows.rs`, I suppose it would make sense to do it with more of them.

agraven:
Awaiting #1264 for implementation of `is_gui_window`.

rust_src/src/windows.rs
```diff
@@ -357,6 +649,17 @@ impl From<LispWindowValidOrSelected> for LispWindowRef {
     }
 }
 
+// Avoid name conflicts
+mod sys {
+    use super::LispWindowRef;
+    use crate::remacs_sys::Lisp_Window;
+
+    #[no_mangle]
+    pub extern "C" fn window_body_width(window: *mut Lisp_Window, pixelwise: bool) -> i32 {
```
shaleh: 1) The usual convention is to name the `lisp_fn` as `foo_lisp` to avoid this.
2) Why not implement this in terms of `window_body_width_lisp` instead of replicating the code? It would be nice if the `extern "C"` functions were simply shallow wrappers onto the Rust code. That way we can remove them as we stop needing them but the code is already written. The Rust code can use the full Rust version.

rust_src/src/windows.rs
```diff
@@ -1206,4 +1509,39 @@ pub fn set_window_redisplay_end_trigger(
...
+/// WINDOW must be a live window and defaults to the selected one.  Optional
+/// argument PIXELWISE non-nil means return the height of WINDOW's text area
+/// in pixels.  The return value does not include the mode line or header
+/// line or any horizontal divider.
+///
+/// If PIXELWISE is nil, return the largest integer smaller than WINDOW's
+/// pixel height divided by the character height of WINDOW's frame.  This
+/// means that if a line at the bottom of the text area is only partially
+/// visible, that line is not counted.
+#[lisp_fn]
```
shaleh: Both of these are `min = "0"`.

rust_src/src/windows.rs
```diff
@@ -91,6 +134,19 @@ impl LispWindowRef {
         self.mini()
     }
 
+    // Equivalent to MINI_NON_ONLY_P
+    /// True if W is a minibuffer window on a frame that contains at least
+    /// one other window
+    pub fn is_minibuffer_non_only(self) -> bool {
+        self.is_minibuffer() && self.prev.is_not_nil()
```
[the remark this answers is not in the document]
agraven: Because you still need to do the `is_minibuffer()` check

rust_src/src/windows.rs
agraven: All right, I fixed the names.

Regarding 2, implementing it in terms of `window_body_width_lisp` seems unnecessary, as it requires converting the argument to a `LispObject` and then to an `ExternalPtr`. This would also run the `LiveOrSelected`/`decode_live_window` check, which the original code does not. Besides, it's more or less a shallow wrapper already, once it's no longer used by the C code it can be dropped without affecting the rust code.

rust_src/src/windows.rs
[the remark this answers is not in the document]
shaleh: Ah, good point. Read that too fast.

rust_src/src/windows.rs
```diff
@@ -228,6 +406,113 @@ impl LispWindowRef {
...
+
+    pub fn has_vertical_scroll_bar(self) -> bool {
+        self.has_vertical_scroll_bar_left() || self.has_vertical_scroll_bar_right()
+    }
+
+    // Equivalent to WINDOW_HAS_HORIZONTAL_SCROLL_BAR
+    /// True if horizontal scroll bars are currently enabled for the window.
+    /// Horizontal scrollbars only exist when a toolkit is enabled.
+    pub fn has_horizontal_scroll_bar(self) -> bool {
+        #[cfg(feature = "window-system")]
```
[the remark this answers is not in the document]
shaleh: Good to know. Thanks.

rust_src/src/windows.rs
shaleh: You would not call `Fwindow_body_width` but instead `window_body_width_lisp`. Then you only need to make a `WindowLiveOrSelected` from a `LispWindowRef` which is another name for an `ExternalPtr`.

However, the C function does not require a `LiveOrSelected` window. So we are better off implementing the C function and the Rust lisp_fn in terms of a shared function that can be called after each performs their type marshalling.

Yes this looks like extra work. But the C interfaces should be kept as light as possible. We have not been as mindful of this in the past and have built up some tech debt.

agraven: Does the `body_width` method on `LispWindowRef` not fullfil the purpose of a shared function?

shaleh: When we have ported some more functions, we will eventually remove the need for the C `window_body_width`. Will we also remove the need for its functionality? Or will all users switch to the method. If that is the plan, you are right we do not need it.

I am bringing this up mostly to reinforce this process as we continue porting.

How long will this function be useful? Is it really just a stop gap or will we need a Rust version once it is gone? Ideally  as we port not just the DEFUN but the underlying support code we try to make a cohesive application that for now has C side channels.

agraven: There will definitely be a need for the method's functionality in the future. And once the C function is no longer used, it can be dropped without requiring rust code changes, as it is just a wrapper for the `body_width` method.

PR #1273, opened 2019-01-22
port frame-list

brotzeit:


rust_src/src/frames.rs
```diff
@@ -593,4 +607,26 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
...
+/// Return a list of all live frames.
+/// The return value does not include any tooltip frame.
+#[lisp_fn]
+pub fn frame_list() -> LispObject {
+    #[cfg(feature = "window-system")]
+    {
+        let mut list: LispObject = Qnil;
+        for_each_frame!(f => {
+            if !f.has_tooltip() {
+                    list = LispObject::cons(LispObject::from(f), list);
```
shaleh: This is no longer the correct style.

`list = (f, list).into();` should be all you need.

rust_src/src/frames.rs
```diff
@@ -507,13 +521,13 @@ pub fn next_frame(frame: LispFrameOrSelected, miniframe: LispObject) -> LispFram
     // a valid candidate will be returned regardless of its position.
     while passed < 2 {
         for_each_frame!(f => {
-	    if passed > 0 {
-	        let tmp = unsafe { candidate_frame(f.into(), frame_obj, miniframe) };
-	        if !tmp.is_nil() {
+        if passed > 0 {
```
shaleh: Are you using the correct `rustfmt` here? Why would it drop the indentation??

agraven: The change stems from a tab literal being changed to four spaces, so the visible difference is a quirk of tabs being displayed as 8 spaces wide. I don't know how the tab literals got in there in the first place.

shaleh: That was likely me hacking in vi while I had a busted remacs binary :-)

shaleh: My merge of #1240 should resolve this for everyone.

shaleh: You should rebase. This issue is merged.

rust_src/src/frames.rs
```diff
@@ -614,17 +628,50 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
...
+}
 
-    Qnil
+/// Return a list of all live frames.
+/// The return value does not include any tooltip frame.
+#[lisp_fn]
+pub fn frame_list() -> LispObject {
+    #[cfg(feature = "window-system")]
+    {
+        let mut list: LispObject = Qnil;
```
shaleh: This is another place where a `fold` would also work. Either form is fine.

rust_src/src/frames.rs
```diff
@@ -614,17 +628,50 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
...
+#[lisp_fn]
+pub fn visible_frame_list() -> LispObject {
+    let mut list: LispObject = Qnil;
+    for_each_frame!(f => {
+            if f.is_visible() {
+                    list = (f, list).into();
+            }
+        }
+        );
+    list
```
shaleh: `visible_frame_list` and `frame_list` are the same code with the only difference being the predicate used to decide which elements are included in the list. Perhaps the right thing to do is factor that code out.

```
fn filter_frame_list(predicate: impl Fn(LispFrameRef) -> bool) -> LispObject {
    // or the fold version...
    for_each_frame!(f => {
            if predicate(f) {
                list = (f, list).into();
            }
        });
}

pub fn visible_frame_list() -> LispObject {
    filter_frame_list(|f| f.is_visible())
}

pub fn frame_list() -> LispObject {
    #[cfg(feature = "window-system")]
    {
        let mut list = filter_frame_list(|f| !f.has_tooltip());
        // Reverse list for consistency with the !HAVE_WINDOW_SYSTEM case.
        unsafe { Fnreverse(list) }
    }
    #[cfg(not(feature = "window-system"))]
    {
        unsafe { Fcopy_sequence(Vframe_list) }
    }
}
```

brotzeit: I think we should use `for_each_frame` consistently or change it to something else. So I will take this one.

shaleh: You asked about folds recently. So I pointed these out. The for each is fine. And sometimes it reads better. These two versions are nearly identical.

shaleh:
This is ready. But it needs a merge or rebase to get rid of the unrelated change.

shaleh:
@brotzeit this should be ready to merge once you land the other PR and rebase this one.

brotzeit:
Ready.

PR #1277, opened 2019-01-22
Use impl notation for generics

nickdrozd:


nickdrozd:
@shaleh You had some concerns previously about the readability of generic functions. What do you think of this style? I just learned about it.

shaleh:
Reminder for others, this syntax is new with Rust 2018 edition. https://rust-lang-nursery.github.io/edition-guide/rust-2018/trait-system/impl-trait-for-returning-complex-types-with-ease.html

> However, there's also an important technical difference between T: Trait and impl Trait here. When you write the former, you can specify the type of T at the call site with turbo-fish syntax as with foo::<usize>(1). In the case of impl Trait, if it is used anywhere in the function definition, then you can't use turbo-fish at all. Therefore, you should be mindful that changing both from and to impl Trait can constitute a breaking change for the users of your code.

So we need to be mindful. If we ever need to say `foo::<SomeType>::eq(bar)` then we should stick with the `T: trait` syntax. Here for the eq it is highly unlikely that we ever want turbofish.

nickdrozd:
I don't understand the turbofish stuff, could you explain what it does? Do we use it anywhere?

shaleh:
From a section of Rust docs:

> Because collect() is so general, it can cause problems with type inference. As such, collect() is one of the few times you'll see the syntax affectionately known as the 'turbofish': ::<>. This helps the inference algorithm understand specifically which collection you're trying to collect into.

nickdrozd:
With `impl` notation, the type is anonymous, and it can't be referenced (say, with a turbofish). But in these cases `T` wasn't being used anyway. Is there a concern that some type reference might be added later?

shaleh:
For `eq` no, impl is likely ok. However, ut would not work for the `isa` function that was proposed. When making the decision of which style we need to consider this.

I'd also like to know how it is implemented internally. The type generic style is like C++ templates and a new function is made behind the scenes for each type. Is the impl Trait style done the same way? Or is it done with Box behind the scenes? We are pushing casts pretty hard. They need to be cheap.

nickdrozd:
Do we know what the hotspots are in GNU Emacs? It could be that we only need to worry the cost of casting in a few places, and that those places haven't been ported yet (the garbage collector, the bytecode interpreter, display of some kind, etc). But that's just a possibility. We really should get profiling figured out.

shaleh:
I know that until we can inline functions we are likely more expensive than C. We don't get inline until we are self hosting on Rust. Yes, profiling is a good idea.

nickdrozd:
Added a few more

Issue #1278, opened 2019-01-23
make-docfile fails when arguments contain extra commas 


krobelus:
This can be reproduced by applying below patch and running `make`.
It can also be observed for lisp_fns that have tuple arguments, say `(LispObject, LispObject)`.

The problem seems to be that `docfile.rs` splits the source code of arguments by ','.
We could solve this by the logic from `docfile.rs` to the `lisp_fn` macro.
If that's takes some time, I'd suggest fixing it now by having `lisp_fn` throw an error whenever there are extra occurrences of characters that the docstring generation uses to split (looks like these are `():,`.

```diff
@@ -149,7 +149,7 @@ pub extern "C" fn invalidate_current_column() {
 ///
 /// The return value is the column where the insertion ends.
 #[lisp_fn(min = "1", intspec = "NIndent to column: ")]
-pub fn indent_to(column: EmacsInt, minimum: Option<EmacsInt>) -> EmacsInt {
+pub fn indent_to(column: EmacsInt, /* , */ minimum: Option<EmacsInt>) -> EmacsInt {
     let buffer = ThreadState::current_buffer_unchecked();
     let tab_width = unsafe { sanitize_tab_width(buffer.tab_width_.force_fixnum()) } as EmacsInt;
     let arg_minimum = minimum.unwrap_or(0);
```

agraven:
@shaleh It seems to be an issue with tuple arguments as well, which is definitely a problem that needs to be handled.

krobelus:
Apart from the tuple case I don't think there will be a reason for having extra commas.
Comments should not appear within a signature - I just used them while debugging. Since adding comments can give unexpected results I'd rather ban them explicitly.

PR #1279, opened 2019-01-23
docfile.rs: panic on invalid function signatures

krobelus:
(I understand this is not an ideal solution but better than the status quo imho)

Some function signatures can not be handled correctly by docfile.
(See #1278)

Panic with a helpful message when such a case might occur.
We check for two conditions:

- If there is a comment within the signature, then we always
  fail. Comments might contain dangerous characters and they shouldn't
  be here anyway.

- If the signature contains consecutive commas (or colons), it is
  definitely not handled correctly (this handles the issue with
  tuples).

### DIFF
--- a/rust_src/remacs-lib/docfile.rs
+++ b/rust_src/remacs-lib/docfile.rs
@@ -43,10 +43,12 @@ pub unsafe extern "C" fn scan_rust_file(
     let mut attribute = String::new();
 
     let mut line_iter = fp.lines();
+    let mut line_number = 0;
 
     let mut in_lisp_fn = false;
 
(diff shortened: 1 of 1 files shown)


shaleh:
Thanks for making this to demonstrate the problem.

We currently have parsing logic used for the `lisp_fn` based on the `syn` crate. A solution would be to extend our code to parse the function declaration as well. In theory this mostly exists since we have the body to apply already.

shaleh:
I have a half done PR to move us to the curren Syn crate. The parsing is a little more obvious at that point. I will try to get it to a state I can share it over the next few days.

PR #1281, opened 2019-01-23
Set up 'workspace' section in cargo config

### LINKED ISSUE #1254: Set up a 'workspace' in our Cargo config
This will enable more cargo tools to work with our crates.

https://learning-rust.github.io/docs/d5.workspaces.html

daichimukai:
remacs-lib, remacs-util and remacs-macros are now members of the remacs workspace. These crates share the same Cargo.lock, output directory and profiles. We drop `lto = true` option until we start to self-host this project as a Rust crate.

Closes #1254.


agraven:
We want to track the new shared Cargo.lock, right?

daichimukai:
@agraven Yes we do, and in my understand the shared lock file is `rust_src/Cargo.lock` which we are already tracking!

agraven:
Ah, I forgot we're already tracking that file. Whoops.

daichimukai:
Fixed a conflict.

PR #1284, opened 2019-01-24
port daemonp

brotzeit:


rust_src/src/emacs.rs
```diff
@@ -34,4 +38,21 @@ pub fn invocation_directory() -> LispObject {
     unsafe { Fcopy_sequence(globals.Vinvocation_directory) }
 }
 
+/// Return non-nil if the current emacs process is a daemon.
+/// If the daemon was given a name argument, return that name.
+#[lisp_fn]
+pub fn daemonp() -> LispObject {
+    unsafe {
+        if is_daemon() {
+            if !daemon_name.is_null() {
```
shaleh: Flipping the check so it does not need to be negated I think will read better.

    if daemon_name.is_null() {
        Qt
    } else {
        build_string(daemon_name)
    }

daichimukai: How about avoiding the nesting?

```rust
if is_daemon() && !daemon_name.is_null() {
    return build_string(daemon_name);
}

is_daemon().into()
```

shaleh: The only negative there is the double call to `is_daemon`.

shaleh:
My understanding is we cannot inline anything currently. If you specify it the compiler will ignore you. All of the Rust code is compiled into a library which prevents the inline analysis. We used to specify inline attributes and stopped.

daichimukai:
@shaleh Interesting. Thanks for information.

PR #1292, opened 2019-01-27
fix x-focus-frame names

brotzeit:


rust_src/src/frames.rs
```diff
@@ -523,13 +523,13 @@ pub fn next_frame(frame: LispFrameOrSelected, miniframe: LispObject) -> LispFram
     // a valid candidate will be returned regardless of its position.
     while passed < 2 {
         for_each_frame!(f => {
-            if passed > 0 {
-                let tmp = unsafe { candidate_frame(f.into(), frame_obj, miniframe) };
-                if !tmp.is_nil() {
+        if passed > 0 {
```
daichimukai: Are these changes what you intended? It seems odd for me :confused:

I ran `rustfmt` for original code and changed one, and for both codes `rustfmt` changed nothing. I'm wondering why.

daichimukai:
Sorry for that I'm not sure about the context of this PR, but is this intended to resolve a collision of the name `x_frame_focus` in C and Rust codes?

brotzeit:
@daichimukai Yeah, now it follows our naming conventions.

rust_src/src/frames.rs
brotzeit: Hmm. It seems to be the right version since the travis build doesn't fail.

Did you remove and add rustfmt-preview after we updated the toolchain ?

shaleh: @brotzeit I think the error is on your side. I resolved this spacing issue when I updated the compiler version.

brotzeit: I don't get it...

brotzeit:
Done.

PR #1295, opened 2019-01-27
Port file-directory-p. Closes: #1238

m-novikov:


rust_src/src/fileio.rs
```diff
@@ -117,6 +119,23 @@ pub fn file_exists_p(filename: LispStringRef) -> bool {
     }
 }
 
+/// Return t if FILENAME names an existing directory.
+/// Symbolic links to directories count as directories.
+/// See `file-symlink-p' to distinguish symlinks.
+#[lisp_fn(name = "file-directory-p", c_name = "file_directory_p")]
+pub fn file_directory_p_lisp(filename: LispStringRef) -> bool {
+    let absname = unsafe { expand_and_dir_to_file(filename.into()) };
+    let handler = find_file_name_handler(absname.into(), Qfile_directory_p);
```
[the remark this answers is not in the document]
m-novikov: This is what I gen without into
```rust
128 |     let handler = find_file_name_handler(absname, Qfile_directory_p);
    |                                          ^^^^^^^ expected struct `lisp::ExternalPtr`, found struct `lisp::LispObject`
```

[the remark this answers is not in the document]
shaleh: @shanavas786 this `into` is converting from LispObject to a native Rust type. Same as using `LispStringRef::from(absname)`.

rust_src/src/fileio.rs
```diff
@@ -117,6 +119,23 @@ pub fn file_exists_p(filename: LispStringRef) -> bool {
...
+
+    if handler.is_not_nil() {
+        call!(handler, Qfile_directory_p, absname.into()).into()
+    } else {
+        unsafe { file_directory_p(encode_file_name(absname.into()).into()) }
+    }
+}
+
+def_lisp_sym!(Qfile_directory_p, "file-directory-p");
```
shaleh: :+1:

rust_src/src/fileio.rs
```diff
@@ -117,6 +119,23 @@ pub fn file_exists_p(filename: LispStringRef) -> bool {
...
+/// Return t if FILENAME names an existing directory.
+/// Symbolic links to directories count as directories.
+/// See `file-symlink-p' to distinguish symlinks.
+#[lisp_fn(name = "file-directory-p", c_name = "file_directory_p")]
+pub fn file_directory_p_lisp(filename: LispStringRef) -> bool {
+    let absname = unsafe { expand_and_dir_to_file(filename.into()) };
+    let handler = find_file_name_handler(absname.into(), Qfile_directory_p);
+
+    if handler.is_not_nil() {
+        call!(handler, Qfile_directory_p, absname.into()).into()
```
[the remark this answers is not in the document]
m-novikov: Same here:
```
expected struct `lisp::ExternalPtr`, found struct `lisp::LispObject`
```

rust_src/src/fileio.rs
m-novikov: @shaleh what does `def_lisp_sym!` actually do?
Because I still must add import of `remacs_sys::Qfile_directory_p` at the top, otherwise compilation fails.

shaleh: Yeah, this could use some attention.

`def_lisp_sym` is a do nothing macro. What happens is that during the build we have code in the build.rs tool which looks for these and generates the right definitions. These end up in rust_src/generated/globals.rs. Right now, globals.rs is imported by `remacs_sys`. So even though it is no longer a C define it still comes from the external source. Which can be a little confusing.

shaleh: https://github.com/Wilfred/remacs/blob/a145bff51035a6d4764ed5e17b084378049fb0ef/rust_src/remacs-lib/docfile.rs#L164 is where this is happening. On line 171 the name is added to the 'globals' and then at the end that becomes `globals.rs`.

daichimukai:
:tada:

PR #1296, opened 2019-01-28
Port window-lines-pixel-dimensions

### LINKED ISSUE #1246: Port window-lines-pixel-dimensions
```c
DEFUN ("window-lines-pixel-dimensions", Fwindow_lines_pixel_dimensions, Swindow_lines_pixel_dimensions, 0, 6, 0,
       doc: /* Return pixel dimensions of WINDOW's lines.
The return value is a list of the x- and y-coordinates of the lower
right corner of the last character of each line.  Return nil if the
current glyph matrix of WINDOW is not up-to-date.

Optional argument WINDOW specifies the window whose lines' dimensions
shall be returned.  Nil or omitted means to return the dimensions for
the selected window.

FIRST, if non-nil, specifies the index of the first line whose
dimensions shall be returned.  If FIRST is nil and BODY is non-nil,
start with the first text line of WINDOW.  Otherwise, start with the
first line of WINDOW.

LAST, if non-nil, specifies the last line whose dimensions shall be
returned.  If LAST is nil and BODY is non-nil, the last line is the last
line of the body (text area) of WINDOW.  Otherwise, last is the last
line of WINDOW.

INVERSE, if nil, means that the y-pixel value returned for a specific
line specifies the distance in pixels from the left edge (body edge if
BODY is non-nil) of WINDOW to the right edge of the last glyph of that
line.  INVERSE non-nil means that the y-pixel value returned for a
specific line specifies the distance in pixels from the right edge of
the last glyph of that line to the right edge (body edge if BODY is
non-nil) of WINDOW.

LEFT non-nil means to return the x- and y-coordinates of the lower left
corner of the leftmost character on each line.  This is the value that
should be used for buffers that mostly display text from right to left.

If LEFT is non-nil and INVERSE is nil, this means that the y-pixel value
returned for a specific line specifies the distance in pixels from the
left edge of the last (leftmost) glyph of that line to the right edge
(body edge if BODY is non-nil) of WINDOW.  If LEFT and INVERSE are both
non-nil, the y-pixel value returned for a specific line specifies the
distance in pixels from the left edge (body edge if BODY is non-nil) of
WINDOW to the left edge of the last (leftmost) glyph of that line.

Normally, the value of this function is not available while Emacs is
busy, for example, when processing a command.  It should be retrievable
though when run from an idle timer with a delay of zero seconds.  */)
  (Lisp_Object window, Lisp_Object first, Lisp_Object last, Lisp_Object body, Lisp_Object inverse, Lisp_Object left)
{
  struct window *w = decode_live_window (window);
  struct buffer *b;
  struct glyph_row *row, *end_row;
  int max_y = NILP (body) ? WINDOW_PIXEL_HEIGHT (w) : window_text_bottom_y (w);
  Lisp_Object rows = Qnil;
  int window_width = NILP (body) ? w->pixel_width : window_body_width (w, true);
  int header_line_height = WINDOW_HEADER_LINE_HEIGHT (w);
  int subtract = NILP (body) ? 0 : header_line_height;
  bool invert = !NILP (inverse);
  bool left_flag = !NILP (left);

  if (noninteractive || w->pseudo_window_p)
    return Qnil;

  CHECK_BUFFER (w->contents);
  b = XBUFFER (w->contents);

  /* Fail if current matrix is not up-to-date.  */
  if (!w->window_end_valid
      || windows_or_buffers_changed
      || b->clip_changed
      || b->prevent_redisplay_optimizations_p
      || window_outdated (w))
    return Qnil;

  if (NILP (first))
    row = (NILP (body)
	   ? MATRIX_ROW (w->current_matrix, 0)
	   : MATRIX_FIRST_TEXT_ROW (w->current_matrix));
  else if (NUMBERP (first))
    {
      CHECK_RANGED_INTEGER (first, 0, w->current_matrix->nrows);
      row = MATRIX_ROW (w->current_matrix, XINT (first));
    }
  else
    error ("Invalid specification of first line");

  if (NILP (last))

    end_row = (NILP (body)
	       ? MATRIX_ROW (w->current_matrix, w->current_matrix->nrows)
	       : MATRIX_BOTTOM_TEXT_ROW (w->current_matrix, w));
  else if (NUMBERP (last))
    {
      CHECK_RANGED_INTEGER (last, 0, w->current_matrix->nrows);
      end_row = MATRIX_ROW (w->current_matrix, XINT (last));
    }
  else
    error ("Invalid specification of last line");

  while (row <= end_row && row->enabled_p
	 && row->y + row->height < max_y)
    {

      if (left_flag)
	{
	  struct glyph *glyph = row->glyphs[TEXT_AREA];

	  rows = Fcons (Fcons (make_number
			       (invert
				? glyph->pixel_width
				: window_width - glyph->pixel_width),
			       make_number (row->y + row->height - subtract)),
			rows);
	}
      else
	rows = Fcons (Fcons (make_number
			     (invert
			      ? window_width - row->pixel_width
			      : row->pixel_width),
			     make_number (row->y + row->height - subtract)),
		      rows);
      row++;
    }

  return Fnreverse (rows);
}
```

agraven:
Closes #1246.

I think the way I ported `CHECK_RANGED_INTEGER` may not have been ideal, I think it's supposed to work on any `LispNumber`, so I should probably either change it to an impl on `LispNumber` or a macro.

I'm also not quite happy with the way the iteration over glyph rows directly copies the c code's way of increasing the value of one pointer until it surpasses the value of an end pointer, but constructing a vec from raw parts would require getting the indices of the rows in the arrays, and I'm not sure I want to add functions for getting both the indices and pointers of rows.

Also, I think we should consider moving the glyph types and functions into their own file, I'm thinking either `dispextern.rs` or `glyphs.rs`


rust_src/src/windows.rs
```diff
@@ -1536,4 +1592,131 @@ pub fn window_body_width_lisp(window: LispWindowLiveOrSelected, pixelwise: bool)
...
+            || windows_or_buffers_changed > 0
+            || buffer.clip_changed()
+            || buffer.prevent_redisplay_optimizations_p()
+            || window.is_outdated()
+        {
+            return Qnil;
+        }
+    }
+
+    let mut row = if let Some(first) = first {
```
shaleh: Why use `if let Some() ... else ...` instead of a `match`? I find that when there is an else, the `match` form is a better fit.

rust_src/src/windows.rs
```diff
@@ -1536,4 +1592,131 @@ pub fn window_body_width_lisp(window: LispWindowLiveOrSelected, pixelwise: bool)
...
+    let mut rows = Qnil;
+    while row.as_ptr() <= end_row.as_ptr() && row.enabled_p() && row.y + row.height < max_y {
+        if left {
+            let glyph = unsafe { &*row.glyphs[TEXT_AREA as usize] };
+            let width = if inverse {
+                glyph.pixel_width as i32
+            } else {
+                window_width - glyph.pixel_width as i32
+            };
+            rows = (LispObject::cons(width, row.y + row.height - subtract), rows).into();
```
shaleh: The last line of the if and else are the same. Can this be moved to be common after the if/else?

I think you can write `rows = ((width, row.y + row.height - subtract), rows).into();`. No?

rust_src/src/numbers.rs
```diff
@@ -158,6 +159,27 @@ impl IsLispNatnum for EmacsInt {
     }
 }
 
+pub trait CheckRange {
+    fn check_range(self, from: EmacsInt, to: EmacsInt);
+}
+
+impl CheckRange for EmacsInt {
```
shaleh: Instead of making this a trait, why not have it be a simple function?

```
pub fn check_range(num: impl Into<EmacsInt>, from: impl Into<EmacsInt>, to: impl Into<EmacsInt>) {
    let num: EmacsInt = num.into();
    let from: EmacsInt = from.into();
    let to: EmacsInt = to.into();
    if !(from <= self && self <= to) {
        args_out_of_range!(
            self,
            if from < 0 && from < MOST_NEGATIVE_FIXNUM {
                MOST_NEGATIVE_FIXNUM
            } else {
                from
            },
            cmp::min(to, MOST_POSITIVE_FIXNUM)
        )
    }
}
```

Since the arguments are all generic, you do not need to call `into` everywhere. Would any type implement this function differently?

rust_src/src/numbers.rs
```diff
@@ -1,6 +1,7 @@
 //! Functions operating on numbers.
 
 use rand::{Rng, SeedableRng, StdRng};
```
shaleh: While you are here, can you move the `rand` import to its own line after the `std` imports?

shaleh:
A `glyphs.rs` sounds good to me.

rust_src/src/windows.rs
agraven: I was on the fence on which to use, but after rewriting it I can see `match` is the nicer option

rust_src/src/numbers.rs
agraven: Sure thing.

rust_src/src/windows.rs
agraven: Yep, you can. My mistake when trying to do something like this was calling `.into()` on the inner tuple.

rust_src/src/numbers.rs
agraven: That's definitely the nicest way of doing it. Thanks for the pointer.

shaleh: Hat tip to @nickdrozd who has been spear heading this lately.

rust_src/src/windows.rs
```diff
@@ -1536,4 +1539,133 @@ pub fn window_body_width_lisp(window: LispWindowLiveOrSelected, pixelwise: bool)
...
+            };
+            rows = ((width, row.y + row.height - subtract), rows).into();
+        } else {
+            let width = if inverse {
+                window_width - row.pixel_width
+            } else {
+                row.pixel_width
+            };
+            rows = ((width, row.y + row.height - subtract), rows).into();
+        }
```
shaleh:
```
        let width = if left {
            let glyph = unsafe { &*row.glyphs[TEXT_AREA as usize] };
            if inverse {
                glyph.pixel_width as i32
            } else {
                window_width - glyph.pixel_width as i32
            }
        } else {
            if inverse {
                window_width - row.pixel_width
            } else {
                row.pixel_width
            }
        };
        rows = ((width, row.y + row.height - subtract), rows).into();
```

This is a powerful Rust idiom that is a little foreign to those not accustomed to "everything is an expression" style.

agraven: Ah yes, there's a lot of opportunities for tidying things up with expressiveness that are easy to miss. Fixed.

shaleh:
The `glyphs.rs` transition should be its own PR. It helps when we are `git bisect` looking for bug culprits.

shaleh:
This is looking close. I need to get the C and Rust side by side to make sure everything is right. If someone else can get to it before I do, great.

rust_src/src/windows.rs
```diff
@@ -1536,4 +1592,132 @@ pub fn window_body_width_lisp(window: LispWindowLiveOrSelected, pixelwise: bool)
...
+pub fn window_lines_pixel_dimensions(
+    window: LispWindowLiveOrSelected,
+    first: Option<LispNumber>,
+    last: Option<LispNumber>,
+    body: bool,
+    inverse: bool,
+    left: bool,
+) -> LispObject {
+    let window: LispWindowRef = window.into();
+    let buffer: LispBufferRef = window.contents.into();
```
shaleh: This is good. The order we check types is important. If we checked the buffer before the window the wrong error can be raised.

rust_src/src/windows.rs
```diff
@@ -510,6 +518,22 @@ impl LispWindowRef {
...
+    /// Return the bottom boundary y-position for text lines in window W.
+    /// This is the first y position at which a line cannot start.
+    /// It is relative to the top of the window.
+    ///
+    /// This is the height of W minus the height of a mode line, if any.
+    pub fn text_bottom_y(mut self) -> i32 {
+        self.pixel_height
+            - self.bottom_divider_width()
+            - self.scroll_bar_area_height()
+            - if self.wants_mode_line() {
```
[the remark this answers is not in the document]
agraven: I'm pretty sure there isn't a way of doing it in the standard library, but I suppose it would be possible to implement a method on bool that returns one of two values depending on the bools value. Something like `predicate.true_or(value, false_value)`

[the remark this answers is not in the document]
agraven: This is a case where I miss C's ternary operator. It's perfectly suited for simple checks like this

rust_src/src/windows.rs
```diff
@@ -1536,4 +1592,132 @@ pub fn window_body_width_lisp(window: LispWindowLiveOrSelected, pixelwise: bool)
...
+            }
+        } else {
+            if inverse {
+                window_width - row.pixel_width
+            } else {
+                row.pixel_width
+            }
+        };
+        rows = ((width, row.y + row.height - subtract), rows).into();
+        let ptr = unsafe { row.as_mut().add(1) };
```
shaleh: The incrementing of the contained pointer feels like functionality the `ExternalPtr` type should make easier. These two lines could be something like `row.ptr_add(1)`. Maybe a follow on PR?

shaleh: I opened #1317

PR #1297, opened 2019-01-29
port nconc

shanavas786:


shaleh:
When working on porting a DEFUN I find it is helpful to leave the original behind, copy it, and name the Rust port "rust-foo". Now, I can write the unit tests against the C version.

```
(ert-deftest test-nconc ()
  (should (equal (nconc nil) (rust-nconc nil)))
  (should (equal (nconc 5) (rust-nconc 5)))
  (should (equal (nconc '(1 . 2) 3) (rust-nconc '(1 . 2) 3))) ; => (1 . 3)
  (should (equal (nconc '(1 . 2) '(3 4 . 5) 6) (rust-nconc '(1 . 2) '(3 4 . 5) 6))) ; => (1 3 4 . 6)
)
```

Once that works, I remove the C version and rewrite the tests to have expected values.


While `nconc` may look like a fold. It really isn't. The key is the end of the C loop:

```
      tem = args[argnum + 1];
      Fsetcdr (tail, tem);
      if (NILP (tem))
       args[argnum + 1] = tail;
```

The loop mutates the array it is iterating over. So the next `elt` might be the `tail` that was just updated. It is not building a completely new Cons list. Only `val` is outside the loop. All of the other items are local to each iteration. The ported Rust looks more like this which still looks like the original C because of the mutating nature of the code. The alternative would be a stack implementation that either pop's elements or puts on a new one.

```
    let mut val = Qnil;

    let len = args.len();

    for i in 0..len {
        let elt = args[i];

        if elt.is_nil() {
            continue;
        }

        if val.is_nil() {
            val = elt;
        }

        // every item is checked at least this far
        if (i + 1) == len {
            break;
        }

        // now the Cons check
        let cons: LispCons = elt.into();

        // FOR_EACH_TAIL only enables the circular checks.
        let mut tail = cons.iter_tails(LispConsEndChecks::off, LispConsCircularChecks::on).last().unwrap();
        let next = args[i + 1]; // peek at the next element in the array
        tail.set_cdr(next);  // since tail is a Cons we can use its method.
        if next.is_nil() {
            args[i + 1] = tail.into(); // this mutates the next element to be examined.
        }
    }

    val
```

shaleh:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Rearrangement.html

shaleh:
Since I kinda ghostwrote this, I will abstain from the reviews.

daichimukai:
kick travis

PR #1298, opened 2019-01-29
Throw an error if pkg-config is missing

### LINKED ISSUE #611: Add about `pkgconfig` thing to README?
I think this related to issue #155 , #189.

I don't know details about this.

I've got error message like this when I trying ./configure
>Maybe some development libraries/packages are missing?
>If you don't want to link with them give
>     --with-gnutls=no
>as options to configure

I found [this](https://stackoverflow.com/questions/41898925/compiling-filezilla-on-osx) on SO, and I can run remacs now.

`brew install pkgconfig`
`pkg-config --cflags gnutls`

agraven:
Explicitly check whether `pkg-config` can be found in path. I have a suspicion this check is superfluous, as `PKG_PROG_PKG_CONFIG` may already check for this. I have yet to confirm this, though.
 
This should close #611.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -197,6 +197,12 @@ archlibdir='${libexecdir}/remacs/${version}/${configuration}'
 etcdocdir='${datadir}/remacs/${version}/etc'
 gamedir='${localstatedir}/games/remacs'
 
+dnl Check for the presence of pkg-config
+AC_CHECK_TOOL([PKG_CONFIG], [pkg-config])
+if test x$PKG_CONFIG = x ; then
+  AC_MSG_ERROR([pkg-config not in path. You may need to install it.])
+fi
+
 dnl Special option to disable the most of other options.
 AC_ARG_WITH(all,
 [AS_HELP_STRING([--without-all],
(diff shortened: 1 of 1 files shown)


agraven:
As I mentioned in #611, I've confirmed that the existing invocation of `PKG_PROG_PKG_CONFIG` makes this check unnecessary. I'm inclined to close this PR, and the original issue. Thoughts?

shaleh:
An experiment.

Move pkg_config aside (or rename it). Run remacs configure. What happens?

shaleh:
On MacOS, when I `mv pkg-config hidden-pkg-config` the configure script does not fail because it cannot find `pkg-config`. Instead if fails because it cannot find the GnuTLS package. So it appears that we do need some improvement here.

agraven:
And is this before or after the addition of this change?

agraven:
It fails with the change, as expected. I realized that the `AC_CHECK_TOOL` invocation is redundant, and that `PKG_PROG_PKG_CONFIG` fails without throwing an error by default.

configure.ac
```diff
@@ -1592,6 +1592,9 @@ pre_PKG_CONFIG_CFLAGS=$CFLAGS
 pre_PKG_CONFIG_LIBS=$LIBS
 
 PKG_PROG_PKG_CONFIG(0.9.0)
```
shaleh: Perfect. This is a small change that the Emacs folk might be interested in as well.

Issue #1287, opened 2019-01-25
Wrong type argument: windowp, nil


manfredlotz:
In org mode when defining a new task and then calling org-schedule I get the message
    Wrong type argument: windowp, nil

Instead a calendar buffer should show up where I can select the schedule date I want.

manfredlotz:
In *Messages* the full message text is:

```
window-edges: Wrong type argument: windowp, nil
```

shaleh:
https://www.gnu.org/software/emacs/manual/html_node/eintr/debug_002don_002dentry.html may be able to help you identify the culprit.

shaleh:
Any chance this is related to your window fix @agraven?

agraven:
I doubt it, the error is `windowp` and not `window-valid-p`

shaleh:
I think we only have two candidates on the Rust side: From<LispObject> for LispWindowRef and window::as_minibuffer_or_error.

manfredlotz:
I tried to debug but didn't find anything which was helpful

agraven:
Can you give us the output in the `*Backtrace*` buffer after setting `debug-on-error` to `t` and then reproducing the error?

manfredlotz:
When first debugging I tried (debug-on-enrty org-schedule). I had to type 'd' to often.

Ok, no only debug-on-error t and I get:

```
Debugger entered--Lisp error: (wrong-type-argument windowp nil)
  window-body-height(#<window 3 on *Backtrace*> t)
  window-edges(nil t)
  window-inside-edges()
  calendar-update-mode-line()
  calendar-generate-window(1 2019)
  calendar-basic-setup(nil)
  calendar()
  org-read-date(nil to-time nil nil nil nil)
  org-add-planning-info(scheduled nil closed)
  #f(compiled-function () #<bytecode 0x2af88cd>)()
  org--deadline-or-schedule(nil scheduled nil)
  org-schedule(nil)
  funcall-interactively(org-schedule nil)
  call-interactively(org-schedule nil nil)
  command-execute(org-schedule)
```

nickdrozd:
It also triggers when doing autocomplete in the minibuffer, which actually makes this a pretty severe usability issue.

agraven:
Looks like the culprit is `window-body-height`. Is the error reproducible in versions prior to the function being ported?

shaleh:
`window-body-height` is quite simple. Also, the window is clearly not nil. This is quite interesting.

[...]

PR #1301, opened 2019-01-30
Fix translation of nils to windows.

shaleh:
Closes: #1287

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1,6 +1,6 @@
 //! Functions operating on windows.
 
-use std::{cmp, ptr};
+use std::{cmp, fmt, ptr};
 
 use libc::c_int;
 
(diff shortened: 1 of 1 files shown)


rust_src/src/windows.rs
```diff
@@ -75,14 +75,6 @@ impl LispWindowRef {
         LispGlyphMatrixRef::new(self.current_matrix)
     }
 
-    pub fn prev(self) -> Self {
-        self.prev.into()
-    }
-
-    pub fn next(self) -> Self {
-        self.next.into()
-    }
```
agraven: Could these not be preserved but changed to `Option<Self>`?

nickdrozd:
Is there a Lisp-level test that could be added to verify the behavior (ie that fails before this fix and passes after)?

rust_src/src/windows.rs
shaleh: In theory. Or turned into iterators.

But for now, all we need is the EQ check.

shaleh:
I do not know how to write it. But yes, this should be testable.

Split-window can trigger it. The requirement is a window that has prev set to nil instead of a valid window. The test needs to cleanup afterwards and not leave dangling windows.

rust_src/src/windows.rs
shaleh: When dealing with LispObject the right mental model is to treat it like a  Option<LispVal>. Unless the code has already ensured it cannot be nil a LispObject must be treated as None until proven otherwise.

shaleh:
Because it is a significant failure I am inclined to merge it and then add tests.

PR #1302, opened 2019-01-30
Add LispSymbolOrString type

### LINKED ISSUE #1209: Implement LispSymbolOrString type

agraven:
Making a draft to make sure I get the type right before I make functions use it. I'm not exactly sure which conversions should do the type check and which should skip them.

Also, when converting this to a LispSymbolRef, should I throw an error if the value is a string or something else?

EDIT: Just discovered `Fmake_sym` exists. Should I use that for converting strings to symbols?

Will close #1209.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -635,7 +635,7 @@ impl From<LispBufferOrName> for Option<LispBufferRef> {
             LispBufferOrName::Name(name) => {
                 let tem = unsafe { Vbuffer_alist }
                     .iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off)
-                    .find(|&item| string_equal(car(item), name.into()));
+                    .find(|&item| string_equal(car(item), name));
 
                 cdr(tem.into()).as_buffer()
             }
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -609,7 +609,7 @@ impl From<LispBufferOrName> for Option<LispBufferRef> {
             LispBufferOrName::Name(name) => {
                 let tem = unsafe { Vbuffer_alist }
                     .iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off)
-                    .find(|&item| string_equal(car(item).into(), name.into()));
+                    .find(|&item| string_equal(car(item), name));
 
                 cdr(tem.into())
             }
@@ -876,7 +876,7 @@ fn get_truename_buffer_1(filename: LispSymbolOrString) -> LispObject {
     LiveBufferIter::new()
         .find(|buf| {
             let buf_truename = buf.truename();
-            buf_truename.is_string() && string_equal(buf_truename.into(), filename)
+            buf_truename.is_string() && string_equal(buf_truename, filename)
         })
         .into()
 }
@@ -950,7 +950,7 @@ pub fn get_file_buffer(filename: LispStringRef) -> Option<LispBufferRef> {
     } else {
         LiveBufferIter::new().find(|buf| {
             let buf_filename = buf.filename();
-            buf_filename.is_string() && string_equal(buf_filename.into(), filename.into())
+            buf_filename.is_string() && string_equal(buf_filename, filename)
         })
     }
 }
@@ -1117,7 +1117,7 @@ pub fn erase_buffer() {
 /// is first appended to NAME, to speed up finding a non-existent buffer.
 #[lisp_fn(min = "1")]
 pub fn generate_new_buffer_name(name: LispStringRef, ignore: LispObject) -> LispStringRef {
-    if (ignore.is_not_nil() && string_equal(name.into(), ignore.into()))
+    if (ignore.is_not_nil() && string_equal(name, ignore))
         || get_buffer(LispBufferOrName::Name(name.into())).is_none()
     {
         return name;
@@ -1144,7 +1144,7 @@ pub fn generate_new_buffer_name(name: LispStringRef, ignore: LispObject) -> Lisp
         let mut s = format!("<{}>", suffix_count);
         local_unibyte_string!(suffix, s);
         let candidate = unsafe { concat2(basename, suffix) };
-        if string_equal(candidate.into(), ignore.into())
+        if string_equal(candidate, ignore)
             || get_buffer(LispBufferOrName::Name(candidate)).is_none()
         {
             return candidate.into();
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -1143,11 +1143,11 @@ pub fn generate_new_buffer_name(name: LispStringRef, ignore: LispObject) -> Lisp
     loop {
         let mut s = format!("<{}>", suffix_count);
         local_unibyte_string!(suffix, s);
-        let candidate = unsafe { concat2(basename, suffix) };
+        let candidate = unsafe { concat2(basename, suffix) }.force_string();
         if string_equal(candidate, ignore)
-            || get_buffer(LispBufferOrName::Name(candidate)).is_none()
+            || get_buffer(LispBufferOrName::Name(candidate.into())).is_none()
         {
-            return candidate.into();
+            return candidate;
         }
         suffix_count += 1;
     }
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -22,8 +22,8 @@ use crate::{
     lists::{car, cdr, list, member},
     lists::{LispConsCircularChecks, LispConsEndChecks},
     marker::{build_marker, marker_buffer, marker_position_lisp, set_marker_both, LispMarkerRef},
-    multibyte::LispStringRef,
     multibyte::{multibyte_length_by_head, string_char},
+    multibyte::{LispStringRef, LispSymbolOrString},
     numbers::MOST_POSITIVE_FIXNUM,
     remacs_sys::{
         allocate_misc, bset_update_mode_line, buffer_local_flags, buffer_local_value,
@@ -609,7 +609,7 @@ impl From<LispBufferOrName> for Option<LispBufferRef> {
             LispBufferOrName::Name(name) => {
                 let tem = unsafe { Vbuffer_alist }
                     .iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off)
-                    .find(|&item| string_equal(car(item), name));
+                    .find(|&item| string_equal(car(item).into(), name.into()));
 
                 cdr(tem.into())
             }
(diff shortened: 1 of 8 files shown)


shaleh:
It will help to identify the use cases in the code where this type would help.

Are the functions that are passed a Symbol or a String and want a Symbol? The string is converted to a Symbol somehow along the way.

Are there functions that are passed a Symbol or String and need a String? The symbol's name or value is substituted.

What are the error cases in the code?

Once these are identified the rest of the design should fall out. Or it may not make sense to add the type after all.

rust_src/src/obarray.rs
```diff
@@ -204,9 +205,10 @@ pub extern "C" fn intern_driver(
 /// A second optional argument specifies the obarray to use;
 /// it defaults to the value of `obarray'.
 #[lisp_fn(min = "1")]
-pub fn intern_soft(name: LispObject, obarray: Option<LispObarrayRef>) -> LispObject {
+pub fn intern_soft(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> LispObject {
     let obarray = obarray.unwrap_or_else(LispObarrayRef::global);
     let tem = obarray.lookup(name);
+    let name: LispObject = name.into();
```
shaleh: The only reason to make a LispObject is to answer "is symbol and does my name match". This seems like something the new `LispSymbolOrString` could answer instead of falling back to LispObject.

rust_src/src/obarray.rs
```diff
@@ -60,9 +60,10 @@ impl LispObarrayRef {
     /// symbol with that name in this `LispObarrayRef`. If Emacs is loading Lisp
     /// code to dump to an executable (ie. `purify-flag` is `t`), the symbol
     /// name will be transferred to pure storage.
-    pub fn intern(&self, string: LispStringRef) -> LispObject {
+    pub fn intern(&self, string: impl Into<LispSymbolOrString>) -> LispObject {
         let string = string.into();
         let tem = self.lookup(string);
+        let string: LispObject = string.into();
```
shaleh: Instead of doing this here. What if the elseif/else below is re-written?

```
} else {
    let string_copy: LispObject = if unsafe { globals.Vpurify_flag }.is_not_nil() {
        // When Emacs is running lisp code to dump to an executable,
        // make use of pure storage.
        unsafe { Fpurecopy(string.into()) }
    } else {
        string.into()
    };
    intern_driver(string_copy, self.into(), tem)
}
```

rust_src/src/multibyte.rs
```diff
@@ -312,6 +313,59 @@ impl LispObject {
...
+impl From<LispSymbolOrString> for LispStringRef {
+    fn from(s: LispSymbolOrString) -> Self {
+        match s.0.as_symbol() {
+            Some(symbol) => symbol.symbol_name().into(),
+            None => s.0.into(),
+        }
+    }
+}
+
+impl From<LispStringRef> for LispSymbolOrString {
```
shaleh: Another option is to make `LispSymbolOrString` an enum containing `String(LispStringRef)` and `Symbol(LispSymbolRef)`. It is not clear which option is preferable. Something to experiment with.

rust_src/src/multibyte.rs
```diff
@@ -312,6 +313,59 @@ impl LispObject {
     }
 }
 
+#[derive(Copy, Clone, PartialEq)]
+pub struct LispSymbolOrString(LispObject);
+
+impl LispSymbolOrString {
+    pub fn is_string(self) -> bool {
```
shaleh: If this were an enum, this would be a match call.

rust_src/src/buffers.rs
```diff
@@ -1144,7 +1144,7 @@ pub fn generate_new_buffer_name(name: LispStringRef, ignore: LispObject) -> Lisp
         let mut s = format!("<{}>", suffix_count);
         local_unibyte_string!(suffix, s);
         let candidate = unsafe { concat2(basename, suffix) };
-        if string_equal(candidate, ignore)
+        if string_equal(candidate.into(), ignore.into())
             || get_buffer(LispBufferOrName::Name(candidate)).is_none()
```
shaleh: See of `LispBufferOrName` uses the `Name` enum value. We know `candidate` is a string at this point. No need to go through the tests for is Symbol.

shaleh:
I left some comments in the PR.

rust_src/src/dired_unix.rs
```diff
@@ -108,7 +108,7 @@ impl LispObjectExt for LispObject {
         if self.is_nil() {
             "NOTstring".to_string()
         } else {
-            let idf_sym_s = self.symbol_or_string_as_string();
+            let idf_sym_s: LispStringRef = LispSymbolOrString::from(*self).into();
```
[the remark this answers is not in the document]
shaleh: How about `self.symbol_or_string().into()`?

agraven:
Is possible/viable that the arguments for `string_equal` and the like have the type `Into<LispSymbolOrString>`? It would remove the need for quite a few `into()` calls

agraven:
I'm inclined to do it the second way, as it means that I won't have to add into() to (more or less) every existing invocation of string_equal.

agraven:
Oh, that was a mistake, I meant `Fmake_symbol`, though looking at it further it looks like `obarray::intern` is what I want.

rust_src/src/buffers.rs
agraven: What tests for is symbol are you referring to?

shaleh: Object (which we know is a string) -> LispSymbolOrString -> LispStringRef 

In the From<LispObject> there is this: `if o.is_string() || o.is_symbol() {` and then in the From<LispSymbolOrString> for LispStringRef there is this: `match s.0.as_symbol()`. Even after passing `is_string` we again try to make the value a symbol. And enum type would avoid some of this duplication or potentially do it more cheaply.

shaleh:
```
diff --git a/rust_src/src/lib.rs b/rust_src/src/lib.rs
index 896891097e..32d46a8aad 100644
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -18,6 +18,7 @@
 #![feature(slice_patterns)]
 #![feature(specialization)]
 #![feature(stmt_expr_attributes)]
+#![feature(type_alias_enum_variants)]
 #![feature(untagged_unions)]
 
 extern crate errno;
diff --git a/rust_src/src/multibyte.rs b/rust_src/src/multibyte.rs
index c275c513cc..2d02a40688 100644
--- a/rust_src/src/multibyte.rs
+++ b/rust_src/src/multibyte.rs
@@ -315,57 +315,77 @@ impl LispObject {
 
     // We can excuse not using an option here because extracting the value checks the type
     pub fn as_symbol_or_string(self) -> LispSymbolOrString {
-        LispSymbolOrString(self)
+        self.into()
     }
 }
 
 #[derive(Copy, Clone, PartialEq)]
-pub struct LispSymbolOrString(LispObject);
+pub enum LispSymbolOrString {
+    String(LispStringRef),
+    Symbol(LispSymbolRef),
+}
 
 impl LispSymbolOrString {
     pub fn is_string(self) -> bool {
-        self.0.is_string()
+        match self {
+            LispSymbolOrString::String(_) => true,
+            _ => false,
+        }
     }
 
     pub fn is_symbol(self) -> bool {
-        self.0.is_symbol()
+        match self {
+            LispSymbolOrString::Symbol(_) => true,
+            _ => false,
+        }
     }
 }
 
 impl From<LispSymbolOrString> for LispObject {
     fn from(s: LispSymbolOrString) -> Self {
-        s.0
+        match s {
+            LispSymbolOrString::String(s) => s.into(),
+            LispSymbolOrString::Symbol(sym) => sym.into(),
+        }
     }
 }
 
 impl From<LispSymbolOrString> for LispStringRef {
     fn from(s: LispSymbolOrString) -> Self {
-        match s.0.as_symbol() {
-            Some(symbol) => symbol.symbol_name().into(),
-            None => s.0.into(),
+        match s {
+            LispSymbolOrString::String(s) => s,
+            LispSymbolOrString::Symbol(sym) => sym.symbol_name().into(),
         }
     }
 }
 
 impl From<LispStringRef> for LispSymbolOrString {
     fn from(s: LispStringRef) -> Self {
-        Self(s.into())
+        Self::String(s)
     }
 }
 
 impl From<LispSymbolOrString> for LispSymbolRef {
     fn from(s: LispSymbolOrString) -> Self {
-        match s.0.as_symbol() {
-            Some(symbol) => symbol,
-            None => LispObarrayRef::global().intern(s).into(),
+        match s {
+            LispSymbolOrString::String(s) => LispObarrayRef::global().intern(s).into(),
+            LispSymbolOrString::Symbol(sym) => sym
         }
     }
 }
 
+impl From<LispSymbolRef> for LispSymbolOrString {
+    fn from(s: LispSymbolRef) -> Self {
+        Self::Symbol(s)
+    }
+}
+
 impl From<LispObject> for LispSymbolOrString {
-        if o.is_string() || o.is_symbol() {
-            Self(o)
+        if let Some(s) = o.as_string() {
+            Self::String(s)
+        } else if let Some(sym) = o.as_symbol() {
+            Self::Symbol(sym)
         } else {
             wrong_type!(Qstringp, o)
         }
@@ -374,7 +394,7 @@ impl From<LispObject> for LispSymbolOrString {
 
 impl PartialEq<LispObject> for LispSymbolOrString {
     fn eq(&self, other: &LispObject) -> bool {
-        self.0.eq(*other)
+        (*other).eq(LispObject::from(*self))
     }
 }
 ```

**Comment by agraven** [comment]
The enum way of doing it definitely seems smarter, as the type of the value is known as soon as a `LispSymbolOrString` is instantiated, which, as you mentioned, makes some conversions either cheaper or free.

File: rust_src/src/multibyte.rs [hunk 7]
```diff
@@ -310,6 +312,91 @@ impl LispObject {
     pub fn empty_unibyte_string() -> LispStringRef {
         LispStringRef::from(unsafe { empty_unibyte_string })
     }
+
+    // We can excuse not using an option here because extracting the value checks the type
+    // TODO: this is false with the enum model, change this
```
shaleh: Still valid right? The LispObject will either pass `as_string()`, `as_symbol()`, or return `error!()` right?

agraven: It will, but its behavior has changed slightly. Now it's the during the conversion from `LispObject` to `LispSymbolOrString` the potential error happens, instead of during the conversion from `LispSymbolOrString` to `LispStringRef` or `LispSymbolRef`.

shaleh: Order of errors matters unfortunately. Trace the execution to be sure the Rust code fails where the C code would have. Which Rust function the error happens inside of is not important but when is.

agraven: As it's currently invoked errors happen when expected, my primary concern is that it doesn't follow the convention of return an option like the other `as_type()` methods, which may lead to misuse. Leaving it as it is would be fine, but it could potentially cause confusion or tech-debt down the line.

shaleh: Ah, yes. Our use of `as_foo()` naming is not really inline with the Rust norm.

https://rust-lang-nursery.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

Clippy has a warning about this as well based on the type of `self` but I cannot find it at that moment.

rust_src/src/obarray.rs
shaleh: @agraven what are your thoughts on this refactor?

agraven: After looking at it again it does seem nicer. I think I'll include it.

PR #1304, opened 2019-01-31
Refactor buffer_objfwd support.

shaleh:
Adds `as_buffer_objfwd()` which returns an `Option<Lisp_Buffer_Objfwd>`.
This simplifies the common use of the object. Remove `if_chain` now
that the code is simpler.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -3,6 +3,7 @@
 use std::sync::Mutex;
 use std::{self, mem, ptr};
 
+use field_offset::FieldOffset;
 use libc::{self, c_char, c_int, c_uchar, c_void, ptrdiff_t};
 
 use rand::{Rng, StdRng};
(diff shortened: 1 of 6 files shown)


rust_src/src/symbols.rs
```diff
@@ -6,9 +6,10 @@ use std::fmt::{Debug, Formatter};
 use remacs_macros::lisp_fn;
 
 use crate::{
-    buffers::LispBufferLocalValueRef,
+    buffers::per_buffer_idx_from_field_offset,
```
shanavas786: these are unused right ?

rust_src/src/buffers.rs
```diff
@@ -1153,10 +1154,16 @@ pub fn generate_new_buffer_name(name: LispStringRef, ignore: LispObject) -> Lisp
     }
 }
 
+pub unsafe fn per_buffer_idx_from_field_offset(
+    offset: FieldOffset<Lisp_Buffer, LispObject>,
+) -> isize {
+    unsafe { *offset.apply_ptr_mut(&mut buffer_local_flags) }.to_fixnum_unchecked() as isize
```
shanavas786: can't we use `per_buffer_idx` here ?

shaleh: With effort. The key issue is the difference in pointer math in Rust versus C.

In C, the idiom is `foo = ptr + num_bytes;` but in Rust the code looks like this `foo = ptr.offset(num_things);`. The difference is the C math is done in raw bytes. The Rust equivalent in C looks like this `PtrType *foo = ptr + (num_things * sizeof(PtrType));`.

Despite the variable name `offset`, it should really be `count`. I will update that in my refactor PR. I forgot last night.

The `Lisp_Fwd` type in Rust uses the `FieldOffset` type so the `apply_ptr` method is more idiomatic. Otherwise we would reimplement it by using `get_byte_offset()` and dividing it by size of `LispObject` to pass to `ptr.offset()`.

rust_src/src/symbols.rs
shaleh: yeah, I pull this PR off of your branch after hacking on it locally.

rust_src/src/buffers.rs
shanavas786: Got it.

PR #1311, opened 2019-02-01
update readme

brotzeit:


README.md
```diff
@@ -235,11 +235,11 @@ different.
 # Progress
 
 At this point we focus on porting lisp functions from C to Rust.
-Currently there are 555 functions in Rust and 911 in C (January 2019).
+Currently there are 586 functions in Rust and 879 in C (February 2019).
```
daichimukai: Out of curiosity, had we lost one lisp function? or just a typo?

brotzeit: Yeah those numbers weren't and still aren't correct. I should take a closer look at it. I just found out there are only 580 lisp functions in rust =)

agraven:
Solid progress :tada:

PR #1312, opened 2019-02-01
Update to latest rand crate

### LINKED ISSUE #1242: Update to latest `rand` crate
remacs currently uses `StdRng` because `ThreadRng` crashes for us. The latest version of the crate no longer exposes the `reseed` functionality we are using without using an implementation of the `ReseedingRng`.

Do we need it? Can we simplify this?
Does the newer `ThreadRng` work? Can we make it work?

daichimukai:
My strategy is:

* Use `ThreadRng` if we can. This is automatically reseeded after generating 32MiB random data. I have not tested this code on macOS yet. I just checked only it could be built using travis. May crash on runtime. Please help testing if you are interested in :wink:
* In `number.rs`, there are something to do with the seed, which seems to not changed for `ThreadRng`. We continue to use static variable.

Closes #1242.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5116,6 +5116,11 @@ LIB_REMACS="$RUST_DEPS"
 
 LDFLAGS_REMACS="-L../$srcdir/rust_src/target/\${CARGO_BUILD_DIR}"
 
+# On macOS, rand crate needs '_SecRandomCopyBytes' symbol in Security framework.
+if test "${opsys}" = "darwin"; then
+LDFLAGS_REMACS="$LDFLAGS_REMACS -framework Security"
+fi
+
 LDFLAGS="$LDFLAGS $LDFLAGS_REMACS"
 
 AC_SUBST(LIB_REMACS)
--- a/rust_src/src/numbers.rs
+++ b/rust_src/src/numbers.rs
@@ -3,7 +3,7 @@
 use std::cmp;
 use std::sync::Mutex;
 
-use rand::{Rng, SeedableRng, StdRng};
+use rand::{rngs::StdRng, FromEntropy, Rng, SeedableRng};
 
 use remacs_macros::lisp_fn;
 
@@ -18,7 +18,7 @@ use crate::{
 };
 
 lazy_static! {
-    static ref RNG: Mutex<StdRng> = Mutex::new(StdRng::new().unwrap());
+    static ref RNG: Mutex<StdRng> = Mutex::new(StdRng::from_entropy());
 }
 
 // Largest and smallest numbers that can be represented as fixnums in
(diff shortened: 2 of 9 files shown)


daichimukai:
I finally tested this patch on my old macbook air. It seems going happy. The following code could be evaluated without crash.

```elisp
(generate-new-buffer " tmp")
;; generate-new-buffer-name should use thread_rng() in this situation.
(generate-new-buffer-name " tmp")
```

configure.ac
```diff
@@ -5113,6 +5113,10 @@ LIB_REMACS="$RUST_DEPS"
 
 LDFLAGS_REMACS="-L../$srcdir/rust_src/target/\${CARGO_BUILD_DIR}"
```
shaleh: Can you add a comment here for why `Security` is needed?

rust_src/src/numbers.rs
```diff
@@ -294,10 +295,13 @@ pub fn number_or_marker_p(object: LispObject) -> bool {
 pub fn random(limit: LispObject) -> LispObject {
     let mut rng = RNG.lock().unwrap();
     if limit.is_t() {
-        *rng = StdRng::new().unwrap();
+        *rng = StdRng::from_entropy();
     } else if let Some(s) = limit.as_string() {
-        let values: Vec<usize> = s.as_slice().iter().map(|&x| x as usize).collect();
-        rng.reseed(&values);
```
shaleh: Do you need the `seed` array here? Can you not pass in a 32 element slice from `values` like `&values[0..32]`?

daichimukai: As I recall, compiler complained that `from_seed()` had expected `[u8; 32]` but found `&[u8]`. I couldn't get better way to cast these types...

shaleh: I see what you mean. Oh well.

daichimukai:
Resolved conflict and added some tests for elisp functions which use rand crate.

shaleh:
@daichimukai are you able to update this?

daichimukai:
@shaleh I don't have a smart idea yet on how to make `[u8; 32]` from `LispStringRef`. Deleting `values` vector may better?

```rust
let mut seed = [0; 32];
let size = s.len_bytes() as usize;
let slice = s.as_slice();
for i in 0..32 {
    seed[i] = slice[i % size];
}
*rng = StdRng::from_seed(seed);
```

shaleh:
The seed needs an array. No big deal. If you add the comment in configure.ac this is ready I think.

daichimukai:
I did that when i have rebased. i forgot to mention it.

shaleh:
My mistake @daichimukai I missed it.

daichimukai:
Thanks.

PR #1319, opened 2019-02-03
Fix error on describe-function for adviced builtin function

### LINKED ISSUE #822: (describe-function 'rename-buffer) errors
This seems to be due to `subr-lang`. See https://github.com/Wilfred/remacs/pull/821

daichimukai:
The aim of this PR is to resolve #822.

As pointed out in comments on previous issues, `subr-lang` gets `wrong-type-argument` for adviced builtin function because `subr-lang` expects an argument is a subroutine but adviced function is no longer regarded as subroutine.

This small patch tries to fix this error giving to `subr-lang` the real function definition being detached from advice part. This seems already prepared in the function concerned using `help-fns--analyze-function`.

### DIFF
--- a/lisp/help-fns.el
+++ b/lisp/help-fns.el
@@ -677,7 +677,7 @@ Returns a list of the form (REAL-FUNCTION DEF ALIASED REAL-DEF)."
 	;; but that's completely wrong when the user used load-file.
 	(princ (format-message " in `%s'"
                                (if (eq file-name 'C-source)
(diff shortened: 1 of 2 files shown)


daichimukai:
@jaccarmac What I intended in changes of test is to turn the meaning of this test from "check `describe-function` fails with `rename-buffer`" into "check `describe-function` goes well". I agree `car` is usually not adviced. I selected `car` and `rename-buffer` as representatives of non-adviced and adviced built-in function respectively.

PR #1320, opened 2019-02-03
Port rename-buffer.

jaccarmac:


rust_src/src/buffers.rs
```diff
@@ -1228,4 +1230,67 @@ pub extern "C" fn rust_syms_of_buffer() {
...
+            // equivalent to
+            // (rename-buffer (generate-new-buffer-name NEWNAME)).
+            if unique.is_nil() && (tem == current_buffer) {
+                return current_buffer.name_.into();
+            }
+            if unique.is_not_nil() {
+                generate_new_buffer_name(newname, current_buffer.name_)
+            } else {
+                error!("Buffer name `{}' is in use", newname)
+            }
```
shaleh: This can be restructured right?

```
           if unique.is_nil() {
               if tem == current_buffer {
                   return current_buffer.name_.into();
               }
               error!("Buffer name `{}' is in use", newname);
           } else { 
               generate_new_buffer_name(newname, current_buffer.name_)
           }
```

I think this version reads better.

```
           if unique.is_not_nil() {
               generate_new_buffer_name(newname, current_buffer.name_)
           } else { 
               if tem == current_buffer {
                   return current_buffer.name_.into();
               }
               error!("Buffer name `{}' is in use", newname);
           }
```

jaccarmac: /Technically/ this version has a little more overhead since we always generate a new buffer name in the common case. Think I'll quickly fix that as well.

PR #1321, opened 2019-02-03
Make the type guarantees in LispBufferOrName a little stronger.

jaccarmac:
(#1320, but with a slightly more strict `LispBufferOrName`)

This did require removing the derivation of `Eq` on the type, but no
changes outside the `buffers` module and trivial changes inside.

Maybe useful to the discussion in #1302 re: the benefits of wrapper
types vs. enums of various specificity.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -563,10 +563,10 @@ impl LispBufferLocalValueRef {
     }
 }
 
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq)]
 pub enum LispBufferOrName {
-    Buffer(LispObject),
-    Name(LispObject),
+    Buffer(LispBufferRef),
+    Name(LispStringRef),
 }
 
 impl LispBufferOrName {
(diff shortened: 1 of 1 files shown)


rust_src/src/buffers.rs
```diff
@@ -575,18 +577,18 @@ impl LispBufferOrName {
...
-            LispBufferOrName::Name(n) => n,
+            LispBufferOrName::Buffer(b) => b.into(),
+            LispBufferOrName::Name(n) => n.into(),
         }
     }
 }
 
 impl From<LispObject> for LispBufferOrName {
     fn from(v: LispObject) -> Self {
         if v.is_string() {
```
shaleh: Since you want the LispString type, why not use `if let Some(s) = v.as_string()` style? No need to check the string type twice which will happen with the `is_string` and then `Into<LispString>` calls.

Same with the LispBuffer below.

rust_src/src/buffers.rs
```diff
@@ -598,9 +600,9 @@ impl From<LispObject> for Option<LispBufferOrName> {
         if v.is_nil() {
             None
         } else if v.is_string() {
```
shaleh: `as_string` and `as_buffer` make sense here too.

rust_src/src/buffers.rs
```diff
@@ -610,11 +612,11 @@ impl From<LispObject> for Option<LispBufferOrName> {
 impl From<LispBufferOrName> for Option<LispBufferRef> {
     fn from(v: LispBufferOrName) -> Self {
         let buffer = match v {
-            LispBufferOrName::Buffer(b) => b,
+            LispBufferOrName::Buffer(b) => b.into(),
```
shaleh: Shouldn't this become `return Some(b)`?

jaccarmac:
Sounds good. Wanted to get that merged before tackling low-hanging fruit on this one, thanks for the early look.

PR #1323, opened 2019-02-03
Port condition-case

### LINKED ISSUE #1306: Port condition-case
```c
DEFUN ("condition-case", Fcondition_case, Scondition_case, 2, UNEVALLED, 0,
       doc: /* Regain control when an error is signaled.
Executes BODYFORM and returns its value if no error happens.
Each element of HANDLERS looks like (CONDITION-NAME BODY...)
where the BODY is made of Lisp expressions.

A handler is applicable to an error
if CONDITION-NAME is one of the error's condition names.
If an error happens, the first applicable handler is run.

The car of a handler may be a list of condition names instead of a
single condition name; then it handles all of them.  If the special
condition name `debug' is present in this list, it allows another
condition in the list to run the debugger if `debug-on-error' and the
other usual mechanisms says it should (otherwise, `condition-case'
suppresses the debugger).

When a handler handles an error, control returns to the `condition-case'
and it executes the handler's BODY...
with VAR bound to (ERROR-SYMBOL . SIGNAL-DATA) from the error.
\(If VAR is nil, the handler can't access that information.)
Then the value of the last BODY form is returned from the `condition-case'
expression.

See also the function `signal' for more info.
usage: (condition-case VAR BODYFORM &rest HANDLERS)  */)
  (Lisp_Object args)
{
  Lisp_Object var = XCAR (args);
  Lisp_Object bodyform = XCAR (XCDR (args));
  Lisp_Object handlers = XCDR (XCDR (args));

  return internal_lisp_condition_case (var, bodyform, handlers);
}
```

jubnzv:
Fix #1306

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -1330,8 +1330,8 @@ pub fn signal(error_symbol: LispObject, data: LispObject) -> ! {
 #[lisp_fn(min = "2", unevalled = "true")]
 pub fn condition_case(args: LispCons) -> LispObject {
     let (var, consq) = args.into();
-    let (bodyform, handlers_) = consq.into();
-    unsafe { internal_lisp_condition_case(var, bodyform, handlers_) }
+    let (bodyform, handlers) = consq.into();
+    unsafe { internal_lisp_condition_case(var, bodyform, handlers) }
 }
 
 include!(concat!(env!("OUT_DIR"), "/eval_exports.rs"));
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -21,8 +21,9 @@ use crate::{
     remacs_sys::{
         backtrace_debug_on_exit, build_string, call_debugger, check_cons_list, do_debug_on_call,
         do_one_unbind, eval_sub, find_symbol_value, funcall_lambda, funcall_subr, globals,
-        grow_specpdl, internal_catch, list2, maybe_gc, maybe_quit, record_in_backtrace,
-        record_unwind_save_match_data, signal_or_quit, specbind, COMPILEDP, MODULE_FUNCTIONP,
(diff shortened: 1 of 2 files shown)


rust_src/src/eval.rs
```diff
@@ -1301,4 +1302,36 @@ pub fn signal(error_symbol: LispObject, data: LispObject) -> ! {
...
+/// and it executes the handler's BODY...
+/// with VAR bound to (ERROR-SYMBOL . SIGNAL-DATA) from the error.
+/// \(If VAR is nil, the handler can't access that information.)
+/// Then the value of the last BODY form is returned from the `condition-case'
+/// expression.
+///
+/// See also the function `signal' for more info.
+/// usage: (condition-case VAR BODYFORM &rest HANDLERS)
+#[lisp_fn(min = "2", unevalled = "true")]
+pub fn condition_case(args: LispCons) -> LispObject {
```
shaleh: @nickdrozd look  at how nicely this turned out.

rust_src/src/eval.rs
```diff
@@ -1301,4 +1302,36 @@ pub fn signal(error_symbol: LispObject, data: LispObject) -> ! {
...
+    let (var, consq) = args.into();
+    let (bodyform, handlers_) = consq.into();
```
shaleh: Why the underscore on handlers? I do not see a name conflict.
```suggestion
    let (bodyform, handlers) = consq.into();
```

rust_src/src/eval.rs
```diff
@@ -1301,4 +1302,36 @@ pub fn signal(error_symbol: LispObject, data: LispObject) -> ! {
...
+    unsafe { internal_lisp_condition_case(var, bodyform, handlers_) }
```
shaleh:
```suggestion
    unsafe { internal_lisp_condition_case(var, bodyform, handlers) }
```

Issue #1286, opened 2019-01-25
docker-compose build failed


Qquanwei:
* os: windows 10
* docker-compose version 1.23.2, build 1110ad01
* docker version 18.09.1, build 4c52b90

stage:

1. `git clone https://github.com/Wilfred/remacs.git`
2.  `cd remacs`
3.  `docker-compose -f ./docker-compose.yml build`

```
Building remacs
Step 1/4 : FROM ubuntu:xenial
 ---> b0ef3016420a
Step 2/4 : RUN apt-get update && apt-get install -y     automake     build-essential     clang     curl     libclang-dev     libgif-dev     libgnutls-dev     libgtk-3-dev     libjpeg-dev     libncurses5-dev     libtiff-dev     libxml2-dev     libxpm-dev     libxt-dev     texinfo
 ---> Using cache
 ---> 87c78df00e9d
Step 3/4 : ENV PATH "/root/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 ---> Using cache
 ---> 5b334d50800c
Step 4/4 : RUN curl https://sh.rustup.rs -o rustup.sh &&     sh rustup.sh         --default-host x86_64-unknown-linux-gnu         --default-toolchain nightly-2018-11-03 -y &&     rustup default nightly-2018-11-03
 ---> Using cache
 ---> b366a1971414
Successfully built b366a1971414
Successfully tagged remacs-build:latest
```

4. `docker-compose -f ./docker-compose.yml up`  hang on for my terminator, so I use docker without compose
5.  `docker run -v /d/git/remacs:/remacs -w /remacs --name remacs-playaround -it remacs-build`

```shell
root@c30eb96fd3d6:/remacs# ./autogen.sh
Checking Rust toolchain install ...
.emacs currently requires Rust toolchain version nightly-2019-01-19
'.n 'rustup install nightly-2019-01-19
root@c30eb96fd3d6:/remacs#
```

It not output `You can now run './configure'. and dirtionary not `configure` file`.

daichimukai:
I reproduced similar issue. This seems to be due to incompatible default toolchain of Rust compiler specified in Dockerfile. 

After changing default toolchain version I could successfully run autogen.sh. Please try: 
```
docker-compose exec remacs rustup default nightly-2019-01-19
```
or something equivalent.

agraven:
Is the issue still re

Qquanwei:
[...]

PR #1324, opened 2019-02-04
Sync Rust version of Dockerized environment

daichimukai:
#1286 may have caused by incompatible version of Rust toolchain in Dockerfile.

### DIFF
--- a/docker/fedora/Dockerfile
+++ b/docker/fedora/Dockerfile
@@ -23,6 +23,6 @@ ENV PATH "/root/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sb
 RUN curl https://sh.rustup.rs -o rustup.sh && \
     sh rustup.sh \
         --default-host x86_64-unknown-linux-gnu \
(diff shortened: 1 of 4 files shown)


shanavas786:
Did this change fix docker build issues ?

I doubt we need to add `clang libclang-dev` to dependencies as well

daichimukai:
@shanavas786 Is there build issue on docker env and not enough what is in Dockerfile already? At least I could build docker image and, after building Remacs, run `docker-compose exec remacs ./src/remacs -Q`.

shanavas786:
It was just a doubt as `clang` is missing from `Dockerfile` for `fedora`

shaleh:
@shanavas786 I think that is more because the Fedora Docker has seen less attention.

shaleh:
In a perfect world where we had tons of resources, we would run all of the docker images in Travis. Perhaps not on every push but on every merge. We would also have a working Windows env too.

daichimukai:
I opened an issue for build dep on Fedora. Thanks for noticing :+1:

PR #1325, opened 2019-02-04
Add transformation methods to ExternalPtr

### LINKED ISSUE #1317: Update ExternalPtr make incrementing/decrementing the pointer easier
We have this code in the project:
```
let ptr = unsafe { row.as_mut().add(1) };
row.replace_ptr(ptr);
```

That could be something like `row.ptr_add(1)` (maybe there is a better name). We likely need a `ptr_sub` as well. `impl Add` or `impl Sub` are appealing, but not idiomatic Rust for pointers.

agraven:
This makes it simpler to modify the underlying pointer of an `ExternalPtr`. Instances of `replace_ptr` have been replaced with the new methods where appropriate.

Closes #1317.

### DIFF
--- a/rust_src/src/lisp.rs
+++ b/rust_src/src/lisp.rs
@@ -125,17 +125,17 @@ impl<T> ExternalPtr<T> {
         self.0 = ptr;
     }
 
-    pub fn ptr_offset(&mut self, size: isize) {
+    pub unsafe fn ptr_offset(&mut self, size: isize) {
         let ptr = self.0.offset(size);
         self.replace_ptr(ptr);
     }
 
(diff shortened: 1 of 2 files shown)


rust_src/src/lisp.rs
```diff
@@ -124,6 +124,21 @@ impl<T> ExternalPtr<T> {
     pub fn replace_ptr(&mut self, ptr: *mut T) {
         self.0 = ptr;
     }
```
shaleh: As Clippy says, these should all be unsafe methods.

agraven: Fixed.

Issue #1260, opened 2019-01-18
Port kill-process


brotzeit:
```c
DEFUN ("kill-process", Fkill_process, Skill_process, 0, 2, 0,
       doc: /* Kill process PROCESS.  May be process or name of one.
See function `interrupt-process' for more details on usage.  */)
  (Lisp_Object process, Lisp_Object current_group)
{
  process_send_signal (process, SIGKILL, current_group, 0);
  return process;
}
```

jubnzv:
Not so good as first issue, because we should also port `port_send_process` [defined](https://github.com/emacs-mirror/emacs/blob/master/src/process.c#L6606) as `static`. :)

I will take it and other `process_send_process` wrappers from this file.

jubnzv:
It's challenging task for me, I'll try it.

PR #1326, opened 2019-02-04
[WIP] Port process_send_signal and wrappers

jubnzv:
I started working on porting [process_send_signal](https://github.com/emacs-mirror/emacs/blob/master/src/process.c#L6618) and all related wrappers like kill-process (#1260). I need some time to figure out with Rust basics.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4589,10 +4589,12 @@ esac
 AH_TEMPLATE(SIGNALS_VIA_CHARACTERS, [Make process_send_signal work by
 "typing" a signal character on the pty.])
 
+CARGO_DEFAULT_FEATURES=""
 case $opsys in
   dnl Perry Smith <pedz@ddivt1.austin.ibm.com> says this is correct for AIX.
   aix4-2 | cygwin | gnu | dragonfly | freebsd | netbsd | openbsd | darwin )
     AC_DEFINE(SIGNALS_VIA_CHARACTERS, 1)
+    CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"signals-via-characters\", "
     ;;
 
   dnl 21 Jun 06: Eric Hanchrow <offby1@blarg.net> says this works.
@@ -4722,6 +4724,7 @@ case $opsys in
     dnl subprocesses the usual way.  But TIOCSIGNAL does work for PTYs,
     dnl and this is all we need.
     AC_DEFINE(TIOCSIGSEND, TIOCSIGNAL, [Some platforms redefine this.])
+    CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"tiocsigsend\", "
     ;;
 esac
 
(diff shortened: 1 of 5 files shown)


jubnzv:
For now I got stuck with two questions:

1. How should I work with C `#ifdef`feature macroses? Should I append `CARGO_DEFAULT_FEATURES` variable in `configure.ac`, or may be there are more elegant solutions?
2. How can I port [global static variable](https://github.com/emacs-mirror/emacs/blob/master/src/process.c#L234) defined in C source? I can't just remove it from C file, because there are other functions that using it. Should I disable `static` flag?

Maybe someone can give me advice?

shaleh:
Yep, you have hit the messy bits for sure.

Things we have done for `static`s include:
  - Port it to Rust and make it a `lazy_static!` like https://github.com/Wilfred/remacs/blob/c6c547dced9691e8cd6c24823d8fbca91e81f03b/rust_src/src/numbers.rs#L20
 - cheat 100% and drop the `static`. Make sure it is initialized somewhere because statics in C are all set to zero.
 - leave the C variable and write get/set functions which are used in Rust.

shaleh:
The C `ifdef` can be troublesome. Some of them adding a cargo feature is the right approach. For others looking into if for our supported systems there is a need. For some of it, I fully expect we may need to have something like a remacs-libc crate which extends the existing libc crate. You can look at the dired module and the need for dired_unix and dired_windows as examples for how we are currently handling this in tree.

The honest truth is to date we have been skipping the complicated things. At some point we will have to start dealing with them.

shaleh:
Only OS we care about
 - Linux
 - BSD (any of them)
 - MacOS
 - recent Windows

shaleh:
Thank you for this work. I am about to complicate it in the hopes of seeking simplicity.

The project may be better off if this code was moved into a separate crate. Why? To make testing easier. This new crate would only have OS level interactions and no Lisp types.

Also, there is a lot in this PR. Let's tease it apart into smaller one or two function PRs. Yeah, this will take longer. But it will help us debug any problems along the way.

jubnzv:
@shaleh yes I agree, there are too many changes for a single PR.

I suggest make `process_send_signal` non-static as @daichimukai [adviced](https://github.com/Wilfred/remacs/issues/1260#issuecomment-460051733) and first port simple wrappers of this function. Then I can port [kill-buffer-process](https://github.com/emacs-mirror/emacs/blob/master/src/process.c#L7888) which is used in other parts of codebase.

Next step I'll port some used OS-level functions in separate crate. For example, there are many logic in [send_process](https://github.com/emacs-mirror/emacs/blob/master/src/process.c#L6265) which is used everywhere in `process.c`. I suppose we should start implementing them in new crate, and add other functionality later when we need it.

And finally I can port `process_send_signal` using separate crate for OS interaction functions.

Does it make sense?

P.S. I'm sorry for my bad English, I'm working on improving it.

shaleh:
@jubnvz your English is good. No need to apologize.

Your plan is a good one.

jubnzv:
@shaleh thank you, I'll do it.

PR #1328, opened 2019-02-05
Port get-file-char

### LINKED ISSUE #1300: Port get-file-char
```c
DEFUN ("get-file-char", Fget_file_char, Sget_file_char, 0, 0, 0,
       doc: /* Don't use this yourself.  */)
  (void)
{
  if (!infile)
    error ("get-file-char misused");
  return make_number (readbyte_from_stdio ());
}
```

agraven:
I was on the fence on whether or not to port `readbyte_from_stdio`, as it relies on a lot of behaviour that's expressed much more elegantly in c than rust, but I think i managed to get it right.

Closes #1300.

### DIFF
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -163,27 +163,30 @@ pub unsafe fn defvar_per_buffer_offset(
 /// If read over network is interrupted, keep trying until read succeeds.
 #[no_mangle]
 pub unsafe extern "C" fn readbyte_from_stdio() -> i32 {
+    let file = &mut *infile;
+
     // If infile has lookahead, use stored value
-    if (*infile).lookahead > 0 {
-        let idx = (*infile.sub(1)).lookahead as usize;
-        return (*infile).buf[idx].into();
+    if file.lookahead > 0 {
+        file.lookahead -= 1;
+        return file.buf[file.lookahead as usize].into();
     }
 
-    let instream = (*infile).stream;
+    let instream = file.stream;
 
     block_input();
 
     // Interrupted read have been observed while reading over the network.
-    let mut c;
-    while {
-        c = getc_unlocked(instream);
-        c == libc::EOF && errno().0 == libc::EINTR && ferror_unlocked(instream) > 0
-    } {
-        unblock_input();
-        maybe_quit();
-        block_input();
-        clearerr_unlocked(instream);
-    }
+    let c = loop {
+        let c = getc_unlocked(instream);
+        if c == libc::EOF && errno().0 == libc::EINTR && ferror_unlocked(instream) > 0 {
+            unblock_input();
+            maybe_quit();
+            block_input();
+            clearerr_unlocked(instream);
+        } else {
+            break c;
+        }
+    };
 
     unblock_input();
 
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -5,6 +5,8 @@ use libc;
 use std::ffi::CString;
 use std::ptr;
 
+use errno::errno;
+
 use remacs_macros::lisp_fn;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


rust_src/src/lread.rs
```diff
@@ -155,6 +159,41 @@ pub unsafe fn defvar_per_buffer_offset(
...
+        let idx = (*infile.sub(1)).lookahead as usize;
+        return (*infile).buf[idx].into();
+    }
+
+    let instream = (*infile).stream;
+
+    block_input();
+
+    // Interrupted read have been observed while reading over the network.
+    let mut c;
```
shaleh: An alternative implementation that might be more idiomatic Rust is:

    loop {
        c = getc_unlocked(instream);
        if c == libc::EOF && errno().0 == libc::EINTR && ferror_unlocked(instream) > 0 {
            unblock_input();
            maybe_quit();
            block_input();
            clearerr_unlocked(instream);
        } else {
            break;
        }
    }

If we factor out the check....

    fn is_interrupted_read(stream: TheStreamType, value: i32) -> bool {
        c == libc::EOF && errno().0 == libc::EINTR && ferror_unlocked(stream) > 0
    }

    loop {
        c = getc_unlocked(instream);
        if not is_interrupted_read(instream, c) {
            break;
        }
        unblock_input();
        maybe_quit();
        block_input();
        clearerr_unlocked(instream);
    }

hmmm

    fn interruptable_read(stream: TheStreamType) -> i32 {
        loop {
            let c = getc_unlocked(stream);
            if c == libc::EOF && errno().0 == libc::EINTR && ferror_unlocked(instream) > 0 {
                unblock_input();
               maybe_quit();
               block_input();
               clearerr_unlocked(instream);
           } else {
                return c;
           }
        }
    }

    let c = interruptable_read(instream);
    
    unblock_input();

    if c != libc::EOF {
        c
    } else {
        -1
    }

rust_src/src/lread.rs
```diff
@@ -155,6 +159,41 @@ pub unsafe fn defvar_per_buffer_offset(
     }
 }
 
+/// Read a byte from stdio. If it has lookahead, use the stored value.
+/// If read over network is interrupted, keep trying until read succeeds.
+#[no_mangle]
+pub unsafe extern "C" fn readbyte_from_stdio() -> i32 {
+    // If infile has lookahead, use stored value
+    if (*infile).lookahead > 0 {
+        let idx = (*infile.sub(1)).lookahead as usize;
```
shaleh: This does not have the same behavior. The C code is decrementing `lookahead` not `infile`. It should read:

    if (*infile).lookahead > 0 {
        (*infile).lookahead -= 1;
        return (*infile).buf[(*infile).lookahead as usize];
    }

agraven: That makes way more sense, I thought the way I read it seemed off.

agraven: Not having to do pointer math also means I can convert the pointer to a reference, which means I won't have to dereference it explicitly so many times.

rust_src/src/lread.rs
agraven: The loop is definitely a more readable approach, but I'm inclined to leave the code in the function body as it's not used anywhere else, and I consider the slight readability trade-off worth having the code in one place.

shaleh: Fair enough.

rust_src/src/lread.rs
```diff
@@ -155,6 +159,44 @@ pub unsafe fn defvar_per_buffer_offset(
...
+    if file.lookahead > 0 {
+        file.lookahead -= 1;
+        return file.buf[file.lookahead as usize].into();
+    }
+
+    let instream = file.stream;
+
+    block_input();
+
+    // Interrupted read have been observed while reading over the network.
```
shaleh: Nice. I was going to suggest this format but wrote out the function/return version instead.

rust_src/src/lread.rs
```diff
@@ -155,6 +159,44 @@ pub unsafe fn defvar_per_buffer_offset(
     }
 }
 
+/// Read a byte from stdio. If it has lookahead, use the stored value.
+/// If read over network is interrupted, keep trying until read succeeds.
+#[no_mangle]
+pub unsafe extern "C" fn readbyte_from_stdio() -> i32 {
```
shaleh: Definite improvement in readability. :+1:

rust_src/src/lread.rs
```diff
@@ -192,6 +234,15 @@ pub fn read(stream: LispObject) -> LispObject {
     }
 }
 
+/// Don't use this yourself
```
shaleh:
```suggestion
/// Don't use this yourself.
```

The end punctuation was dropped.

PR #1332, opened 2019-02-07
Added necessary libraries for fedora

### LINKED ISSUE #1327: On build dependency of Fedora docker image
As @shanavas786  pointed out in #1324, Fedora's Dockerfile seems to be lacking `clang` and `libclang-dev` (these are package names for Ubuntu, it may differ for Fedora) dependency. Investigate if we need to add them and we can build Remacs for Fedora image.

manithree:
Also, yum is deprecated and may be removed in Fedora 30, so change to
dnf.

Closes #1327

### DIFF
--- a/docker/fedora/Dockerfile
+++ b/docker/fedora/Dockerfile
@@ -1,6 +1,6 @@
 FROM fedora:latest
 
-RUN yum install -y \
+RUN dnf install -y \
     git \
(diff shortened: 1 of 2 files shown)


shaleh:
Putting the Issue number in the description like I just did will automatically close the ticket when the Pull Request is merged.

PR #1335, opened 2019-02-08
Port reverse

### LINKED ISSUE #1247: Port reverse
```c
DEFUN ("reverse", Freverse, Sreverse, 1, 1, 0,
       doc: /* Return the reversed copy of list, vector, or string SEQ.
See also the function `nreverse', which is used more often.  */)
  (Lisp_Object seq)
{
  Lisp_Object new;

  if (NILP (seq))
    return Qnil;
  else if (CONSP (seq))
    {
      new = Qnil;
      FOR_EACH_TAIL (seq)
	new = Fcons (XCAR (seq), new);
      CHECK_LIST_END (seq, seq);
    }
  else if (VECTORP (seq))
    {
      ptrdiff_t i, size = ASIZE (seq);

      new = make_uninit_vector (size);
      for (i = 0; i < size; i++)
	ASET (new, i, AREF (seq, size - i - 1));
    }
  else if (BOOL_VECTOR_P (seq))
    {
      ptrdiff_t i;
      EMACS_INT nbits = bool_vector_size (seq);

      new = make_uninit_bool_vector (nbits);
      for (i = 0; i < nbits; i++)
	bool_vector_set (new, i, bool_vector_bitref (seq, nbits - i - 1));
    }
  else if (STRINGP (seq))
    {
      ptrdiff_t size = SCHARS (seq), bytes = SBYTES (seq);

      if (size == bytes)
	{
	  ptrdiff_t i;

	  new = make_uninit_string (size);
	  for (i = 0; i < size; i++)
	    SSET (new, i, SREF (seq, size - i - 1));
	}
      else
	{
	  unsigned char *p, *q;

	  new = make_uninit_multibyte_string (size, bytes);
	  p = SDATA (seq), q = SDATA (new) + bytes;
	  while (q > SDATA (new))
	    {
	      int ch, len;

	      ch = STRING_CHAR_AND_LENGTH (p, len);
	      p += len, q -= len;
	      CHAR_STRING (ch, q);
	    }
	}
    }
  else
    wrong_type_argument (Qsequencep, seq);
  return new;
}
```

agraven:
Closes #1247.

The code doing the actual reversing for each of the different types could be methods that this function calls. I'm not sure which is preferable.

### DIFF
--- a/rust_src/src/fns.rs
+++ b/rust_src/src/fns.rs
@@ -1,6 +1,6 @@
 //! Random utility Lisp functions.
 
-use std::ptr;
+use std::{ptr, slice};
 
 use libc;
 
@@ -14,16 +14,20 @@ use crate::{
     lists::{assq, car, get, mapcar1, member, memq, put},
     lists::{LispCons, LispConsCircularChecks, LispConsEndChecks},
     minibuf::read_from_minibuffer,
-    multibyte::LispStringRef,
+    multibyte::{string_char_and_length, write_codepoint, LispStringRef},
     numbers::LispNumber,
     obarray::loadhist_attach,
     objects::equal,
     remacs_sys::Vautoload_queue,
-    remacs_sys::{concat as lisp_concat, globals, message1, redisplay_preserve_echo_area},
+    remacs_sys::{
+        concat as lisp_concat, globals, make_uninit_bool_vector, make_uninit_multibyte_string,
+        make_uninit_string, make_uninit_vector, message1, redisplay_preserve_echo_area,
+    },
     remacs_sys::{EmacsInt, Lisp_Type},
     remacs_sys::{Fdiscard_input, Fload, Fx_popup_dialog},
     remacs_sys::{
-        Qfuncall, Qlistp, Qnil, Qprovide, Qquote, Qrequire, Qsubfeatures, Qt, Qyes_or_no_p_history,
+        Qfuncall, Qlistp, Qnil, Qprovide, Qquote, Qrequire, Qsequencep, Qsubfeatures, Qt,
+        Qyes_or_no_p_history,
     },
     symbols::LispSymbolRef,
     threads::c_specpdl_index,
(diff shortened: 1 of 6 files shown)


rust_src/src/fns.rs
```diff
@@ -274,6 +278,67 @@ pub fn concat(args: &mut [LispObject]) -> LispObject {
     }
 }
 
+/// Return the reversed copy of list, vector, or string SEQ.
+/// See also the function `nreverse', which is used more often.
+#[lisp_fn]
+pub fn reverse(seq: LispObject) -> LispObject {
+    if seq.is_nil() {
+        return Qnil;
+    } else if let Some(cons) = seq.as_cons() {
```
shaleh: This looks like a classic 'fold'. Also, I think `iter_cars` works here. `iter_tails` returns Cons cells and `iter_cars` is `XCAR(cons)` over `iter_tails`. Why the `into` at the end? `new` is already a LispObject right?

```
let new = cons.iter_cars(LispConsEndChecks::on, LispConsCircularChecks::on).fold(Qnil, |cons, elt| {
    LispObject::cons(elt, cons)
});
```

rust_src/src/multibyte.rs
```diff
@@ -583,6 +583,30 @@ pub fn multibyte_char_at(slice: &[c_uchar]) -> (Codepoint, usize) {
     }
 }
 
+/// Same as STRING_CHAR_AND_LENGHT
```
shaleh:
```suggestion
/// Same as STRING_CHAR_AND_LENGTH
```
Typo

rust_src/src/multibyte.rs
```diff
@@ -583,6 +583,30 @@ pub fn multibyte_char_at(slice: &[c_uchar]) -> (Codepoint, usize) {
...
+        }
+        3 => {
+            let cp = (((head & 0x0F) as Codepoint) << 12)
+                | (((*ptr.add(1) & 0x3F) as Codepoint) << 6)
+                | (*ptr.add(2) & 0x3F) as Codepoint;
+            (cp, 3)
+        }
+        _ => {
+            let mut len = 0;
+            let cp = string_char(ptr, ptr::null_mut(), &mut len);
```
shaleh: `string_char` should be migrated to a shallow C wrapper and a Rust function which returns the tuple. That can be another PR.

rust_src/src/fns.rs
```diff
@@ -274,6 +278,67 @@ pub fn concat(args: &mut [LispObject]) -> LispObject {
     }
 }
 
+/// Return the reversed copy of list, vector, or string SEQ.
+/// See also the function `nreverse', which is used more often.
+#[lisp_fn]
+pub fn reverse(seq: LispObject) -> LispObject {
+    if seq.is_nil() {
+        return Qnil;
```
```suggestion
        Qnil
```

You don't need the return since you made this all one if/else construct.

shaleh:
re: methods or not.

If the methods existed, would the code benefit from calling them directly instead of calling `Freverse` or is this mostly called from lisp? If this is mostly lisp then the methods might be overkill for now. I could see them being beneficial if we could write Rust unit tests for the data types.

rust_src/src/fns.rs
agraven: Yeah, that looks like it would work. Would remove the need for a `new` as well.

rust_src/src/fns.rs
agraven: Fixed

rust_src/src/multibyte.rs
agraven: Thought something looked off about the way I wrote that. Thanks.

rust_src/src/fns.rs
shaleh: If we wanted to be cute we could even write it as `(elt, cons).into()`. I am ok with the explicit LispObject::cons().

agraven: Yeah, I think the explicitness makes it easier to read.

agraven:
The function is used very sparsely in C code (4 times), and not at all in rust code, so at least for now it should probably be left in the function, and can be moved out to methods if the need arises.

shaleh:
`nreverse` is used far more. Is there a way to unify the code and support both? The difference is `nreverse` is destructive and `reverse` returns a new copy.

It looks like String and Cons are the two candidates for turning into methods once `nreverse` is also ported.

agraven:
I suppose you could make reverse just call nreverse on a copy of its argument. I was actually just about to port nreverse since I was already at it, I'll look into the possibilities.

shaleh:
Agreed. I wonder if an internal function which tasks `source` and `destination` where destination is sometimes the `make_uninit...` would work.

PR #1337, opened 2019-02-08
[WIP] Port eval-sub

shaleh:


rust_src/src/eval.rs
```diff
@@ -1334,4 +1330,158 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+// Actual work for evaluating a sub-expression happens here.
+fn eval_sub_1(
+    original_fun: LispObject,
+    original_args: LispObject,
+    count: libc::ptrdiff_t,
+) -> (LispObject, bool) {
+    let mut keep_going = true;
+    let mut val = Qnil;
+
+    while keep_going {
```
nickdrozd: The `while` can be changed to `loop` and `keep_going` can be dropped. The `keep_going = false` line can be replaced with `break` or `return (val, false)`.

rust_src/src/eval.rs
```diff
@@ -1334,4 +1330,158 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+pub extern "C" fn eval_sub(form: LispObject) -> LispObject {
+    if let Some(symbol) = form.as_symbol() {
+        let env = unsafe { globals.Vinternal_interpreter_environment };
+
+        // Look up its binding in the lexical environment.
+        // We do not pay attention to the declared_special flag here, since we
+        // already did that when let-binding the variable.
+        let lex_binding = env.map_or(Qnil, |e| assq(form, e));
+
+        if let Some((_, cdr)) = lex_binding.into() {
```
nickdrozd: In general I'd like to start using more descriptive names for cons elements.

rust_src/src/eval.rs
shaleh: I tried that originally and was unhappy with it. If you have time, would you give it a go and either share a diff or directly edit this PR?

rust_src/src/eval.rs
shaleh: Agreed. It is sometimes hard to name them.

rust_src/src/eval.rs
nickdrozd:
```diff
modified   rust_src/src/eval.rs
@@ -1407,10 +1407,9 @@ fn eval_sub_1(
     original_args: LispObject,
     count: libc::ptrdiff_t,
 ) -> (LispObject, bool) {
-    let mut keep_going = true;
     let mut val = Qnil;
 
-    while keep_going {
+    loop {
         // Optimize for no indirection.
         let mut fun = original_fun;
         if let Some(symbol) = fun.as_symbol() {
@@ -1478,10 +1477,8 @@ fn eval_sub_1(
             }
         }
 
-        keep_going = false;
+        return (val, false);
     }
-
-    (val, false)
 }
 
 include!(concat!(env!("OUT_DIR"), "/eval_exports.rs"));
```

I think that should do it.

PR #1340, opened 2019-02-10
Port window_next_sibling and window_prev_sibling

mehakun:


rust_src/src/windows.rs
```diff
@@ -1252,6 +1252,24 @@ pub fn window_next_buffers(window: LispWindowLiveOrSelected) -> LispObject {
     win.next_buffers
 }
```
shaleh: Close. Both lisp_fns need `min = "0"` because the DEFUN says "0, 1, 0" which means min 0, max 1, not interactive.

mehakun: oh, yeah, thx. Read porting guide poorly

shaleh: In rust_src/admin there is a script called rustize-emacs. C-u M-| is a good way to call it.

mehakun: added `min`'s

PR #1343, opened 2019-02-11
Port `process_send_signal' wrappers

### LINKED ISSUE #1260: Port kill-process

jubnzv:
Port four small functions that wraps around [process_send_signal](https://github.com/Wilfred/remacs/blob/master/src/process.c#L6149) function.

This will close #1260

### DIFF
--- a/rust_src/src/process.rs
+++ b/rust_src/src/process.rs
@@ -584,21 +584,24 @@ pub fn quit_process(process: LispObject, current_group: LispObject) -> LispObjec
 /// of incoming traffic.
 #[lisp_fn(min = "0")]
 pub fn stop_process(process: LispObject, current_group: LispObject) -> LispObject {
-    let mut p_ref: LispProcessRef = get_process(process).into();
-    let process_type = p_ref.ptype();
-    if process_type.eq(Qnetwork) || process_type.eq(Qserial) || process_type.eq(Qpipe) {
-        unsafe {
-            delete_read_fd(p_ref.infd);
-            p_ref.command = Qt;
+    match get_process(process).as_process() {
+        Some(mut p_ref) => {
+            let process_type = p_ref.ptype();
+            if process_type.eq(Qnetwork) || process_type.eq(Qserial) || process_type.eq(Qpipe) {
+                unsafe {
+                    delete_read_fd(p_ref.infd);
+                    p_ref.command = Qt;
+                }
+            }
         }
-        return process;
-    }
-
-    #[cfg(windows)]
-    error!("No SIGTSTP support");
-    #[cfg(not(windows))]
-    unsafe {
-        process_send_signal(process, libc::SIGTSTP, current_group, false);
+        None => {
+            #[cfg(windows)]
+            error!("No SIGTSTP support");
+            #[cfg(not(windows))]
+            unsafe {
+                process_send_signal(process, libc::SIGTSTP, current_group, false);
+            }
+        }
     }
 
     process
--- a/rust_src/src/process.rs
+++ b/rust_src/src/process.rs
@@ -13,7 +13,7 @@ use crate::{
     remacs_sys::{
         add_process_read_fd, current_thread, delete_read_fd, emacs_get_tty_pgrp, list1,
         list_system_processes, process_send_signal, send_process, setup_process_coding_systems,
-        update_status, Fmapcar, STRING_BYTES,
+        tcflush, update_status, Fmapcar, STRING_BYTES,
     },
     remacs_sys::{pvec_type, EmacsInt, Lisp_Process, Lisp_Type, Vprocess_alist},
     remacs_sys::{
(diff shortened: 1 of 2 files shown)


rust_src/src/process.rs
```diff
@@ -556,6 +556,77 @@ pub fn internal_default_interrupt_process(
...
+
+/// Stop process PROCESS. May be process or name of one.
+/// See function `interrupt-process' for more details on usage.
+/// If PROCESS is a network or serial or pipe connection, inhibit handling
+/// of incoming traffic.
+#[lisp_fn(min = "0")]
+pub fn stop_process(process: LispObject, current_group: LispObject) -> LispObject {
+    let mut p_ref: LispProcessRef = get_process(process).into();
+    let process_type = p_ref.ptype();
+    if process_type.eq(Qnetwork) || process_type.eq(Qserial) || process_type.eq(Qpipe) {
```
shaleh: This makes me miss Python's `in` operator. I want to re-write this as `if process_type in [Qnetwork, Qserial, Qpipe]`.

rust_src/src/process.rs
```diff
@@ -556,6 +556,77 @@ pub fn internal_default_interrupt_process(
...
+    let mut p_ref: LispProcessRef = process.into();
+    let process_type = p_ref.ptype();
+    if process_type.eq(Qnetwork) || process_type.eq(Qserial) || process_type.eq(Qpipe) {
+        unsafe {
+            if p_ref.command.eq(Qt)
+                && p_ref.infd >= 0
+                && (!p_ref.filter.eq(Qt) || p_ref.status.eq(Qlisten))
+            {
+                add_process_read_fd(p_ref.infd);
+                tcflush(p_ref.infd, libc::TCIFLUSH);
```
shaleh: What about Windows support?

rust_src/src/process.rs
```diff
@@ -556,6 +556,77 @@ pub fn internal_default_interrupt_process(
...
+            }
+        }
+        p_ref.command = Qnil;
+        return process;
+    }
+
+    unsafe {
+        process_send_signal(process, libc::SIGCONT, current_group, false);
```
shaleh: How are you planning on handling the SIGCONT and SIGSTOP for platforms without them? Or are they supported on MacOS, Linux, BSD, and Windows?

rust_src/src/process.rs
jubnzv: Yes, Python's `in` would look great here. :)
But I don't see the point in separate function because there are some others conditions with these values in other unported functions in process.c. We can't consider all of them in single function (unless define macros, but I don't think it's worth it).

rust_src/src/process.rs
jubnzv: Really I have no any experience with Windows systems. I'll check their signals capabilities a bit later.

shaleh: Microsoft nows makes Win10 VirtualBox images available for download. You can pull it down, add in Rust and try to build.

shaleh: We used to use `appveyor` to run a Windows check for us like Travis does for Linux and MacOS but it was very unreliable.

jubnzv:
Windows is so bad, it's really hard to make it build C/Rust sources properly. After some experiments with mingw cross-compilation and Windows VM. I make it works, but its really horrible.

`SIGTSTP` and `SIGCONT` are [unavailable](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=vs-2017) in Windows, but Linux, BSDs (as least FreeBSD, OpenBSD and DragonFlyBSD) and MacOS supports them as well. So I can check that compatibility with `#[cfg(windows)]` macro.

I'm not sure about [PurgeComm](https://github.com/remacs/remacs/blob/master/src/process.c#L6349) in Rust. I drop `continue-process` that use it from this PR, because I'm not ready to work with any Windows-related things for now.

shaleh:
Thank you for trying.

We do not hear much from Windows users. But it is hard to say if that is due to it not working or there being a lack of interest.

rust_src/src/process.rs
```diff
@@ -556,6 +556,52 @@ pub fn internal_default_interrupt_process(
...
+    process
+}
+
+/// Stop process PROCESS. May be process or name of one.
+/// See function `interrupt-process' for more details on usage.
+/// If PROCESS is a network or serial or pipe connection, inhibit handling
+/// of incoming traffic.
+#[lisp_fn(min = "0")]
+pub fn stop_process(process: LispObject, current_group: LispObject) -> LispObject {
+    let mut p_ref: LispProcessRef = get_process(process).into();
```
brotzeit: Do we need a `nil` check here ? The c code also checks with `PROCESSP`.

brotzeit: I think this doesn't work like the c implementation. This throws an error, but it shouldn't.

brotzeit: `process_send_signal` at the end of the function will throw the error so I guess that doesn't matter.

brotzeit:
Hmm ok, there are some missing macros.

rust_src/src/process.rs
agraven: You could do `[Qnetwork, Qserial, Qpipe].contains(process_type)`

brotzeit:
Should be good.

rust_src/src/process.rs
shaleh: Changing when errors are thrown could break some lisp.

shaleh: Do this check in a `match` instead would mimic the original flow more closely.

```
match get_process(process).as_process() {
    Some(p_ref) => {
        let process_type = p_ref.ptype();
        if process_type.eq(Qnetwork) || process_type.eq(Qserial) || process_type.eq(Qpipe) {
        unsafe {
            delete_read_fd(p_ref.infd);
            p_ref.command = Qt;
        }
        process
    },
    None => {
        #[cfg(windows)]
        error!("No SIGTSTP support");
        #[cfg(not(windows))]
        unsafe {
            process_send_signal(process, libc::SIGTSTP, current_group, false);
        }
    
        process
    }
}

brotzeit: I don't see the difference, but I like it more.

PR #1344, opened 2019-02-12
Refactor validate region

shaleh:
The C function takes two pointers. In Rust, we can return a tuple instead. Less unsafe is good.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -813,26 +813,28 @@ pub fn overlay_properties(overlay: LispOverlayRef) -> LispObject {
 
 #[no_mangle]
 pub unsafe extern "C" fn validate_region(b: *mut LispObject, e: *mut LispObject) {
-    let start = *b;
-    let stop = *e;
+    let (begin, end) = validate_region_rust(*b, *e);
+    *b = begin.into();
+    *e = end.into();
+}
 
-    let mut beg = start.as_fixnum_coerce_marker_or_error();
+pub fn validate_region_rust(start: LispObject, stop: LispObject) -> (isize, isize) {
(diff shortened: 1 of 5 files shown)


rust_src/src/buffers.rs
```diff
@@ -814,26 +814,28 @@ pub fn overlay_properties(overlay: LispOverlayRef) -> LispObject {
...
 pub unsafe extern "C" fn validate_region(b: *mut LispObject, e: *mut LispObject) {
-    let start = *b;
-    let stop = *e;
+    let (begin, end) = validate_region_rust(*b, *e);
+    *b = begin.into();
+    *e = end.into();
+}
 
-    let mut beg = start.as_fixnum_coerce_marker_or_error();
+pub fn validate_region_rust(start: LispObject, stop: LispObject) -> (isize, isize) {
```
nickdrozd: Can the inputs be made to be more specific than `LispObject`? Same for all the callers.

shaleh: I did not think we had a match for "fixnum or coerced marker".

shaleh: I need to research this more. There is `CHECK_FIXNUM_COERCE_MARKER` and `CHECK_NUMBER_COERCE_MARKER` in the original Emacs source. We have a representation for the NUMBER version called `LispNumber`. We do not have a similar type for the FIXNUM variant currently. There are matching `Qinteger_or_marker_p` and `Qnumber_or_marker_p` as well.

I need to look and see what the Emacs source was expecting and then evaluate if a more explicit type is possible.

Wilfred: I believe the fixnum variant is new, now that upstream has introduced arbitrary sized integers.

PR #1345, opened 2019-02-12
port font-put

brotzeit:


rust_src/src/fonts.rs
```diff
@@ -265,4 +312,33 @@ pub fn font_at_lisp(
...
+///
+/// If FONT is a font-entity or a font-object, KEY must not be the one
+/// accepted by `font-spec'.
+#[lisp_fn]
+pub fn font_put(font: LispObject, prop: LispObject, val: LispObject) -> LispObject {
+    unsafe {
+        let idx = get_font_prop_index(prop);
+
+        if idx >= 0 && idx < font_property_index::FONT_EXTRA_INDEX as i32 {
+            LispFontSpecRef::from(font);
```
daichimukai: So, is this type checking? I was confused at first sight why we need this :confused:

brotzeit: Yeah, we removed the or_error methods. I agree, it's a little confusing.

brotzeit: Maybe we should throw an error explicitly in those cases.

daichimukai: Hmm, i'm thinking that `TryFrom` trait may help such cases for readability. For example, implementing this for `LispFontSpecRef` like (not tested at all):

```rust
impl TryFrom<LispObject> for LispFontSpecRef {
    type Error = LispObject;
    
    fn from(o: LispObject) -> Result<Self, Self::Error> {
            if v.is_pseudovector(pvec_type::PVEC_FONT) && o.is_font_spec() {
                Ok(unsafe { mem::transmute(o) })
            } else {
                Err(Qfont_spec)
            }
        })
    }
}
```

and defining `check_type`  macro:

```rust
macro_rules! check_type {
    ($obj:ident, $t:ty) => {
        if let Err(expect) = $t::try_from($obj) {
            wrong_type!(expect, $obj);
        }
    };
}
```

will allow us to check type with something like:

```rust
check_type!(font, LispFontSpec);
```

Well, beyond the aim of this PR.

shaleh: Or, the `check_type` macro could use the `lisp_type` information and not do any real work at all.

We can start by adding `check_foo` methods to `LispObject`.  The code above could be `obj.check_font()`

shaleh: Another option. All of these array/vector types could implement a VectorAccess trait and provide get/set methods. Then you could write something like:
```
let font: LispFontRef = obj.into();
font.vector_set(idx, font_prop_validate(idx, Qnil, val));
```

Wilfred:
Toggling build to test Travis.

shaleh:
@brotzeit would you like to revisit this?

Issue #1245, opened 2019-01-16
Move to a Remacs organisation


Wilfred:
Remacs has grown way beyond the first patches I've made :)

I'd like to make it an organisation, to reflect the community that we have around the project. Any objections / opinions?

shaleh:
Yes please.

harryfei:
Yeah, that's good.

shanavas786:
Good move :+1:

Wilfred:
Moved! :)

I've created a 'remacs' organisation, and invited all the current collaborators to be members.

Please comment if this has broken anything. I suspect Travis will need fiddling.

In general I love us giving merge power to anyone, but we should also have a few additional admins beyond me. @shaleh are you happy being an admin?

Wilfred:
OK, Travis is running: https://travis-ci.org/remacs/remacs

The only issue I'm aware of is updating the Travis badge in the readme.

Wilfred:
Annoyingly, I can't see the state of old Travis builds here: https://travis-ci.org/Wilfred/remacs or individual builds like https://travis-ci.org/Wilfred/remacs/builds/492195097

Wilfred:
Based on https://help.github.com/articles/repository-permission-levels-for-an-organization/ I think 'write' permission is the right level to enable people to merge PRs. We should still be requiring an approval of PRs according to https://github.com/remacs/remacs/settings/branch_protection_rules/551755

shaleh:
Yeah, I can admin. We can huddle on gitter and work out the people able to be staff.

Wilfred:
Done.

I don't think there's anything left to do here, but feel free to open new issues and tag me if you see anything.

PR #1348, opened 2019-02-13
Update Travis badge for new GitHub URL

Wilfred:
Now that Remacs is a separate organisation (see #1245), the README needs updating.


Wilfred:
Excellent, looks like Travis is picking up new PRs here.

PR #1351, opened 2019-02-15
Complete the port of casetab.c

shaleh:
Add a `LispCaseTable` which is a type wrapper around `LispCharTableRef`.
Implement the bare minimum of needed pass through methods.


rust_src/src/casetab.rs
```diff
@@ -65,18 +231,81 @@ pub fn standard_case_table() -> LispObject {
...
 #[lisp_fn]
 pub fn set_standard_case_table(table: LispObject) -> LispObject {
-    unsafe { set_case_table(table, true) }
+    set_case_table(table, true)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn init_casetab_once() {
+    def_lisp_sym!(Qcase_table, "case-table");
+    Fput(Qcase_table, Qchar_table_extra_slots, 3.into());
```
nickdrozd: `put` has been ported (although it would require casting the symbols to `LispSymbolRef`).

rust_src/src/casetab.rs
```diff
@@ -3,31 +3,185 @@
...
+        }
+    }
+}
+
+// Set elements of char-table TABLE for C to C itself.  C may be a
+// cons specifying a character range.  In that case, set characters in
+// that range to themselves.  This is done only when ELT is a
+// character.  This is called in map_char_table.
+extern "C" fn set_identity(table: LispObject, c: LispObject, elt: LispObject) {
+    if elt.is_natnum() {
```
nickdrozd:
```suggestion
    if !elt.is_natnum() {
        return;
    }
```

That would save a level of indentation in the rest of the function.

rust_src/src/casetab.rs
```diff
@@ -3,31 +3,185 @@
...
+
+// Permute the elements of TABLE (which is initially an identity
+// mapping) so that it has one cycle for each equivalence class
+// induced by the translation table on which map_char_table is
+// operated.
+extern "C" fn shuffle(table: LispObject, c: LispObject, elt: LispObject) {
+    if let Some(idx) = elt.as_natnum() {
+        let char_table: LispCharTableRef = table.into();
+
+        let (from, to): (EmacsInt, EmacsInt) = match c.into() {
```
nickdrozd: So what is `c` here? A `LispCons` or an `EmacsInt`?

nickdrozd: Inclusive or

rust_src/src/casetab.rs
shaleh: Exactly. I did not see the benefit of the Symbol transform here. :shrug:

rust_src/src/casetab.rs
shaleh: True. I was ambivalent. This is probably the right thing to do.

rust_src/src/casetab.rs
shaleh: It is either sadly.

shaleh: I went to rename it and could not give it a better name than "value". Sigh.

PR #1352, opened 2019-02-16
Refine LispCaseTable

shaleh:
This builds on the previous casetab.c PR. Now `LispCaseTable` is
responsible for more work and the conversion routines are more
fleshed out. My plan is to work through the chartable.c next
which will allow for more Rust-y code here and in chartable.rs.


rust_src/src/casetab.rs
```diff
@@ -3,100 +3,126 @@
...
-            map_char_table(Some(set_canon), Qnil, table, table);
-        }
+        let (up, canon, eqv) = table.extras();
 
-        if eqv.is_nil() {
-            eqv = Fmake_char_table(Qcase_table, Qnil);
-            map_char_table(Some(set_identity), Qnil, canon, eqv);
-            map_char_table(Some(shuffle), Qnil, canon, eqv);
-            set_char_table_extras(table, 2, eqv);
+        if (up.is_nil() || up.is_char_table())
```
nickdrozd: I realize this is copied from elsewhere, but I think this is equivalent to 

```rust
up.is_maybe_char_table()
&& canon.is_maybe_char_table()
&& eqv.is_maybe_char_table()
```

where `x.is_maybe_char_table()` is equivalent to `x.is_nil() || x.is_char_table()`.

rust_src/src/casetab.rs
```diff
@@ -3,35 +3,200 @@
...
+// translated ELT by UP and DOWN char-tables.  This is done only when
+// ELT is a character.  The char-tables CANON, UP, and DOWN are in
+// TABLE.
+extern "C" fn set_canon(table: LispObject, range: LispObject, elt: LispObject) {
+    if let Some(idx) = elt.as_natnum() {
+        let case_table: LispCaseTable = table.into();
+
+        let (up, canon, _) = case_table.extras();
+
+        let up_table: LispCharTableRef = up.into();
```
nickdrozd: Is this conversion guaranteed? It looks like `.extras()` can also return nil objects.

rust_src/src/casetab.rs
shaleh: Nice. I had not seen that style. But totally accurate.

shaleh: A recurring frustration with Emacs source is the places where nil is ok and where nil is not ok.

rust_src/src/casetab.rs
shaleh: Hmmm. This is run to set the values.

This is the current version in Emacs:

```
set_canon:
  Lisp_Object up = XCHAR_TABLE (case_table)->extras[0];
  Lisp_Object canon = XCHAR_TABLE (case_table)->extras[1];

  if (FIXNATP (elt))
    Fset_char_table_range (canon, range, Faref (case_table, Faref (up, elt)));
```

If `up` is nil, that call to `Faref` will throw `wrong_type(Qarrayp)`. What that means is my code likely changes the error message. But now it is a useful error instead of the message from `aref`.

If we want to update this we should do it on my previous PR, which I based this one on.

shaleh: I would like to land the casetab.c PR and rebase this one. I posted it as a place holder.

shaleh:
@nickdrozd, I gave this PR a fresh look when I rebased it. The current version is a slimmed down from the original. I will do another pass once this one lands.

rust_src/src/casetab.rs
shaleh: Reading this again, I think the boolean is a little more complex. I believe the proper way to read this is:
```
up.is_maybe_char_table()
&& ((canon.is_nil() && eqv.is_nil())
        || (canon.is_char_table() && eqv.is_maybe_char_table()))
```

I like this idea of "is_maybe". I wonder if we can cheaply generalize it to `is_maybe!(type, value)`? Then we could have `is_maybe!(LispCons, obj)` too and likely `is_maybe!(LispString, obj)`.

nickdrozd: Okay, so it's not possible for canon to be a char table but eqv be nil. What a horrible expression.

nickdrozd: It would probably be better long-term to figure out a way to cast them as `Option<LispCharTableRef>` than come up with a new macro. I don't know exactly how that would work.

shaleh: A `Result<Option<Thing>>` would work. `Err` for completely wrong type, `None` for nil, and `Some` for the real value.

shaleh: Or perhaps we need/want a 3 way type to handle that.

Err
None
Value

But using the existing Result type means we can use `try_into`. Things to experiment with.

Completely off topic but relevant still. Places where we hits walls with using C are in part due to the use of setjmp/longjmp in the C code I think. So we step into undefined territory (literally). As we transition to pure Rust we will need to work out how to handle this. `lisp_fn` returns a Result type and having the macro generated wrapper handle it might be a way to maintain the current behavior but in a Rust manner. The `Err` value could track the depth and what not instead of only being a String.

rust_src/src/alloc.rs
```diff
@@ -60,6 +79,36 @@ pub fn bool_vector(args: &mut [LispObject]) -> LispObject {
...
+/// PURPOSE should be a symbol.  If it has a `char-table-extra-slots'
+/// property, the property's value should be an integer between 0 and 10
+/// that specifies how many extra slots the char-table has.  Otherwise,
+/// the char-table has no extra slot.
+#[lisp_fn(min = "1")]
+pub fn make_char_table(purpose: LispSymbolRef, init: LispObject) -> LispCharTableRef {
+    let n = get(purpose, Qchar_table_extra_slots);
+    let n_extras: EmacsUint = match n.into() {
+        None => 0,
+        Some(x) => {
```
agraven: I think it would be nicer to use a match guard here, like so:
```rust
Some(x) if x > 10 => args_out_of_range!(n, Qnil),
Some(x) => x
```

rust_src/src/casetab.rs
```diff
@@ -3,100 +3,138 @@
...
+    }
+
     pub fn extras(&self) -> (LispObject, LispObject, LispObject) {
         let extras = unsafe { self.0.extras.as_slice(3) };
         (extras[0], extras[1], extras[2])
     }
 
-    pub fn from_char_table(table: LispCharTableRef) -> Self {
-        Self(table)
+    pub fn set_extras(&mut self, idx: usize, value: LispObject) {
```
agraven: Renaming this method to `set_extra` would make its purpose more clear. Extra being plural implies all the extras are being set. While on the subject, would it make sense to have a method taking a tuple that sets all of the extras?

rust_src/src/casetab.rs
```diff
@@ -3,100 +3,138 @@
...
+        let sym_case_table = LispSymbolRef::from(Qcase_table);
 
-    unsafe {
         if up.is_nil() {
-            up = Fmake_char_table(Qcase_table, Qnil);
-            map_char_table(Some(set_identity), Qnil, table, up);
-            map_char_table(Some(shuffle), Qnil, table, up);
-            set_char_table_extras(table, 0, up);
+            let up_table = make_char_table(sym_case_table, Qnil);
+            up = LispObject::from(up_table);
```
agraven: Is the type annotation here necessary? I think it could be removed in line 80 as well (`canon = LispObject::from(tem)`)

rust_src/src/casetab.rs
```diff
@@ -3,100 +3,138 @@
...
     }
+}
 
-    table
+// Allow for the transformation of char tables into case tables when their
+// `purpose` field is set to `Qcase_table`. Do not check the `extras` fields.
+// This way a new char table can be transformed into a case table.
+impl From<LispObject> for Option<LispCaseTable> {
+    fn from(obj: LispObject) -> Self {
+        Option::<LispCharTableRef>::from(obj).and_then(LispCaseTable::from_char_table)
```
agraven: It feels like it would make sense to require `is_proper` to be true for the conversion to succeed, I think the "unchecked" version should have a separate method.

rust_src/src/casetab.rs
```diff
@@ -233,24 +257,51 @@ pub unsafe extern "C" fn get_canonical_case_table() -> LispObject {
...
+pub fn set_standard_case_table(table: LispObject) -> LispCaseTable {
+    let case_table = LispCaseTable::from(table);
+    if !case_table.is_proper() {
+        wrong_type!(Qcase_table, table);
+    }
+    set_standard_case_table_1(case_table);
+    case_table
+}
+
+fn set_standard_case_table_1(mut case_table: LispCaseTable) {
```
agraven: Making `is_proper` required for conversion using `From` would remove the need for two functions here.

rust_src/src/chartable.rs
```diff
@@ -156,6 +161,13 @@ impl LispCharTableRef {
     pub fn set_unchecked(self, idx: isize, value: LispObject) {
         unsafe { CHAR_TABLE_SET(self.into(), idx as i32, value) };
     }
+
+    pub fn set_extras(&mut self, idx: usize, value: LispObject) {
```
agraven: It would make sense to replace `set_char_table_extras` with a wrapper around this method.

shaleh:
Thanks for the review @agraven. Things to ponder.

rust_src/src/chartable.rs
shaleh: Better I think to port the C code over first. No sense in C -> Rust -> C here.

agraven:
Since this PR hasn't seen any activity for quite a while, I'd like to either overtake writing it or redo it in smaller pieces from scratch.

shaleh:
I actually have a significant chunk in a local branch. Let me push the mess up and you are welcome to finesse it.

agraven:
@shaleh do you still have the local changes?

shaleh:
yes. blah. Sorry about that.

shaleh:
@agraven I have pushed everything I have to my remacs fork. Happy picking. Ping me if you have any questions.

agraven:
On which branch? I don't see any recent pushes on the chartables-updates one

shaleh:
@agraven in my shaleh/remacs look at port-foo function which match C names. I was trying to port it piece by piece.

PR #1354, opened 2019-02-22
Port scroll_command and add InteractiveNumericPrefix as a new helper type.

shaleh:
Implement `InteractiveNumericPrefix` type.
This type represents when "p" or "P" is used as the interactive argument for a DEFUN.


rust_src/src/windows.rs
```diff
@@ -1456,8 +1466,48 @@ pub fn scroll_up(arg: LispObject) {
 /// If ARG is the atom `-', scroll upward by nearly full screen.
 /// When calling from a program, supply as argument a number, nil, or `-'.
 #[lisp_fn(min = "0", intspec = "^P")]
-pub fn scroll_down(arg: LispObject) {
-    unsafe { scroll_command(arg, -1) };
+pub fn scroll_down(arg: Option<InteractiveNumericPrefix>) {
+    scroll_command(arg, -1);
+}
+
+// Scroll selected window up or down.  If N is nil, scroll upward by a
```
nickdrozd: Should be `///`?

shaleh: My understanding is the triple quote is for API level docs. Most of the internal functions we only use double slash. Did I misunderstand? Or are you suggesting this is useful enough to be raised up to the API docs?

rust_src/src/interactive.rs
```diff
@@ -7,22 +7,76 @@ use crate::{
...
+    pub fn unwrap(self) -> EmacsInt {
+        match self {
+            Self::Number(value) => value,
+            Self::Object(raw) => {
+                if raw.is_nil() {
+                    1
+                } else if raw.eq(Qminus) {
+                    -1
+                } else if raw.is_integer() {
+                    raw.as_fixnum_or_error()
```
brotzeit: Didn't we drop this kind of methods ?

shaleh: Yes, we do usually. I left this one and the one in the next `else if` for clarity. Easy enough to switch them out for `into()` if that is the review consensus.

PR #1359, opened 2019-02-28
Update rust to nightly-2019-02-27

shanavas786:
I am not sure should we address clippy warnings on redundant closures or just allow them.
It seems to me that the current style is more readable and concise than the suggested one.


shaleh:
Good question @shanavas786. The `into` ones could be replaced by From which would be a little more readable in that style. Clippy is mentioning this because the compiler is making a closure it does not need to make so by removing it the code can be a little more efficient.

rust_src/src/process.rs
```diff
@@ -169,7 +171,7 @@ pub fn process_id(process: LispProcessRef) -> Option<EmacsInt> {
 /// deleted or killed.
 #[lisp_fn]
 pub fn get_buffer_process(buffer_or_name: Option<LispBufferOrName>) -> Option<LispProcessRef> {
-    get_buffer_process_internal(buffer_or_name.and_then(|b| b.into()))
+    get_buffer_process_internal(buffer_or_name.and_then(Into::into))
```
shanavas786: Looks a bit odd. I couldn't find another way of doing this.

daichimukai:
By the way, we probably should also update Dockerfile when we change rust version.

shaleh:
@daichimukai could you add a CI test to validate that the version in the Dockerfile is up to date? Or should we stop shipping the file and instead move to a Dockerfile.in style for all of them and the user has to 'build' them?

rust_src/src/process.rs
shaleh: `Option<LispProcessRef>::from` would work right?

Perhaps this is a place we have used Convert too much and it should be `LispBufferOrName::to_process()`.

rust_src/src/buffers.rs
```diff
@@ -690,7 +691,7 @@ impl LispObject {
     }
 
     pub fn as_live_buffer(self) -> Option<LispBufferRef> {
-        self.as_buffer().and_then(|b| b.as_live())
```
nickdrozd: Is this new? I've always hated having to create a closure just to call a method with no arguments.

shanavas786:
Updated dockerfiles.

daichimukai:
@shaleh Sorry for late reply. It seems to me that both ways have positive and negative side, and I'm not sure which is better. This PR is merged so let's discuss that on an another issue.

PR #1360, opened 2019-03-01
Port window-normal-size

th0114nd:
https://github.com/remacs/remacs/issues/1358

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -25,10 +25,8 @@ use crate::{
         update_mode_lines, window_list_1, window_menu_bar_p, window_tool_bar_p,
         windows_or_buffers_changed, wset_redisplay,
     },
-    remacs_sys::{
-        face_id, glyph_matrix, glyph_row, pvec_type, vertical_scroll_bar_type, EmacsInt, Lisp_Type,
-        Lisp_Window,
-    },
+    remacs_sys::{face_id, glyph_matrix, glyph_row, pvec_type, vertical_scroll_bar_type},
+    remacs_sys::{EmacsDouble, EmacsInt, Lisp_Type, Lisp_Window},
     remacs_sys::{Fcopy_alist, Fnreverse},
     remacs_sys::{
         Qceiling, Qfloor, Qheader_line_format, Qleft, Qmode_line_format, Qnil, Qnone, Qright, Qt,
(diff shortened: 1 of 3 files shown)


rust_src/src/windows.rs
```diff
@@ -713,6 +713,40 @@ pub extern "C" fn decode_any_window(window: LispObject) -> LispWindowRef {
...
+/// horizontal combination equals 1.0.
+///
+/// The normal sizes of windows are used to restore the proportional sizes
+/// of windows after they have been shrunk to their minimum sizes; for
+/// example when a frame is temporarily made very small and afterwards gets
+/// re-enlarged to its previous size.
+#[lisp_fn(min = "0")]
+pub fn window_normal_size(
+    window: LispWindowValidOrSelected,
+    horizontal: LispObject,
```
shaleh: `horizontal` could be a boolean. The if/else below becomes `if !horizontal.....` or you could flip the if/else and use `if horizontal`.
It is not always safe to use a LispObject as a boolean. But on this case its only use is the nil check.

test/rust_src/src/windows-tests.el
```diff
@@ -95,3 +95,13 @@
   (set-window-parameter (selected-window) 'test 'test)
   (should (consp (window-parameters)))
   (should (consp (window-parameters (selected-window)))))
+
```
shaleh: Excellent. Thank you for the tests.

rust_src/src/windows.rs
```diff
@@ -713,6 +713,40 @@ pub extern "C" fn decode_any_window(window: LispObject) -> LispWindowRef {
...
+) -> EmacsDouble {
+    let win: LispWindowRef = window.into();
+    EmacsDouble::from(if horizontal.is_nil() {
+        win.normal_lines
+    } else {
+        win.normal_cols
+    })
+}
```
shaleh: I might have written this in the following style.

```
let value = if horizontal {
win.normal_cols
} else {
win.normal_lines
};
value.into() // or EmacsDouble::from(value)

File: rust_src/src/windows.rs [hunk 4]
```diff
@@ -26,8 +26,8 @@ use crate::{
         windows_or_buffers_changed, wset_redisplay,
     },
     remacs_sys::{
-        face_id, glyph_matrix, glyph_row, pvec_type, vertical_scroll_bar_type, EmacsInt, Lisp_Type,
-        Lisp_Window,
+        face_id, glyph_matrix, glyph_row, pvec_type, vertical_scroll_bar_type, EmacsDouble,
+        EmacsInt, Lisp_Type, Lisp_Window,
```
shaleh: We usually try to import types separately from functions. This is all done to cut down on merge conflicts.

test/rust_src/src/windows-tests.el
th0114nd: NP

rust_src/src/windows.rs
th0114nd: Done

rust_src/src/windows.rs
th0114nd: I think these are types, but I separated the remacs binding ypes from the c struct types

brotzeit:
:+1:

PR #1365, opened 2019-03-08
Port nreverse

agraven:
This does not use any of the code reuse suggested in #1335.

I was not able to successfully implement the reversing of cons lists using iterators, though I think it may to make it work using a peekable iterator, ~~I will add a commit if I manage to do so~~ done.


shanavas786:
triggering build

rust_src/src/fns.rs
```diff
@@ -335,6 +335,50 @@ pub fn reverse(seq: LispObject) -> LispObject {
...
+    } else if seq.is_string() {
+        return reverse(seq);
+    } else if let Some(cons) = seq.as_cons() {
+        let mut iter = cons
+            .iter_tails(LispConsEndChecks::on, LispConsCircularChecks::on)
+            .peekable();
+        let mut prev = Qnil;
+        while let Some(tail) = iter.next() {
+            // Invoked to advance the underlying iterator
+            iter.peek();
```
shaleh: From the Rust docs (https://doc.rust-lang.org/std/iter/struct.Peekable.html):

> Returns a reference to the next() value without advancing the iterator.

That call to `peek()` is not doing anything.

rust_src/src/fns.rs
```diff
@@ -335,6 +335,50 @@ pub fn reverse(seq: LispObject) -> LispObject {
...
+            tail.set_cdr(prev);
+            prev = tail.into();
+        }
+        seq = prev;
+    } else if let Some(mut vec) = seq.as_vector() {
+        let len = vec.len() / 2;
+        let (left, right) = vec.as_mut_slice().split_at_mut(len);
+        for (a, b) in left.iter_mut().zip(right.iter_mut().rev()) {
+            let c = *a;
```
shaleh: You can use `std::mem::swap` here.

rust_src/src/lists.rs
```diff
@@ -31,6 +31,12 @@ impl LispObject {
         }
     }
```
shaleh: This is correct but looks wrong. Is there a reason not to use `list.into()` instead of passing in the list?

shaleh: Why is this function defined?

rust_src/src/fns.rs
agraven: This contradicts the documentation in [Iterator::peekable](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.peekable), which states:

> Note that the underlying iterator is still advanced when peek is called for the first time: In order to retrieve the next element, next is called on the underlying iterator, hence any side effects (i.e. anything other than fetching the next value) of the next method will occur.

And from looking at how the code actually behaves, peek does invoke side effects, as the iterator is correctly advanced before the cdr is set.

agraven: Looking into it further, I think the documentation for the peek method is poorly worded, I think when it says "without advancing the Iterator", it's referring to the wrapping iterator, and not the underlying one.

rust_src/src/lists.rs
agraven: It was used in the cons reversing implementation that didn't use iterators. It replaces the C macro `CHECK_LIST_END`

rust_src/src/fns.rs
agraven: Fixed

rust_src/src/fns.rs
shaleh: Weird. I would not expect `peek` to change the iterator. What a horrible name.

rust_src/src/lists.rs
shaleh: Ah, I see. This is currently part of the iterator code. Does it need to be refactored out into a stand alone function?

agraven: I guess it's only necessary if there's places where the iterators can't be used to implement a thing that needs to check if something is the end of a list, which I'm not sure there is, but I could be wrong

rust_src/src/fns.rs
shaleh: Wait.

```
let xs = [1, 2, 3];

let mut iter = xs.iter().peekable();

// peek() lets us see into the future
assert_eq!(iter.peek(), Some(&&1));
assert_eq!(iter.peek(), Some(&&1));
```

How can `peek` be advancing the iterator if the same value is there?? The examples on the site match my intuition. Do we have a bug in how our iterator works?

agraven:
No it works as intended. peek() only advances the underlying iterator the
first time it's invoked after a call to next()

fre. 8. mar. 2019 22.40 skrev Sean Perry <notifications@github.com>:

rust_src/src/fns.rs
agraven: No, it works as intended. peek() only advances the underlying iterator the first time it is invoked after a call to next()

shaleh: Does this playground match your expectation? https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=7c0334c9597396569f02bbd9d3d50e10

shaleh: My understanding is this loop:
```
let mut it = thing.iter();
loop {
    match it.next() {
        Some(value) => { some action on value },
        None => break,
    }
}
```
and this loop:
```
let mut it = thing.iter().peekable();
loop {
    match it.peek() {
        Some(p) => { println!("Peek work: {}", p); },
        None => { println!("Nothing else to see."); },
    };
    match it.next() {
        Some(value) => { some action on value },
        None => break,
    };
}
```

Should yield the same values and essentially be identical.

agraven: Yes, the playground matches my expectation. And the reason that invoking peek() changes behaviour is that it advances the underlying iterator before the cdr is changed, effectively caching which next value to use.

If peek() is not invoked, every time next() is called it would advance the iterator to the new cdr, which is the cons cell the used to be the car of the current one.

The caching behaviour mirrors how the C code defines a `next = XCDR(tail)` value at the start of every loop, and invokes `tail = next` at the end of the loop (or rather, the last part of the for loop. but it's functionally the same).

agraven:
I have discovered something strange about the behavior of list reversing. It behaves as expected the first time it's invoked on a list, but second time only the (originally) first element is in the list:
```
(nreverse '(one two three))
⇒ (three two one)

(let ((A '(one two three))) (nreverse A) (nreverse A))
⇒ (one)
```
This is the case for both the implementation that's a more or less direct clone of the c code from the first commit of this PR, and for the implementation using iterators. This is very frustrating to debug, as `nreverse` is invoked so often it's nigh impossible to break on a specific invocation of it.

shaleh:
What I do is leave the C version in place and give then new Rust version a different name. nreverse2 or rust-nreverse. This way you can verify your implementation. Even write unit tests comparing results.

agraven:
It behaves as it's supposed to. My mistake was not realizing that A still points to the same cons cell after the nreverse invocation.

agraven:
Kicking the build made the error stop happening. I suggest merging the PR.

shaleh:
Overall I am positive @agraven. However, I still dislike the peekable based code. That code feels like there is hidden magic in it. The call to `peek()` that is not saving the result but moving the iterator feels wrong. You should be able to call peek a hundred times and not change the underlying data.

agraven:
I have two ways of doing it. The first one is the following, using an iterator and two "caching" variables.
```rust
let mut iter = cons.iter_tails(LispConsEndChecks::on, LispConsCircularChecks::on);
let mut prev2 = Qnil;
let mut prev = Qnil;
while let Some(tail) = iter.next() {
    if let Some(prev) = prev.as_cons() {
        prev.set_cdr(prev2);
    }
    prev2 = prev;
    prev = tail.into();
}
prev.as_cons().map(|last| last.set_cdr(prev2));
seq = prev;
```
This is by far the least efficient solution, as it needs to do a type check every iteration.

The other is the original implementation, a direct translation of the c code, which means it doesn't use iterators and is thus it less idiomatic, but it should be at least as efficient as the c code.
```rust
let mut prev = Qnil;
let mut tail = seq;
let mut next;
while tail.is_cons() {
	next = tail.force_cons().cdr();

	if next == seq {
		circular_list(seq);
	}
	tail.force_cons().set_cdr(prev);
	prev = tail;
	tail = next;
}
tail.check_list_end(seq);
seq = prev;
```

agraven:
I added a commit using the `PutBack` solution.

brotzeit:
Looks quite nice.

PR #1367, opened 2019-03-09
Port window-fringes and set-window-fringes 

razzmatazz:


test/rust_src/src/windows-tests.el
```diff
@@ -105,3 +105,34 @@
...
+
+(ert-deftest window-fringes ()
+  (let ((w1 (selected-window))
+        (w2 (split-window)))
+    (set-window-fringes w1 4 4 nil)
+    (set-window-fringes w2 16 16 t)
+    (should (equal `(4 4 nil) (window-fringes w1)))
+    (should (equal `(4 4 nil) (window-fringes nil)))
+    (should (equal `(16 16 t) (window-fringes w2)))))
+
```
shaleh: None of these check the return value of `set-window-fringes`. Can there be a test which validates the nil return?

shaleh:
Beware, running tests within remacs using M-x ert sometimes yields different results from running the tests via `make check`. We suspect this has to do with running interactive versus running without a controlling window/shell.

razzmatazz:
For me, it seems `make install` to be working OK when I run it from an `ansi-term` buffer, but not from a `shell`. Also, running make install from iterm2 works as well. So *maybe* there could be some relation to terminal caps. I tried playing with environment vars for a bit in order to figure wheres is the problem to no avail.

razzmatazz:
@shaleh Yeah, the problem is that window-fringes and set-window-fringes requires graphical UI to be working. And that is what is making these tests fail for me since they run with emacs on a tty, right?

See https://github.com/remacs/remacs/blob/dc2ea22ddfa1ba077178ffabeff8590aea910242/src/window.c#L5789

Now, should I remove/disable the tests?

razzmatazz:
@shaleh I have made it so those tests are skiped when run on a tty in the latest commit.

test/rust_src/src/windows-tests.el
razzmatazz: Sure, forgot about this. I have added this code now in:
 - https://github.com/remacs/remacs/pull/1367/commits/d265174a2fac343a81e3b1a7e8bc671bd8658534

razzmatazz:
@shaleh I tried running this on an emacs in a tty and I get the behaviour where `(set-window-fringes)` calls return `nil` all the time (on regular emacs 26.1 from macOS brew). And `(window-fringes)` always return `(0 0 nil)` whereas on macOS UI (with spacemacs as default setup) I get `(8 8 nil)`

shaleh:
https://www.gnu.org/software/emacs/manual/html_node/emacs/Fringes.html#Fringes you are correct, fringes are only active on GUI frames.

PR #1369, opened 2019-03-10
Port key-binding

### LINKED ISSUE #1236: Port key-binding
```c
DEFUN ("key-binding", Fkey_binding, Skey_binding, 1, 4, 0,
       doc: /* Return the binding for command KEY in current keymaps.
KEY is a string or vector, a sequence of keystrokes.
The binding is probably a symbol with a function definition.

Normally, `key-binding' ignores bindings for t, which act as default
bindings, used when nothing else in the keymap applies; this makes it
usable as a general function for probing keymaps.  However, if the
optional second argument ACCEPT-DEFAULT is non-nil, `key-binding' does
recognize the default bindings, just as `read-key-sequence' does.

Like the normal command loop, `key-binding' will remap the command
resulting from looking up KEY by looking up the command in the
current keymaps.  However, if the optional third argument NO-REMAP
is non-nil, `key-binding' returns the unmapped command.

If KEY is a key sequence initiated with the mouse, the used keymaps
will depend on the clicked mouse position with regard to the buffer
and possible local keymaps on strings.

If the optional argument POSITION is non-nil, it specifies a mouse
position as returned by `event-start' and `event-end', and the lookup
occurs in the keymaps associated with it instead of KEY.  It can also
be a number or marker, in which case the keymap properties at the
specified buffer position instead of point are used.
  */)
  (Lisp_Object key, Lisp_Object accept_default, Lisp_Object no_remap, Lisp_Object position)
{
  Lisp_Object value;

  if (NILP (position) && VECTORP (key))
    {
      Lisp_Object event;

      if (ASIZE (key) == 0)
	return Qnil;

      /* mouse events may have a symbolic prefix indicating the
	 scrollbar or mode line */
      event = AREF (key, SYMBOLP (AREF (key, 0)) && ASIZE (key) > 1 ? 1 : 0);

      /* We are not interested in locations without event data */

      if (EVENT_HAS_PARAMETERS (event) && CONSP (XCDR (event)))
	{
	  Lisp_Object kind = EVENT_HEAD_KIND (EVENT_HEAD (event));
	  if (EQ (kind, Qmouse_click))
	    position = EVENT_START (event);
	}
    }

  value = Flookup_key (Fcons (Qkeymap, Fcurrent_active_maps (Qt, position)),
		       key, accept_default);

  if (NILP (value) || INTEGERP (value))
    return Qnil;

  /* If the result of the ordinary keymap lookup is an interactive
     command, look for a key binding (ie. remapping) for that command.  */

  if (NILP (no_remap) && SYMBOLP (value))
    {
      Lisp_Object value1;
      if (value1 = Fcommand_remapping (value, position, Qnil), !NILP (value1))
	value = value1;
    }

  return value;
}
```

agraven:
This PR adds an Event newtype for LispObject to replace the EVENT_ macros.

Closes #1236.


src/keymap.c
```diff
@@ -1154,78 +1154,6 @@ like in the respective argument of `key-binding'.  */)
...
-    {
-      Lisp_Object value1;
-      if (value1 = Fcommand_remapping (value, position, Qnil), !NILP (value1))
-	value = value1;
-    }
-
-  return value;
-}
-
-/* GC is possible in this function if it autoloads a keymap.  */
```
brotzeit: This comment belongs to the function below right ?

daichimukai: There is an exactly same comment above the function so git (and we) have no way to determine which one is removed from the code...

rust_src/src/keyboard.rs
```diff
@@ -32,6 +34,64 @@ use crate::{
 #[cfg(feature = "window-system")]
 use crate::remacs_sys::cancel_hourglass;
 
+#[derive(Clone, Copy, Debug)]
+pub struct Event(LispObject);
+
```
shaleh: Several of these functions would be more efficient as `match` against as_cons instead of using if and then force. Also, this style would make for easier debugging.

rust_src/src/keymap.rs
```diff
@@ -12,7 +12,7 @@ use crate::{
     data::{aref, fset, indirect_function, set},
     eval::{autoload_do_load, unbind_to},
     indent::indent_to,
-    keyboard::lucid_event_type_list_p,
+    keyboard::{self, lucid_event_type_list_p},
```
shaleh: In lists you used a different import style for the module itself. We should pick on style and stick with it. I have no preference

agraven: I'm sticking with the style that doesn't use `self`, then.

rust_src/src/keyboard.rs
agraven: Fixed.

rust_src/src/keymap.rs
```diff
@@ -673,4 +675,80 @@ pub fn copy_keymap(keymap: LispObject) -> LispObject {
...
+    if value.is_nil() || value.is_integer() {
+        return Qnil;
+    }
+
+    // If the result of the ordinary keymap lookup is an interactive
+    // command, look for a key binding (i.e. remapping) for that command.
+    if !no_remap && value.is_symbol() {
+        let value = unsafe { Fcommand_remapping(value, position, Qnil) };
+        if value.is_not_nil() {
+            return value;
```
brotzeit: Would it make sense to use `match` to avoid the additional variable. If not, I would at least rename the variable. I think it's easier to read if we don't shadow the variable here.

rust_src/src/keyboard.rs
```diff
@@ -32,6 +34,66 @@ use crate::{
...
+    // Note that the C macro expects an event head, this method expects an event which it exctracts the head from itself.
+    /// Extracts the kind from an event's head.
+    pub fn head_kind(self) -> LispObject {
+        lists::get(self.head().into(), Qevent_kind)
+    }
+
+    // replaces EVENT_START
+    /// Extracts the starting position from a composite event.
+    pub fn start(self) -> LispObject {
+        car_safe(cdr_safe(self.0))
```
nickdrozd: These could be a little clearer with pattern matching. Something like

```rust
    match self.0.into() {
        Some((_, (appropriate_name, _)) => appropriate_name,
        None => Qnil,
    }
```

Is that right? Probably not, but you get the idea. That way the structure of the object is easier to see.

Similar for all other uses of car / cdr. Really, there are few reasons for them to be used in Rust at all.

shaleh: The issue with pattern matching is we need a car and then a cdr and either may not be there. So the code above would need to be something like:
```
match self.0.into() {
        Some(((_, appropriate_name), _)) => appropriate_name,
        None => Qnil,
}
```

I am not sure that works. Since it needs to evaluate to None when any of the internal transformations do not work.

rust_src/src/keymap.rs
agraven: I don't think match is able to help here, but renaming the variable does make things a lot more clear.

rust_src/src/keyboard.rs
agraven: I'm starting to doubt this works for method works for substituting several car/cdr calls at all unless we add some more From impl's, and even then it might not.

Calling `.into()` once makes the compiler complain that the inner tuple is not a `LispObject`, and calling it twice makes the type of the first into call uninferrable, and adding  any type annotations does not fix it.

nickdrozd: I get it mixed up sometimes -- we can convert arbitrarily nested tuples in LispObjects, but not vice versa. It would be nice though.

PR #1377, opened 2019-03-13
Port get-buffer-create

### LINKED ISSUE #1331: Finish / re-implement get-buffer-create
This #786 needs updating. It was written before bindgen and before we added `LispBufferOrName`.

Options:
 - @nitishch finishes it
 - someone else makes a new PR
 - someone does the port but pastes it into the existing PR.

agraven:
Hoo boy, this ended up being a big one. I'm creating a separate PR for the `reset_local_variables` function.

I'm on the fence on whether the check for an existing buffer of the supplied name should be done in the method or only in the lisp function. The latter prevents you from shooting yourself in the foot, but the former is more flexible. Alternatively the method in it's current form could be changed to a private one called something like init_new, and the uniqueness check could be done in create_new.

There may also be a smarter of accessing `QSFundamental`, but bindgen refuses to expose it even if I declare it in an extern. I'm alse not entirely happy about keeping the bit of c code in what's now `reset_per_buffer_values`, but I also don't entirely trust trying to use a [rust version of offsetof](https://stackoverflow.com/questions/38819569/how-to-get-a-pointer-to-a-containing-struct-from-a-pointer-to-a-member) on a struct defined in C. Both of these concerns are about `reset_local_variables` though, so I suppose they should be discussed in the method's PR.

Will close #1331.


rust_src/src/buffers.rs
```diff
@@ -105,6 +110,90 @@ pub type LispBufferRef = ExternalPtr<Lisp_Buffer>;
 pub type LispOverlayRef = ExternalPtr<Lisp_Overlay>;
 
 impl LispBufferRef {
+    pub fn create_new(name: LispStringRef) -> LispBufferRef {
+        if name.is_empty() {
+            error!("Empty string for buffer name is not allowed");
+        }
+
+        let mut b = unsafe { ExternalPtr::new(allocate_buffer()) };
```
shanavas786: why not `LispBufferRef::new` ? here

rust_src/src/buffers.rs
```diff
@@ -105,6 +110,90 @@ pub type LispBufferRef = ExternalPtr<Lisp_Buffer>;
...
+        b.width_table_ = Qnil;
+        b.set_prevent_redisplay_optimizations_p(true);
+
+        // An ordinary buffer normally doesn't need markers to handle BEGV and ZV.
+        b.pt_marker_ = Qnil;
+        b.begv_marker_ = Qnil;
+        b.zv_marker_ = Qnil;
+
+        let name: LispStringRef = unsafe { Fcopy_sequence(name.into()) }.force_string();
+        unsafe { name.u.s }.intervals = ptr::null_mut();
```
shanavas786: it would be better to create a `LispStringRef::set_intervals` as we may need this often.

rust_src/src/buffers.rs
agraven: Because it would conflict with `ExternalPtr`'s `new`

agraven: Though following the standard of a lot of other functions that initialize objects, it would probably be more in line with convention to name it `make_new` or just `make`

rust_src/src/buffers.rs
agraven: Done.

rust_src/src/buffers.rs
```diff
@@ -105,6 +110,90 @@ pub type LispBufferRef = ExternalPtr<Lisp_Buffer>;
...
+
+        // An ordinary buffer normally doesn't need markers to handle BEGV and ZV.
+        b.pt_marker_ = Qnil;
+        b.begv_marker_ = Qnil;
+        b.zv_marker_ = Qnil;
+
+        let mut name: LispStringRef = unsafe { Fcopy_sequence(name.into()) }.force_string();
+        name.set_intervals(ptr::null_mut());
+        b.name_ = name.into();
+        b.undo_list_ = if name.byte_at(0) != b' ' { Qnil } else { Qt };
```
shaleh: Would adding a `bool_as_lisp` function help make our checks read better?

`b.undo_list_ = bool_as_lisp(name.byte_at(0) == b' ');`

shanavas786: may be add a trait with method `as_lisp` which in turn calls `into` ??

shaleh: My point was that sometimes the explicit naming makes the code easier to read. `as_lisp` is no better than `into` really.

shanavas786:
triggering

brotzeit:
`set_per_buffer_value` got lost.

PR #1380, opened 2019-03-13
Port system-groups

yshalenyk:
Port `system-groups`. Windows implemention is to discuss, I think.


rust_src/src/dired_unix.rs
```diff
@@ -765,3 +765,19 @@ pub fn get_users() -> LispObject {
 
     list(&unames)
 }
+
+pub fn get_groups() -> LispObject {
+    let mut group_names = Vec::new();
+    let mut done = false;
+
```
shaleh: A more idiomatic approach might be:
```
loop {
    let gr = unsafe { getgrent() };
    if gr.is_null() {
        break;
    }
    group_names.push(unsafe { build_string((*gr).gr_name) });
}
```

rust_src/src/dired_windows.rs
```diff
@@ -39,3 +39,7 @@ pub fn get_users() -> LispObject {
     unames.push(get_user_real_login_name());
     list(&mut unames)
 }
+
+pub fn get_groups() -> LispObject {
+    Qnil
```
shaleh: This matches the `IFDEF` from the C code.

test/rust_src/src/dired-tests.el
```diff
@@ -35,3 +35,14 @@
         (should (= (length (system-users)) 1)))
     (progn
       (should (>= (length (system-users)) 1)))))
+
+(ert-deftest test-system-groups ()
+  (should-error (eval '(system-groups 'rms)) :type 'wrong-number-of-arguments)
+  ;; The result should be a list of >= 1 group name(s) on all Unix and GNU systems.
+  ;; Windows should be a ?????
```
shaleh: Windows should be nil.

rust_src/src/dired_unix.rs
shaleh: A more idiomatic Rust library would let you write:
```
while let Some(gr) = getgrent() {
    // use gr here
}
```

But libc exposes the C API mostly.

yshalenyk: Good catch, thank you. Looks much nicer.

test/rust_src/src/dired-tests.el
yshalenyk: Thank you, I've fixed this test

test/rust_src/src/dired-tests.el
```diff
@@ -35,3 +35,14 @@
         (should (= (length (system-users)) 1)))
     (progn
       (should (>= (length (system-users)) 1)))))
+
+(ert-deftest test-system-groups ()
+  (should-error (eval '(system-groups 'rms)) :type 'wrong-number-of-arguments)
```
nickdrozd: No need for `eval` here. `(should-error (system-groups 'rms) ...)` should work fine.

test/rust_src/src/dired-tests.el
```diff
@@ -35,3 +35,14 @@
...
+  ;; The result should be a list of >= 1 group name(s) on all Unix and GNU systems.
+  ;; Windows should be a nil
+  (should (eq 'cons (type-of (system-groups))))
```
nickdrozd: This won't be true on Windows. `(system-groups)` will return nil, which is not a cons.

I would do these tests this way:

```emacs-lisp
(let ((system-groups (system-groups)))
  (if (eq system-type 'windows-nt)
      (should-not system-groups)
    (should system-groups)
    (should (listp system-groups))
    (should (< 0 (length system-groups)))))
```

rust_src/src/dired_unix.rs
```diff
@@ -765,3 +765,16 @@ pub fn get_users() -> LispObject {
...
 }
+
+pub fn get_groups() -> LispObject {
+    let mut group_names = Vec::new();
+    loop {
+        let gr = unsafe { getgrent() };
+        if gr.is_null() {
+            break;
+        }
+        group_names.push(unsafe { build_string((*gr).gr_name) });
```
shaleh: What about the `DECODE_SYSTEM` call from the original C?

rust_src/src/coding.rs
```diff
@@ -98,4 +98,16 @@ pub fn encode_file_name(fname: LispStringRef) -> LispStringRef {
     unsafe { c_encode_file_name(fname.into()) }.into()
 }
 
+/// Implements DECODE_SYSTEM macro
+/// Decode the string `input_string` using the specified coding system
+/// for system functions, if any.
+pub fn decode_system(input_string: LispStringRef) -> LispStringRef {
```
shaleh: `into()` is used in Rust to transition from one type to another aka casting. The From and Into traits are related to each other.

```
let foo = LispObject::from(1);
// and
let bar: LispObject = 1.into();
```
are equivalent. If you manually transform the calls to into() in this function you will see `LispStringRef::from(input_string)` which is already a string. Clippy points these out for you.

rust_src/src/coding.rs
```diff
@@ -98,4 +98,16 @@ pub fn encode_file_name(fname: LispStringRef) -> LispStringRef {
...
+    let local_coding_system: LispObject = unsafe { globals.Vlocale_coding_system }.into();
+    if !local_coding_system.is_nil() {
```
shaleh: How about flipping this so it reads `if coding system is nil { input_string } else { code_convert_string_norecord(...) }`.
Does that read as well or better?

yshalenyk: Sounds legit to me, fixed now

rust_src/src/coding.rs
yshalenyk: ok, thank you... really need to develop "run clippy" habit :)

shaleh: Clippy can be frustrating because it runs a build. So you end up `cargo clean; cargo clippy` otherwise it fools you into believing everything is fine.

brotzeit:
Looks good ?

rust_src/src/dired_unix.rs
brotzeit: Do we need a wrapper for `getgrent` to apply @shaleh 's proposal ?

PR #1384, opened 2019-03-14
Port select-frame

yshalenyk:


rust_src/src/frames.rs
```diff
@@ -661,4 +664,23 @@ pub fn visible_frame_list() -> LispObject {
...
+/// this function is called.  If you are using a window system, the
+/// previously selected frame may be restored as the selected frame
+/// when returning to the command loop, because it still may have
+/// the window system's input focus.  On a text terminal, the next
+/// redisplay will display FRAME.
+///
+/// This function returns FRAME, or nil if FRAME has been deleted.
+#[lisp_fn(min = "1")]
+pub fn select_frame(frame: LispObject, norecord: LispObject) -> LispObject {
+    unsafe { do_switch_frame(frame.into(), 1, 0, norecord.into()) }
```
shaleh: Why are `frame` and `norecord` being cast?

rust_src/src/frames.rs
```diff
@@ -661,4 +664,23 @@ pub fn visible_frame_list() -> LispObject {
...
+/// The selection of FRAME lasts until the next time the user does
+/// something to select a different frame, or until the next time
+/// this function is called.  If you are using a window system, the
+/// previously selected frame may be restored as the selected frame
+/// when returning to the command loop, because it still may have
+/// the window system's input focus.  On a text terminal, the next
+/// redisplay will display FRAME.
+///
+/// This function returns FRAME, or nil if FRAME has been deleted.
+#[lisp_fn(min = "1")]
```
shaleh: This misses the interactive "e". Look at rust_src/admin/rustize-emacs.

rust_src/src/frames.rs
yshalenyk: My mistake, fixed now...

rust_src/src/frames.rs
yshalenyk: Thanks, cool trick :)

shaleh:
Would you like to give porting `do_switch_frame` a go?

shaleh:
Typically, we would port `frame` as a `LispFrameRef`. In this case, it would immediately be cast back into a LispObject to call the C function `do_switch_frame`. So there is little value. However, if the C function is also ported then the Rust types have more usefulness.

yshalenyk:
ok, I will try to do it.

yshalenyk:
Sorry for the delay, folks.  I've added an initial implementation of `do_switch_frame`, it's still incomplete but I think a little review would help ;)

rust_src/src/frames.rs
```diff
@@ -31,9 +35,18 @@ impl LispFrameRef {
         !self.terminal.is_null()
     }
 
+    pub fn focus_frame(self) -> LispObject {
```
shaleh: We typically avoid getter routines unless they are required.

rust_src/src/frames.rs
```diff
@@ -31,9 +35,18 @@ impl LispFrameRef {
...
+        self.focus_frame.into()
+    }
+
     // Awaiting Wilfred#1264
```
shaleh: Is this comment still valid? If not, let's drop it too.

rust_src/src/frames.rs
```diff
@@ -195,6 +208,115 @@ impl From<LispFrameLiveOrSelected> for LispFrameRef {
...
+    mut frame: LispObject,
+    track: i32,
+    for_deletion: i32,
+    norecord: LispObject,
+) -> LispObject {
+    let sf: LispFrameRef = selected_frame();
+
+    // If FRAME is a switch-frame event, extract the frame we should
+    // switch to.
+    // TODO: this could be probably better
```
shaleh: Something like this should work. Giving the variables descriptive names helps the reader (which may be you next time....).

```
if let Some((event_type (event_frame, _))) = frame.as_cons() {
    if event_type.eq(Qswitch_frame) {
        frame = event_frame;
    }
}
```

rust_src/src/frames.rs
```diff
@@ -195,6 +208,115 @@ impl From<LispFrameLiveOrSelected> for LispFrameRef {
...
+    if let Some(event) = frame.as_cons() {
+        let (head, tail) = event.into();
+        if head.eq(Qswitch_frame) {
+            if let Some(f) = tail.as_cons() {
+                frame = f.car().into();
+            }
+        }
+    }
+
+    // This used to say CHECK_LIVE_FRAME, but apparently it's possible for
```
shaleh: We should update comments as needed to keep them relevant to the Rust port. Since `CHECK_LIVE_FRAME` is no longer used this could say "This used to check for a live frame,....".

rust_src/src/frames.rs
```diff
@@ -195,6 +208,115 @@ impl From<LispFrameLiveOrSelected> for LispFrameRef {
...
+        // If a frame's focus has been redirected toward the currently
+        // selected frame, we should change the redirection to point to the
+        // newly selected frame.  This means that if the focus is redirected
+        // from a minibufferless frame to a surrogate minibuffer frame, we
+        // can use `other-window' to switch between all the frames using
+        // that minibuffer frame, and the focus redirection will follow us
+        // around.
+
+        // Apply it only to the frame we're pointing to.
+        #[cfg(feature = "window-system")]
```
shaleh: I believe this can be written as `if cfg(feature = "window-system") {`

rust_src/src/frames.rs
```diff
@@ -195,6 +208,115 @@ impl From<LispFrameLiveOrSelected> for LispFrameRef {
...
+//
+// FOR_DELETION non-zero means that the selected frame is being
+// deleted, which includes the possibility that the frame's terminal
+// is dead.
+//
+// The value of NORECORD is passed as argument to Fselect_window.
+#[no_mangle]
+pub extern "C" fn do_switch_frame(
+    mut frame: LispObject,
+    track: i32,
```
shaleh: It looks like `track` is really `bool`. Does that work?

rust_src/src/frames.rs
```diff
@@ -195,6 +208,115 @@ impl From<LispFrameLiveOrSelected> for LispFrameRef {
...
+        // around.
+
+        // Apply it only to the frame we're pointing to.
+        #[cfg(feature = "window-system")]
+        {
+            if track != 0 && f.is_gui_window() {
+                let xfocus = unsafe { x_get_focus_frame(f.as_mut()) };
+                let focus: Option<LispFrameRef> =
+                    xfocus.as_frame().and_then(|f| f.focus_frame().into());
+                if (focus.map(|f| f.eq(&sf)).is_some())
```
shaleh: perhaps an `if let Some()` here as well?

rust_src/src/frames.rs
```diff
@@ -195,6 +208,115 @@ impl From<LispFrameLiveOrSelected> for LispFrameRef {
...
+                }
+            }
+        }
+        unsafe {
+            current_frame = frame;
+            // TODO: need to expose variable last_nonminibuf_frame
+            // if ! sf.minibuffer_window().is_minibuffer_only() {
+            //     last_nonminibuf_frame = sf.as_mut();
+            // }
+            Fselect_window(f.selected_window.into(), norecord);
```
shaleh: `Fselect_window` has been ported. Use `select_window_lisp`.

rust_src/src/frames.rs
```diff
@@ -661,6 +783,25 @@ pub fn visible_frame_list() -> LispObject {
...
+/// this function is called.  If you are using a window system, the
+/// previously selected frame may be restored as the selected frame
+/// when returning to the command loop, because it still may have
+/// the window system's input focus.  On a text terminal, the next
+/// redisplay will display FRAME.
+///
+/// This function returns FRAME, or nil if FRAME has been deleted.
+#[lisp_fn(min = "1", intspec = "e")]
+pub fn select_frame(frame: LispObject, norecord: LispObject) -> LispObject {
+    do_switch_frame(frame, 1, 0, norecord)
```
shaleh: This can become `do_switch_frame_rust(frame, true, false, norecord).into()` once you make the Rust version.

rust_src/src/frames.rs
yshalenyk: PR still open, I've pulled code from it to simplify merge...

rust_src/src/frames.rs
yshalenyk: seems working simply with bools

rust_src/src/frames.rs
yshalenyk: I've tried, not really. Travis builds failing ):

rust_src/src/frames.rs
```diff
@@ -34,9 +41,12 @@ impl LispFrameRef {
         !self.terminal.is_null()
     }
 
-    // Awaiting Wilfred#1264
+    // Dragged from https://github.com/remacs/remacs/pull/1264
```
shaleh: This comment can simply be dropped. Once #1264 is merged this comment will have no meaning.

yshalenyk:
I wonder how to deal with Travis builds errors. Basically, it's complaining about unused `track`  argument, which is not used without "window-system" feature.

shaleh:
This means you do not have enough `cfg` protection. Which leads to `track` being exposed when it is not needed/used.

agraven:
I would recommend at least for now to just rename `track` to `_track`.

rust_src/src/frames.rs
```diff
@@ -691,4 +698,135 @@ pub extern "C" fn get_frame_param(frame: LispFrameRef, prop: LispObject) -> Lisp
...
+        #[cfg(feature = "window-system")]
+        {
+            if _track && target_frame.is_gui_window() {
+                let xfocus = unsafe { x_get_focus_frame(target_frame.as_mut()) };
+
+                if let Some(focus) = xfocus.as_frame().and_then(|f| f.focus_frame.as_frame()) {
+                    if focus.eq(&current_frame_ref)
+                        || target_frame
+                            .minibuffer_window
+                            .eq(current_frame_ref.selected_window)
```
brotzeit: Do we need the `NILP (focus)` from the C code ?

rust_src/src/frames.rs
```diff
@@ -691,4 +698,135 @@ pub extern "C" fn get_frame_param(frame: LispFrameRef, prop: LispObject) -> Lisp
...
+            }
+        }
+        let minibuf_window: LispWindowRef = current_frame_ref.minibuffer_window.into();
+        unsafe {
+            current_frame = frame;
+            if minibuf_window.is_minibuffer_only() {
+                last_nonminibuf_frame = current_frame_ref.as_mut();
+            }
+        }
+        select_window_lisp(target_frame.selected_window, norecord);
```
brotzeit: Don't we need this:
```c
#ifdef HAVE_WINDOW_SYSTEM
  if (!frame_ancestor_p (f, sf))
#endif
    internal_last_event_frame = Qnil;
```

rust_src/src/frames.rs
shaleh: Yes, we likely do. Which means this code needs to be a little more complex. Which means that `if let Some()` needs to be refactor and broken up.

shaleh:
@yshalenyk are you interested in picking this back up?

yshalenyk:
@shaleh yes, I've rebased on top of current master and tweaked code based on review comments. CI failed I will fix it this evening

rust_src/src/frame.rs
```diff
@@ -811,4 +821,151 @@ pub fn frame_char_width(_frame: LispFrameLiveOrSelected) -> i32 {
...
+
+            let top_frame = unsafe { (*tty).top_frame };
+
+            if !frame.eq(top_frame) {
+                // TODO: refactor this code
+                unsafe {
+                    if let Some(mut tf) = top_frame.as_frame() {
+                        SET_FRAME_VISIBLE(tf.as_mut(), 2);
+                    }
+                    SET_FRAME_VISIBLE(target_frame.as_mut(), 1);
```
agraven: The comments from the C code could be copied over here

rust_src/src/frame.rs
```diff
@@ -857,96 +857,103 @@ pub extern "C" fn do_switch_frame(
     // This used to check for a live frame, but apparently it's possible for
     // a switch-frame event to arrive after a frame is no longer live,
     // especially when deleting the initial frame during startup.
-    if let Some(mut target_frame) = frame.as_live_frame() {
-        if target_frame.eq(&current_frame_ref) {
```
agraven: I think this one got lost in the translation to a match statement. It could be added back my adding a
```rust
    Some(target_frame) if target_frame.eq(&current_frame_ref) => return frame,
```
branch to the start of the match

PR #1391, opened 2019-03-21
Port frame-face-alist

### LINKED ISSUE #1376: Port frame-face-alist
```c
DEFUN ("frame-face-alist", Fframe_face_alist, Sframe_face_alist,
       0, 1, 0,
       doc: /* Return an alist of frame-local faces defined on FRAME.
For internal use only.  */)
  (Lisp_Object frame)
{
  return decode_live_frame (frame)->face_alist;
}
```

take-cheeze:
closes #1376

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -661,4 +661,11 @@ pub fn visible_frame_list() -> LispObject {
     filter_frame_list(LispFrameRef::is_visible)
 }
 
+/// Return an alist of frame-local faces defined on FRAME.
+/// For internal use only.
+#[lisp_fn]
+pub fn frame_face_alist(frame: LispObject) -> LispObject {
+    frame.as_live_frame_or_error().face_alist
+}
+
 include!(concat!(env!("OUT_DIR"), "/frames_exports.rs"));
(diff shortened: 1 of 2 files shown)


rust_src/src/frames.rs
```diff
@@ -661,4 +661,11 @@ pub fn visible_frame_list() -> LispObject {
     filter_frame_list(LispFrameRef::is_visible)
 }
```
shaleh: Welcome to the project.

The structure of this port is correct. However, it misses a couple of finer details.

1) `0, 1, 0,` in the original DEFUN means "min arguments 0, max arguments 1, not interactive". That last 0 is read kind of like a NULL. If the function was interactive it would look like `0, 1, "p"`. Because the min is different from the max, the `lisp_fn` macro needs to have a min parameter. `#[lisp_fn(min = "0")]`.
2) Instead of used `as_live_frame_or_error` we have created types to represent this use case. This means that `frame` should be defined as `LispFrameLiveOrSelected` instead of `LispObject`.

Look at the definition of `frame_position` in this file and compare it to https://github.com/emacs-mirror/emacs/blob/c569cceb2d334564d320d2b4098c855db7eb88a0/src/frame.c#L3574 which is what it looked like originally.

take-cheeze: @shaleh 
Thanks for the review. Was a good tutorial to port.
I've fixed as you pointed out:
https://github.com/remacs/remacs/pull/1391/commits/3d121927d97a1d719b8649bd38160168966f8db0

PR #1392, opened 2019-03-22
Port window-pixel-left

ereslibre:
Fixes: https://github.com/remacs/remacs/issues/1374

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1850,6 +1850,14 @@ pub fn window_lines_pixel_dimensions(
     unsafe { Fnreverse(rows) }
 }
 
+/// Return left pixel edge of window WINDOW.
+/// WINDOW must be a valid window and defaults to the selected one.
+#[lisp_fn(min = "0")]
+pub fn window_pixel_left(window: LispWindowValidOrSelected) -> i32 {
+    let window: LispWindowRef = window.into();
+    window.pixel_left
+}
+
 /// Return top pixel edge of window WINDOW.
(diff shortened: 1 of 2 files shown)


rust_src/src/windows.rs
```diff
@@ -1850,6 +1850,14 @@ pub fn window_lines_pixel_dimensions(
     unsafe { Fnreverse(rows) }
 }
 
+/// Return left pixel edge of window WINDOW.
+/// WINDOW must be a valid window and defaults to the selected one.
+#[lisp_fn(min = "0")]
+pub fn window_pixel_left(window: LispWindowValidOrSelected) -> EmacsInt {
+    let window: LispWindowRef = window.into();
+    window.pixel_left.into()
```
shaleh: Perhaps this could simply return `pixel_left` and have a return type of `i32`. If we need to have an `EmacsInt` the value can be cast at point of use or we can come back and update this function.

Good work. This PR is ready to use as is.

ereslibre: I just started to look into the code so at a very early stage; I see that https://github.com/remacs/remacs/blob/8a1e30df09a5ccba2f111f375b3dcd2f802133d1/rust_src/src/windows.rs#L1856-L1859 might need to be adapted as well then (in a follow-up PR maybe?)

ereslibre:
Didn't have the chance to get to the computer yet, will tackle at soon as possible.

brotzeit:
I just wanted to trigger the build.

ereslibre:
PR updated, thanks for the feedback!

PR #1394, opened 2019-03-22
Port lock-buffer and unlock-buffer

### LINKED ISSUE #928: Port lock-buffer and unlock-buffer
```c
DEFUN ("lock-buffer", Flock_buffer, Slock_buffer,
       0, 1, 0,
       doc: /* Lock FILE, if current buffer is modified.
FILE defaults to current buffer's visited file,
or else nothing is done if current buffer isn't visiting a file.

If the option `create-lockfiles' is nil, this does nothing.  */)
  (Lisp_Object file)
{
  if (NILP (file))
    file = BVAR (current_buffer, file_truename);
  else
    CHECK_STRING (file);
  if (SAVE_MODIFF < MODIFF
      && !NILP (file))
    lock_file (file);
  return Qnil;
}

DEFUN ("unlock-buffer", Funlock_buffer, Sunlock_buffer,
       0, 0, 0,
       doc: /* Unlock the file visited in the current buffer.
If the buffer is not modified, this does nothing because the file
should not be locked in that case.  */)
  (void)
{
  if (SAVE_MODIFF < MODIFF
      && STRINGP (BVAR (current_buffer, file_truename)))
    unlock_file (BVAR (current_buffer, file_truename));
  return Qnil;
}
```

liff:
Fixes #928. Introduces `filelock.rs`.

Added `LispBufferRef.modified_since_save()` since grepping for `SAVE_MODIFF` in `src` shows it could be useful in other places as well.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -362,6 +362,11 @@ impl LispBufferRef {
         unsafe { (*self.text).chars_modiff }
     }
 
+    /// Check if buffer was modified since its file was last read or saved.
+    pub fn modified_since_save(self) -> bool {
+        self.modifications_since_save() < self.modifications()
+    }
+
     pub fn overlay_modifications(self) -> EmacsInt {
         unsafe { (*self.text).overlay_modiff }
     }
@@ -1127,7 +1132,7 @@ pub fn buffer_file_name(buffer: LispBufferOrCurrent) -> LispObject {
 #[lisp_fn(min = "0")]
 pub fn buffer_modified_p(buffer: LispBufferOrCurrent) -> bool {
     let buf: LispBufferRef = buffer.into();
-    buf.modifications_since_save() < buf.modifications()
+    buf.modified_since_save()
 }
 
 /// Return the name of BUFFER, as a string.
--- a/rust_src/src/filelock.rs
+++ b/rust_src/src/filelock.rs
@@ -4,8 +4,8 @@ use remacs_macros::lisp_fn;
 
 use crate::{
     lisp::LispObject,
-    remacs_sys::{lock_file, unlock_file},
     remacs_sys::Qstringp,
+    remacs_sys::{lock_file, unlock_file},
     threads::ThreadState,
 };
 
--- a/rust_src/src/filelock.rs
+++ b/rust_src/src/filelock.rs
@@ -1,8 +1,11 @@
+//! Lock files for editing.
+
 use remacs_macros::lisp_fn;
 
 use crate::{
     lisp::LispObject,
-    remacs_sys::{lock_file, unlock_file, Qstringp},
+    remacs_sys::{lock_file, unlock_file},
+    remacs_sys::Qstringp,
     threads::ThreadState,
 };
 
--- a/rust_src/src/filelock.rs
+++ b/rust_src/src/filelock.rs
@@ -1,0 +1,43 @@
+use remacs_macros::lisp_fn;
+
+use crate::{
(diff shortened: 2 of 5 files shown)


rust_src/src/filelock.rs
```diff
@@ -0,0 +1,43 @@
+use remacs_macros::lisp_fn;
```
shaleh: Look at the other files and add a similar comment to the start of this one.

rust_src/src/filelock.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use crate::{
+    lisp::LispObject,
+    remacs_sys::{lock_file, unlock_file, Qstringp},
```
shaleh: import `Qfoo` items on their own.

test/rust_src/src/filelock-tests.el
```diff
@@ -0,0 +1,46 @@
+;;; filelock-tests.el --- Tests for filelock.rs
```
shaleh: Bravo for adding tests!

rust_src/src/filelock.rs
liff: Stole comment from `src/filelock.c`.

rust_src/src/filelock.rs
```diff
@@ -0,0 +1,46 @@
...
+    remacs_sys::Qstringp,
+    threads::ThreadState,
+};
+
+/// Lock FILE, if current buffer is modified.
+/// FILE defaults to current buffer's visited file,
+/// or else nothing is done if current buffer isn't visiting a file.
+///
+/// If the option `create-lockfiles' is nil, this does nothing.
+#[lisp_fn(min = "0", name = "lock-buffer")]
```
shaleh: Good job handling the min and adding a name. However, the name is not needed here. We only use the `foo_lisp` style naming scheme when there is a conflicting function with the same name as the DEFUN. Is there a `lock_buffer` or `unlock_buffer` function in the code?

rust_src/src/buffers.rs
```diff
@@ -362,6 +362,11 @@ impl LispBufferRef {
         unsafe { (*self.text).chars_modiff }
     }
 
+    /// Check if buffer was modified since its file was last read or saved.
+    pub fn modified_since_save(self) -> bool {
```
shaleh: Good call. This is a useful addition.

rust_src/src/filelock.rs
liff: Dropped from `lock_buffer`.

A public `unlock_buffer`, however, [exists](https://github.com/remacs/remacs/blob/11d0cb791daf6a440599c4ab1f9b1bb2db8b3b2d/src/filelock.c#L792).

rust_src/src/filelock.rs
shaleh: Perfect 👍

rust_src/src/filelock.rs
```diff
@@ -0,0 +1,46 @@
+//! Lock files for editing.
+
+use remacs_macros::lisp_fn;
+
+use crate::{
+    lisp::LispObject,
+    remacs_sys::{lock_file, unlock_file},
+    remacs_sys::Qstringp,
```
shaleh: Thanks. We do this in an effort to cut down on git conflicts as we add new entries.

shaleh:
`(setq rust-format-on-save t)` in your config will ensure rustfmt is always run.

liff:
Oops, rustfmted.

Issue #1393, opened 2019-03-22
make type error for ARM aarch64 u8/i8


MarcusE1W:
Hi.

I am trying to compile remacs on a Pinebook with ARM processor running Manjaro ARM for aarch64 architecture.

It goes well for quite some time but then during the `make` process the following error appears:

(this is just a short section of the log)
```
.
.
   Compiling remacs v0.1.0 (/home/mmw/local/remacs/rust_src)
   Compiling remacs-macros v0.1.0 (/home/mmw/local/remacs/rust_src/remacs-macros)
error[E0308]: mismatched types
  --> src/callproc.rs:62:31
   |
62 |         unsafe { build_string(NULL_DEVICE.as_ptr() as *const i8) }.into()
   |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u8, found i8
   |
   = note: expected type `*const u8`
              found type `*const i8`

error[E0308]: mismatched types
  --> src/callproc.rs:69:13
   |
69 |             encoded_file.const_data_ptr() as *const i8,
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u8, found i8
   |
   = note: expected type `*const u8`
              found type `*const i8`

error[E0308]: mismatched types
  --> src/callproc.rs:78:17
   |
78 |                 "Opening process input file".as_ptr() as *const i8,
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u8, found i8
   |
   = note: expected type `*const u8`
              found type `*const i8`
.
.
.
```
The actual list of errors is longer and other file are involved, but I guess they are follow up errors? If required I could maybe get the full log.

Any idea how to fix that ?

shaleh:
Ah, the classic "char" is "unsigned char" on some ARM platforms.

In the remacs port we need to consistently use a data type that matches the platform. Clearly `i8` is not it. Can you try `libc::c_char` and see if that helps. If it does we can switch out any place where a C char pointer is used.

MarcusE1W:
I can try. 
How would I do that?
Unfortunately I don"t know much about Rust or programming.

jamcha-aa:
[...]

PR #1397, opened 2019-03-25
Fix ARM aarch64 error

jamcha-aa:
Fix errors of [#1393](https://github.com/remacs/remacs/issues/1393).

### DIFF
--- a/rust_src/src/callproc.rs
+++ b/rust_src/src/callproc.rs
@@ -59,14 +59,14 @@ pub fn call_process_lisp(args: &mut [LispObject]) -> LispObject {
             ThreadState::current_buffer_unchecked().directory_.into(),
         )
     } else {
-        unsafe { build_string(NULL_DEVICE.as_ptr() as *const i8) }.into()
+        unsafe { build_string(NULL_DEVICE.as_ptr() as *const libc::c_char) }.into()
     };
 
     let encoded_file = encode_file_name(infile);
 
     let filefd = unsafe {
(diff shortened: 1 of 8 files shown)


shaleh:
An annoyance with the current `rustup` is it does not automatically update installed components.
We need to run this whenever the version in `rust-toolchain changes: `rustup component add rustfmt-preview`. Otherwise you have the wrong version of rustfmt. Also, set `(setq rust-format-on-save t)` in your (r)emacs config and rust mode will call rustfmt whenever you save.

BTW, rustup is smart and will try to run the correct version of rustfmt based on the configuration of your current project. So do not worry about having conflicting versions installed as long as all were installed with `rustup component add`.

shaleh:
The change looks reasonable though. Let's see what @MarcusE1W says.

shaleh:
@jamcha-aa I hand edited the PR so the rest of the tests can run. If you need to make updates, be sure to pull first.

BTW, it is good practice to make a branch for each PR instead of committing to master in your own repo. This lets you work on multiple items and reduces the chance of messing up your local tree.

jamcha-aa:
Thanks @shaleh  I will split the branch and develop it before making future PRs.

MarcusE1W:
Happy to report that the 'jamcha-aa:master' branch works on my Pinebook.

jamcha-aa:
Thanks for merging. I'd be happy to test out PRs for supporting ARM architectures.

PR #1400, opened 2019-03-26
[WIP] Port trace-redisplay

### LINKED ISSUE #1014: Port trace-redisplay
```c
DEFUN ("trace-redisplay", Ftrace_redisplay, Strace_redisplay, 0, 1, "P",
       doc: /* Toggle tracing of redisplay.
With ARG, turn tracing on if and only if ARG is positive.  */)
  (Lisp_Object arg)
{
  if (NILP (arg))
    trace_redisplay_p = !trace_redisplay_p;
  else
    {
      arg = Fprefix_numeric_value (arg);
      trace_redisplay_p = XINT (arg) > 0;
    }

  return Qnil;
}
```

liff:
Fixes #1014.

Added Rust features for all of the Emacs checking flags while at it. Let me know if they are not (going to be) needed and I’ll just drop them.

### DIFF
--- a/rust_src/build.rs
+++ b/rust_src/build.rs
@@ -127,9 +127,15 @@ impl<'a> ModuleParser<'a> {
     }
 
     pub fn run(&mut self, in_file: impl BufRead) -> Result<ModuleData, BuildError> {
+        lazy_static! {
+            static ref FEATURE_RE: Regex =
+                Regex::new(r#"^#\[cfg\(feature *= *"([^"]+)"\)\]"#).unwrap();
+        }
+
         let mut mod_data = ModuleData::new(self.info.clone());
         let mut reader = in_file.lines();
         let mut has_include = false;
+        let mut require_feature: Option<String> = None;
 
         while let Some(next) = reader.next() {
             let line = next?;
@@ -152,6 +158,11 @@ impl<'a> ModuleParser<'a> {
                     }
                 } else {
                     self.fail(1, "unexpected end of file");
+                }
+            } else if line.starts_with("#[cfg") {
+                match FEATURE_RE.captures(&line) {
+                    Some(caps) => require_feature = Some(caps[1].to_string()),
+                    _ => (),
                 }
             } else if line.starts_with("#[lisp_fn") {
                 let line = if line.ends_with("]") {
@@ -193,11 +204,18 @@ impl<'a> ModuleParser<'a> {
                     let line = next?;
 
                     if let Some(func) = self.parse_c_export(&line, name)? {
-                        mod_data.lisp_fns.push(func);
+                        if require_feature
+                            .map(|f| is_feature_activated(&f))
+                            .unwrap_or(true)
+                        {
+                            mod_data.lisp_fns.push(func);
+                        }
                     }
                 } else {
                     self.fail(1, "unexpected end of file");
                 }
+
+                require_feature = None;
             } else if line.starts_with("include!(concat!(env!(\"OUT_DIR\"),") {
                 has_include = true;
             } else if line.starts_with("/*") && !line.ends_with("*/") {
@@ -207,6 +225,8 @@ impl<'a> ModuleParser<'a> {
                         break;
                     }
                 }
+            } else {
+                require_feature = None;
             }
         }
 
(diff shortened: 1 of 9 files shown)


rust_src/src/xdisp.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Toggle tracing of redisplay.
+/// With ARG, turn tracing on if and only if ARG is positive.
+#[cfg(feature = "glyph-debug")]
+#[lisp_fn(min = "0", intspec = "P")]
+pub fn trace_redisplay(arg: Option<EmacsInt>) {
+    unsafe {
+        match arg {
+            None => trace_redisplay_p = !trace_redisplay_p,
+            Some(n) if n > 0 => trace_redisplay_p = true,
```
shaleh: While the guard is interesting, could this not be written more simply as `Some(n) => trace_redisplay_p = n > 0,`?

rust_src/src/xdisp.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use remacs_macros::lisp_fn;
+
+use crate::remacs_sys::{trace_redisplay_p, EmacsInt};
+
+/// Toggle tracing of redisplay.
+/// With ARG, turn tracing on if and only if ARG is positive.
+#[cfg(feature = "glyph-debug")]
+#[lisp_fn(min = "0", intspec = "P")]
+pub fn trace_redisplay(arg: Option<EmacsInt>) {
```
shaleh:
```suggestion
pub fn trace_redisplay(arg: InteractiveNumericPrefix) {
```

We have a type called `InteractiveNumericPrefix` which replaces what used to be `Fprefix_numeric_value`.

rust_src/src/xdisp.rs
liff: Indeed, fixed, thanks.

rust_src/src/xdisp.rs
liff: Changed. Should now match original function more closely (adjusted tests).

liff:
I added some ad-hoc logic to deal with exporting conditionally compiled lisp functions. It does very simple parsing of `#[cfg` lines and so may lead to surprises if one tries to use anything more complicated than `#[cfg(feature = "abc")]`.

Any suggestions for better ways of handling that?

rust_src/build.rs
```diff
@@ -127,9 +127,15 @@ impl<'a> ModuleParser<'a> {
     }
 
     pub fn run(&mut self, in_file: impl BufRead) -> Result<ModuleData, BuildError> {
```
shaleh: Can this be extended to include clippy warnings as well? Because I do not think they work currently.

shaleh:
We should update one of the Travis builds to enable the compiler option. Edit the .travis.yml as part of this commit and it will immediately apply.

shaleh:
This all looks really good @liff

rust_src/build.rs
liff: I don’t quite understand what you mean. Can you elaborate a bit?

liff:
> We should update one of the Travis builds to enable the compiler option. Edit the .travis.yml as part of this commit and it will immediately apply.

You mean add `--enable-checking=glyps` for all current jobs or add new jobs with that flag set?

liff:
Came up with slightly different (better?) way of handling the `cfg` attributes. Now the build script just captures them as-is and passes them through to the exported entries.

test/rust_src/src/xdisp-tests.el
```diff
@@ -0,0 +1,18 @@
...
+(require 'ert)
+
+(ert-deftest xdisp-tests--trace-redisplay-base ()
+  "Check (trace-redisplay) base cases.
+   Only run when configured with --enable-checks=glyphs or equivalent."
+  (when (fboundp 'trace-redisplay)
+    (should-error (trace-redisplay 1 2))
+    (should-error (trace-redisplay nil 2))
+
+    (should (eq nil (trace-redisplay)))
```
nickdrozd: You can use `(should-not ...)` instead of `(should (eq nil ...))`.

shaleh:
> > We should update one of the Travis builds to enable the compiler option. Edit the .travis.yml as part of this commit and it will immediately apply.
> 
> You mean add `--enable-checking=glyps` for all current jobs or add new jobs with that flag set?

Maybe pick one of the current tests and include it there?

nickdrozd:
What's left on this one? It looks like the `glyph-debug` stuff is what https://github.com/remacs/remacs/pull/1366 needs.

brotzeit:
This looks really good...

brotzeit:
@liff Thanks!

PR #1401, opened 2019-03-27
Port window-left-column

### LINKED ISSUE #1399: Port window-left-column
```c
DEFUN ("window-left-column", Fwindow_left_column, Swindow_left_column, 0, 1, 0,
       doc: /* Return left column of window WINDOW.
This is the distance, in columns, between the left edge of WINDOW and
the left edge of the frame's window area.  For instance, the return
value is 0 if there is no window to the left of WINDOW.

WINDOW must be a valid window and defaults to the selected one.  */)
  (Lisp_Object window)
{
  return make_number (decode_valid_window (window)->left_col);
}
```

aclonegeek:
Closes #1399

### DIFF
--- a/test/rust_src/src/windows-tests.el
+++ b/test/rust_src/src/windows-tests.el
@@ -24,13 +24,13 @@
   (should(eq (window-dedicated-p (selected-window)) nil)))
 
 (ert-deftest window-dedicated-set-not-nil()
-  ;; set widnow to dedicated
+  ;; set window to dedicated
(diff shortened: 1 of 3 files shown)


test/rust_src/src/windows-tests.el
```diff
@@ -113,6 +113,14 @@
     (should (= (window-pixel-top) (window-pixel-top w1)))
     (should-not (= (window-pixel-top w1) (window-pixel-top w2)))))
```
shaleh: Tests! Good work!

PR #1404, opened 2019-03-27
Rollup to 767780169b

brotzeit:


src/xdisp.c
```diff
@@ -10144,23 +10144,21 @@ include the height of both, if present, in the return value.  */)
      directionality, and regions that begin and end in text of the
      same directionality.  */
   it.bidi_p = false;
```
shanavas786: `git diff master 767780169b src/xdisp.c` gives some more changes.

```
--- a/src/xdisp.c
+++ b/src/xdisp.c
@@ -9598,7 +9598,6 @@ move_it_to (struct it *it, ptrdiff_t to_charpos, int to_x, int to_y, int to_vpos
       it->current_x = line_start_x;
       line_start_x = 0;
       it->hpos = 0;
-      it->line_number_produced_p = false;
       it->current_y += it->max_ascent + it->max_descent;
       ++it->vpos;
       last_height = it->max_ascent + it->max_descent;
...
@@ -13888,7 +13885,7 @@ redisplay_internal (void)
   if (!fr->glyphs_initialized_p)
     return;
 
-#if defined (USE_GTK) || defined (HAVE_NS)
+#if defined (USE_X_TOOLKIT) || defined (USE_GTK) || defined (HAVE_NS)
   if (popup_activated ())
     return;
 #endif
...
@@ -17542,7 +17539,7 @@ redisplay_window (Lisp_Object window, bool just_this_one_p)
 
       if (FRAME_WINDOW_P (f))
        {
-#if defined (HAVE_NTGUI) \
+#if defined (USE_X_TOOLKIT) || defined (HAVE_NTGUI) \
     || defined (HAVE_NS) || defined (USE_GTK)
          redisplay_menu_p = FRAME_EXTERNAL_MENU_BAR (f);
 #else
...
@@ -22489,11 +22486,6 @@ Value is the new character position of point.  */)
                    new_pos += (row->reversed_p ? -dir : dir);
                  else
                    new_pos -= (row->reversed_p ? -dir : dir);
-                 new_pos = clip_to_bounds (BEGV, new_pos, ZV);
-                 /* If we didn't move, we've hit BEGV or ZV, so we
-                    need to signal a suitable error.  */
-                 if (new_pos == PT)
-                   break;
                }
              else if (BUFFERP (g->object))
                new_pos = g->charpos;
...
```
and
```
-#if defined (USE_GTK)
+#if defined (USE_X_TOOLKIT) || defined (USE_GTK)
```
at some more places

brotzeit: Oh. Thanks for taking a look at it.

brotzeit: Hmm I can't find it. This should be the same like `git diff 3959892419 767780169b` right ?

shanavas786: Note sure.
Well, these things don't show up on merging. So its good to go :+1:

brotzeit: Your last rollup was to 3959892419. So the diff would be between 3959892419 + 1 and 767780169b.

shanavas786:
@brotzeit anything pending on this ?

brotzeit:
I wanted to give others the chance to look at it. But it seems to be fine I guess.

brotzeit:
I will merge this when I'm home.

PR #1406, opened 2019-03-27
Replace `clippy::cyclomatic_complexity` with `clippy::cognitive_complexity`

### LINKED ISSUE #1402: lint `clippy::cyclomatic_complexity` has been renamed to `clippy::cognitive_complexity`
The code needs to be updated for the new Clippy version.

Bira:
Resolves #1402 

This lint rule has been renamed in the latest Clippy version

These are the only two references to the name I found by grepping the code.

I tested it by temporarily removing these lines and running `cargo clippy` again. This caused cognitive complexity warnings to be displayed when they weren't before.

I'm pretty new to both Rust and the emacs source code, but this is something I figured I could handle :).


brotzeit:
Thanks!

PR #1413, opened 2019-04-01
Port get-text-property

Bira:
This should solve #1363

I followed the documentation on this one, though I'm not yet at the point where I can port the two C functions called by get-text-property.

The test is copied from `test/src/textprop-tests.el`.


test/rust_src/src/textprop-tests.el
```diff
@@ -0,0 +1,72 @@
+;;; textprop-tests.el --- Test suite for text properties.
```
shaleh: Did you forget to remove the original code?

rust_src/src/textprop.rs
```diff
@@ -19,4 +22,13 @@ pub fn get_char_property(position: EmacsInt, prop: LispObject, object: LispObjec
     unsafe { get_char_property_and_overlay(position.into(), prop, object, ptr::null_mut()) }
 }
 
+/// Return the value of POSITION's property PROP, in OBJECT.
+/// OBJECT should be a buffer or a string; if omitted or nil, it defaults
+/// to the current buffer.
+/// If POSITION is at the end of OBJECT, the value is nil.
+#[lisp_fn(min = "2")]
+pub fn get_text_property(position: EmacsInt, prop: LispObject, object: LispObject) -> LispObject {
+    unsafe { textget(Ftext_properties_at(position.into(), object), prop) }
```
shaleh: Do you want to try porting `text-properties-at` as well?

Bira:
I've deleted the copied tests and replaced them with a simple smoke test for get-text-property, which passes.

rust_src/src/textprop.rs
Bira: Yes! I think I'll do it in another pull request, though, since it's likely to take me longer than this one (I'm a newbie :)).

shaleh:
My apologies that my guidance was not clear. Copying the tests was the right choice. Be sure to delete the original location when you do.

Bira:
Ah, I see. Sorry about the confusion! I'll fix this ASAP.

Bira:
I believe this is ready to be reviewed again. I moved the textprop tests to the new location, while keeping the new smoke test I added there.

PR #1414, opened 2019-04-03
Port variable_binding_locus

shanavas786:


shanavas786:
```
Test xref-elisp-test-find-defs-defun-el-defvar-c condition:
    (ert-test-failed
     ((should
       (looking-at
	(or expected-source ...)))
      :form
      (looking-at "DEFVAR_LISP (\"abbrev-mode\"")
      :value nil))
```
But `DEFVAR_LISP ("abbrev-mode")` has not been removed in this change.
or am I missing something ?

rust_src/src/lisp.rs
```diff
@@ -72,6 +72,10 @@ impl LispObject {
     pub fn from_float(v: EmacsDouble) -> Self {
         unsafe { make_float(v) }
     }
+
+    pub fn fallthrough() -> Self {
```
shaleh: `fallthrough` has been misinterpreted. This is a bit of C idiom. The intent is to continue into the next `switch` branch.

```
switch(value) {
    case FOO: {
        do stuff....
        break; // stop processing here
    }
    case BAR: {
        do stuff....
        FALLTHROUGH;
    }
    case THING: {
        some stuff.....
        break;
    }
}
```

If the `BAR` condition is met both the BAR code and the THING code are run. `FALLTHROUGH` is a meaningless no-op value that allows the developer to visually see that the intent is to fall through. Without it, one has to spot the CASE branches missing their `break` statements.

I suspect this is where your bugs are happening, the extra code is not running.

rust_src/src/symbols.rs
```diff
@@ -487,6 +489,46 @@ pub fn local_variable_p(mut symbol: LispSymbolRef, buffer: LispBufferOrCurrent)
...
+/// If the current binding is global (the default), the value is nil.
+#[lisp_fn]
+pub fn variable_binding_locus(mut symbol: LispSymbolRef) -> LispObject {
+    // Make sure the current binding is actually swapped in.
+    unsafe {
+        symbol.find_value();
+    }
+    symbol = symbol.get_indirect_variable();
+
+    match symbol.get_redirect() {
```
shaleh: Continuing my comment about `fallthrough` I think this has to be written as if/else statements with explicit returns OR the `symbol_localized` code is copied into the fallthrough else that is currently written. Either way, yuck.

rust_src/src/lisp.rs
shanavas786: Oops!!
Thanks a lot :)

rust_src/src/symbols.rs
shanavas786: absorbing `symbol_localized` into an inner function looks somewhat cleaner.

rust_src/src/symbols.rs
```diff
@@ -487,6 +489,48 @@ pub fn local_variable_p(mut symbol: LispSymbolRef, buffer: LispBufferOrCurrent)
...
+    }
+    symbol = symbol.get_indirect_variable();
+
+    fn localized_handler(sym: LispSymbolRef) -> LispObject {
+        // For a local variable, record both the symbol and which
+        // buffer's or frame's value we are saving.
+        let blv = unsafe { sym.get_blv() };
+
+        if local_variable_p(sym, LispBufferOrCurrent::Current) {
+            current_buffer().into()
```
shaleh:
```suggestion
            current_buffer()
```
Clippy identifies this as an unneeded conversion.

shanavas786:
Done

shaleh:
@shanavas786 the Mac builds are segfaulting. Give this another look.

shaleh:
I think the Mac builds are simply broken again. Not related to this change.

PR #1417, opened 2019-04-05
Port set-window-hscroll

### LINKED ISSUE #1408: Port set-window-hscroll
```c
DEFUN ("set-window-hscroll", Fset_window_hscroll, Sset_window_hscroll, 2, 2, 0,
       doc: /* Set number of columns WINDOW is scrolled from left margin to NCOL.
WINDOW must be a live window and defaults to the selected one.
Clip the number to a reasonable value if out of range.
Return the new number.  NCOL should be zero or positive.

Note that if `automatic-hscrolling' is non-nil, you cannot scroll the
window so that the location of point moves off-window.  */)
  (Lisp_Object window, Lisp_Object ncol)
{
  CHECK_NUMBER (ncol);
  return set_window_hscroll (decode_live_window (window), XINT (ncol));
}
```

emlun:
Fixes #1408.

I'm not sure I have the best approach to computing `hscroll_max`, but at least it should be safe - unless I'm mistaken, the only reason the `try_from` call should fail is if `MOST_POSITIVE_FIXNUM` is too big for an `isize`. `std::cmp::max(MOST_POSITIVE_FIXNUM, isize::max_value())` is not applicable since the two arguments are of different types.

### DIFF
--- a/test/rust_src/src/windows-tests.el
+++ b/test/rust_src/src/windows-tests.el
@@ -152,3 +152,30 @@
     (should (set-window-fringes w1 1 1 nil))
     (should-not (set-window-fringes w1 1 1 nil))
     (should (set-window-fringes w1 1 2 nil))))
+
+(ert-deftest window-hscroll ()
+  "Effectively tests both `window-hscroll' (the getter) and
+  `set-window-hscroll' (the setter)."
+
+  ;; Can we change hscroll on this window?
+  (set-window-hscroll nil 42)
+  (should (= (window-hscroll) 42))
+  (should (= (window-hscroll nil) 42))
+  (set-window-hscroll nil 7)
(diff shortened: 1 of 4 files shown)


test/rust_src/src/windows-tests.el
```diff
@@ -152,3 +152,30 @@
     (should (set-window-fringes w1 1 1 nil))
     (should-not (set-window-fringes w1 1 1 nil))
     (should (set-window-fringes w1 1 2 nil))))
+
```
shaleh: Bravo!

emlun: Thanks! Note that the tests don't cover the `set_prevent_redisplay_optimizations_p` and `set_suspend_auto_hscroll` side effects, because I don't know what those mean...

PR #1420, opened 2019-04-07
Rollup to b1d7e58520

brotzeit:


brotzeit:
Should be good.

brotzeit:
I'd like to do some more merges, but I need a little help =)

PR #1426, opened 2019-04-09
Port text-properties-at

Bira:
This is my initial attempt at porting the text-properties-at function. 

The code as it is right now causes a failure in `test/lisp/emacs-lisp/ert-x-tests.el`, in the tests for `propertized-string`:

```
Test ert-propertized-string condition:
    (args-out-of-range
     #("abcd" 1 2
       (a b)
       2 4
       (c t))
     4 5)
   FAILED  2/9  ert-propertized-string
```

I'm pretty sure I'm doing something wrong in the call to validate_interval_range, but unfortunately my knowledge isn't great enough to figure out what the right way to call it would be. Any help on that would be appreciated :).

Also, is there a way to run a single lisp test file?

### DIFF
--- a/rust_src/src/textprop.rs
+++ b/rust_src/src/textprop.rs
@@ -47,8 +47,9 @@ pub fn text_properties_at(mut position: LispObject, mut object: LispObject) -> L
         object = ThreadState::current_buffer().into();
     }
 
+    let position_pointer = &mut position as *mut LispObject;
     let i: *mut Lisp_Interval =
-        unsafe { validate_interval_range(object, &mut position, &mut position, false) };
+        unsafe { validate_interval_range(object, position_pointer, position_pointer, false) };
 
     if ptr::eq(i, ptr::null_mut()) {
         return Qnil;
@@ -59,12 +60,11 @@ pub fn text_properties_at(mut position: LispObject, mut object: LispObject) -> L
         // it means it's the end of OBJECT.
         // There are no properties at the very end,
         // since no character follows.
-        let position_isize = EmacsInt::from(position) as isize;
-        if position_isize == ((*i).total_length + (*i).position) {
-            return Qnil;
+        if EmacsInt::from(position) == (((*i).total_length + (*i).position) as EmacsInt) {
+            Qnil
+        } else {
+            (*i).plist
         }
-
-        (*i).plist
     }
 }
 
--- a/rust_src/src/textprop.rs
+++ b/rust_src/src/textprop.rs
@@ -8,8 +8,10 @@ use crate::{
     lisp::LispObject,
     numbers::LispNumber,
     remacs_sys::Ftext_properties_at,
-    remacs_sys::Qt,
-    remacs_sys::{get_char_property_and_overlay, set_text_properties, textget},
+    remacs_sys::{
+        get_char_property_and_overlay, set_text_properties, textget, validate_interval_range,
+    },
+    remacs_sys::{EmacsInt, Lisp_Interval, Qnil, Qt, ThreadState},
 };
 
 /// Return the value of POSITION's property PROP, in OBJECT.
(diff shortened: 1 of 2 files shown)


rust_src/src/textprop.rs
```diff
@@ -28,7 +29,39 @@ pub fn get_char_property(position: EmacsInt, prop: LispObject, object: LispObjec
...
+    unsafe { textget(text_properties_at(position.into(), object), prop) }
+}
+
+/// Return the list of properties of the character at POSITION in OBJECT.
+/// If the optional second argument OBJECT is a buffer (or nil, which means
+/// the current buffer), POSITION is a buffer position (integer or marker).
+/// If OBJECT is a string, POSITION is a 0-based index into it.
+/// If POSITION is at the end of OBJECT, the value is nil.
+#[lisp_fn(min = "1")]
+pub fn text_properties_at(mut position: LispObject, mut object: LispObject) -> LispObject {
```
brotzeit: I think you can make position a `LispNumber`.

Bira: I managed it but ended up breaking a lot more tests than before :). I'm sure it will be possible once validate_interval_range is ported.

rust_src/src/textprop.rs
```diff
@@ -28,7 +29,39 @@ pub fn get_char_property(position: EmacsInt, prop: LispObject, object: LispObjec
...
+    if object.is_nil() {
+        object = ThreadState::current_buffer().into();
+    }
+
+    let position_pointer = &mut position as *mut LispObject;
+    let i: *mut Lisp_Interval =
+        unsafe { validate_interval_range(object, position_pointer, position_pointer, false) };
+
+    if ptr::eq(i, ptr::null_mut()) {
```
shaleh:
```suggestion
    if i.is_null() {
```

`ptr` provides `is_null`.

Bira:
I finally found out what was causing the text failure I was getting here. I was totally wrong about the workings of the LENGTH macro for intervals. I solved this by adding a minimal LispIntervalRef implementation with only the things I needed for this specific case. 

Is there anything else I need to add to that before it can be considered a true "minimal" implementation ready for future expansion?

rust_src/src/textprop.rs
```diff
@@ -30,7 +30,42 @@ pub fn get_char_property(position: LispNumber, prop: LispObject, object: LispObj
...
+/// the current buffer), POSITION is a buffer position (integer or marker).
+/// If OBJECT is a string, POSITION is a 0-based index into it.
+/// If POSITION is at the end of OBJECT, the value is nil.
+#[lisp_fn(min = "1")]
+pub fn text_properties_at(mut position: LispObject, mut object: LispObject) -> LispObject {
+    if object.is_nil() {
+        object = ThreadState::current_buffer().into();
+    }
+
+    let i: LispIntervalRef = LispIntervalRef::new(unsafe {
```
shaleh: `validate_interval_range` is used as a constructor in the C code except in one place where it is only used for the validation. A more appropriate design for `LispIntervalRef` I think would be to add a `new_from_position()` factory method which takes the object and the position (only once). In this use case the final option is always false in the C code.

`let i: LispIntervalRef::new_from_position(object, position);`

This does however lose the "validated" which is likely important for the reader. Should the name be `new_validated_from_position`? `new_validated_position`? `new_position_with_validation`?

Bira: I have it as `from_valid_range` locally, with the validation itself separated into another function that's called internally. It seems to work in that the tests still pass, but I'm not 100% clear on what the XSETFASTINT macro does on the C version. It's being called to assign the value of `*begin` to itself, but I couldn't quite parse the bitwise operations it does. Is that where they increment these numbers by 1?

shaleh: For Rust we can essentially read `XSETFASTINT` as `let x: LispObject = LispObject::from(some_number);`.

Bira:
I've submitted my implementation of `LispIntervalRef::from_valid_range()` for review. It's _almost_ a port of `validate_interval_range` - should it be approved, the actual port could consist of wrapping this method in an extern C function.

Bira:
To clarify my comment from earlier, I didn't want to port `validate_interval_range` itself as part of this PR, but to use the new Rust method for the port of `text_properties_at`. I'm not quite sure the method would work for a general port as-is, since it's missing some of the pointer-wrangling of the C version, and I would rather investigate that in a separate PR. 

Of course, if you think it's better to do the full port in this branch I will work on it.

shaleh:
If that is the case, perhaps the right thing to do is have the method you created simply call the C function and then in a follow in PR you can replace the C call with the fully functioning Rust code.

Bira:
I've removed the LispIntervalRef changes to a separate branch, and implemented a private function with the absolute minimum logic I need to make text-properties-at work correctly. I would be glad to proceed with the LispIntervalRef work in a separate PR.

Bira:
Looks like that full port of validate_interval_range has enough "hidden" issues with it that it's a bit beyond my ability to fix properly. Still, does the code in this branch look good enough to merge?

shaleh:
@Bira do you have time to refresh it?

Bira:
I'll try to do that this week!

Bira:
I've refreshed it!

PR #1427, opened 2019-04-10
Port overlay_recenter

### LINKED ISSUE #433: Port overlay-recenter to Rust

Bira:
Closes #433 

This didn't break any tests locally, so I think I'm on the right track with this function. The compiler caught most of my mistakes :).

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -52,10 +52,11 @@ use crate::{
         Qinhibit_read_only, Qmakunbound, Qnil, Qoverlayp, Qpermanent_local, Qpermanent_local_hook,
         Qt, Qunbound, UNKNOWN_MODTIME_NSECS,
     },
-    remacs_sys::{Fcopy_sequence, Fmake_marker},
+    remacs_sys::{recenter_overlay_lists, Fcopy_sequence, Fmake_marker},
     strings::string_equal,
     textprop::get_text_property,
     threads::{c_specpdl_index, ThreadState},
+    util::clip_to_bounds,
     vectors::LispVectorlikeRef,
 };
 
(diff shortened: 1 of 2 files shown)


rust_src/src/buffers.rs
```diff
@@ -1402,6 +1403,22 @@ fn get_truename_buffer_1(filename: LispSymbolOrString) -> LispObject {
         .into()
 }
 
+/// Recenter the overlays of the current buffer around position POS.
+/// That makes overlay lookup faster for positions near POS (but perhaps slower
+/// for positions far away from POS).
+#[lisp_fn]
+pub fn overlay_recenter(pos: LispObject) -> LispObject {
```
shaleh: I am somewhat certain you can use `LispNumber` here, it uses coerce internally. Then below you use `pos.into()` to get the EmacsInt aka fixnum. Since this only returns Qnil you do not need to specify a return.

```suggestion
pub fn overlay_recenter(pos: LispNumber) {
```

rust_src/src/buffers.rs
```diff
@@ -1402,6 +1403,22 @@ fn get_truename_buffer_1(filename: LispSymbolOrString) -> LispObject {
...
+    let p = clip_to_bounds(
+        std::isize::MIN,
+        pos.as_fixnum_coerce_marker_or_error(),
+        std::isize::MAX,
+    );
+    unsafe {
+        recenter_overlay_lists(ThreadState::current_buffer_unchecked().as_mut(), p);
+    }
+    Qnil
```
shaleh: Not needed.

rust_src/src/buffers.rs
Bira: I had to use `to_fixnum()` on the call to `clip_to_bounds`, because the compiler complained that the `From<numbers::LispNumber>` trait wasn't implemented for i64. Is that the correct way?

PR #1431, opened 2019-04-13
Rollup to 2a1fe08307

brotzeit:


brotzeit:
I wonder if we can just keep `TEST_LOAD_EL ?= yes`...

brotzeit:
@shaleh What do you think ?

shaleh:
I need to look at it. Good work identifying the TEST_LOAD_EL change. I am unlikely to look at this for a couple of days.

brotzeit:
#1472

PR #1433, opened 2019-04-14
fix property functions: change args EmacsInt -> LispNumber

brotzeit:


brotzeit:
This is a bug fix. Would be nice if somebody can approve.

rust_src/src/editfns.rs
```diff
@@ -669,11 +669,11 @@ pub fn constrain_to_field(
             }
                 && (old_pos <= begv
                     || get_char_property(
-                        old_pos,
+                        old_pos.into (),
```
shaleh: Hmm, how did this pass rustfmt?

PR #1434, opened 2019-04-14
Port window-pixel-width-before-size-change (and -height)

Flowneee:
Fixes https://github.com/remacs/remacs/issues/1429

Currently lacks tests due to problem with ert: it doesn't actually change `window-pixel-width-before-size-change` during tests (probably because tests doesn't change frame and thus `window-size-change-functions` not invoked).

### DIFF
--- a/rust_src/src/windows.rs
+++ b/rust_src/src/windows.rs
@@ -1974,10 +1974,9 @@ pub fn set_window_fringes_lisp(
 /// `window-size-change-functions' was run.  It's zero if WINDOW was made
 /// after that.
 #[lisp_fn(min = "0")]
-pub fn window_pixel_width_before_size_change(window: LispWindowValidOrSelected) -> EmacsInt {
-    LispWindowRef::from(window)
-        .pixel_width_before_size_change
-        .into()
+pub fn window_pixel_width_before_size_change(window: LispWindowValidOrSelected) -> i32 {
+    let window: LispWindowRef = window.into();
+    window.pixel_width_before_size_change
 }
 
 /// Return pixel height of window WINDOW before last size changes.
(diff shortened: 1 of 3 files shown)


rust_src/src/windows.rs
```diff
@@ -1967,4 +1967,30 @@ pub fn set_window_fringes_lisp(
...
 
+/// Return pixel width of window WINDOW before last size changes.
+/// WINDOW must be a valid window and defaults to the selected one.
+///
+/// The return value is the pixel width of WINDOW at the last time
+/// `window-size-change-functions' was run.  It's zero if WINDOW was made
+/// after that.
+#[lisp_fn(min = "0")]
+pub fn window_pixel_width_before_size_change(window: LispWindowValidOrSelected) -> EmacsInt {
+    LispWindowRef::from(window)
```
brotzeit: I think we should use `let window: LispWindowRef = window.into();` to make things consistent.

rust_src/src/windows.rs
```diff
@@ -1967,4 +1967,30 @@ pub fn set_window_fringes_lisp(
...
+}
+
+/// Return pixel height of window WINDOW before last size changes.
+/// WINDOW must be a valid window and defaults to the selected one.
+///
+/// The return value is the pixel height of WINDOW at the last time
+/// `window-size-change-functions' was run.  It's zero if WINDOW was made
+/// after that.
+#[lisp_fn(min = "0")]
+pub fn window_pixel_height_before_size_change(window: LispWindowValidOrSelected) -> EmacsInt {
```
brotzeit: You can return i32 in both functions and drop the `into`.

rust_src/src/windows.rs
Flowneee: Makes sense.

rust_src/src/windows.rs
Flowneee: Is there any difference between `i32` and `EmacsInt` (which is i64, iirc) for Remacs? I see that some functions (like `window_new_pixel`) return `EmacsInt`, and some `i32`. Won't there be a problem when negative will be returned?

brotzeit: `EmacsInt` is a `i64`. I didn't take a look at `window_new_pixel`, but I think this was ported some time ago or it passed the review. `lisp_fn` casts the returned `i32` to a `LispObject`.

Flowneee: Thx for the explanation. Fixed.

PR #1435, opened 2019-04-14
Use const fn

nickdrozd:
Suggested by the "nursery" Clippy lint missing_const_for_fn


nickdrozd:
I don't know if this is needed, or if all the changes are correct.

shaleh:
https://doc.rust-lang.org/unstable-book/language-features/const-fn.html for reference.

PR #1437, opened 2019-04-17
Use Self

nickdrozd:
Suggested by the "pedantic" Clippy lint use_self


nickdrozd:
There are a few other places where `Self` could be used, but Clippy doesn't indicate it (eg https://github.com/remacs/remacs/blob/ef9173ee9ac6619f366c0b604ced51103a829cc9/rust_src/src/dired_unix.rs#L48). I don't know if that's deliberate or not, but in any case I'll leave them for now.

PR #1438, opened 2019-04-17
Add comparisons for lisp strings with native types

### LINKED ISSUE #1291: Can we make string comparisons more ergonomic?
It would be nice to allow comparison of LispStringRef and strings or byte slice via PartialEq.

A potential snag is Emacs is not UTF8 compliant internally. This means using Rust's string support tools will not work without us checking and raising errors.

agraven:
I'd like to have a `PartialEq` implementation that works for both `&str` and `&String`, but `PartialEq<Deref<Target=str>>` doesn't do the trick. Any advice here would be appreciated.

And feel free to bikeshed about the test name.

Closes #1291.

### DIFF
--- a/rust_src/src/multibyte.rs
+++ b/rust_src/src/multibyte.rs
@@ -193,9 +193,15 @@ impl PartialEq<&[u8]> for LispStringRef {
     }
 }
 
-impl PartialEq<&str> for LispStringRef {
-    fn eq(&self, other: &&str) -> bool {
+impl<'a> PartialEq<&'a str> for LispStringRef {
+    fn eq(&self, other: &&'a str) -> bool {
         self.as_slice() == other.as_bytes()
+    }
+}
+
+impl PartialEq<String> for LispStringRef {
+    fn eq(&self, other: &String) -> bool {
+        self == &other.as_str()
     }
 }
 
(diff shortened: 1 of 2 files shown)


nickdrozd:
Could you add in at least a few instances where this would be used?

agraven:
Shoot, I can't easily grep my way to finding places where this can be used. I'm gonna do a more thorough search then.

rust_src/src/strings.rs
```diff
@@ -227,6 +227,14 @@ fn str_equality() {
     assert!(!string_equal(string2, string3));
 }
```
shaleh: Let's include some words with letters beyond ASCII in this testing.

shaleh:
Agreed @nickdrozd.

rust_src/src/multibyte.rs
```diff
@@ -193,12 +193,18 @@ impl PartialEq<&[u8]> for LispStringRef {
     }
 }
 
-impl PartialEq<&str> for LispStringRef {
-    fn eq(&self, other: &&str) -> bool {
```
shaleh: Yes! Lifetimes finally showing up. Nice.

PR #1449, opened 2019-04-20
Port lisp_string_width

agraven:
I faced an ancient foe of mine which hath struck me down once before, and emerged victorious.

LispStringRef now has two methods for determining width, one which takes a precision argument, and one which does not, the latter being a convenience function for the former. Feel free to bikeshed about the names.

The `find_composition` wrapper function should probably also live in its own `composite.rs` file, but I considered outside the scope of this PR.

There's also a discussion to be had about what should be `usize` or `isize` where. I went with the approach of making as many variables usize as possible (where it made sense), but this may have caused a few more conversions than necessary.

I might also have missed some rust functions that have the functionality I'm using c functions for.


rust_src/src/multibyte.rs
```diff
@@ -950,3 +1026,45 @@ pub fn char_to_byte8(c: Codepoint) -> u8 {
...
+pub fn find_composition(
+    pos: usize,
+    limit: Option<usize>,
+    object: LispObject,
+) -> Option<(usize, usize, LispObject)> {
+    let pos = pos as isize;
+    let limit = limit.map_or(-1, |l| l as isize);
+    let mut start = 0;
+    let mut end = 0;
+    let mut prop = Qnil;
```
nickdrozd:
```suggestion
    let (mut start, mut end, mut prop) = (0, 0, Qnil);
```

rust_src/src/multibyte.rs
```diff
@@ -88,13 +92,85 @@ impl LispStringRef {
...
+
+            width = match width.checked_add(thiswidth) {
+                Some(w) => w,
+                None => string_overflow(),
+            };
+            i += chars;
+            b += bytes;
+        }
+
+        let n = precision.map(|_| (i, b));
```
nickdrozd: What is this line doing?

rust_src/src/multibyte.rs
```diff
@@ -88,13 +92,85 @@ impl LispStringRef {
...
+        let multibyte = self.len_chars() < self.len_bytes();
+        let distab = unsafe { buffer_display_table() };
+        // Sum width
+        let mut width = 0;
+        // Character index
+        let mut i = 0;
+        // Byte index
+        let mut b = 0;
+
+        while i < len {
```
nickdrozd: Could this be done with more sophisticated control flow? Maybe a for loop with a range or something? That way you could remove the i-mutation statement at the end of the loop. A really slick solution would also get rid of the b-mutation statement. (I don't know how, but that's my feeling.)

rust_src/src/multibyte.rs
```diff
@@ -88,13 +92,85 @@ impl LispStringRef {
...
+            let (chars, bytes, thiswidth) = unsafe {
```
nickdrozd: It might be clearer to calculate `(cmp_id, end)` in one block and then assign `(chars, bytes, thiswidth)` in a separate block. Right now the values and the variables are pretty far away from each other, which makes it a little hard to read.

rust_src/src/multibyte.rs
```diff
@@ -88,13 +92,85 @@ impl LispStringRef {
...
+                        string_char_and_length(self.const_data_ptr().add(b))
+                    } else {
+                        (self.as_slice()[b].into(), 1)
+                    };
+                    (1, bytes, char_width(c as i32, distab) as usize)
+                }
+            };
+
+            // Return if adding character exceeds precision
+            if let Some(precision) = precision {
```
nickdrozd: This early return check just relies on `thiswidth`, right? And that's calculated independently of `chars` and `bytes`? If so, you could calculate `thiswidth` first, then run the early return check, then calculate `chars` and `bytes` and save yourself a little extra work if it does return early.

rust_src/src/multibyte.rs
agraven: It's essentially a more compact way of writing
```rust
let n = if precision.is_some() {
    Some((i, b))
} else {
    None
}
```
It seemed a bit verbose to write it that way for something that simple, but I can add a comment like "n is Some((i, b)) if precision is Some" or just rewrite the line to the code above if it's necessary for clarity.

rust_src/src/multibyte.rs
agraven: That's going to be difficult considering that a single iteration can jump not only several bytes, but also several code points. Anything involving iterators is going to be significantly more clumsy than just mutating an index since it'll require a "skip counter" or something similar, putting us back to square one but uglier.

rust_src/src/multibyte.rs
agraven: That would definitely make things easier to follow.

rust_src/src/multibyte.rs
shaleh: @agraven I think part of the confusion is the alias above where you unwrap `precision` but here you are again using the original Option.

If I read this right, `n` is really a list or nil. The current `map` would be fine if there was a comment explaining why precision was involved in making the list. And maybe giving `n` a better name too. The Emacs C has too many shadowy corners and lack luster variable names are part of the problem.

rust_src/src/multibyte.rs
agraven: I think calculating thiswidth first scatters the core logic of the function too far apart for the slight optimisation to be worth the readability compromise

rust_src/src/multibyte.rs
```diff
@@ -88,13 +92,89 @@ impl LispStringRef {
...
+    /// paying attention to compositions.
+    ///
+    /// With `precision` argument, return the width of longest substring that
+    /// doesn't exceed `precision`, and the number of characters and bytes it
+    /// contains in the returned tuple.
+    pub fn display_width(self, precision: Option<usize>) -> (usize, Option<(usize, usize)>) {
+        // Manually determine if string is unibyte (lets us ignore multibyte
+        // handling in more cases).
+        let len = self.len_chars() as usize;
+        let multibyte = self.len_chars() < self.len_bytes();
```
shaleh: Can we use a method here? That way if the definition of multibyte changes we are not hunting the code base for it?

rust_src/src/multibyte.rs
```diff
@@ -950,3 +1030,43 @@ pub fn char_to_byte8(c: Codepoint) -> u8 {
...
+}
+
+#[no_mangle]
+pub extern "C" fn lisp_string_width(
+    string: LispObject,
+    precision: isize,
+    nchars: *mut isize,
+    nbytes: *mut isize,
+) -> isize {
+    let precision = if precision < 0 {
```
shaleh: An option here is using `TryInto` couple with `usize`.

shaleh:
This looks good @agraven, I need to sit down and really walk the logic.

How well is this covered by elisp tests? Could we write Rust tests?

agraven:
Rust tests should be pretty easy to do with mock strings. There's currently elisp tests in `test/rust_src/src/strings-tests.el` for unibyte strings, tabs, and compositions, however they don't test specifying precision as that functionality is not exported to lisp.

rust_src/src/multibyte.rs
agraven: Maybe, but I'm not sure. The reason the `is_multibyte` method isn't used is so that in cases where strings are marked as multibyte but their contents are still unibyte, we can skip checking for unibyte characters. Maybe a `is_multibyte_content` method that does this check could be added (the actual name is up for bikeshedding of course).

shaleh: If the String is marked "multibyte" shouldn't the code trust that and proceed accordingly?

rust_src/src/multibyte.rs
agraven: Done, that looks much nicer.

agraven:
So I realized rust tests can't be done to this because they can't link to C code, so the current elisp tests will have to suffice.

shaleh:
@agraven is this ready to go?

agraven:
Yes.

PR #1451, opened 2019-04-21
Port clear-charset-maps

### LINKED ISSUE #1174: Port clear-charset-maps
```c
DEFUN ("clear-charset-maps", Fclear_charset_maps, Sclear_charset_maps,
       0, 0, 0,
       doc: /*
Internal use only.
Clear temporary charset mapping tables.
It should be called only from temacs invoked for dumping.  */)
  (void)
{
  if (temp_charset_work)
    {
      xfree (temp_charset_work);
      temp_charset_work = NULL;
    }

  if (CHAR_TABLE_P (Vchar_unify_table))
    Foptimize_char_table (Vchar_unify_table, Qnil);

  return Qnil;
}
```

Bira:
Closes #1174 

I _think_ I'm on the right path here, but I'm not quite there yet. I get this error when running `make`, when trying to link/build the temacs executable:

```
.././rust_src/target/release/libremacs.a(remacs-d72c4967c380b3d7.remacs.254nx1kb-cgu.3.rcgu.o): In function `Fclear_charset_maps':
remacs.254nx1kb-cgu.3:(.text.Fclear_charset_maps+0x4): undefined reference to `temp_charset_work'
```

Part of the reason for this might be that I added a new header file to contain the `temp_charset_work` struct, as I mentioned in issue #1174. I've referenced it in charset.c and wrapper.h, should I add it somewhere else?

### DIFF
--- a/rust_src/src/charset.rs
+++ b/rust_src/src/charset.rs
@@ -4,7 +4,6 @@ use remacs_macros::lisp_fn;
 
 use crate::{
     hashtable::{HashLookupResult, LispHashTableRef},
-    libc::c_int,
     lisp::LispObject,
     remacs_sys::temp_charset_work,
     remacs_sys::Foptimize_char_table,
--- a/rust_src/src/charset.rs
+++ b/rust_src/src/charset.rs
@@ -7,8 +7,8 @@ use crate::{
     hashtable::{HashLookupResult, LispHashTableRef},
     lisp::LispObject,
     remacs_sys::temp_charset_work,
-    remacs_sys::Foptimize_char_table,
     remacs_sys::Qnil,
+    remacs_sys::{xfree, Foptimize_char_table},
     remacs_sys::{Vchar_unify_table, Vcharset_hash_table},
 };
 
@@ -28,6 +28,7 @@ impl LispObject {
 #[lisp_fn]
 pub fn clear_charset_maps() {
     unsafe {
+        xfree(temp_charset_work as *mut libc::c_void);
         temp_charset_work = ptr::null_mut();
 
         if Vchar_unify_table.is_char_table() {
--- a/rust_src/src/charset.rs
+++ b/rust_src/src/charset.rs
@@ -4,8 +4,12 @@ use remacs_macros::lisp_fn;
 
 use crate::{
     hashtable::{HashLookupResult, LispHashTableRef},
+    libc::c_int,
     lisp::LispObject,
-    remacs_sys::Vcharset_hash_table,
+    remacs_sys::temp_charset_work,
+    remacs_sys::Foptimize_char_table,
+    remacs_sys::Qnil,
+    remacs_sys::{Vchar_unify_table, Vcharset_hash_table},
 };
 
 impl LispObject {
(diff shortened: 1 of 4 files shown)


brotzeit:
Unfortunately bindgen doesn't notice new bindings. Try cargo clean in /rust_src.

brotzeit:
I will document it...

Bira:
Looks like I forgot to enable rust debug in the laptop. This is a clearer form of the error message above:

```
.././rust_src/target/debug/libremacs.a(remacs-822eccce6018751f.33so64k5df520t9w.rcgu.o): In function `remacs::charset::clear_charset_maps':
/home/bira/src/remacs/rust_src/src/charset.rs:(.text._ZN6remacs7charset18clear_charset_maps17hfc23dc89244d6510E+0x21): undefined reference to `temp_charset_work'

```

Bira:
I've tried both `cargo clean` and `make clean` before rebuilding, and the error above happens in both cases. I see the structure corresponding to the C struct in the generated bindings, but I guess there's some trouble during the linking stage there.

shaleh:
@Bira please check in what you have. The CI builds are failing because you have unused (aka dead) code.

src/tempcharset.h
```diff
@@ -0,0 +1,54 @@
+static struct
```
shaleh: Here is the point of confusion.

This line of code combined with line 31 is doing two different things. First it is defining a `struct` and second it is making a file local variable from that struct. `static` means "global to this file only, invisible otherwise" with the side effect of initializing the value to 0 (aka NULL in this case since it is a pointer).

Remove the static and the variable and give this struct a proper name like `struct CharseWork`. Then you can declare a `static` instance of it in C. Any Rust code that needs the variable will then need to access it via an accessor function like get_temp_charset_work(). It is fine to assign to the pointer. Once we move the rest of the code using the variable to Rust we can switch over to using Rust's `lazy_static` support.

Bira:
I believe this is ready now. Once again, I apologize for the confusion.

rust_src/src/charset.rs
```diff
@@ -18,6 +21,20 @@ impl LispObject {
     }
 }
 
+/// Internal use only.
+/// Clear temporary charset mapping tables.
+/// It should be called only from temacs invoked for dumping.
+#[lisp_fn]
+pub fn clear_charset_maps() {
+    unsafe {
+        *temp_charset_work = Default::default();
```
Bira: Is this OK, or should I assign `ptr::null_mut()` here instead? The C code freed the memory explicitly and assigned NULL. Would assigning `ptr::null_mut()` in Rust automatically free the memory?

brotzeit: AFAIK C still handles memory, so I guess you have to free it.

Bira: I apologize for the silly question, but I want to be sure I'm doing the right thing.

From what I've read Rust automatically frees memory once the values in question fall out of scope, but this function has several things going on with it that make me less sure this would happen here (this is a raw mutable pointer to a struct shared with C code via FFI).

If I replace the assignment above with `temp_charset_work = ptr:null_mut()`, will Rust still free the memory at the end of this function, or do I have to do something else to ensure the memory is freed?

brotzeit: Don't worry, I still don't understand most of this stuff after almost 2 years =)
But I think you have to do it the same way. Rust doesn't free globals from C.

rust_src/src/charset.rs
```diff
@@ -1,11 +1,15 @@
 //! Basic character set support.
 
 use remacs_macros::lisp_fn;
+use std::ptr;
```
shaleh: Forgive the nitpicking please.

The import order is:
 - std
 - external crates
 - internal crates.

rust_src/src/charset.rs
```diff
@@ -18,6 +22,20 @@ impl LispObject {
     }
 }
 
+/// Internal use only.
+/// Clear temporary charset mapping tables.
+/// It should be called only from temacs invoked for dumping.
+#[lisp_fn]
+pub fn clear_charset_maps() {
+    unsafe {
+        temp_charset_work = ptr::null_mut();
```
shaleh: The original code used `xfree`. Which means this code as written will leak memory. Rust cannot handle the C allocated memory. There are no lifetimes associated with it.

src/tempcharset.h
```diff
@@ -0,0 +1,56 @@
...
+    /* Table mapping Unicode characters to code-indices of the current
+       charset.  The first 0x10000 elements are for BMP (0..0xFFFF),
+       and the last 0x10000 are for SMP (0x10000..0x1FFFF) or SIP
+       (0x20000..0x2FFFF).  Note that there is no charset map that
+       uses both SMP and SIP.  */
+    unsigned short encoder[0x20000];
+  } table;
+} TempCharsetWork;
+
+TempCharsetWork *temp_charset_work;
```
shaleh: Declaring this variable in a header does not proper C guidelines. It needs to be defined in the C file and listed here as `extern`.

rust_src/src/charset.rs
Bira: It took me the longest time to learn xfree is defined in alloc.c and I could call it from here. I'll push it up as soon as the tests finish running locally.

shaleh: https://github.com/Wilfred/deadgrep <-- game changer for hunting around in the source.

rust_src/src/charset.rs
Bira: No worries! I'll fix this. It looks like my editor reordered the lines because I placed them too close together.

shaleh: not your editor, that was rustfmt.

rust_src/src/charset.rs
```diff
@@ -18,6 +23,21 @@ impl LispObject {
     }
 }
 
+/// Internal use only.
+/// Clear temporary charset mapping tables.
+/// It should be called only from temacs invoked for dumping.
+#[lisp_fn]
+pub fn clear_charset_maps() {
+    unsafe {
+        xfree(temp_charset_work as *mut libc::c_void);
```
shaleh: The pointer has to be checked first. Calling free on a NULL is not guaranteed to be safe. This is why the C code has an `if`.

Bira:
I've restored the missing conditional and reordered the imports. Sorry for all these small errors, I'll try my best not to do them again in the future :).

PR #1454, opened 2019-04-24
Port purecopy

nickdrozd:


nickdrozd:
I've been doing "just Rust" for a while, without actually porting anything. Getting back in it with an easy one, with the benefit of getting rid of a bunch of `unsafe` blocks.

rust_src/src/alloc.rs
```diff
@@ -95,4 +96,19 @@ pub fn record(args: &mut [LispObject]) -> LispObject {
...
+/// Does not copy symbols.  Copies strings without text properties.
+#[lisp_fn]
+pub fn purecopy(obj: LispObject) -> LispObject {
+    if unsafe { globals.Vpurify_flag.is_nil() } {
+        obj
+    } else if obj.is_marker() || obj.is_overlay() || obj.is_symbol() {
+        // Can't purify those.
+        obj
+    } else {
+        unsafe { c_purecopy(obj) }
```
shaleh: The C version is still called `purecopy`.

rust_src/src/alloc.rs
```diff
@@ -95,4 +96,19 @@ pub fn record(args: &mut [LispObject]) -> LispObject {
     }
 }
 
+/// Make a copy of object OBJ in pure storage.
+/// Recursively copies contents of vectors and cons cells.
+/// Does not copy symbols.  Copies strings without text properties.
+#[lisp_fn(name = "purecopy", c_name = "purecopy")]
+pub fn purecopy_lisp(obj: LispObject) -> LispObject {
```
shaleh: Clippy complains that both of the arms of the if/else result in `obj`.

nickdrozd: I thought about changing that, but the separate arms don't relate to each other, they just happen to return the same value.

nickdrozd:
Is there another Rust / C naming conflict here? Or am I missing something?

rust_src/src/alloc.rs
shaleh: That is a clippy error, not a warning.

shaleh: You could do this:
```
if unsafe { globals.Vpurify_flag.is_nil() } {
    obj
}

if obj.is_marker() || obj.is_overlay() || obj.is_symbol() {
    // Can't purify those.
    obj
} else {
    unsafe { c_purecopy(obj) }
}
```

PR #1455, opened 2019-04-24
Port get_frame_param

Bira:
I noticed this function was fairly simple and only called functions which already had been ported to Rust, so I made an attempt at porting it.

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -105,10 +105,10 @@ impl LispFrameRef {
         }
     }
 
-    pub fn get_param(self, prop: LispObject) -> Option<LispObject> {
+    pub fn get_param(self, prop: LispObject) -> LispObject {
         match assq(prop, self.param_alist) {
-            Qnil => None,
-            cons => Some(cdr(cons)),
+            Qnil => Qnil,
+            cons => cdr(cons),
         }
     }
 }
@@ -683,9 +683,9 @@ pub fn frame_face_alist(frame: LispFrameLiveOrSelected) -> LispObject {
 /// Return the value of frame parameter PROP in frame FRAME.
 #[no_mangle]
 pub extern "C" fn get_frame_param(frame: LispFrameRef, prop: LispObject) -> LispObject {
-    // I should be possible to use this method directly when we port
+    // It should be possible to use this method directly when we port
     // one of the original function's callers.
-    frame.get_param(prop).unwrap_or(Qnil)
+    frame.get_param(prop)
 }
 
 include!(concat!(env!("OUT_DIR"), "/frames_exports.rs"));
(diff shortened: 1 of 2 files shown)


rust_src/src/frames.rs
```diff
@@ -672,4 +680,12 @@ pub fn frame_face_alist(frame: LispFrameLiveOrSelected) -> LispObject {
     frame_ref.face_alist
 }
 
+/// Return the value of frame parameter PROP in frame FRAME.
+#[no_mangle]
+pub extern "C" fn get_frame_param(frame: LispFrameRef, prop: LispObject) -> LispObject {
+    // I should be possible to use this method directly when we port
```
shaleh: typo

rust_src/src/frames.rs
```diff
@@ -672,4 +680,12 @@ pub fn frame_face_alist(frame: LispFrameLiveOrSelected) -> LispObject {
     frame_ref.face_alist
 }
 
+/// Return the value of frame parameter PROP in frame FRAME.
+#[no_mangle]
+pub extern "C" fn get_frame_param(frame: LispFrameRef, prop: LispObject) -> LispObject {
```
shaleh: This is dangerous. `frame` might be NULL and never validate it. I see that the original C did not either so likely not a problem here. But something we should keep in mind while porting.

rust_src/src/frames.rs
```diff
@@ -103,6 +104,13 @@ impl LispFrameRef {
             0
         }
     }
+
```
shaleh: `LispObject` is already a kind of `Option`. Wrapping it in an `Option` does not buy us match currently. I'd probably write this:

```
map assq(prop, self.param_alist) {
    Qnil => Qnil,
    cons => cdr(cons),
}
```

src/frame.c
```diff
@@ -148,19 +148,6 @@ check_window_system (struct frame *f)
...
-
-Lisp_Object
-get_frame_param (register struct frame *frame, Lisp_Object prop)
-{
-  register Lisp_Object tem;
-
-  tem = Fassq (prop, frame->param_alist);
-  if (EQ (tem, Qnil))
-    return tem;
-  return Fcdr (tem);
```
shaleh: After reading this code again, I wonder if the following is an equivalent function:

```
match assq(prop, frame.param_alist).as_cons() {
    Some(cons) => cons.cdr(),
    None => Qnil,
}
```

rust_src/src/frames.rs
Bira: I'd be glad to handle the NULL case here. Should we also return NULL if the frame is null?

Bira:
I've used the implementation you suggested. As far as I can tell it works fine! 

Should I do something to handle the case where `frame` is null on the get_frame_param? A cursory look through the calls to this function seems to show the calling code performs the null check itself when instantiating a frame, or expects its own caller to perform it when receiving a frame as an argument. I could be wrong there, however.

rust_src/src/frames.rs
shaleh: Let's keep an eye on it. I suspect as the port continues this function won't live long.

PR #1457, opened 2019-04-27
Port frame-font-cache.

### LINKED ISSUE #1398: Port frame-font-cache
```c
DEFUN ("frame-font-cache", Fframe_font_cache, Sframe_font_cache, 0, 1, 0,
       doc: /* Return FRAME's font cache.  Mainly used for debugging.
If FRAME is omitted or nil, use the selected frame.  */)
  (Lisp_Object frame)
{
#ifdef HAVE_WINDOW_SYSTEM

  if (FRAME_WINDOW_P (f))
    return FRAME_DISPLAY_INFO (f)->name_list_element;
  else
#endif
    return Qnil;
}
```

Search for `#[cfg(feature = "window-system")]` in our code base to port this function.

cantsin:
Closes #1398.

Not sure how to add tests here, some pointers would be appreciated.

### DIFF
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -354,4 +354,31 @@ pub fn list_fonts(
     }
 }
 
+/// Return FRAME's font cache.  Mainly used for debugging.
+/// If FRAME is omitted or nil, use the selected frame.
+#[lisp_fn(min = "0")]
+pub fn frame_font_cache(frame: LispFrameLiveOrSelected) -> LispObject {
+    #[cfg(feature = "window-system")]
+    {
+        let frame: LispFrameRef = frame.into();
+        if frame.is_gui_window() {
+            unsafe {
+                #[cfg(feature = "window-system-x11")]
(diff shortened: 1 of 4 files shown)


brotzeit:
Thanks for the PR! I just realized we need to add `FONT_DEBUG` for this function.

cantsin:
Sure, I can add `FONT_DEBUG` into here if that makes sense.

brotzeit:
I actually doubt it makes sense. It isn't defined in configure.ac. Instead it's defined this way:

```c
#ifndef FONT_DEBUG
#define FONT_DEBUG
#endif
```

Looks odd to me ?

brotzeit:
Looks to me like we can remove this completely...

cantsin:
O_o that is weird. Alright then, removing.

Tangential question, how do we sync up with emacs code? Would removing `FONT_DEBUG` make it harder to sync up later?

brotzeit:
I don't think so since the affected files don't get changed frequently.

PR #1463, opened 2019-04-28
Port file-locked-p

liff:
As a sort of continuation to #1394.

Some notes/caveats:
 - No Windows support. remacs doesn’t seem to build on Windows right now so there’s no way to test :frowning_face: 
 - Adds dependency to [systemstat crate](https://crates.io/crates/systemstat), which provides, among other things, system boot time information.
 - Some duplication between `filelock.c` and `filelock.rs` remain, pending ports of [`lock_file`](https://github.com/remacs/remacs/blob/fa68c1f5df7834d063e8b9c7b8a262178e0059ca/src/filelock.c#L640), [`unlock_file`](https://github.com/remacs/remacs/blob/fa68c1f5df7834d063e8b9c7b8a262178e0059ca/src/filelock.c#L725), [`unlock_all_files`](https://github.com/remacs/remacs/blob/fa68c1f5df7834d063e8b9c7b8a262178e0059ca/src/filelock.c#L742) and [`unlock_buffer`](https://github.com/remacs/remacs/blob/fa68c1f5df7834d063e8b9c7b8a262178e0059ca/src/filelock.c#L757).

### DIFF
--- a/rust_src/src/filelock.rs
+++ b/rust_src/src/filelock.rs
@@ -13,12 +13,7 @@ use systemstat::Platform;
 #[cfg(unix)]
 use libc::{ELOOP, O_NOFOLLOW, O_RDONLY};
 #[cfg(unix)]
-use std::{
-    ffi::OsStr,
-    ffi::CString,
-    os::unix::ffi::OsStrExt,
-    os::unix::io::FromRawFd,
-};
+use std::{ffi::CString, ffi::OsStr, os::unix::ffi::OsStrExt, os::unix::io::FromRawFd};
 
 use crate::{
     coding::encode_file_name,
@@ -31,7 +26,6 @@ use crate::{
     threads::ThreadState,
 };
 use LockState::*;
-
 
 /// An arbitrary limit on lock contents length when it is stored in the
 /// contents of the lock file.  8 K should be plenty big enough in practice.
--- a/rust_src/src/filelock.rs
+++ b/rust_src/src/filelock.rs
@@ -2,12 +2,276 @@
 
 use remacs_macros::lisp_fn;
 
+use std::fs::{remove_file, File};
+use std::io::prelude::Read;
+use std::io::Error;
+use std::io::ErrorKind::{InvalidData, InvalidInput, NotFound, PermissionDenied};
+use std::io::Result;
+use std::path::{Path, PathBuf};
+
 use crate::{
+    coding::encode_file_name,
+    editfns::system_name,
+    fileio::expand_file_name,
     lisp::LispObject,
-    remacs_sys::Qstringp,
-    remacs_sys::{lock_file, unlock_file},
+    multibyte::LispStringRef,
+    remacs_sys::{emacs_open, lock_file, maybe_quit, unlock_file},
+    remacs_sys::{Qnil, Qstringp, Qt},
     threads::ThreadState,
 };
+
+/// An arbitrary limit on lock contents length when it is stored in the
+/// contents of the lock file.  8 K should be plenty big enough in practice.
+const MAX_LOCK_INFO: usize = 8 * 1024;
+
+/// The lock info as parsed from a string like `USER@HOST.PID:BOOT_TIME`.
+#[derive(PartialEq, Eq, Debug)]
+struct LockInfo {
+    /// The user that owns the lock.
+    user: String,
+    /// The host on which the file was locked.
+    host: String,
+    /// The ID of the process that locked the file.
+    pid: i32,
+    /// Boot time of the system that locked the file, as Unix epoch seconds.
+    boot_time: Option<i64>,
+}
+
+impl LockInfo {
+    /// Parses a string like `USER@HOST.PID:BOOT_TIME` into a [`LockInfo`].
+    /// Returns [`None`] if the parse fails.
+    fn parse(data: &str) -> Option<LockInfo> {
+        // Treat "\357\200\242" (U+F022 in UTF-8) as if it were ":" (Bug#24656).
+        // This works around a bug in the Linux CIFS kernel client, which can
+        // mistakenly transliterate ':' to U+F022 in symlink contents.
+        // See https://bugzilla.redhat.com/show_bug.cgi?id=1384153
+        let boot_time_sep = |c| c == ':' || c == '\u{F022}';
+
+        // The USER is everything before the last @.
+        let (user, rest) = data.split_at(data.rfind('@')?);
+        let user = user.to_string();
+
+        // The HOST is everything after the '@' to the last '.'.
+        let (at_host, rest) = rest.split_at(rest.rfind('.')?);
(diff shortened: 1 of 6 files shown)


brotzeit:
Wow =)

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+        // See https://bugzilla.redhat.com/show_bug.cgi?id=1384153
+        let boot_time_sep = |c| c == ':' || c == '\u{F022}';
+
+        // The USER is everything before the last @.
+        let (user, rest) = data.split_at(data.rfind('@')?);
+        let user = user.to_string();
+
+        // The HOST is everything after the '@' to the last '.'.
+        let (at_host, rest) = rest.split_at(rest.rfind('.')?);
+        let host = at_host[1..].to_string();
```
shaleh: Why the slicing of `at_host` when `user` above could be used directly?

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+    /// The file is locked by this Emacs process.
+    LockedByUs,
+    /// The file is locked by some other Emacs process identified
+    /// by [`LockInfo`].
+    LockedBy(LockInfo),
+}
+
+/// Opens a file in `O_NOFOLLOW` mode. Returns [`None`] if the open failed
+/// because the `path` refers to a symbolic link.
+#[cfg(unix)]
```
shaleh: Can you add stubs for windows please? That when someone wants Windows to work we have all of the steps laid out for them.

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+fn open_nofollow(path: &Path) -> Result<Option<File>> {
```
shaleh: Instead of including here could you add this to the top with a `cfg(unix)` wrapper?

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+        fd => {
+            let file = unsafe { File::from_raw_fd(fd) };
+            Ok(Some(file))
+        }
+    }
+}
+
+/// Reads the contents of a regular file at `path` up to `max_content_length` bytes.
+/// Returns [`None`] if the `path` is a symbolic link.
+fn read_nofollow(path: &Path, max_content_length: usize) -> Result<Option<String>> {
```
shaleh: I am on the fence. I think saving the result into a variable and returning `Ok(result)` makes it more clear that the only failures here are from the reading and parsing. 

```
let result = match open_nofollow(path)? {
    Some(file) => {
        let mut content = String::with_capacity(max_content_length);
            file.take(max_content_length as u64)
                .read_to_string(&mut content)?;
        Some(content)
    }
    None => None,
};
Ok(result)
```

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+        }
+
+        Err(ref e) if e.kind() == NotFound => Ok(None),
+
+        Err(e) => Err(e),
+    }
+}
+
+/// Returns [`true`] if a process with PID `pid` is running.
+#[cfg(unix)]
```
shaleh: Windows stub here too.

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+        0 => true,
+        // `PermissionDenied` indicates that there _is_ a process but it’s
+        // owned by another user
+        _ => Error::last_os_error().kind() == PermissionDenied,
+    }
+}
+
+/// Returns the system’s last boot time as Unix epoch seconds.
+fn get_boot_time() -> Result<i64> {
+    use systemstat::Platform;
```
shaleh: Why is this import here and not at the top?

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+    let sys = systemstat::System::new();
+    sys.boot_time().map(|t| t.timestamp())
+}
+
+/// Returns [`true`] if there is no boot time in the lock info or if the
+/// boot time is within one second of the system’s boot time.
+fn boot_time_within_one_second(info: &LockInfo) -> bool {
```
shaleh: Great naming and easy to follow logic.

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+    info.boot_time.map_or(true, |boot_time| {
+        let system_boot_time = get_boot_time().unwrap_or(0);
+        (boot_time - system_boot_time).abs() <= 1
+    })
+}
+
+/// Returns the current state of the lock from lock file `path`.
+fn current_lock_owner(path: &Path) -> Result<LockState> {
+    use LockState::*;
```
shaleh: Why the local import?

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+
+    match read_lock_info(path)? {
```
shaleh: This is another example where perhaps saving the result and returning `Ok(result)` is easier to follow.

rust_src/src/filelock.rs
```diff
@@ -43,4 +307,112 @@ pub fn unlock_buffer_lisp() {
...
+
+    match current_lock_owner(&path) {
+        Ok(NotLocked) | Err(_) => Qnil,
+        Ok(LockedByUs) => Qt,
+        Ok(LockedBy(info)) => LispObject::from(info.user.as_str()),
+    }
+}
+
 include!(concat!(env!("OUT_DIR"), "/filelock_exports.rs"));
+
```
shaleh: So nice to see some native test cases.

rust_src/src/filelock.rs
```diff
@@ -2,13 +2,277 @@
...
+        // The HOST is everything after the '@' to the last '.'.
+        let (at_host, rest) = rest.split_at(rest.rfind('.')?);
+        let host = at_host[1..].to_string();
+
+        // The PID is everything from the last '.' to the ':' or equivalent.
+        // NOTE: Unlike GNU Emacs, we do not tolerate negative PID values
+        // or over/underflow.
+        let (pid_str, boot_time_str) =
+            rest.split_at(rest.find(boot_time_sep).unwrap_or(rest.len()));
+        let pid = pid_str[1..].parse::<i32>().ok().filter(|pid| pid > &0)?;
```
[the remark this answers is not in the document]
liff: [`parse()`](https://doc.rust-lang.org/std/primitive.str.html#method.parse) returns a `Result` and [`ok()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) converts it to an `Option` so we just lose the error details, not panic.

liff:
Thanks for the review. Updated per your comments.

brotzeit:
@liff You can merge your PR when you think it's ready.

PR #1464, opened 2019-04-28
port face-color-supported-p

brotzeit:


rust_src/src/xfaces.rs
```diff
@@ -15,4 +21,19 @@ pub fn clear_face_cache_lisp(thoroughly: bool) {
...
 }
 
+/// Return non-nil if COLOR can be displayed on FRAME.
+/// BACKGROUND-P non-nil means COLOR is used as a background.
+/// Otherwise, this function tells whether it can be used as a foreground.
+/// If FRAME is nil or omitted, use the selected frame.
+/// COLOR must be a valid color name.
+#[lisp_fn(min = "1")]
+pub fn color_supported_p(
+    mut color: LispStringRef,
```
shaleh: Should we add const compatible versions of the data access functions so we don't have to mark variables mutable?

brotzeit: I guess there's not much value in it. So why not.

PR #1467, opened 2019-04-30
Port font-get-system-font and font-get-system-normal-font

### LINKED ISSUE #1458: Port font-get-system-font
```c
DEFUN ("font-get-system-font", Ffont_get_system_font, Sfont_get_system_font,
       0, 0, 0,
       doc: /* Get the system default fixed width font. */)
  (void)
{
  return current_mono_font ? build_string (current_mono_font) : Qnil;
}
```

Bira:
There wasn't a xsettings.rs file yet, so I created it. Since both of these functions relied on static C variables, I went with the same strategy I used for clear-charset-maps, creating a new `tempxsettings.h` header to contain their `extern` declarations.

This closes #1458 and also ports font-get-normal-system-font, which was very similar.

### DIFF
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -126,6 +126,7 @@ mod windows;
 mod xdisp;
 mod xfaces;
 mod xml;
+mod xsettings;
 
 #[cfg(all(not(test), target_os = "macos", feature = "unexecmacosx"))]
 use alloc_unexecmacosx::OsxUnexecAlloc;
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -126,6 +126,7 @@ mod windows;
 mod xdisp;
 mod xfaces;
 mod xml;
+#[cfg(feature = "window-system-x11")]
 mod xsettings;
 
 #[cfg(all(not(test), target_os = "macos", feature = "unexecmacosx"))]
--- a/rust_src/build.rs
+++ b/rust_src/build.rs
@@ -355,18 +355,30 @@ fn env_var(name: &str) -> String {
 }
 
 // What to ignore when walking the list of files
-fn ignore(path: &str) -> bool {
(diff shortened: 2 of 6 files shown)


Bira:
Got it, I'll work on that.

Bira:
I've used the same changes introduced with the trace-redisplay port to make this file empty if the "window-system-x11" feature is disabled.

shaleh:
A simpler approach would be to only include it in lib.rs when `cfg(feature = "window-system-x11")` is true. If the file is not included there Cargo won't build it. Add a nice comment at the top of xsettings so people know it is only active when `window-system-x11` is enabled.

Bira:
My initial attempt indeed consisted of adding `cfg(feature = "window-system-x11")` to the `mod xsettings;` line in lib.rs, but that apparently doesn't prevent this file's functions from being added to the generated c_exports.rs during the build process, which causes the build to fail.  Adding the `cfg` to the include macro at the end of the file didn't work either.

rust_src/src/lib.rs
```diff
@@ -126,6 +126,7 @@ mod windows;
 mod xdisp;
 mod xfaces;
 mod xml;
+mod xsettings;
```
brotzeit: I think you can drop the `#[cfg(feature = "window-system-x11")]` in this file and just mark the file with it instead.

brotzeit:
@Bira Yes. I think we have to modify `generate_include_files` to conditionally generate the bindings for this file.

brotzeit:
I can't think of a better solution. Let's see what the others say.

brotzeit:
Unless there are any objections, I would merge this ?

Bira:
:+1:

I'll merge it as soon as it's approved.

brotzeit:
Please squash this PR.

shaleh:
Always squash unless you are merging from Emacs. Really simple.

Issue #1396, opened 2019-03-24
crashes in print_object on macOS


razzmatazz:
I cannot find how to replicate this easily, but remacs crashes for me every day or two with the following stacktrace on macOS 10.14 (mojave)
```
Process:               Remacs [33593]
Path:                  /Users/USER/*/Remacs.app/Contents/MacOS/Remacs
Identifier:            org.gnu.Remacs
Version:               Version 27.0.50 (9.0)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           Remacs [33593]
User ID:               501

Date/Time:             2019-03-24 12:22:24.461 +0200
OS Version:            Mac OS X 10.14.3 (18D109)
Report Version:        12
Anonymous UUID:        F8D20147-A164-0284-0365-0163BC7EDA98

Sleep/Wake UUID:       4A101425-29C7-4455-8775-6C31FE7C3F9F

Time Awake Since Boot: 120000 seconds
Time Since Wake:       9700 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGABRT)
Exception Codes:       KERN_INVALID_ADDRESS at 0x000000023c44e238
Exception Note:        EXC_CORPSE_NOTIFY

VM Regions Near 0x23c44e238:
    CoreAnimation          0000000140000000-0000000143a9b000 [ 58.6M] rw-/rwx SM=PRV  
--> 
    MALLOC_NANO            0000600000000000-0000600008000000 [128.0M] rw-/rwx SM=PRV  

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff5c3d723e __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff5c48dc1c pthread_kill + 285
2   libsystem_c.dylib             	0x00007fff5c2f3096 raise + 26
3   org.gnu.Remacs                	0x00000001000b3a89 terminate_due_to_signal + 153 (emacs.c:391)
4   org.gnu.Remacs                	0x00000001000ceb73 emacs_abort + 19
5   org.gnu.Remacs                	0x000000010019680a ns_term_shutdown + 122
6   org.gnu.Remacs                	0x00000001000b3c76 shut_down_emacs + 262 (emacs.c:2060)
[...]

PR #1472, opened 2019-05-01
Revert "Merge emacs Mojave support" and merge 2a1fe08307

brotzeit:
Let's see if we can get our mac tests green without this cherry pick.


brotzeit:
I will try to remerge the mojave support as soon as possible...

brotzeit:
Hopefully this works...

.travis.yml
```diff
@@ -37,6 +37,11 @@ jobs:
         - travis_wait rustup component add rustfmt-preview
         - ./autogen.sh && ./configure --without-makeinfo --with-x=no --with-ns=no --without-gconf --without-gsettings
       script: ./.travis-format.sh
```
shaleh: Why are you running it twice? To catch mistakes in the code under different `cfg` options? if so, the comment should go and this could be restructured like the unit test runs further down.

brotzeit: There's an error in the mac logs `error: 'rustfmt' is not installed for the toolchain 'nightly-2019-05-01-x86_64-apple-darwin'`. I want to get the mac tests green.

shaleh: Ah. Well, adding this test won't help that really. You need to install rustfmt during the 'test' steps.

brotzeit:
Ok let's merge this. I guess those errors are the reason why we turned off the mac tests for merging.

brotzeit:
Maybe this really fixes #1396...

brotzeit:
Reminder: we are 14 months behind...

brotzeit:
I don't think these 50 commits will make any problems :P

PR #1473, opened 2019-05-02
Fix clippy warnings

agraven:


rust_src/src/filelock.rs
```diff
@@ -67,8 +67,8 @@ impl LockInfo {
         // NOTE: Unlike GNU Emacs, we do not tolerate negative PID values
         // or over/underflow.
         let (pid_str, boot_time_str) =
-            rest.split_at(rest.find(boot_time_sep).unwrap_or(rest.len()));
-        let pid = pid_str[1..].parse::<i32>().ok().filter(|pid| pid > &0)?;
+            rest.split_at(rest.find(boot_time_sep).unwrap_or_else(|| rest.len()));
```
shaleh:
```suggestion
            rest.split_at(rest.find(boot_time_sep).unwrap_or(|| rest.len));
```

I think this works too.

brotzeit: I think `unwrap_or` doesn't work with a closure ?

agraven: Nope, `unwrap_or` only accepts values, `unwrap_or_else` exists specifically to use a closure

shaleh: Right, you don't need the closure. I typo'ed. Since `unwrap_or` takes a zero argument function we can pass one instead of making one with a closure. I _think_ this will work with the method. I know clippy complains about `unwrap_or_else(|| foo())` when it could simply be `unwrap_or(foo)`.

agraven: In that case is should be `unwrap_or(str::len)` (or maybe it's String, I'm not completely sure).

agraven: I just realized that it can't be done by just specifying a function as argument, because with anything involving unwrap_or_something the method has no idea that it's `rest` the function is supposed to be applied to, so its current form is the only one that works.

agraven:
I would suggest merging this.

nickdrozd:
Some new warnings were introduced by recent PRs, but there will always be new warnings. We'll take them as they come.

PR #1474, opened 2019-05-03
update syn for `remacs-macros`

cantsin:
Partially addresses https://github.com/remacs/remacs/issues/1244

It doesn't look like we can upgrade `syn` for the `remacs-util` crate because newer versions of `syn` use `proc_macro` internally and trying to use these related functions outside of `#[proc_macro_attribute]` or `#[proc_macro_derive(...)]` will give us a `procedural macro API is used outside of a procedural macro` error. (We manually parse `lisp_fn` attributes in `remacs_lib/docfile.rs`)

The older version of `syn` (0.11.11) for the `remacs-util` crate works because it uses a fork of `nom` called [synom](https://crates.io/crates/synom) instead of relying on `proc_macro`; as far as I can see, this fork was abandoned a couple of years ago. 

TL;DR: this PR updates `syn` where-ever possible (only `remacs-macros`); we may need to re-write or re-structure `remacs-util` if we want to update it in the future (perhaps it should utilize `nom` directly).

### DIFF
--- a/rust_src/Cargo.lock
+++ b/rust_src/Cargo.lock
@@ -329,7 +329,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "proc-macro2"
-version = "0.3.8"
+version = "0.4.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "unicode-xid 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


brotzeit:
Thanks for the investigation =)

cantsin:
Thanks! What's the etiquette here wrt merging? Can I merge myself after the PR is approved? (Apologies if this was documented somewhere, but I didn't see it in the wiki).

brotzeit:
We let the author do the merge. I usually give others the chance to review my PRs after an approval and wait another few days. But this PR is already 10 days old and I think there won't be another review.

PR #1478, opened 2019-05-05
Port set-buffer-redisplay

### LINKED ISSUE #1448: Port set-buffer-redisplay
```c
DEFUN ("set-buffer-redisplay", Fset_buffer_redisplay,
       Sset_buffer_redisplay, 4, 4, 0,
       doc: /* Mark the current buffer for redisplay.
This function may be passed to `add-variable-watcher'.  */)
  (Lisp_Object symbol, Lisp_Object newval, Lisp_Object op, Lisp_Object where)
{
  bset_update_mode_line (current_buffer);
  current_buffer->prevent_redisplay_optimizations_p = true;
  return Qnil;
}
```

simon-frankau:
Fixes #1448

I'm not entirely sure what to do about testing for this.

Note that this code is very similar to the "else" case of "force_mode_line_update" in buffers.rs, but I'm also not sure if/how that should be factored out.

### DIFF
--- a/rust_src/src/xdisp.rs
+++ b/rust_src/src/xdisp.rs
@@ -1,12 +1,29 @@
 //! Display generation from window structure and buffer text.
 
-// Remove the feature check when this file gains functions
-// that are not guarded by it.
-#[cfg(feature = "glyph-debug")]
 use remacs_macros::lisp_fn;
 
+use crate::lisp::LispObject;
+use crate::remacs_sys::{bset_update_mode_line, Qnil};
+use crate::threads::ThreadState;
 #[cfg(feature = "glyph-debug")]
 use crate::{interactive::InteractiveNumericPrefix, remacs_sys::trace_redisplay_p};
+
+/// Mark the current buffer for redisplay.
+/// This function may be passed to `add-variable-watcher'.
+#[lisp_fn]
+pub fn set_buffer_redisplay(
+    _symbol: LispObject,
+    _newval: LispObject,
+    _op: LispObject,
+    _where: LispObject,
+) -> LispObject {
+    let mut current_buffer = ThreadState::current_buffer_unchecked();
+    unsafe {
+        bset_update_mode_line(current_buffer.as_mut());
+    }
+    current_buffer.set_prevent_redisplay_optimizations_p(true);
+    Qnil
+}
 
 /// Toggle tracing of redisplay.
 /// With ARG, turn tracing on if and only if ARG is positive.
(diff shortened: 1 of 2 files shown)


brotzeit:
Thanks for the PR. I think this is ok.

rust_src/src/xdisp.rs
```diff
@@ -1,13 +1,30 @@
...
+/// Mark the current buffer for redisplay.
+/// This function may be passed to `add-variable-watcher'.
+#[lisp_fn]
+pub fn set_buffer_redisplay(
+    _symbol: LispObject,
+    _newval: LispObject,
+    _op: LispObject,
+    _where: LispObject,
+) -> LispObject {
+    let mut current_buffer = ThreadState::current_buffer_unchecked();
```
shaleh: You are correct this looks pretty close to `force_mode_line_update`. We have choices at this point.
We can ignore it and allow some minor code duplication or we can pull this out into a function or method on LispBufferRef. Then this function becomes something like:

```
let mut current_buffer = ThreadState::current_buffer_unchecked();
current_buffer.set_buffer_redisplay();
```

and the else in `force_mode_line_update` does the same.

rust_src/src/xdisp.rs
```diff
@@ -1,13 +1,30 @@
...
+    unsafe {
+        bset_update_mode_line(current_buffer.as_mut());
+    }
+    current_buffer.set_prevent_redisplay_optimizations_p(true);
+    Qnil
```
brotzeit: You can drop the Qnil and let the function return nothing.

rust_src/src/xdisp.rs
simon-frankau: Thinking about it a bit more, I'm wary of pulling the common code out. The two functions are claiming to do different things - redisplay the buffer, and redisplay the mode line. The fact that they're using the same code in the two cases suggests they're misnamed, or at least do more work than their names suggest.

It feels to me like the correct (eventual) refactor would be to make them actually do different things that match their names. Pulling out the common code into a new function works against that. 

So, I propose to leave in the duplication. What do you think?

rust_src/src/xdisp.rs
simon-frankau: Cheers. I have made this change.

rust_src/src/xdisp.rs
brotzeit: I don't have a problem with that since we don't really have much code duplication with this.

But I just realized that we commented out `current_buffer.set_prevent_redisplay_optimizations_p(true)` in `force_mode_line_update´.

@shaleh what was your intention for doing this ?

shaleh: A comment in the C code questioning its value. This code is a bit too complex and frankly no one, here or on Emacs side I suspect could really tell us what is happening. I commented it out as an experiment. It is fine to put it back.

PR #1480, opened 2019-05-07
Port unintern

### LINKED ISSUE #1468: Port unintern
```c
DEFUN ("unintern", Funintern, Sunintern, 1, 2, 0,
       doc: /* Delete the symbol named NAME, if any, from OBARRAY.
The value is t if a symbol was found and deleted, nil otherwise.
NAME may be a string or a symbol.  If it is a symbol, that symbol
is deleted, if it belongs to OBARRAY--no other symbol is deleted.
OBARRAY, if nil, defaults to the value of the variable `obarray'.
usage: (unintern NAME OBARRAY)  */)
  (Lisp_Object name, Lisp_Object obarray)
{
  register Lisp_Object string, tem;
  size_t hash;

  if (NILP (obarray)) obarray = Vobarray;
  obarray = check_obarray (obarray);

  if (SYMBOLP (name))
    string = SYMBOL_NAME (name);
  else
    {
      CHECK_STRING (name);
      string = name;
    }

  tem = oblookup (obarray, SSDATA (string),
		  SCHARS (string),
		  SBYTES (string));
  if (INTEGERP (tem))
    return Qnil;
  /* If arg was a symbol, don't delete anything but that symbol itself.  */
  if (SYMBOLP (name) && !EQ (name, tem))
    return Qnil;

  /* There are plenty of other symbols which will screw up the Emacs
     session if we unintern them, as well as even more ways to use
     `setq' or `fset' or whatnot to make the Emacs session
     unusable.  Let's not go down this silly road.  --Stef  */
  /* if (EQ (tem, Qnil) || EQ (tem, Qt))
       error ("Attempt to unintern t or nil"); */

  XSYMBOL (tem)->u.s.interned = SYMBOL_UNINTERNED;

  hash = oblookup_last_bucket_number;

  if (EQ (AREF (obarray, hash), tem))
    {
      if (XSYMBOL (tem)->u.s.next)
	{
	  Lisp_Object sym;
	  XSETSYMBOL (sym, XSYMBOL (tem)->u.s.next);
	  ASET (obarray, hash, sym);
	}
      else
	ASET (obarray, hash, make_number (0));
    }
  else
    {
      Lisp_Object tail, following;

      for (tail = AREF (obarray, hash);
	   XSYMBOL (tail)->u.s.next;
	   tail = following)
	{
	  XSETSYMBOL (following, XSYMBOL (tail)->u.s.next);
	  if (EQ (following, tem))
	    {
	      set_symbol_next (tail, XSYMBOL (following)->u.s.next);
	      break;
	    }
	}
    }

  return Qt;
}
```

cantsin:
Closes #1468

### DIFF
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -18,7 +18,7 @@ use crate::{
     eval::unbind_to,
     lisp::LispObject,
     multibyte::{char_resolve_modifier_mask, LispSymbolOrString},
-    obarray::{check_obarray, intern, intern_c_string_1, LispObarrayRef},
+    obarray::{intern, intern_c_string_1, LispObarrayRef},
     remacs_sys,
     remacs_sys::infile,
     remacs_sys::{
@@ -413,8 +413,7 @@ pub fn read_char_exclusive(
 /// usage: (unintern NAME OBARRAY)
 #[lisp_fn(min = "1")]
 pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> bool {
-    let obarray = obarray.unwrap_or_else(LispObarrayRef::global);
-    let obarray: LispObarrayRef = check_obarray(obarray.into()).into();
+    let obarray = obarray.unwrap_or_else(LispObarrayRef::global).check();
 
     let tem = obarray.lookup(name);
     if tem.is_integer() {
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -433,7 +433,7 @@ pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> bo
 
     if symbol == temp {
         match symbol.get_next() {
-            Some(sym) => obarray.set(hash, sym.into()),
+            Some(sym) => obarray.set(hash, sym),
             None => obarray.set(hash, LispObject::from_natnum(0)),
         };
     } else {
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -27,7 +27,7 @@ use crate::{
         symbol_redirect, unblock_input,
     },
     remacs_sys::{globals, EmacsInt},
-    remacs_sys::{Qeval_buffer_list, Qnil, Qread_char, Qstandard_output, Qsymbolp, Qt},
+    remacs_sys::{Qeval_buffer_list, Qnil, Qread_char, Qstandard_output, Qsymbolp},
     symbols::LispSymbolRef,
     threads::{c_specpdl_index, ThreadState},
 };
@@ -412,17 +412,17 @@ pub fn read_char_exclusive(
 /// OBARRAY, if nil, defaults to the value of the variable `obarray'.
 /// usage: (unintern NAME OBARRAY)
 #[lisp_fn(min = "1")]
-pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> LispObject {
+pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> bool {
     let obarray = obarray.unwrap_or_else(LispObarrayRef::global);
     let obarray: LispObarrayRef = check_obarray(obarray.into()).into();
 
     let tem = obarray.lookup(name);
     if tem.is_integer() {
-        return Qnil;
+        return false;
     }
     // If arg was a symbol, don't delete anything but that symbol itself.
     if name.is_symbol() && name != tem {
-        return Qnil;
+        return false;
     }
 
     let mut temp: LispSymbolRef = tem.into();
@@ -450,7 +450,7 @@ pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> Li
         }
     }
 
-    Qt
+    true
 }
 
 include!(concat!(env!("OUT_DIR"), "/lread_exports.rs"));
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -437,14 +437,10 @@ pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> bo
             None => obarray.set(hash, LispObject::from_natnum(0)),
         };
     } else {
-        for tail in symbol.iter() {
-            if let Some(following) = tail.get_next() {
-                if following == temp {
-                    let next = following.get_next();
-                    tail.set_next(next);
-                    break;
-                }
-            }
+        let tail = symbol.iter().find(|elem| elem.get_next() == Some(temp));
+        if let Some(previous) = tail {
+            let next = temp.get_next();
+            previous.set_next(next);
         }
     }
 
--- a/rust_src/src/lread.rs
+++ b/rust_src/src/lread.rs
@@ -17,16 +17,17 @@ use crate::{
     },
     eval::unbind_to,
     lisp::LispObject,
-    multibyte::char_resolve_modifier_mask,
-    obarray::{intern, intern_c_string_1},
+    multibyte::{char_resolve_modifier_mask, LispStringRef, LispSymbolOrString},
+    obarray::{check_obarray, intern, intern_c_string_1, LispObarrayRef},
     remacs_sys,
     remacs_sys::infile,
     remacs_sys::{
-        block_input, build_string, getc_unlocked, maybe_quit, read_filtered_event,
-        read_internal_start, readevalloop, specbind, staticpro, symbol_redirect, unblock_input,
+        block_input, build_string, getc_unlocked, maybe_quit, oblookup,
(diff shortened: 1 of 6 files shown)


rust_src/src/lread.rs
```diff
@@ -404,4 +405,52 @@ pub fn read_char_exclusive(
...
+/// Delete the symbol named NAME, if any, from OBARRAY.
+/// The value is t if a symbol was found and deleted, nil otherwise.
+/// NAME may be a string or a symbol.  If it is a symbol, that symbol
+/// is deleted, if it belongs to OBARRAY--no other symbol is deleted.
+/// OBARRAY, if nil, defaults to the value of the variable `obarray'.
+/// usage: (unintern NAME OBARRAY)
+#[lisp_fn(min = "1")]
+pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> LispObject {
+    let obarray = obarray.unwrap_or_else(LispObarrayRef::global);
+    let obarray: LispObarrayRef = check_obarray(obarray.into()).into();
```
shaleh: Let's make a `check` method on `LispObarrayRef` and have the `check_obarray` function call it.

rust_src/src/lread.rs
```diff
@@ -404,4 +405,52 @@ pub fn read_char_exclusive(
...
 }
 
+/// Delete the symbol named NAME, if any, from OBARRAY.
+/// The value is t if a symbol was found and deleted, nil otherwise.
+/// NAME may be a string or a symbol.  If it is a symbol, that symbol
+/// is deleted, if it belongs to OBARRAY--no other symbol is deleted.
+/// OBARRAY, if nil, defaults to the value of the variable `obarray'.
+/// usage: (unintern NAME OBARRAY)
+#[lisp_fn(min = "1")]
+pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> LispObject {
```
shaleh:
```suggestion
pub fn unintern(name: LispSymbolOrString, obarray: Option<LispObarrayRef>) -> bool {
```

rust_src/src/lread.rs
```diff
@@ -404,4 +405,52 @@ pub fn read_char_exclusive(
...
+    let obarray = obarray.unwrap_or_else(LispObarrayRef::global);
+    let obarray: LispObarrayRef = check_obarray(obarray.into()).into();
+
+    let tem = obarray.lookup(name);
+    if tem.is_integer() {
+        return Qnil;
```
shaleh:
```suggestion
        return false;
```

rust_src/src/lread.rs
```diff
@@ -404,4 +405,52 @@ pub fn read_char_exclusive(
...
+    }
+    // If arg was a symbol, don't delete anything but that symbol itself.
+    if name.is_symbol() && name != tem {
+        return Qnil;
```
shaleh:
```suggestion
        return false;
```

rust_src/src/lread.rs
```diff
@@ -404,4 +405,52 @@ pub fn read_char_exclusive(
...
+    }
+
+    let mut temp: LispSymbolRef = tem.into();
+    temp.set_uninterned();
+
+    let hash = unsafe { oblookup_last_bucket_number };
+
```
shaleh: If we add `get` and `set` methods to `LispObarrayRef` this code becomes simpler to read and reason about.

rust_src/src/lread.rs
```diff
@@ -404,4 +405,52 @@ pub fn read_char_exclusive(
...
+    let mut obarray = LispObject::from(obarray).as_vector_or_error();
+    let symbol: LispSymbolRef = obarray.get(hash).into();
+
+    if symbol == temp {
+        match symbol.get_next() {
```
shaleh: `get_next` is the same as `iter().next()` I am not certain we need it.

rust_src/src/lread.rs
```diff
@@ -404,4 +405,52 @@ pub fn read_char_exclusive(
...
+            Some(sym) => obarray.set(hash, sym.into()),
+            None => obarray.set(hash, LispObject::from_natnum(0)),
+        };
+    } else {
```
shaleh: There should be a clear way to write this which embraces the iterators. Would you mind giving another go at it?

rust_src/src/lread.rs
cantsin: D'oh. Thanks.

rust_src/src/lread.rs
cantsin: Sure thing, I can add wrapper methods if that's what you mean. Actually, now I'm wondering. Since every (valid) obarray is also a vector-like... would it make sense to re-use the macro `impl_vectorlike_ref!` for `LispObarrayRef`?

rust_src/src/lread.rs
cantsin: Sure thing!

rust_src/src/lread.rs
cantsin: Sure, makes sense. I wasn't sure what to do with the corresponding doc comments, so I left them as-is, with `check_obarray`. Let me know if these should be moved to `check` instead.

rust_src/src/lread.rs
cantsin: Due to how we have the Iterator trait set up for `LispSymbolIter`, `iter().next()` returns the current item (which makes sense from a Rust perspective, not so much from C). ironically, here we'd need to do another `.next()` on the iterator to _really_ get the next item. I adjusted the code to `iter().nth(1)` which, admittedly, is still awkward. And I left the `get_next` function in here for symmetry with `set_next`, seems a bit cleaner. Open to other options (no pun intended) ...

rust_src/src/lread.rs
shaleh: If I recall, `LispObarrayRef` is a simple type alias for a `LispObject`. Perhaps it makes more sense for it to be an alias for `LispVector` which would then make adding pass through methods easy. That sounds beyond the scope of this PR but definitely a worthwhile follow on effort.

rust_src/src/lread.rs
cantsin: Revised, should be cleaner now. The `if let ...` is probably unnecessary, but since the original C code silently passes even if the symbol is not found, I decided to leave it in there.

rust_src/src/obarray.rs
```diff
@@ -58,6 +58,37 @@ impl LispObarrayRef {
...
+                unsafe { globals.Vobarray = initial_obarray };
+            }
+            wrong_type!(Qvectorp, self.0);
+        }
+
+        self
+    }
+
+    pub fn get(&self, idx: usize) -> LispSymbolRef {
+        LispObject::from(self).as_vector_or_error().get(idx).into()
```
shaleh:
```suggestion
        LispObject::from(self).force_vector().get(idx).into()
```

We have `force_foo` methods which act like the XFOO C macros. No checks, all trust.

rust_src/src/obarray.rs
```diff
@@ -58,6 +58,37 @@ impl LispObarrayRef {
...
+    }
+
+    pub fn set(&mut self, idx: usize, item: LispObject) {
```
shaleh: Perhaps this could be a generic function on `item`. The first line of the method then calls `into` on `item`.
```suggestion
    pub fn set<O: Into<LispObject>>(&mut self, idx: usize, item: O) {
```

rust_src/src/obarray.rs
```diff
@@ -58,6 +58,37 @@ impl LispObarrayRef {
...
+        let mut vec = LispObject::from(&*self).as_vector_or_error();
```
shaleh: This can use `force_vector` as well.

rust_src/src/symbols.rs
```diff
@@ -187,6 +192,18 @@ impl LispSymbolRef {
         LispSymbolIter { current: self }
     }
 
+    pub fn get_next(self) -> Option<LispSymbolRef> {
```
shaleh: You PR comment above about why `nth` is used would make a good code comment here.

rust_src/src/obarray.rs
cantsin: Ooh, great idea, all this casting can be verbose. Thanks!

rust_src/src/obarray.rs
cantsin: Good to know, thanks. Updated.

rust_src/src/obarray.rs
cantsin: Updated as well.

rust_src/src/symbols.rs
cantsin: Done.

cantsin:
Anything else I should do here? (No rush...)

shaleh:
@cantsin I need to go over it again. Will try to do get to it this evening.

cantsin:
No worries, whenever you can. Thanks.

PR #1484, opened 2019-05-08
update readme

brotzeit:


brotzeit:
I think I know why I got different results for rust and C. The C code always has 3 different functions for the different window systems and we only have one right ?

shaleh:
Perhaps. As we get further in and replace the C how we handle the GUI bits will be a place of research.
It is plausible we end up with `some_func_windows` `some_func_xwindow` etc.

PR #1487, opened 2019-05-09
Port tool-bar-get-system-style

Bira:
This builds on the groundwork laid down by the porting of the other xsettings defuns.

### DIFF
--- a/rust_src/src/xsettings.rs
+++ b/rust_src/src/xsettings.rs
@@ -38,13 +38,10 @@ pub fn font_get_system_font() -> LispObject {
 /// known style.  Otherwise return image.
 #[lisp_fn]
 pub fn tool_bar_get_system_style() -> LispObject {
-    match unsafe { globals.Vtool_bar_style } {
-        tool_bar_style if tool_bar_style.is_nil() => unsafe { current_tool_bar_style },
-        tool_bar_style
-            if [Qimage, Qtext, Qboth, Qboth_horiz, Qtext_image_horiz].contains(&tool_bar_style) =>
-        {
-            tool_bar_style
-        }
+    let style = unsafe { globals.Vtool_bar_style };
+    match style {
+        Qnil => unsafe { current_tool_bar_style },
+        Qimage | Qtext | Qboth | Qboth_horiz | Qtext_image_horiz => style,
         _ => Qimage,
     }
 }
--- a/rust_src/src/xsettings.rs
+++ b/rust_src/src/xsettings.rs
@@ -38,18 +38,14 @@ pub fn font_get_system_font() -> LispObject {
 /// known style.  Otherwise return image.
 #[lisp_fn]
 pub fn tool_bar_get_system_style() -> LispObject {
-    let tool_bar_style = unsafe { globals.Vtool_bar_style };
-    if tool_bar_style == Qimage
-        || tool_bar_style == Qtext
-        || tool_bar_style == Qboth
-        || tool_bar_style == Qboth_horiz
-        || tool_bar_style == Qtext_image_horiz
-    {
+    match unsafe { globals.Vtool_bar_style } {
+        tool_bar_style if tool_bar_style.is_nil() => unsafe { current_tool_bar_style },
         tool_bar_style
-    } else if !tool_bar_style.is_nil() {
-        unsafe { current_tool_bar_style }
-    } else {
-        Qimage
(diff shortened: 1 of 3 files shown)


rust_src/src/xsettings.rs
```diff
@@ -32,4 +33,24 @@ pub fn font_get_system_font() -> LispObject {
     }
 }
 
+/// Get the system tool bar style.
+/// If no system tool bar style is known, return `tool-bar-style' if set to a
+/// known style.  Otherwise return image.
+#[lisp_fn]
+pub fn tool_bar_get_system_style() -> LispObject {
+    let tool_bar_style = unsafe { globals.Vtool_bar_style };
```
shaleh: Try writing this as a match instead.

Bira:
I initially copied the style of the C function because I wasn't sure how to write a concise match where multiple options should have the same result. Is the style I used for the match good, or is there a better way to do it?

rust_src/src/xsettings.rs
```diff
@@ -32,4 +33,20 @@ pub fn font_get_system_font() -> LispObject {
     }
 }
 
+/// Get the system tool bar style.
+/// If no system tool bar style is known, return `tool-bar-style' if set to a
+/// known style.  Otherwise return image.
+#[lisp_fn]
+pub fn tool_bar_get_system_style() -> LispObject {
+    match unsafe { globals.Vtool_bar_style } {
```
shaleh: Does work?

```
match unsafe { globals.Vtool_bar_style } {
  Qnil => unsafe { current_tool_bar_style },
  style @ Qimage | Qtext | Qboth | Qboth_horiz | Qtext_image_horiz => style,
  _ => Qimage,
}
```

Bira: After a bit of experimentation, this is what I have:
```
    match unsafe { globals.Vtool_bar_style } {
        Qnil => unsafe { current_tool_bar_style },
        style @ Qimage
        | style @ Qtext
        | style @ Qboth
        | style @ Qboth_horiz
        | style @ Qtext_image_horiz => style,
        _ => Qimage,
    }
```
Which do you prefer?

shaleh: Yeah. Yuck:
> If you use @ with |, you need to make sure the name is bound in each part of the pattern:

I am inclined to leave the @ binding off and simply return `unsafe { globals.Vtool_bar_style }`.

Bira: I took the bindings off, and added a local `style` variable to the top to be used both in the match and in that return branch.

rust_src/src/xsettings.rs
```diff
@@ -38,13 +38,10 @@ pub fn font_get_system_font() -> LispObject {
...
 #[lisp_fn]
 pub fn tool_bar_get_system_style() -> LispObject {
-    match unsafe { globals.Vtool_bar_style } {
-        tool_bar_style if tool_bar_style.is_nil() => unsafe { current_tool_bar_style },
-        tool_bar_style
-            if [Qimage, Qtext, Qboth, Qboth_horiz, Qtext_image_horiz].contains(&tool_bar_style) =>
-        {
-            tool_bar_style
-        }
+    let style = unsafe { globals.Vtool_bar_style };
```
shaleh: Nicely done.

src/xsettings.c
```diff
@@ -974,24 +974,6 @@ xsettings_get_system_font (void)
...
-known style.  Otherwise return image.  */)
-  (void)
-{
-  if (EQ (Vtool_bar_style, Qimage)
-      || EQ (Vtool_bar_style, Qtext)
-      || EQ (Vtool_bar_style, Qboth)
-      || EQ (Vtool_bar_style, Qboth_horiz)
-      || EQ (Vtool_bar_style, Qtext_image_horiz))
-    return Vtool_bar_style;
-  if (!NILP (current_tool_bar_style))
```
shaleh: I did not look at this closely enough last time since I was focused on the `match`. `current_tool_bar_style` is only returned when it has a value. The nil check in the `match` is incorrect. The port should read:

```
match style {
      Qimage | Qtext | Qboth | Qboth_horiz | Qtext_image_horiz => style,
      _ if  unsafe { current_tool_bar_style.is_nil() } => Qimage,
      _ => unsafe { current_tool_bar_style },
}
```

Bira: Good catch! I pushed a fixed version now. I used another local variable to fetch the value of `current_tool_bar_style` to keep the number of unsafe blocks down.

PR #1489, opened 2019-05-10
Port other-buffer

### LINKED ISSUE #1476: Port other-buffer
```c
DEFUN ("other-buffer", Fother_buffer, Sother_buffer, 0, 3, 0,
       doc: /* Return most recently selected buffer other than BUFFER.
Buffers not visible in windows are preferred to visible buffers, unless
optional second argument VISIBLE-OK is non-nil.  Ignore the argument
BUFFER unless it denotes a live buffer.  If the optional third argument
FRAME specifies a live frame, then use that frame's buffer list instead
of the selected frame's buffer list.

The buffer is found by scanning the selected or specified frame's buffer
list first, followed by the list of all buffers.  If no other buffer
exists, return the buffer `*scratch*' (creating it if necessary).  */)
  (Lisp_Object buffer, Lisp_Object visible_ok, Lisp_Object frame)
{
  Lisp_Object tail = f->buffer_list, pred = f->buffer_predicate;
  Lisp_Object buf, notsogood = Qnil;

  /* Consider buffers that have been seen in the frame first.  */
  for (; CONSP (tail); tail = XCDR (tail))
    {
      buf = XCAR (tail);
      if (candidate_buffer (buf, buffer)
	  /* If the frame has a buffer_predicate, disregard buffers that
	     don't fit the predicate.  */
	  && (NILP (pred) || !NILP (call1 (pred, buf))))
	{
	  if (!NILP (visible_ok)
	      || NILP (Fget_buffer_window (buf, Qvisible)))
	    return buf;
	  else if (NILP (notsogood))
	    notsogood = buf;
	}
    }

  /* Consider alist of all buffers next.  */
  FOR_EACH_LIVE_BUFFER (tail, buf)
    {
      if (candidate_buffer (buf, buffer)
	  /* If the frame has a buffer_predicate, disregard buffers that
	     don't fit the predicate.  */
	  && (NILP (pred) || !NILP (call1 (pred, buf))))
	{
	  if (!NILP (visible_ok)
	      || NILP (Fget_buffer_window (buf, Qvisible)))
	    return buf;
	  else if (NILP (notsogood))
	    notsogood = buf;
	}
    }

  if (!NILP (notsogood))
    return notsogood;
  else
    {
      AUTO_STRING (scratch, "*scratch*");
      buf = Fget_buffer (scratch);
      if (NILP (buf))
	{
	  buf = Fget_buffer_create (scratch);
	  Fset_buffer_major_mode (buf);
	}
      return buf;
    }
}
```

simon-frankau:
Fixed #1476.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -18,7 +18,7 @@ use crate::{
     eval::unbind_to,
     fileio::{expand_file_name, find_file_name_handler},
     fns::{copy_sequence, nconc, nreverse},
-    frames::LispFrameRef,
+    frames::{LispFrameLiveOrSelected, LispFrameRef},
     hashtable::LispHashTableRef,
     lisp::{ExternalPtr, LispMiscRef, LispObject, LispStructuralEqual, LiveBufferIter},
     lists,
@@ -33,7 +33,7 @@ use crate::{
     numbers::{LispNumber, MOST_POSITIVE_FIXNUM},
     obarray::intern,
     remacs_sys::symbol_trapped_write::SYMBOL_TRAPPED_WRITE,
-    remacs_sys::Fmake_marker,
+    remacs_sys::{Fget_buffer_window, Fmake_marker, Fset_buffer_major_mode},
     remacs_sys::{
         alloc_buffer_text, allocate_buffer, allocate_misc, block_input, bset_update_mode_line,
         buffer_fundamental_string, buffer_local_flags, buffer_local_value, buffer_memory_full,
@@ -51,7 +51,7 @@ use crate::{
         buffer_permanent_local_flags, Qafter_string, Qbefore_string, Qbuffer_list_update_hook,
         Qbuffer_read_only, Qbufferp, Qfundamental_mode, Qget_file_buffer, Qinhibit_quit,
         Qinhibit_read_only, Qmakunbound, Qnil, Qoverlayp, Qpermanent_local, Qpermanent_local_hook,
-        Qt, Qunbound, UNKNOWN_MODTIME_NSECS,
+        Qt, Qunbound, Qvisible, UNKNOWN_MODTIME_NSECS,
     },
     strings::string_equal,
     textprop::get_text_property,
@@ -278,6 +278,10 @@ impl LispBufferRef {
     // Check if buffer is live
     pub fn is_live(self) -> bool {
         self.name_.is_not_nil()
+    }
+
+    pub fn is_hidden(self) -> bool {
+        LispStringRef::from(self.name_).byte_at(0) == b' '
     }
 
     pub fn set_pt_both(&mut self, charpos: ptrdiff_t, byte: ptrdiff_t) {
@@ -1841,6 +1845,82 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
     if charpos - 1 != nchars {
         panic!("byte_char_debug_check failed.")
     }
+}
+
+// True if B can be used as 'other-than-BUFFER' buffer.
+fn candidate_buffer(b: LispObject, buffer: LispObject) -> bool {
+    if let Some(buf) = Option::<LispBufferRef>::from(b) {
+        !b.eq(buffer) && buf.is_live() && !buf.is_hidden()
+    } else {
+        false
+    }
+}
+
+/// Return most recently selected buffer other than BUFFER.
+/// Buffers not visible in windows are preferred to visible buffers, unless
+/// optional second argument VISIBLE-OK is non-nil.  Ignore the argument
+/// BUFFER unless it denotes a live buffer.  If the optional third argument
+/// FRAME specifies a live frame, then use that frame's buffer list instead
+/// of the selected frame's buffer list.
+///
+/// The buffer is found by scanning the selected or specified frame's buffer
+/// list first, followed by the list of all buffers.  If no other buffer
+/// exists, return the buffer `*scratch*' (creating it if necessary).
+#[lisp_fn(min = "0")]
+pub fn other_buffer(buffer: LispObject, visible_ok: LispObject, frame: LispObject) -> LispObject {
+    let f: LispFrameRef = LispFrameLiveOrSelected::from(frame).into();
+    let pred = f.buffer_predicate;
+    let mut notsogood = None;
+
+    let frame_bufs = f
+        .buffer_list
+        .iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off);
+    let all_bufs = LiveBufferIter::new().map(|x| x.into());
+    for buf in frame_bufs.chain(all_bufs) {
+        // If the frame has a buffer_predicate, disregard buffers that
+        // don't fit the predicate.
+        if candidate_buffer(buf, buffer)
+            && pred.map_or_else(|| true, |p| call!(p, buf).is_not_nil())
+        {
+            if visible_ok.is_not_nil() || unsafe { Fget_buffer_window(buf, Qvisible) }.is_nil() {
+                return buf;
+            } else {
+                notsogood = notsogood.or(Some(buf));
+            }
+        }
+    }
+
+    if let Some(buf) = notsogood {
+        buf
+    } else {
+        // TODO: This was AUTO_STRING, which doesn't exist yet in Rust.
+        let scratch = new_unibyte_string!("*scratch*");
+        let mut buf = Fget_buffer(scratch);
+        if buf.is_nil() {
+            buf = Fget_buffer_create(scratch);
+            unsafe { Fset_buffer_major_mode(buf) };
+        }
+        buf
+    }
+}
+
+// The following function is a safe variant of Fother_buffer: It doesn't
+// pay attention to any frame-local buffer lists, doesn't care about
+// visibility of buffers, and doesn't evaluate any frame predicates.
+pub fn other_buffer_safely(buffer: LispObject) -> LispObject {
+    for buf in LiveBufferIter::new().map(|x| x.into()) {
+        if candidate_buffer(buf, buffer) {
+            return buf;
+        }
+    }
+
+    let scratch = new_unibyte_string!("*scratch*");
+    let mut buf = Fget_buffer(scratch);
+    if buf.is_nil() {
+        buf = Fget_buffer_create(scratch);
+        unsafe { Fset_buffer_major_mode(buf) };
+    }
+    buf
 }
 
 include!(concat!(env!("OUT_DIR"), "/buffers_exports.rs"));
--- a/test/rust_src/src/buffers-tests.el
+++ b/test/rust_src/src/buffers-tests.el
@@ -7,11 +7,13 @@
 (ert-deftest test-buffer-base-buffer-indirect ()
   (let* ((base (get-buffer-create "base"))
          (ind-buf (make-indirect-buffer base "indbuf")))
-    (should (eq (buffer-base-buffer ind-buf) base))))
+    (should (eq (buffer-base-buffer ind-buf) base))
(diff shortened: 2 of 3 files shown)


rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,80 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
     }
 }
 
+// True if B can be used as 'other-than-BUFFER' buffer.
+fn candidate_buffer(b: LispObject, buffer: LispObject) -> bool {
+    if let Some(buf) = Option::<LispBufferRef>::from(b) {
```
simon-frankau: It looks like the preferred style is to use X.into(), but I can't see a way to make that work with the Rust type system here. Any suggestions?

test/rust_src/src/buffers-tests.el
```diff
@@ -7,11 +7,13 @@
 (ert-deftest test-buffer-base-buffer-indirect ()
   (let* ((base (get-buffer-create "base"))
          (ind-buf (make-indirect-buffer base "indbuf")))
-    (should (eq (buffer-base-buffer ind-buf) base))))
+    (should (eq (buffer-base-buffer ind-buf) base))
```
simon-frankau: I've tried to make all the tests self-contained - previously, they weren't so tests were relying on state from other tests. Knowing what buffers are present is pretty much needed to test `other-buffer` cleanly.

rust_src/src/buffers.rs
shaleh: `From` is ok when required.

test/rust_src/src/buffers-tests.el
```diff
@@ -104,7 +122,33 @@
...
+(ert-deftest test-other-buffer-many ()
+  (let* ((msgs (get-buffer "*Messages*"))
+         (buf1 (get-buffer-create "1"))
+         (buf2 (get-buffer-create "2"))
+         (buf3 (get-buffer-create "3"))
+         (other (other-buffer msgs)))
+    (should (string= (buffer-name other) "1"))
+    (kill-buffer buf1)
+    (kill-buffer buf2)
+    (kill-buffer buf3)))
```
simon-frankau: I'm not quite sure how much testing to throw in. Fully testing `other-buffer`'s functionality would require writing a fair chunk of tests (exercising the predicates, visibility, frame-local buffer lists, etc.), and my elisp knowledge is pretty weak. If you think it's worth it, I can add more, but I don't want to create a huge pile of tests for a single function if that looks silly.

rust_src/src/buffers.rs
shaleh: This could be written as a match to be more idiomatic Rust.

```
match b.as_buffer() {
    Some(buf) => !b.eq(buffer) && buf.is_live() && !buf.is_hidden(),
    None => false
}
```

You might be able to use `match b.into()` here but I suspect the type checker won't allow it. `as_buffer` might end up being deprecated but it is ok to use it for now.

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,80 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
     }
 }
 
+// True if B can be used as 'other-than-BUFFER' buffer.
+fn candidate_buffer(b: LispObject, buffer: LispObject) -> bool {
```
shaleh: Can this function replace the C function?

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,80 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+pub fn other_buffer(buffer: LispObject, visible_ok: LispObject, frame: LispObject) -> LispObject {
+    let f: LispFrameRef = LispFrameLiveOrSelected::from(frame).into();
+    let pred = f.buffer_predicate;
+    let mut notsogood = None;
+
+    let frame_bufs = f
+        .buffer_list
+        .iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off);
+    let all_bufs = LiveBufferIter::new().map(|x| x.into());
+    for buf in frame_bufs.chain(all_bufs) {
```
shaleh: Why use `chain` here instead of `map` on the `frame_bufs` iter?

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,80 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+        {
+            if visible_ok.is_not_nil() || unsafe { Fget_buffer_window(buf, Qvisible) }.is_nil() {
+                return buf;
+            } else {
+                notsogood = notsogood.or(Some(buf));
+            }
+        }
+    }
+
+    if let Some(buf) = notsogood {
```
shaleh: `let Some()` with an else is a `match`.

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,80 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+#[lisp_fn(min = "0")]
+pub fn other_buffer(buffer: LispObject, visible_ok: LispObject, frame: LispObject) -> LispObject {
+    let f: LispFrameRef = LispFrameLiveOrSelected::from(frame).into();
+    let pred = f.buffer_predicate;
+    let mut notsogood = None;
+
+    let frame_bufs = f
+        .buffer_list
+        .iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off);
+    let all_bufs = LiveBufferIter::new().map(|x| x.into());
```
shaleh: This `into` might be better written using a `From` so the intent is clearer.

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,80 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+            unsafe { Fset_buffer_major_mode(buf) };
+        }
+        buf
+    }
+}
+
+// The following function is a safe variant of Fother_buffer: It doesn't
+// pay attention to any frame-local buffer lists, doesn't care about
+// visibility of buffers, and doesn't evaluate any frame predicates.
+pub fn other_buffer_safely(buffer: LispObject) -> LispObject {
```
shaleh: Is anything able to use this?

test/rust_src/src/buffers-tests.el
shaleh: Are there tests in the expensive suite? The more tests we have the more flexibility we have to experiment with editor internals.

shaleh: It is ok to add simple ones and open an Issue for "feature X needs more testing".

rust_src/src/buffers.rs
simon-frankau: Thanks.

simon-frankau: Thanks, done.

rust_src/src/buffers.rs
simon-frankau: Done.

rust_src/src/buffers.rs
simon-frankau: Yes. I've just `pub extern "C"`'d it.

rust_src/src/buffers.rs
simon-frankau: Done.

rust_src/src/buffers.rs
simon-frankau: Or rather, it was only used by other_buffer and other_buffer_safely, and I've made the latter visible to C, so the C implementation of both of those can go.

rust_src/src/buffers.rs
simon-frankau: I've made it replace the C implementation.

test/rust_src/src/buffers-tests.el
simon-frankau: I couldn't see any further tests of other-buffer. I've added a test for the predicate, but the part with visibility requires much more understanding of how to test frames/windows than I currently have.

simon-frankau:
I think I've made all the changes I intended to do here, and would appreciate someone taking another pass. (I'm also happy to make further changes as required.)

I'm also new to github review workflows, so now I'll try pressing the "re-request review" button and see if it does what I expect. :)

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,81 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+/// optional second argument VISIBLE-OK is non-nil.  Ignore the argument
+/// BUFFER unless it denotes a live buffer.  If the optional third argument
+/// FRAME specifies a live frame, then use that frame's buffer list instead
+/// of the selected frame's buffer list.
+///
+/// The buffer is found by scanning the selected or specified frame's buffer
+/// list first, followed by the list of all buffers.  If no other buffer
+/// exists, return the buffer `*scratch*' (creating it if necessary).
+#[lisp_fn(min = "0")]
+pub fn other_buffer(buffer: LispObject, visible_ok: LispObject, frame: LispObject) -> LispObject {
```
shaleh:
```suggestion
pub fn other_buffer(buffer: LispObject, visible_ok: LispObject, frame: LispFrameLiveOrSelected) -> LispObject {
```

Followed by:
`let f: LispFrameRef = frame.into();`.

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,81 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+// pay attention to any frame-local buffer lists, doesn't care about
+// visibility of buffers, and doesn't evaluate any frame predicates.
+#[no_mangle]
+pub extern "C" fn other_buffer_safely(buffer: LispObject) -> LispObject {
+    for buf in LiveBufferIter::new().map(|x| x.into()) {
+        if candidate_buffer(buf, buffer) {
+            return buf;
+        }
+    }
+
```
shaleh: This block looks like a shared function.

rust_src/src/buffers.rs
simon-frankau: Thanks. Done.

rust_src/src/buffers.rs
simon-frankau: Oh, yes. Thanks.

rust_src/src/buffers.rs
```diff
@@ -1843,4 +1847,81 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+        Some(buf) => buf,
+        None => get_scratch_buf(),
+    }
+}
+
+// The following function is a safe variant of Fother_buffer: It doesn't
+// pay attention to any frame-local buffer lists, doesn't care about
+// visibility of buffers, and doesn't evaluate any frame predicates.
+#[no_mangle]
+pub extern "C" fn other_buffer_safely(buffer: LispObject) -> LispObject {
```
shaleh: Do we want to take this further into Rust style and make it:
```
let found = LiveBufferIter::new().map(|x| x.into()).find(|buf| candidate_buffer(buf, buffer));
match found {
    Some(buf) => buf,
    None => get_scratch_buf(),
}
```

?

simon-frankau: This looks better to me, so I've added it.

Sorry for delay - work's been distracting!

agraven:
Some tests are failing, it looks like buffers aren't being named correctly for some reason. An example:
```
(ert-test-failed
  ((should (string= (buffer-name other) "1"))
      :form (string= "test-move-overlay-in-buffer-add-to-buffer" "1")
```

PR #1497, opened 2019-05-13
restore ported C code, wrap in #ifndef PORTED_TO_RUST

shaleh:
The goal here is to make creating and reviewing merges from emacs upstream easier. This is an experiment at the moment.


brotzeit:
Maybe we should start again by restoring all code from emacs(6695c1be51) and then put our rust code in step by step. This way we can also see where the mac builds stopped working.

brotzeit:
I think it's more important for now that we get our tests green than to port more new code. Get to a stable point again.

brotzeit:
I'm also trying to find a solution for this. I will give you an update when I got something that works.

shaleh:
Closing this one, brotzeit has the right idea.

PR #1499, opened 2019-05-14
Port defvar

Flowneee:
Closes https://github.com/remacs/remacs/issues/1475

### DIFF
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -18,7 +18,7 @@ use crate::{
     multibyte::LispStringRef,
     obarray::loadhist_attach,
     objects::equal,
-    remacs_sys::specbind_tag::*,
+    remacs_sys::specbind_tag,
     remacs_sys::{
         backtrace_debug_on_exit, build_string, call_debugger, check_cons_list, do_debug_on_call,
         do_one_unbind, eval_sub, funcall_lambda, funcall_subr, globals, grow_specpdl,
@@ -52,7 +52,7 @@ pub unsafe extern "C" fn record_unwind_protect(
     let unwind = (*ThreadState::current_thread().m_specpdl_ptr)
         .unwind
         .as_mut();
-    unwind.set_kind(SPECPDL_UNWIND);
+    unwind.set_kind(specbind_tag::SPECPDL_UNWIND);
     unwind.func = function;
     unwind.arg = arg;
     grow_specpdl();
@@ -66,7 +66,7 @@ pub unsafe extern "C" fn record_unwind_protect_ptr(
     let unwind = (*ThreadState::current_thread().m_specpdl_ptr)
         .unwind_ptr
         .as_mut();
-    unwind.set_kind(SPECPDL_UNWIND_PTR);
+    unwind.set_kind(specbind_tag::SPECPDL_UNWIND_PTR);
     unwind.func = function;
     unwind.arg = arg;
     grow_specpdl();
@@ -80,7 +80,7 @@ pub unsafe extern "C" fn record_unwind_protect_int(
     let unwind = (*ThreadState::current_thread().m_specpdl_ptr)
         .unwind_int
         .as_mut();
-    unwind.set_kind(SPECPDL_UNWIND_INT);
+    unwind.set_kind(specbind_tag::SPECPDL_UNWIND_INT);
     unwind.func = function;
     unwind.arg = arg;
     grow_specpdl();
@@ -91,7 +91,7 @@ pub unsafe extern "C" fn record_unwind_protect_void(function: Option<unsafe exte
     let unwind = (*ThreadState::current_thread().m_specpdl_ptr)
         .unwind_void
         .as_mut();
-    unwind.set_kind(SPECPDL_UNWIND_VOID);
+    unwind.set_kind(specbind_tag::SPECPDL_UNWIND_VOID);
     unwind.func = function;
     grow_specpdl();
 }
--- a/rust_src/src/eval.rs
+++ b/rust_src/src/eval.rs
@@ -12,7 +12,7 @@ use crate::{
         defalias, default_boundp, fset, indirect_function, indirect_function_lisp, set, set_default,
     },
     lisp::is_autoload,
-    lisp::{ExternalPtr, LispObject, LispSubrRef},
+    lisp::{LispObject, LispSubrRef, SpecbindingRef},
     lists::{assq, car, cdr, get, memq, nth, put},
     lists::{LispCons, LispConsCircularChecks, LispConsEndChecks},
     multibyte::LispStringRef,
@@ -23,8 +23,8 @@ use crate::{
         backtrace_debug_on_exit, build_string, call_debugger, check_cons_list, do_debug_on_call,
         do_one_unbind, eval_sub, funcall_lambda, funcall_subr, globals, grow_specpdl,
         internal_catch, internal_lisp_condition_case, list2, maybe_gc, maybe_quit,
-        record_in_backtrace, record_unwind_save_match_data, signal_or_quit, specbind, specbinding,
-        COMPILEDP, MODULE_FUNCTIONP,
+        record_in_backtrace, record_unwind_save_match_data, signal_or_quit, specbind, COMPILEDP,
+        MODULE_FUNCTIONP,
     },
     remacs_sys::{pvec_type, EmacsInt, Lisp_Compiled, Set_Internal_Bind},
     remacs_sys::{Fapply, Fcons, Fdefault_value, Fload},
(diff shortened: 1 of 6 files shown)


rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,143 @@ pub fn condition_case(args: LispCons) -> LispObject {
     unsafe { internal_lisp_condition_case(var, bodyform, handlers) }
 }
 
+type SpecbindingRef = ExternalPtr<specbinding>;
+
```
shaleh: We have made similar functions into member functions of the type. Does that makes sense to do for `SpecBindingRef` as well? Look at `LispBufferRef` as an example.

rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,143 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+    let mut binding = SpecbindingRef::new(std::ptr::null_mut());
+    let mut pdl = SpecbindingRef::new(current_thread.m_specpdl_ptr);
+
+    while pdl > specpdl {
+        unsafe {
+            pdl.ptr_sub(1);
+        }
+        match pdl.kind() {
+            ref x if [SPECPDL_LET_DEFAULT, SPECPDL_LET].contains(x) => {
```
shaleh: Why use a guard here instead of `SPECPDL_LET_DEFAULT | SPECPDL_LET => { ... }`?

rust_src/src/eval.rs
Flowneee: > We have made similar functions into member functions of the type.

Can you explain about what functions you talking? If about `specpdl_*`, I made them as functions because they used from C code, so I don't think that make them into methods make sense.

rust_src/src/eval.rs
Flowneee: I actually don't remember, but now I see that you're right. Fixed

rust_src/src/eval.rs
shaleh: Use by C code *today*. As we port the code to Rust will they continue to be used as functions or could we be operating on the specpdl type itself? It is a common pattern in the port to have `extern "C"` functions which act as wrappers around the Rust types and methods.

Flowneee: Hm, you right. I'm gonna do that a little bit later.

shaleh: What we have been aiming for is the smallest possible `extern "C"` functions which simply translate the C needs into Rust and the Rust result back into C. This way the Rust code can use the desired Rust idioms and style. This ends up looking like:

```
pub extern "C" fn foo(buffer: LispObject) -> LispObject {
    let result = foo_rust(buffer.into());
    result.into()
}

pub fn foo_rust(buffer: LispBufferRef) -> int32 {
    // all of the original C logic here
}
```

rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,135 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+
+    if tail.is_not_nil() {
+        if LispCons::from(tail).length() > 2 {
+            error!("Too many arguments");
+        }
+
+        let sym: LispSymbolRef = sym_obj.into();
+        let tem = default_boundp(sym);
+
+        /* Do it before evaluating the initial value, for self-references.  */
```
agraven: Comments should be with two slashes, as in `// Do it before...`

rust_src/src/eval.rs
Flowneee: Okay, I made `specpdl_*` into methods of `SpecbindingRef` type and moved `SpecbindingRef` (and its impl) to `lisp.rs`, because `specbinding` originally was defined in `lisp.h`. I also, as you suggested, split `default_toplevel_binding` into `_rust` and `extern "C"` functions.

rust_src/src/eval.rs
Flowneee: Done

rust_src/src/eval.rs
Flowneee: @shaleh Hi. Can you review changes again?

rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,132 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+pub fn defvar(args: LispCons) -> LispObject {
+    let (sym_obj, tail) = args.into();
+
+    if tail.is_not_nil() {
+        if LispCons::from(tail).length() > 2 {
+            error!("Too many arguments");
+        }
+
+        let sym: LispSymbolRef = sym_obj.into();
+        let tem = default_boundp(sym);
```
agraven: I would recommend giving this variable a more meaningful name, e.g. `has_default`.

rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,132 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+                tem = purecopy(tem);
+            }
+            put(sym, Qvariable_documentation, tem);
+        }
+        loadhist_attach(sym_obj);
+    } else if unsafe { globals.Vinternal_interpreter_environment }.is_not_nil()
+        && sym_obj
+            .as_symbol()
+            .map(|x| !x.get_declared_special())
+            .unwrap_or(false)
```
agraven: You can shorten the last two methods to `.map_or(false, |x| !x.get_declared_special())`

rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,132 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+    {
+        // A simple (defvar foo) with lexical scoping does "nothing" except
+        // declare that var to be dynamically scoped *locally* (i.e. within
+        // the current file or let-block).
+        unsafe {
+            globals.Vinternal_interpreter_environment =
+                Fcons(sym_obj, globals.Vinternal_interpreter_environment);
```
agraven: The preferred way of writing this is `LispObject::cons(sym_obj, globals.Vinternal_interpreter_environment)`

rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,132 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+pub extern "C" fn specpdl_old_value(pdl: SpecbindingRef) -> LispObject {
+    pdl.specpdl_old_value()
+}
+
+#[no_mangle]
+pub extern "C" fn set_specpdl_old_value(mut pdl: SpecbindingRef, val: LispObject) {
+    pdl.set_specpdl_old_value(val)
+}
+
+pub fn default_toplevel_binding_rust(symbol: LispSymbolRef) -> SpecbindingRef {
```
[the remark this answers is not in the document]
Flowneee: Done

rust_src/src/lisp.rs
```diff
@@ -279,6 +287,27 @@ impl From<LispObject> for Option<LispSubrRef> {
...
+        debug_assert!(self.kind() >= SPECPDL_LET);
+        unsafe { self.let_.as_ref().symbol }.into()
+    }
+
+    pub fn specpdl_old_value(&self) -> LispObject {
+        debug_assert!(self.kind() >= SPECPDL_LET);
+        unsafe { self.let_.as_ref().old_value }
+    }
+
+    pub fn set_specpdl_old_value(&mut self, val: LispObject) {
```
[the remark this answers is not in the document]
Flowneee: Done

rust_src/src/eval.rs
```diff
@@ -1338,4 +1338,132 @@ pub fn condition_case(args: LispCons) -> LispObject {
...
+/// The optional argument DOCSTRING is a documentation string for the
+/// variable.
+///
+/// To define a user option, use `defcustom' instead of `defvar'.
+/// usage: (defvar SYMBOL &optional INITVALUE DOCSTRING)
+#[lisp_fn(min = "1", unevalled = "true")]
+pub fn defvar(args: LispCons) -> LispObject {
+    let (sym_obj, tail) = args.into();
+
+    if tail.is_not_nil() {
```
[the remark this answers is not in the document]
Flowneee: Done

rust_src/src/eval.rs
Flowneee: Done

rust_src/src/eval.rs
Flowneee: Done

rust_src/src/eval.rs
Flowneee: Done

rust_src/src/lisp.rs
```diff
@@ -20,13 +20,15 @@ use crate::{
     lists::{list, memq, CarIter, LispCons, LispConsCircularChecks, LispConsEndChecks},
     multibyte::LispStringRef,
     process::LispProcessRef,
+    remacs_sys::specbind_tag::*,
```
shaleh: Why not import `specbind_tag` and use it `specbind_tag::SPECPDL_LET`?

`*` imports have a way of biting you at some point.

rust_src/src/symbols.rs
```diff
@@ -16,9 +16,10 @@ use crate::{
     },
     frames::selected_frame,
     hashtable::LispHashTableRef,
-    lisp::{ExternalPtr, LispObject, LispStructuralEqual},
+    lisp::{ExternalPtr, LispObject, LispStructuralEqual, SpecbindingRef},
     lists::LispCons,
     multibyte::LispStringRef,
+    remacs_sys::specbind_tag::*,
```
shaleh: Same comment as lisp.rs.

rust_src/src/symbols.rs
```diff
@@ -212,6 +214,32 @@ impl LispSymbolRef {
...
+    pub fn default_toplevel_binding_rust(&self) -> SpecbindingRef {
+        let current_thread = ThreadState::current_thread();
+        let specpdl = SpecbindingRef::new(current_thread.m_specpdl);
+
+        let mut binding = SpecbindingRef::new(std::ptr::null_mut());
+        let mut pdl = SpecbindingRef::new(current_thread.m_specpdl_ptr);
+
+        while pdl > specpdl {
+            unsafe {
+                pdl.ptr_sub(1);
```
shaleh: where is `ptr_sub` coming from? I know of `ptr::sub`. That does not change the pointer but instead returns a new one. So it would be `pdl = pdl.sub(1);`

agraven: That's coming from a method on `ExternalPtr`, it's the one(s) for modifying the pointer I added.

rust_src/src/lisp.rs
Flowneee: I don't think that there will be another `SPECPDL_LET` but agreed, better to be sure. Fixed

rust_src/src/symbols.rs
Flowneee: Done

rust_src/src/symbols.rs
Flowneee: Thx @agraven for explanation. I believe this should be okay then.

Flowneee:
@shaleh Hi. If everything is okay, can you merge?

rust_src/src/lisp.rs
shaleh: I still see the '*' import here.

rust_src/src/lisp.rs
```diff
@@ -279,6 +287,27 @@ impl From<LispObject> for Option<LispSubrRef> {
     }
 }
```
shaleh: Should this be in symbols.rs?

rust_src/src/lisp.rs
Flowneee: Whoops, my bad. Forgot to stage file. Fixed.

rust_src/src/lisp.rs
Flowneee: I don't think so. Original `specbinding` struct defined is `lisp.h`, so I think it makes more sense if we place `SpecbindingRef` in `lisp.rs` as well.

PR #1502, opened 2019-05-18
Make Codepoint a newtype

### LINKED ISSUE #1501: Discussion: Make Codepoint a newtype?
I propose making `multibyte::Codepoint` a `#[repr(transparent)]` newtype instead of a type alias, as it makes it easier to define methods on it. An example of a method could be a port of `char_width` from `character.c`

agraven:
Lots of functions have been made methods on Codepoint instead. There
have also been made lots of changes around the code to accommodate the
type change.

Closes #1501.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -366,13 +366,13 @@ impl LispBufferRef {
             0
         };
 
-        unsafe {
-            let (c, _) = multibyte_char_at(slice::from_raw_parts(
+        let (c, _) = multibyte_char_at(unsafe {
+            slice::from_raw_parts(
                 self.beg_addr().offset(offset + n - self.beg_byte()),
                 MAX_MULTIBYTE_LENGTH,
-            ));
-            c
-        }
+            )
+        });
+        c
     }
 
     pub fn multibyte_characters_enabled(self) -> bool {
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -1,9 +1,9 @@
 //! Functions operating on buffers.
 
-use std::{self, iter, mem, ops, ptr};
+use std::{self, iter, mem, ops, ptr, slice};
 
 use field_offset::FieldOffset;
-use libc::{self, c_char, c_int, c_uchar, c_void, ptrdiff_t};
+use libc::{self, c_char, c_uchar, c_void, ptrdiff_t};
 
 use rand::{thread_rng, Rng};
 
@@ -28,8 +28,9 @@ use crate::{
         build_marker, build_marker_rust, marker_buffer, marker_position_lisp, set_marker_both,
         LispMarkerRef, MARKER_DEBUG,
     },
-    multibyte::{multibyte_chars_in_text, multibyte_length_by_head, string_char},
-    multibyte::{LispStringRef, LispSymbolOrString},
+    multibyte::MAX_MULTIBYTE_LENGTH,
+    multibyte::{multibyte_char_at, multibyte_chars_in_text, multibyte_length_by_head},
+    multibyte::{Codepoint, LispStringRef, LispSymbolOrString},
     numbers::{LispNumber, MOST_POSITIVE_FIXNUM},
     obarray::intern,
     remacs_sys::symbol_trapped_write::SYMBOL_TRAPPED_WRITE,
@@ -347,18 +348,18 @@ impl LispBufferRef {
 
     /// Return character at byte position POS.  See the caveat WARNING for
     /// FETCH_MULTIBYTE_CHAR below.
-    pub fn fetch_char(self, n: ptrdiff_t) -> c_int {
+    pub fn fetch_char(self, n: ptrdiff_t) -> Codepoint {
         if self.multibyte_characters_enabled() {
             self.fetch_multibyte_char(n)
         } else {
-            c_int::from(self.fetch_byte(n))
+            Codepoint::from(self.fetch_byte(n))
         }
     }
 
     /// Return character code of multi-byte form at byte position POS.  If POS
     /// doesn't point the head of valid multi-byte form, only the byte at
     /// POS is returned.  No range checking.
-    pub fn fetch_multibyte_char(self, n: ptrdiff_t) -> c_int {
+    pub fn fetch_multibyte_char(self, n: ptrdiff_t) -> Codepoint {
         let offset = if n >= self.gpt_byte() && n >= 0 {
             self.gap_size()
         } else {
@@ -366,11 +367,11 @@ impl LispBufferRef {
         };
 
         unsafe {
-            string_char(
+            let (c, _) = multibyte_char_at(slice::from_raw_parts(
                 self.beg_addr().offset(offset + n - self.beg_byte()),
-                ptr::null_mut(),
-                ptr::null_mut(),
-            )
+                MAX_MULTIBYTE_LENGTH,
+            ));
+            c
         }
     }
 
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -17,12 +17,6 @@ impl LispObject {
         self.as_fixnum()
             .map_or(false, |i| 0 <= i && i <= EmacsInt::from(MAX_CHAR))
     }
-
-    /// Check if Lisp object is a character or not and return the codepoint
-    /// Similar to CHECK_CHARACTER
-    pub fn as_character_or_error(self) -> Codepoint {
-        Codepoint::from(self)
-    }
 }
 
 /// True iff byte starts a character in a multibyte form.
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -8,7 +8,6 @@ use crate::{
     lisp::LispObject,
     multibyte::{Codepoint, MAX_CHAR},
     remacs_sys::EmacsInt,
-    remacs_sys::Qcharacterp,
     threads::ThreadState,
 };
 
@@ -22,10 +21,7 @@ impl LispObject {
     /// Check if Lisp object is a character or not and return the codepoint
     /// Similar to CHECK_CHARACTER
     pub fn as_character_or_error(self) -> Codepoint {
-        if !self.is_character() {
-            wrong_type!(Qcharacterp, self)
-        }
-        Codepoint::from(self.as_fixnum().unwrap() as u32)
+        Codepoint::from(self)
     }
 }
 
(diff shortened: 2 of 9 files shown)


rust_src/src/multibyte.rs
```diff
@@ -42,32 +43,285 @@ use crate::{
...
-pub const MAX_CHAR: Codepoint = (1 << char_bits::CHARACTERBITS as usize) - 1;
+pub const MAX_CHAR: u32 = (1 << char_bits::CHARACTERBITS as usize) - 1;
+
+// Maximum character codes for several encoded lengths
+/// Maximum value for a single byte codepoint
+pub const MAX_1_BYTE_CHAR: u32 = 0x7F;
+
+/// Minimum value for a two byte codepoint
+pub const MIN_2_BYTE_CHAR: u32 = 0x80;
+/// Minimum value for a two byte codepoint
```
db48x: s/Minimum/Maximum/

rust_src/src/multibyte.rs
```diff
@@ -42,32 +43,285 @@ use crate::{
...
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct Codepoint(u32);
+
+impl Codepoint {
+    // Equivalent to BYTE8_TO_CHAR
+    /// Create a codepoint from a raw byte. If non-ascii, byte8 encode it.
+    pub fn from_raw(byte: u8) -> Codepoint {
+        match Codepoint::from(byte) {
+            cp if cp.is_ascii() => cp,
+            cp => Codepoint::from(cp.0 + 0x3F_FF00),
```
db48x: Should use the constant `BYTE8_OFFSET` which you added

rust_src/src/multibyte.rs
```diff
@@ -42,32 +43,285 @@ use crate::{
...
+    // Equivalent to CHAR_STRING
+    /// Write the codepoint to the given slice. The slice needs to be big
+    /// enough to hold the resulting bytes. Returns the amount of bytes written
+    pub fn write_to(self, to: &mut [u8]) -> usize {
+        let cp: u32 = self.into();
+        if cp <= MAX_1_BYTE_CHAR.into() {
+            to[0] = cp as u8;
+            1
+        } else if cp <= MAX_2_BYTE_CHAR {
+            // note: setting later bytes first to avoid multiple bound checks
```
db48x: +1

db48x: although you might move the comment up a level, since it applies to most of the branches.

rust_src/src/multibyte.rs
```diff
@@ -42,32 +43,285 @@ use crate::{
...
+            return Codepoint::from(cp);
+        }
+        let ascii = (cp & 0x7F) as u8;
+        // For Meta, Shift, and Control modifiers, we need special care.
+        if cp & char_bits::CHAR_SHIFT != 0 {
+            let unshifted = cp & !char_bits::CHAR_SHIFT;
+            // Shift modifier is valid only with [A-Za-z].
+            // Shift modifier for control characters and SPC is ignored.
+            if (ascii >= b'A' && ascii <= b'Z') || ascii <= b' ' {
+                cp = unshifted;
```
db48x: This logic is super confusing. But probably not more confusing than the original C.

agraven: I'll try to see if I can refactor it to make it more readable, but I didn't write it, only made it a method, so I'm equally confused.

rust_src/src/multibyte.rs
agraven: +1 as in "this is good" or "change the 2 to a 3"?

rust_src/src/multibyte.rs
agraven: Fixed

rust_src/src/multibyte.rs
db48x: Yea, I'm not going to make you rewrite it. I just can't tell if it's correct when looking at it. If the copies were faithful then it's all probably fine.

rust_src/src/editfns.rs
```diff
@@ -226,16 +223,15 @@ pub fn position_bytes(position: LispNumber) -> Option<EmacsInt> {
...
 pub fn insert_byte(byte: EmacsInt, count: Option<EmacsInt>, inherit: bool) {
-    if byte < 0 || byte > 255 {
-        args_out_of_range!(byte, 0, 255)
-    }
+    check_range(byte, 0, 255);
+    let byte = byte as u8;
     let buf = ThreadState::current_buffer_unchecked();
-    let toinsert = if byte >= 128 && buf.multibyte_characters_enabled() {
-        EmacsInt::from(raw_byte_codepoint(byte as c_uchar))
+    let toinsert = if buf.multibyte_characters_enabled() {
```
agraven: Note that I removed the is_ascii check because `Codepoint::from_raw` already does, making it redundant

rust_src/src/buffers.rs
```diff
@@ -347,30 +348,30 @@ impl LispBufferRef {
...
+    pub fn fetch_multibyte_char(self, n: ptrdiff_t) -> Codepoint {
         let offset = if n >= self.gpt_byte() && n >= 0 {
             self.gap_size()
         } else {
             0
         };
 
         unsafe {
-            string_char(
+            let (c, _) = multibyte_char_at(slice::from_raw_parts(
```
nickdrozd:
```suggestion
            let (c, _) = multibyte_char_at(unsafe {slice::from_raw_parts(
```

rust_src/src/character.rs
```diff
@@ -23,10 +21,7 @@ impl LispObject {
     /// Check if Lisp object is a character or not and return the codepoint
     /// Similar to CHECK_CHARACTER
     pub fn as_character_or_error(self) -> Codepoint {
```
nickdrozd: Is there any reason to keep this around?

agraven: I don't think so

rust_src/src/character.rs
```diff
@@ -83,25 +78,23 @@ pub fn char_or_string_p(object: LispObject) -> bool {
 
 /// Convert the byte CH to multibyte character.
 #[lisp_fn]
-pub fn unibyte_char_to_multibyte(ch: LispObject) -> LispObject {
-    let c = ch.as_character_or_error();
-    if c >= 0x100 {
-        error!("Not a unibyte character: {}", c);
+pub fn unibyte_char_to_multibyte(ch: Codepoint) -> Codepoint {
```
nickdrozd: I don't understand how this stuff works, but would it be possible to further break `Codepoint` into `Unibyte` and `Multibyte` types? Or something like that?

rust_src/src/cmds.rs
```diff
@@ -341,7 +338,7 @@ fn internal_self_insert(mut c: Codepoint, n: usize) -> EmacsInt {
         // or before a tab if it doesn't use the whole width of the tab.  */
         if overwrite == Qoverwrite_mode_binary {
             chars_to_delete = n as usize;
-        } else if c != '\n' as Codepoint && c2 != '\n' as Codepoint {
+        } else if c != Codepoint::from('\n') && c2 != Codepoint::from('\n') {
```
nickdrozd: Are the conversion needed here?

rust_src/src/cmds.rs
```diff
@@ -413,15 +410,15 @@ fn internal_self_insert(mut c: Codepoint, n: usize) -> EmacsInt {
...
+        let mc = if current_buffer.multibyte_characters_enabled() && c.is_single_byte() {
+            c.unibyte_to_char()
         } else {
             c
         };
         let mut string = unsafe { Fmake_string(n.into(), mc.into(), Qnil) };
         if spaces_to_insert > 0 {
             let tem =
-                unsafe { Fmake_string(spaces_to_insert.into(), (' ' as Codepoint).into(), Qnil) };
+                unsafe { Fmake_string(spaces_to_insert.into(), Codepoint::from(' ').into(), Qnil) };
```
nickdrozd: Is there a way to avoid the double conversion?

rust_src/src/cmds.rs
agraven: Nope, that's a leftover from before i implented `PartialEq<char>`

rust_src/src/cmds.rs
agraven: `b' '.into()` would work

rust_src/src/multibyte.rs
```diff
@@ -42,32 +43,291 @@ use crate::{
...
+            MIN_3_BYTE_CHAR..=MAX_3_BYTE_CHAR => 3,
+            MIN_4_BYTE_CHAR..=MAX_4_BYTE_CHAR => 4,
+            MIN_5_BYTE_CHAR..=MAX_5_BYTE_CHAR => 5,
+            MIN_BYTE8_CHAR..=MAX_BYTE8_CHAR => 2,
+            _ => invalid_character(self),
+        }
+    }
+
+    /// Get the integer value of the codepoint
+    pub const fn val(self) -> u32 {
```
nickdrozd: AFAICT, `val` is only used for conversions, eg `cp.val() as c_int`. If possible, I think it would be cleaner to drop this and just implement the remaining conversions directly.

rust_src/src/character.rs
agraven: It would be possible but I think it would create more problems than it would solve, especially when it comes to byte8 handling

rust_src/src/cmds.rs
```diff
@@ -378,18 +375,18 @@ fn internal_self_insert(mut c: Codepoint, n: usize) -> EmacsInt {
         }
         hairy = 2;
     }
-    synt = unsafe { syntax_property(c as i32, true) };
+    synt = unsafe { syntax_property(c.val() as i32, true) };
 
     let previous_char = if current_buffer.multibyte_characters_enabled() {
-        preceding_char() as Codepoint
+        Codepoint::from(preceding_char() as u32)
```
nickdrozd: Double conversions here too

nickdrozd:
Great job. As usual, I support anything that makes for shorter / fewer lines.

rust_src/src/multibyte.rs
agraven: Those conversions would be fallible though, and anything involving tryfrom would just introduce more complexity. Adding as_i32 style methods would work though, if that's what you're thinking of

rust_src/src/cmds.rs
agraven: I don't think there's a good way of avoiding a double here, as I still don't feel comfortable adding conversions which could fail (or saturate, I suppose).

PR #1503, opened 2019-05-19
Port make-marker [WIP?]

### LINKED ISSUE #1494: Port make-marker
```c
DEFUN ("make-marker", Fmake_marker, Smake_marker, 0, 0, 0,
       doc: /* Return a newly allocated marker which does not point at any place.  */)
  (void)
{
  register Lisp_Object val;
  register struct Lisp_Marker *p;

  val = allocate_misc (Lisp_Misc_Marker);
  p = XMARKER (val);
  p->buffer = 0;
  p->bytepos = 0;
  p->charpos = 0;
  p->next = NULL;
  p->insertion_type = 0;
  p->need_adjustment = 0;
  return val;
}
```

sehnsucht13:
This should hopefully close #1494.

I am quite new to rust so I am not too sure if the choices I made when it came to porting `XMARKER` are correct but hopefully this is enough to solve this issue. I also tried thinking of some tests but was not too sure as to how to test this exactly. Either way, I am looking forward to the code review. Thanks in advance!

### DIFF
--- a/rust_src/src/alloc.rs
+++ b/rust_src/src/alloc.rs
@@ -1,14 +1,16 @@
 //! Storage allocation and gc
 
 use remacs_macros::lisp_fn;
+use std::ptr;
 
 use crate::{
     lisp::{ExternalPtr, LispObject},
+    marker::LispMarkerRef,
     remacs_sys::globals,
     remacs_sys::Lisp_Type::Lisp_Vectorlike,
     remacs_sys::{
-        allocate_record, bool_vector_fill, bool_vector_set, bounded_number,
-        make_uninit_bool_vector, purecopy as c_purecopy,
+        allocate_misc, allocate_record, bool_vector_fill, bool_vector_set, bounded_number,
+        make_uninit_bool_vector, purecopy as c_purecopy, Lisp_Misc_Type,
     },
     remacs_sys::{EmacsInt, EmacsUint},
 };
@@ -110,6 +112,21 @@ pub fn purecopy(obj: LispObject) -> LispObject {
     } else {
         unsafe { c_purecopy(obj) }
     }
+}
+
+/// Return a newly allocated marker which does not point to any place.
+#[lisp_fn]
+pub fn make_marker() -> LispMarkerRef {
+    let mut marker = unsafe { allocate_misc(Lisp_Misc_Type::Lisp_Misc_Marker) }.force_marker();
+
+    // Set the properties of the marker to nothing
+    marker.set_buffer(ptr::null_mut());
+    marker.set_charpos(0isize);
+    marker.set_bytepos(0isize);
+    marker.set_insertion_type(false);
+    marker.set_need_adjustment(false);
+
+    marker
 }
 
 include!(concat!(env!("OUT_DIR"), "/alloc_exports.rs"));
--- a/rust_src/src/marker.rs
+++ b/rust_src/src/marker.rs
@@ -126,6 +126,14 @@ impl LispObject {
     pub fn as_marker(self) -> Option<LispMarkerRef> {
         self.into()
     }
+
+    pub unsafe fn to_marker_unchecked(self) -> LispMarkerRef {
+        LispMarkerRef::new(self.get_untaggedptr() as *mut Lisp_Marker)
+    }
+
+    pub fn force_marker(self) -> LispMarkerRef {
+        unsafe { self.to_marker_unchecked() }
+    }
 }
 
 impl LispMiscRef {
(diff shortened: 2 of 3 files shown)


rust_src/src/alloc.rs
```diff
@@ -112,4 +114,20 @@ pub fn purecopy(obj: LispObject) -> LispObject {
     }
 }
 
+/// Return a newly allocated marker which does not point to any place.
+#[lisp_fn(c_name = "make_marker", name = "make-marker")]
+pub fn make_marker_lisp() -> LispObject {
```
agraven: You can simply name the function `make_marker` and remove the `c_name` and `name` from the `lisp_fn`, those are only in case the desired name is already taken.

rust_src/src/alloc.rs
```diff
@@ -112,4 +114,20 @@ pub fn purecopy(obj: LispObject) -> LispObject {
...
+    let val: LispObject = unsafe { allocate_misc(Lisp_Misc_Type::Lisp_Misc_Marker) };
+    let mut p: LispMarkerRef = val.as_marker().unwrap();
```
agraven: These two lines can be written as
```rust
let mut p = unsafe { allocate_misc(Lisp_Misc_Type::Lisp_Misc_Marker) }.force_marker();
```
If `force_marker` doesn't all ready exist, add these methods to the first `impl LispObject` you see in `marker.rs`:
```rust
    pub unsafe fn to_marker_unchecked(self) -> LispMarkerRef {
        LispMarkerRef::new(self.get_untaggedptr() as *mut Lisp_Marker)
    }

    pub fn force_marker(self) -> LispMarkerRef {
        unsafe { self.to_marker_unchecked() }
    }
```

rust_src/src/alloc.rs
```diff
@@ -112,4 +114,20 @@ pub fn purecopy(obj: LispObject) -> LispObject {
...
+
+    // Set the properties of the marker to nothing
+    p.set_buffer(ptr::null_mut());
+    p.set_charpos(0 as isize);
+    p.set_bytepos(0 as isize);
```
agraven: The `as isize` can be removed. (If I'm mistaken and it gives an error if you don't specify the type, then it's more correct to write `0isize`)

sehnsucht13: Thank you, I will take care of the changes shortly!

rust_src/src/alloc.rs
sehnsucht13: There is only one thing I am unsure of regarding this. Should I change the return type of the function to `LispMarkerRef`? Thanks in advance!

sehnsucht13:
Also an additional question, I originally built this with

rust_src/src/alloc.rs
agraven: Yep, generally we want return types to be as specific as possible

rust_src/src/alloc.rs
```diff
@@ -112,4 +114,21 @@ pub fn purecopy(obj: LispObject) -> LispObject {
     }
 }
 
+/// Return a newly allocated marker which does not point to any place.
+#[lisp_fn]
+pub fn make_marker() -> LispMarkerRef {
+    //let val: LispObject = unsafe { allocate_misc(Lisp_Misc_Type::Lisp_Misc_Marker) };
+    //let mut p: LispMarkerRef = val.as_marker().unwrap();
```
agraven: Some commented out lines got left in, please remove them

rust_src/src/alloc.rs
```diff
@@ -112,4 +114,21 @@ pub fn purecopy(obj: LispObject) -> LispObject {
...
+    let mut p = unsafe { allocate_misc(Lisp_Misc_Type::Lisp_Misc_Marker) }.force_marker();
```
agraven: This is a slightly pedantic point, but it would be nice if you could give `p` a slightly more semantically meaningful name like `mark` or `marker`.

rust_src/src/alloc.rs
```diff
@@ -112,4 +114,21 @@ pub fn purecopy(obj: LispObject) -> LispObject {
...
+
+    // Set the properties of the marker to nothing
```
agraven: The added comment is much appreciated, like with the variable names, the original C code can be quite obtuse sometimes.

sehnsucht13:
Thanks again for all of your suggestions and patience. They have all been added now. I have also squashed everything into one commit so the tree does not get filled with random junk. I figured that this is not complex enough to warrant multiple commits.

PR #1507, opened 2019-05-26
[WIP] Port kill-buffer

simon-frankau:
A somewhat literal translation of kill-buffer to Rust.

There's a lot of unsafe code, including plenty of pointer work on linked lists. Compared to e.g. accessors,  I don't see a particularly pleasant way to wrap up the modifications to these structures to hide the pointer-handling away in reusable helper functions (at least, that doesn't feel too forced).

I think useful next steps would be to port the other (no longer static) functions called by kill-buffer over to Rust, and perhaps break apart kill_buffer into smaller functions rather than one giant do-everything function, but I'd appreciate thoughts on what we have so far before going further.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -14,6 +14,7 @@ use crate::{
     character::char_head_p,
     chartable::LispCharTableRef,
     data::Lisp_Fwd,
+    dispnew::is_interactive,
     editfns::{point, widen},
     eval::unbind_to,
     fileio::{expand_file_name, find_file_name_handler},
@@ -33,31 +34,42 @@ use crate::{
     numbers::{LispNumber, MOST_POSITIVE_FIXNUM},
     obarray::intern,
     remacs_sys::symbol_trapped_write::SYMBOL_TRAPPED_WRITE,
-    remacs_sys::Fmake_marker,
     remacs_sys::{
-        alloc_buffer_text, allocate_buffer, allocate_misc, block_input, bset_update_mode_line,
-        buffer_fundamental_string, buffer_local_flags, buffer_local_value, buffer_memory_full,
-        buffer_window_count, del_range, delete_all_overlays, globals, last_per_buffer_idx,
-        lookup_char_property, make_timespec, marker_position, modify_overlay,
-        notify_variable_watchers, per_buffer_default, recenter_overlay_lists,
-        set_buffer_internal_1, set_per_buffer_value, specbind, unblock_input, unchain_both,
-        unchain_marker, update_mode_lines, windows_or_buffers_changed,
+        all_buffers, alloc_buffer_text, allocate_buffer, allocate_misc, block_input,
+        bset_undo_list, bset_update_mode_line, bset_width_table, buffer_fundamental_string,
+        buffer_intervals, buffer_local_flags, buffer_local_value, buffer_memory_full,
+        buffer_window_count, clear_charpos_cache, del_range, delete_all_overlays, do_yes_or_no_p,
+        frames_discard_buffer, free_buffer_text, free_region_cache, globals, internal_delete_file,
+        kill_buffer_processes, kill_buffer_xwidgets, last_per_buffer_idx, lookup_char_property,
+        make_timespec, marker_position, minibuf_window, modify_overlay, notify_variable_watchers,
+        per_buffer_default, recenter_overlay_lists, record_unwind_protect,
+        replace_buffer_in_windows, replace_buffer_in_windows_safely, run_hook,
+        save_excursion_restore, save_excursion_save, set_buffer_internal, set_buffer_internal_1,
+        set_buffer_intervals, set_interval_object, set_per_buffer_value, specbind,
+        swap_out_buffer_local_variables, thread_check_current_buffer, unblock_input, unchain_both,
+        unchain_marker, unlock_buffer, update_mode_lines, windows_or_buffers_changed,
+        SPECPDL_INDEX,
     },
     remacs_sys::{
         buffer_defaults, equal_kind, pvec_type, EmacsInt, Lisp_Buffer, Lisp_Buffer_Local_Value,
-        Lisp_Misc_Type, Lisp_Overlay, Lisp_Type, Vbuffer_alist, Vrun_hooks,
+        Lisp_Marker, Lisp_Misc_Type, Lisp_Overlay, Lisp_Type, Vbuffer_alist, Vrun_hooks,
     },
     remacs_sys::{
         buffer_permanent_local_flags, Qafter_string, Qbefore_string, Qbuffer_list_update_hook,
         Qbuffer_read_only, Qbufferp, Qfundamental_mode, Qget_file_buffer, Qinhibit_quit,
-        Qinhibit_read_only, Qmakunbound, Qnil, Qoverlayp, Qpermanent_local, Qpermanent_local_hook,
-        Qt, Qunbound, UNKNOWN_MODTIME_NSECS,
+        Qinhibit_read_only, Qkill_buffer_hook, Qkill_buffer_query_functions, Qmakunbound, Qnil,
+        Qoverlayp, Qpermanent_local, Qpermanent_local_hook, Qt, Qunbound, UNKNOWN_MODTIME_NSECS,
+    },
+    remacs_sys::{
+        Fdelq, Fformat, Fmake_marker, Fother_buffer, Frassq, Frun_hook_with_args_until_failure,
+        Fsymbol_value,
     },
     strings::string_equal,
     textprop::get_text_property,
     threads::{c_specpdl_index, ThreadState},
     util::clip_to_bounds,
     vectors::LispVectorlikeRef,
+    windows::LispWindowRef,
 };
 
 pub const BEG: ptrdiff_t = 1;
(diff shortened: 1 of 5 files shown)


rust_src/src/buffers.rs
```diff
@@ -1843,4 +1855,272 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+/// the buffer is not killed.  The hook `kill-buffer-hook' is run before the
+/// buffer is actually killed.  The buffer being killed will be current
+/// while the hook is running.  Functions called by any of these hooks are
+/// supposed to not change the current buffer.
+///
+/// Any processes that have this buffer as the `process-buffer' are killed
+/// with SIGHUP.  This function calls `replace-buffer-in-windows' for
+/// cleaning up all windows currently displaying the buffer to be killed.
+#[lisp_fn(min = "0", intspec = "bKill buffer: ")]
+pub fn kill_buffer(buffer_or_name: LispObject) -> LispObject {
```
agraven: As far as I can tell you only return `Qnil` or `Qt`, so you can make the return type `bool` and return `true` or `false`

agraven:
As this PR hasn't seen activity for quite some time I've taken the liberty of continuing it

simon-frankau:
Oh, wow, thank you! This PR fell off my radar as I've been busy with work. I'd love to get back onto it, but I doubt that's realistic any time soon. If you can push it forwards, that'd be awesome. Sorry for leaving it dangling.

agraven:
That's quite all right! Everyone has obligations from time to time 🙂. And there's a solid amount of groundwork to build on, so finishing it won't be too difficult.

agraven:
This function is somewhat long, I've managed to stare myself blind reading it over and over. I've used the rust versions of functions where their C exports were previously used, and I've added some iterator types to make some iterations idiomatic.

There are a couple very simple C functions that are invoked that could potentially be ported as well, but I refrained from doing so to keep things from blowing too far out of scope.

rust_src/src/buffers.rs
```diff
@@ -1982,4 +2030,268 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+/// while the hook is running.  Functions called by any of these hooks are
+/// supposed to not change the current buffer.
+///
+/// Any processes that have this buffer as the `process-buffer' are killed
+/// with SIGHUP.  This function calls `replace-buffer-in-windows' for
+/// cleaning up all windows currently displaying the buffer to be killed.
+#[lisp_fn(min = "0", intspec = "bKill buffer: ")]
+pub fn kill_buffer(buffer_or_name: Option<LispBufferOrName>) -> bool {
+    let mut b = match buffer_or_name {
+        Some(buf_or_name) => LispBufferRef::from(buf_or_name),
```
shaleh: If the line above is instead:
`let mut b: LispBufferRef =...` that would allow `into` to work here and match the rest of the code better.

rust_src/src/buffers.rs
```diff
@@ -1982,4 +2030,268 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+
+    // First run the query functions; if any query is answered no,
+    // don't kill the buffer.
+    if !run_hook_with_args_until_failure(&mut [Qkill_buffer_query_functions]) {
+        return unbind_to(count, Qnil).into();
+    }
+
+    // Query if the buffer is still modified.
+    if is_interactive() && b.filename().is_not_nil() && b.modified_since_save() {
+        // TODO: This was AUTO_STRING.
```
shaleh: What is the TODO action here?

agraven: This comment was left by simon-frankau. I believe the intention was to communicate that if there was a more suitable way of mirroring that it should be used, but if there is one I'm not aware of it.

rust_src/src/buffers.rs
```diff
@@ -1982,4 +2030,268 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+    if LispObject::from(b) == LispWindowRef::from(unsafe { minibuf_window }).contents {
+        return false;
+    }
+
+    // When we kill an ordinary buffer which shares its buffer text
+    // with indirect buffer(s), we must kill indirect buffer(s) too.
+    // We do it at this stage so nothing terrible happens if they
+    // ask questions or their hooks get errors.
+    if b.base_buffer.is_null() && b.indirections > 0 {
+        for other in LispBufferIter::new() {
```
shaleh: I like the idea of this iterator. But maybe we can come up with a name for the iteration over the global buffer list instead of the Iterator automatically walking that list.

agraven: So something like renaming the constructor to something like `all` or `global_list` or `from_global`?

rust_src/src/buffers.rs
```diff
@@ -1982,4 +2030,268 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+    // Avoid trouble for buffers that are already dead.
+    if !b.is_live() || unsafe { thread_check_current_buffer(b.as_mut()) } {
+        return false;
+    }
+
+    // Run hooks with the buffer to be killed the current buffer.
+    let count = unsafe { SPECPDL_INDEX() };
+
+    unsafe {
+        record_unwind_protect(Some(save_excursion_restore), save_excursion_save());
```
shaleh: Out of scope for the PR, but long term I would love to see this protection wrapped in an object so the code could become something like:
```
let unwind = UnwindProtection::start();
.......
if some_reason:
    return unwind.complete();

// finish the function
return blah; // this will also call unwind.complete automatically.
```

rust_src/src/buffers.rs
```diff
@@ -1982,4 +2030,268 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+    if !b.is_live() {
+        return true;
+    }
+
+    // We have no more questions to ask.  Verify that it is valid
+    // to kill the buffer.  This must be done after the questions
+    // since anything can happen within do_yes_or_no_p.
+
+    // Don't kill the minibuffer now current.
+    if LispObject::from(b) == LispWindowRef::from(unsafe { minibuf_window }).contents {
```
shaleh: This one reads odd. On the left we have a LispObject and on the right we have a LispWindowRef which is not a LispObject.

rust_src/src/buffers.rs
```diff
@@ -1982,4 +2030,268 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
...
+        kill_buffer_xwidgets(b.into());
+    }
+
+    // Killing buffer processes may run sentinels which may have killed
+    // our buffer.
+    if !b.is_live() {
+        return true;
+    }
+
+    // These may run Lisp code and into infinite loops (if someone
```
shaleh: Is this our grammar issue or was it in the original code?

agraven: That was in the original, most comments have been copied verbatim

rust_src/wrapper.h
```diff
@@ -61,3 +61,4 @@
 #endif
 #include "window.h"
 #include "xgselect.h"
+#include "xwidget.h"
```
shaleh: Why is this here?

src/buffer.c
```diff
@@ -5260,7 +5262,7 @@ Functions running this hook are, `get-buffer-create',
   defsubr (&Sbuffer_local_variables);
   defsubr (&Sset_buffer_modified_p);
   defsubr (&Sother_buffer);
-  defsubr (&Skill_buffer);
+//  defsubr (&Skill_buffer);
```
shaleh: This should simply be deleted.

src/buffer.c
```diff
@@ -947,6 +945,9 @@ compact_buffer (struct buffer *buffer)
     }
 }
 
+extern Lisp_Object Fkill_buffer(Lisp_Object buffer_or_name);
```
shaleh: This should not be needed. The compilation process is supposed to generate this.

rust_src/src/buffers.rs
shaleh: Ideally the LispBufferIter constructor could be given a list of buffers and there would be a function or constructor/factory that would give the iterator over the global list.

shaleh:
@agraven thanks for picking this up. This function is a great example of the meandering, old-school C code. When we can refactor all of these into idiomatic code it will be nice nice nice.

rust_src/src/buffers.rs
agraven: Alongside the added safety benefit it would also make it more mentally intuitive

rust_src/src/buffers.rs
agraven: It's a byproduct of having to access a `LispObject` member of the `minibuf_window`, which is also stored as a `LispObject`. I've added a function for accessing `minibuf` that converts it to a `LispWindowRef` in my local tree, which makes this look far more readable.

rust_src/wrapper.h
agraven: If it's removed `kill_buffer_xwidgets` is not added to `remacs_sys`

rust_src/src/buffers.rs
shaleh: Let's drop the word 'and'.

rust_src/src/minibuf.rs
```diff
@@ -23,8 +23,13 @@ use crate::{
     symbols::symbol_value,
     textprop::get_char_property,
     threads::{c_specpdl_index, ThreadState},
+    windows::LispWindowRef,
 };
 
+pub fn minibuffer_window() -> LispWindowRef {
```
shaleh: :+1:

rust_src/src/buffers.rs
agraven: Perhaps a similar thing could be done with input locking as well, i.e. also having a `InputLock::start()` and `input_lock.release()`

rust_src/src/buffers.rs
```diff
@@ -1149,6 +1171,41 @@ impl From<LispBufferOrCurrent> for LispBufferRef {
...
+        Self {
+            current: Some(buffer)
+        }
+    }
+
+    /// Iterate over all buffers, i.e. start at the buffer defined by `all_buffers`.
+    pub fn all_buffers() -> Self {
+        unsafe {
+            Self {
+                current: LispBufferRef::from_ptr(all_buffers as *mut c_void),
```
agraven: Note that the constructor iterating all buffers doesn't rely on the one for starting at a specified point, as I felt safer accounting for the possibility of `all_buffers` being null.

agraven:
Whoops, didn't realize rustfmt wasn't set up correctly

Issue #1425, opened 2019-04-09
bootstrap-emacs segmentation fault on macOS 10.13.6


npassaro:
Executed the build with:
```
$ ./autogen.sh && ./configure --enable-rust-debug && make clean install CARGO_FLAGS="-vv" RUSTFLAGS="-Zunstable-options --cfg MARKER_DEBUG"
```

And the output:
```
Loading /Users/psn1av/projects/remacs/lisp/tooltip.el (source)...
Finding pointers to doc strings...
Finding pointers to doc strings...done
Dumping under the name remacs
--- List of All Regions ---
   address     size prot maxp
--- List of Regions to be Dumped ---
   address     size prot maxp
--- Header Information ---
Magic = 0xfeedfacf
CPUType = 16777223
CPUSubType = -2147483645
FileType = 0x2
NCmds = 32
SizeOfCmds = 4608
Flags = 0x00800085
Highest address of load commands in input file: 0x101a43000
Lowest offset of all sections in __TEXT segment:   0x3010
--- List of Load Commands in Input File ---
# cmd              cmdsize name                address     size
0 LC_SEGMENT_64          72 __PAGEZERO                0 0x100000000
1 LC_SEGMENT_64         952 __TEXT           0x100000000 0xa79000
                           __text           0x100003010 0x889ed8
                           __stubs          0x10088cee8    0xf4e
                           __stub_helper    0x10088de38   0x1992
                           __const          0x10088f7d0  0x88398
                           __cstring        0x100917b70  0x1569a
                           __gcc_except_tab 0x10092d20c  0x18acc
                           __objc_methname  0x100945cd8   0x3921
                           __objc_classname 0x1009495f9    0x113
                           __objc_methtype  0x10094970c   0x1c26
                           __unwind_info    0x10094b334  0x14458
                           __eh_frame       0x10095f790 0x119868
2 LC_SEGMENT_64        1752 __DATA           0x100a79000 0x47f000
                           __nl_symbol_ptr  0x100a79000     0x10
                           __got            0x100a79010    0x280
                           __la_symbol_ptr  0x100a79290   0x1468
                           __const          0x100a7a700  0x5e148
                           __cfstring       0x100ad8848    0x880
                           __objc_classlist 0x100ad90c8     0x78
                           __objc_catlist   0x100ad9140      0x8
                           __objc_protolist 0x100ad9148     0x28
                           __objc_imageinfo 0x100ad9170      0x8
[...]

PR #1508, opened 2019-06-01
ThreadRNG causes segfaults.

shaleh:
likely because C owns the threads. Fixes MacOS compilation.

Closes: #1425


brotzeit:
Out of curiosity. How did you identify the source of the bug ?

shaleh:
lldb, the stack trace showed the failure arising from the call to ThreadRNG.

PR #1510, opened 2019-06-06
Port char-resolve-modifiers

### LINKED ISSUE #1462: Port char-resolve-modifiers
```c
DEFUN ("char-resolve-modifiers", Fchar_resolve_modifiers,
       Schar_resolve_modifiers, 1, 1, 0,
       doc: /* Resolve modifiers in the character CHAR.
The value is a character with modifiers resolved into the character
code.  Unresolved modifiers are kept in the value.
usage: (char-resolve-modifiers CHAR)  */)
  (Lisp_Object character)
{
  EMACS_INT c;

  CHECK_NUMBER (character);
  c = XINT (character);
  return make_number (char_resolve_modifier_mask (c));
}
```

aclonegeek:
Closes #1462

### DIFF
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -6,7 +6,7 @@ use remacs_macros::lisp_fn;
 
 use crate::{
     lisp::LispObject,
-    multibyte::{Codepoint, MAX_CHAR},
+    multibyte::{char_resolve_modifier_mask, Codepoint, MAX_CHAR},
     remacs_sys::EmacsInt,
     threads::ThreadState,
 };
@@ -90,6 +90,15 @@ pub fn multibyte_char_to_unibyte(ch: Codepoint) -> EmacsInt {
     } else {
         ch.to_byte8().map(EmacsInt::from).unwrap_or(-1)
     }
+}
+
+/// Resolve modifiers in the character CHAR.
+/// The value is a character with modifiers resolved into the character
+/// code.  Unresolved modifiers are kept in the value.
+/// usage: (char-resolve-modifiers CHAR)
+#[lisp_fn]
+pub fn char_resolve_modifiers(character: LispObject) -> EmacsInt {
+    char_resolve_modifier_mask(character.into())
 }
 
 include!(concat!(env!("OUT_DIR"), "/character_exports.rs"));
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -96,7 +96,7 @@ pub fn multibyte_char_to_unibyte(ch: Codepoint) -> EmacsInt {
 /// The value is a character with modifiers resolved into the character
 /// code.  Unresolved modifiers are kept in the value.
 /// usage: (char-resolve-modifiers CHAR)
-#[lisp_fn]
+#[lisp_fn(min = "1")]
 pub fn char_resolve_modifiers(character: LispObject) -> EmacsInt {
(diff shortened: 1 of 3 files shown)


rust_src/src/character.rs
```diff
@@ -92,4 +92,13 @@ pub fn multibyte_char_to_unibyte(ch: Codepoint) -> EmacsInt {
     }
 }
 
+/// Resolve modifiers in the character CHAR.
+/// The value is a character with modifiers resolved into the character
+/// code.  Unresolved modifiers are kept in the value.
+/// usage: (char-resolve-modifiers CHAR)
+#[lisp_fn(min = "1")]
```
shaleh: No need to specify a `min` value in this case since min and max are the same.

test/rust_src/src/character-tests.el
```diff
@@ -0,0 +1,12 @@
+;;; character-tests.el ---  -*- lexical-binding: t; -*-
+
+(require 'ert)
+
+(ert-deftest char-resolve-modifiers ()
+    (should (equal (char-resolve-modifiers -1) 4294967295))
```
shaleh: Let's use the hex value to make it clear that this is a "full" integer.
```suggestion
    (should (equal (char-resolve-modifiers -1) #xffffffff))
```

aclonegeek:
Thanks for the review @shaleh. I've made the suggested changes.

rust_src/src/character.rs
```diff
@@ -92,4 +92,13 @@ pub fn multibyte_char_to_unibyte(ch: Codepoint) -> EmacsInt {
     }
 }
 
+/// Resolve modifiers in the character CHAR.
+/// The value is a character with modifiers resolved into the character
+/// code.  Unresolved modifiers are kept in the value.
+/// usage: (char-resolve-modifiers CHAR)
+#[lisp_fn]
+pub fn char_resolve_modifiers(character: LispObject) -> EmacsInt {
```
shaleh:
```suggestion
pub fn char_resolve_modifiers(character: CodePoint) -> EmacsInt {
```

If we did this, we would not need to perform the transformation inside the function. What do you think?

agraven: If both the argument and return type was changed, the method could simply be called directly, i.e. replacing the function body with `character.resolve_modifier_mask()`

aclonegeek: If we change the argument to a `Codepoint`, then we can no longer handle negative numbers being passed in, which emacs currently accepts. Is this desired? If so, I'll do as @agraven suggests.

agraven: If I understood the discussion in gitter correctly, we're currently imitating emacs 26's behaviour of not accepting negative number, and handling negative numbers was introduced in emacs 27. If you want to test this you can temporarily rename your port to something new, and keep the old function alongside it. You can then compare their outputs.

shaleh: @agraven the question is, when we catch up and need to support negative values can we continue to use `Codepoint` or will we need a new type? If so, perhaps we should consider migrating to that type now even if we do not yet support negative values?

aclonegeek: @shaleh Can this PR be merged as is (assuming no other changes need to be made)? We should track the potential `Codepoint` change as a separate issue IMO.

aclonegeek: @shaleh Any updates as to the status of this PR? I know you've been busy!

PR #1371, opened 2019-03-11
Port list-fonts

agraven:
Closes #1143.

PR #1512, opened 2019-06-08
fix list-fonts

### LINKED ISSUE #1444: Address boundary error
Trying to start remacs 31231ecd3b7cca5860bfe0388e425d6926a5790a I get this
```
Fatal error 11: Segmentation fault
Backtrace:
/home/torstein/progs/remacs/src/remacs[0x500e04]
/home/torstein/progs/remacs/src/remacs[0x4e9584]
/home/torstein/progs/remacs/src/remacs[0x4ff6ee]
/home/torstein/progs/remacs/src/remacs[0x4ff8e8]
/home/torstein/progs/remacs/src/remacs[0x4ff959]
/usr/lib64/libpthread.so.0(+0x13070)[0x7f9f1a7f7070]
/home/torstein/progs/remacs/src/remacs[0x5a32e9]
/home/torstein/progs/remacs/src/remacs[0x5a4671]
/home/torstein/progs/remacs/src/remacs[0x5a6be9]
/home/torstein/progs/remacs/src/remacs[0x55cd8d]
/home/torstein/progs/remacs/src/remacs[0x5f4062]
/home/torstein/progs/remacs/src/remacs[0x5f3b65]
/home/torstein/progs/remacs/src/remacs[0x5cd446]
/home/torstein/progs/remacs/src/remacs[0x57a5c0]
/home/torstein/progs/remacs/src/remacs[0x54ec5a]
/home/torstein/progs/remacs/src/remacs[0x5cd3ce]
/home/torstein/progs/remacs/src/remacs[0x57a5c0]
/home/torstein/progs/remacs/src/remacs[0x54ec5a]
/home/torstein/progs/remacs/src/remacs[0x5cd3ce]
/home/torstein/progs/remacs/src/remacs[0x57a5c0]
/home/torstein/progs/remacs/src/remacs[0x54f2c4]
/home/torstein/progs/remacs/src/remacs[0x56ab95]
/home/torstein/progs/remacs/src/remacs[0x56b37c]
/home/torstein/progs/remacs/src/remacs[0x5cd446]
/home/torstein/progs/remacs/src/remacs[0x57a5c0]
/home/torstein/progs/remacs/src/remacs[0x5cd3ce]
/home/torstein/progs/remacs/src/remacs[0x57a5c0]
/home/torstein/progs/remacs/src/remacs[0x5cd3ce]
/home/torstein/progs/remacs/src/remacs[0x57a5c0]
/home/torstein/progs/remacs/src/remacs[0x54f5c6]
/home/torstein/progs/remacs/src/remacs[0x54f02d]
/home/torstein/progs/remacs/src/remacs[0x5cc3b1]
fish: Job 2, “~/progs/remacs/src/remacs  $argv” terminated by signal SIGSEGV (Address boundary error)
````
Fedora 29

### PR DESCRIPTION
**Pull request by ngortheone**
Closes #1444.

Fixes segmentation fault introduced in https://github.com/remacs/remacs/pull/1371
and discussed in https://github.com/remacs/remacs/issues/1444

The crash was caused by applying an unnecessary tag to the pointer during `LispFontSpecRef -> LispObject` conversion

##### Repro: 
evaluate `(list-fonts (font-spec :size 10))`
Remacs returns `nil` (or crashes without this fix)
Emacs returns a list of fonts

### DIFF
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -173,7 +173,13 @@ pub type LispFontSpecRef = ExternalPtr<Lisp_Font_Spec>;
 
 impl From<LispFontSpecRef> for LispObject {
     fn from(f: LispFontSpecRef) -> Self {
-        Self::from_C(f.as_ptr() as EmacsInt)
+        // Use this if use solution by by @cantsin
+        // conflicts with solution by @ngortheone
+        Self::tag_ptr(f, Lisp_Type::Lisp_Vectorlike)
+
+        // Solution by @ngortheone
+        // conflicts with solution by @cantsin
+        // Self::from_C(f.as_ptr() as EmacsInt)
     }
 }
 
@@ -190,7 +196,12 @@ impl From<LispObject> for Option<LispFontSpecRef> {
     fn from(o: LispObject) -> Self {
         o.as_vectorlike().and_then(|v| {
             if v.is_pseudovector(pvec_type::PVEC_FONT) && o.is_font_spec() {
-                Some(unsafe { mem::transmute(o) })
+                // use this if using solution by @ngortheone
+                // Some(unsafe { mem::transmute(o) })
+
+                // Solution by @cantsin
+                // conflicts with solution by @ngortheone
+                Some(unsafe { mem::transmute(v) })
             } else {
                 None
             }
@@ -325,23 +336,25 @@ pub fn list_fonts(
     let mut frame: LispFrameRef = frame.into();
 
     let n = match num {
-        Some(n) if n < 0 => return Qnil,
+        Some(n) if n <= 0 => return Qnil,
         Some(n) => n,
         None => 0,
     } as usize;
 
     let list = unsafe { font_list_entities(frame.as_mut(), font_spec.into()) };
+
     match list.into() {
         Some((car, cdr)) if cdr.is_nil() => match car.as_vector() {
             Some(vec) if vec.len() == 1 => return list!(vec.get(0)),
             _ => (),
         },
-        _ => return Qnil,
+        Some(_) => (),
+        None => return Qnil,
     }
 
     let vec = match prefer {
-        None => unsafe { font_sort_entities(list, prefer.into(), frame.as_mut(), 0) },
-        Some(_) => vconcat_entity_vectors(list.into()),
+        None => vconcat_entity_vectors(list.into()),
+        Some(_) => unsafe { font_sort_entities(list, prefer.into(), frame.as_mut(), 0) },
     }
     .force_vector();
 
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -173,7 +173,7 @@ pub type LispFontSpecRef = ExternalPtr<Lisp_Font_Spec>;
 
 impl From<LispFontSpecRef> for LispObject {
     fn from(f: LispFontSpecRef) -> Self {
-        Self::tag_ptr(f, Lisp_Type::Lisp_Vectorlike)
+        Self::from_C(f.as_ptr() as EmacsInt)
     }
 }
 
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -173,13 +173,7 @@ pub type LispFontSpecRef = ExternalPtr<Lisp_Font_Spec>;
 
 impl From<LispFontSpecRef> for LispObject {
     fn from(f: LispFontSpecRef) -> Self {
-        // Use this if use solution by by @cantsin
-        // conflicts with solution by @ngortheone
         Self::tag_ptr(f, Lisp_Type::Lisp_Vectorlike)
-
-        // Solution by @ngortheone
-        // conflicts with solution by @cantsin
-        // Self::from_C(f.as_ptr() as EmacsInt)
     }
 }
 
@@ -196,11 +190,6 @@ impl From<LispObject> for Option<LispFontSpecRef> {
     fn from(o: LispObject) -> Self {
         o.as_vectorlike().and_then(|v| {
             if v.is_pseudovector(pvec_type::PVEC_FONT) && o.is_font_spec() {
-                // use this if using solution by @ngortheone
-                // Some(unsafe { mem::transmute(o) })
-
-                // Solution by @cantsin
-                // conflicts with solution by @ngortheone
                 Some(unsafe { mem::transmute(v) })
             } else {
                 None
@@ -354,7 +343,7 @@ pub fn list_fonts(
 
     let vec = match prefer {
         None => vconcat_entity_vectors(list.into()),
-        Some(_) => unsafe { font_sort_entities(list, prefer.into(), frame.as_mut(), 0) },
+        Some(f) => unsafe { font_sort_entities(list, f.into(), frame.as_mut(), 0) },
     }
     .force_vector();
 

### CODE REVIEW

**Comment by ngortheone** [comment]
Other outstanding differences that I am going to look into:

>  @shaleh commented 2 days ago
>Perhaps here: https://github.com/remacs/remacs/pull/1371/files#diff-52189e04d12e2d229393c7f81d59d74eL4195

>The C code returns Qnil if num has a value of <= 0 but the ported Rust
https://github.com/remacs/remacs/pull/1371/files#diff-75082f76acd744000705115c0b909719R327
checks only for < 0 and keeps going when the value is 0.

>Also https://github.com/remacs/remacs/pull/1371/files#diff-52189e04d12e2d229393c7f81d59d74eL4208 It looks like the logic was inverted in the Rust port. When prefer has a value font_sort_entities should be called but Remacs calls it when prefer is None.

**Comment by ngortheone** [comment]
cc @jamcha-aa @black-puppydog @tsoernes

**Comment by cantsin** [comment]
Beat me to it! The following diff makes spacemacs load for me:

```diff
diff --git a/rust_src/src/fonts.rs b/rust_src/src/fonts.rs
index 4e31c0d907..cce6bb7d15 100644
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -190,7 +190,7 @@ impl From<LispObject> for Option<LispFontSpecRef> {
     fn from(o: LispObject) -> Self {
         o.as_vectorlike().and_then(|v| {
             if v.is_pseudovector(pvec_type::PVEC_FONT) && o.is_font_spec() {
-                Some(unsafe { mem::transmute(o) })
+                Some(unsafe { mem::transmute(v) })
             } else {
                 None
             }
@@ -340,8 +340,8 @@ pub fn list_fonts(
     }

     let vec = match prefer {
-        None => unsafe { font_sort_entities(list, prefer.into(), frame.as_mut(), 0) },
-        Some(_) => vconcat_entity_vectors(list.into()),
+        None => vconcat_entity_vectors(list.into()),
+        Some(_) => unsafe { font_sort_entities(list, prefer.into(), frame.as_mut(), 0) },
     }
     .force_vector();

```

Not totally sure about the first section (your solution may work better), but the second is definitely a transposed logic error.

ngortheone:
@cantsin 
Second section matches what @shaleh mentioned
>Also https://github.com/remacs/remacs/pull/1371/files#diff-52189e04d12e2d229393c7f81d59d74eL4208 It looks like the logic was inverted in the Rust port. When prefer has a value font_sort_entities should be called but Remacs calls it when prefer is None.

I took your second section and built Remacs with it. It still gives me an error on the front page like:
```
 Warnings:
- Cannot find any of the specified fonts (DejaVu Sans Mono)! Font settings may not be correct.      
```

Do you observe something similar?

ngortheone:
I tested your whole solution, instead of mine, still same error. There are more logic errors out there...

ngortheone:
both your 
```
-                Some(unsafe { mem::transmute(o) })
+                Some(unsafe { mem::transmute(v) })
```
and mine 

```
- Self::tag_ptr(f, Lisp_Type::Lisp_Vectorlike)
+ Self::from_C(f.as_ptr() as EmacsInt)
```
achieve the same thing in the end. Mine seems to be shorter (performs less work) but yours may be more correct / in the spirit of Remacs.

This is my first PR in Remacs codebase, so I honestly don't know which way is better. I'd love to hear some opinions..

cantsin:
@ngortheone Hmm, where does that warning pop up for you? My `*Messages*` has the following: 

```
Loading /home/james/remacs-build/spacemacs-master/init.el (source)...
Loading /home/james/remacs-build/spacemacs-master/core/core-load-paths.el (source)...done
Loading /home/james/.spacemacs...done
Setting the font...
PNG warning: Interlace handling should be turned on when using png_read_image
Open the quickhelp.
Loading /home/james/remacs-build/spacemacs-master/layers/+completion/ivy/packages.el (source)...done
Loading /home/james/remacs-build/spacemacs-master/layers/+completion/ivy/layers.el (source)...done
Loading /home/james/remacs-build/spacemacs-master/layers/+lang/emacs-lisp/packages.el (source)...done
Loading /home/james/remacs-build/spacemacs-master/layers/+distributions/spacemacs/layers.el (source)...
Loading /home/james/remacs-build/spacemacs-master/layers/+distributions/spacemacs-base/packages.el (source)...done

<... snip, a lot of elisp files being loaded ... >
```

(But I do believe that there are more logic errors out there!)

ngortheone:
It is right on the front of *spacemacs* page
DeepinScreenshot_select-area_20190607232833

cantsin:
Ah, nope. I don't get that error.



I am using a stock unmodified spacemacs, though (I do not use spacemacs myself) and the font under point appears to be `Source Code Pro`. (`what-cursor-position` via `C-u C-x =`)

ngortheone:
Does `(list-fonts (font-spec :size 10))` give you something other than `nil`?

cantsin:
`(list-fonts (font-spec :size 10))` does give me `nil` yes! So yes, something else is still awry here.

ngortheone:
I bet if you set this in your `.spacemacs` config you will get the same error
```
   dotspacemacs-default-font '("DejaVu Sans Mono"
                               :size 13
                               :weight normal
                               :width normal)
```
Source Code Pro appears to be default fallback font for spacemacs and it does not throw the error, If I set it my error goes away

ngortheone:
I fixed it!

cantsin:
Great, I confirm your fix works!

rust_src/src/fonts.rs
```diff
@@ -325,23 +336,29 @@ pub fn list_fonts(
...
-            _ => (),
-        },
-        _ => return Qnil,
+        None => return Qnil,
+        Some((car, cdr)) => {
+            let car_size = match car.as_vector() {
+                Some(v) => v.len(),
+                None => 0,
+            };
+            if (cdr.is_nil() && car_size == 1) {
```
cantsin: If you haven't already, please check the `rustfmt` output. These parentheses are unnecessary:

```
warning: unnecessary parentheses around `if` condition
   --> src/fonts.rs:341:16
    |
341 |             if (cdr.is_nil() && car_size == 1) {
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove these parentheses
    |
    = note: #[warn(unused_parens)] on by default
```

ngortheone:
This had a problem. 
```rust
    Some((car, cdr)) if cdr.is_nil() => match car.as_vector() {
            Some(vec) if vec.len() == 1 => return list!(vec.get(0)),
            _ => (),
        },
        _ => return Qnil,
```
It performed an early exit with Qnil. When original implementation just kept going. I did not realize this in the beginning, something in this code just didn't felt right.
I guess it can be fixed by
```rust
        None => return Qnil,
```

rust_src/src/fonts.rs
ngortheone: ack

rust_src/src/fonts.rs
```diff
@@ -190,7 +196,12 @@ impl From<LispObject> for Option<LispFontSpecRef> {
     fn from(o: LispObject) -> Self {
         o.as_vectorlike().and_then(|v| {
             if v.is_pseudovector(pvec_type::PVEC_FONT) && o.is_font_spec() {
```
shaleh: `transmute(o)` tries to turn the LispObject directly into a pointer.
`transmute(v)` takes the Lisp_Vectorlike pointer and turns it into a LispFontSpecRef pointer.

Since we already made the V using it makes sense to me. I would lean towards @cantsin's solution.

rust_src/src/fonts.rs
```diff
@@ -325,23 +336,25 @@ pub fn list_fonts(
...
-        _ => return Qnil,
+        Some(_) => (),
+        None => return Qnil,
     }
 
     let vec = match prefer {
-        None => unsafe { font_sort_entities(list, prefer.into(), frame.as_mut(), 0) },
-        Some(_) => vconcat_entity_vectors(list.into()),
+        None => vconcat_entity_vectors(list.into()),
+        Some(_) => unsafe { font_sort_entities(list, prefer.into(), frame.as_mut(), 0) },
```
shaleh: Would this version work?

`Some(font_spec)  => unsafe { font_sort_entities(list, font_spec.into(), frame.as_mut(), 0) },`

We already know that `prefer` is a `Some` the `into` call is going to perform these checks again.

ngortheone: `font_spec` conflicts with already existing var. But otherwise seems to be working

ngortheone:
Removed [WIP]
I think this is ready to go!

black-puppydog:
This seems to work for me :) :tada:

ngortheone:
I have noticed one small thing. Icons in Treemacs are small in Remacs, while normal size in Emacs. 
Not sure if it is connected with this PR..

ngortheone:
Does not seem like it is related to this PR.
Icons are small even before #1371
Maybe a regression that was introduced earlier?

ngortheone:
also `treemacs-resize-icons` does not work,
https://github.com/Alexander-Miller/treemacs/issues/165

While it works on emacs 26.2 and 27

ngortheone:
I will open a separate issue for it. Does not look related to this one, but definitely Remacs-specific

rust_src/src/fonts.rs
agraven: I prefer @cantsin's solution as well. And I can confirm that using `o` instead of `v` was a typo.

PR #1516, opened 2019-06-11
Port frame-char-height and frame-char-width

### LINKED ISSUE #1495: Port frame-char-height and frame-char-width
```c
DEFUN ("frame-char-height", Fframe_char_height, Sframe_char_height,
       0, 1, 0,
       doc: /* Height in pixels of a line in the font in frame FRAME.
If FRAME is omitted or nil, the selected frame is used.
For a terminal frame, the value is always 1.  */)
  (Lisp_Object frame)
{
#ifdef HAVE_WINDOW_SYSTEM
  struct frame *f = decode_any_frame (frame);

  if (FRAME_WINDOW_P (f))
    return make_number (FRAME_LINE_HEIGHT (f));
  else
#endif
    return make_number (1);
}

DEFUN ("frame-char-width", Fframe_char_width, Sframe_char_width,
       0, 1, 0,
       doc: /* Width in pixels of characters in the font in frame FRAME.
If FRAME is omitted or nil, the selected frame is used.
On a graphical screen, the width is the standard width of the default font.
For a terminal screen, the value is always 1.  */)
  (Lisp_Object frame)
{
#ifdef HAVE_WINDOW_SYSTEM
  struct frame *f = decode_any_frame (frame);

  if (FRAME_WINDOW_P (f))
    return make_number (FRAME_COLUMN_WIDTH (f));
  else
#endif
    return make_number (1);
}
```

aclonegeek:
Closes #1495


aclonegeek:
WIP because I'd like to add tests, but I'm running into an issue that I haven't yet had much time to look into. It seems like the problem is that the corresponding `frame.c` rust file is called `frames.rs` instead of `frame.rs`. I'll look into it more tomorrow.

This is the error that I get in case anyone is curious (after creating `remacs/test/rust_src/src/frame-tests.el`):
```
$ make frame-tests
make[1]: Entering directory '/home/randy/dev/remacs/test'
make[1]: *** No rule to make target '../rust_src/src/frame.rs', needed by 'rust_src/src/frame-tests.log'.  Stop.
make[1]: Leaving directory '/home/randy/dev/remacs/test'
make: *** [Makefile:233: rust_src/src/frame-tests] Error 2
```

aclonegeek:
The filename `frames.rs` was indeed the culprit as suspected, so it's now called `frame.rs` to match its C counterpart. The tests I added aren't super useful, but there's not much that can be done without a graphic display.

shaleh:
You could add code to test/manual/ directory to test when in GUI mode.

rust_src/src/frame.rs
```diff
@@ -691,4 +691,47 @@ pub extern "C" fn get_frame_param(frame: LispFrameRef, prop: LispObject) -> Lisp
...
+        } else {
+            1
+        }
+    }
+    #[cfg(not(feature = "window-system"))]
+    {
+        1
+    }
+}
+
```
agraven: I'd recommend compressing this a bit if by changing the body of the if to `return frame_ref.line_height` and dropped the else and the second cfg, i.e.:
```rust
        if frame_ref.is_gui_window() {
            return frame_ref.line_height;
        }
    }
    1
}
```

`frame_char_width` can be shortened in a similar way.

aclonegeek: @agraven Thanks for the suggestion - that's a good idea.I'll update the PR when I get a chance.

The module is being renamed because it doesn't match the c file counterpart like the rest of the files do (and should). shaleh's suggestion was specifically regarding testing this change when in GUI mode.

agraven: Ah, apologies, I misunderstood.

aclonegeek: @agraven To take your suggestion further, shouldn't it be possible to completely drop all the `cfg`s? It seems to me that `is_gui_window()` would simply return false when there is no window system, as it would be in terminal mode (or initial, whatever that is).

agraven: That's true, I forgot that function wasn't behind a `cfg`. And I suspect initial is before either the terminal or the gui interface is initialized, but I'm not entirely sure.

shaleh: @aclonegeek that sounds reasonable, you might need to touch some other code.

shaleh: The main reason behind the C #ifdef logic is it is faster and yields a smaller binary. There is no reason to interrogate the frame if there is no GUI support.

aclonegeek: @shaleh Good point. I think we should keep `cfg` then, but only one of them is necessary (in this context). I'll apply @agraven's suggestion.

PR #1517, opened 2019-06-11
port current-minor-mode-maps

### LINKED ISSUE #1470: Port current-minor-mode-maps
```c
DEFUN ("current-minor-mode-maps", Fcurrent_minor_mode_maps, Scurrent_minor_mode_maps, 0, 0, 0,
       doc: /* Return a list of keymaps for the minor modes of the current buffer.  */)
  (void)
{
  Lisp_Object *maps;
  int nmaps = current_minor_maps (0, &maps);

  return Flist (nmaps, maps);
}
```

sehnsucht13:
Should hopefully close #1470

### DIFF
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -781,7 +781,7 @@ pub fn current_minor_mode_maps() -> LispObject {
     let mut maps: *mut LispObject = ptr::null_mut();
     unsafe {
         let num_of_maps = current_minor_maps(ptr::null_mut(), &mut maps);
-        Flist(num_of_maps, maps)
+        list!(num_of_maps, maps)
     }
 }
 include!(concat!(env!("OUT_DIR"), "/keymap_exports.rs"));
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -29,7 +29,7 @@ use crate::{
     },
     remacs_sys::{char_bits, current_global_map as _current_global_map, globals, EmacsInt},
     remacs_sys::{
-        Fcommand_remapping, Fcurrent_active_maps, Fevent_convert_list, Flist, Fmake_char_table,
+        Fcommand_remapping, Fcurrent_active_maps, Fevent_convert_list, Fmake_char_table,
         Fset_char_table_range, Fterpri,
     },
     remacs_sys::{
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -17,7 +17,7 @@ use crate::{
     keyboard,
     keyboard::lucid_event_type_list_p,
     lisp::LispObject,
-    lists::{nth, setcdr},
+    lists::{nth, setcdr, Flist},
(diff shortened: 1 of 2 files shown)


rust_src/src/keymap.rs
```diff
@@ -773,4 +773,15 @@ pub fn apropos_internal(regexp: LispStringRef, predicate: LispObject) -> LispObj
...
 }
 
+/// Return a list of keymaps for the minor modes of the current buffer.
+#[lisp_fn]
+pub fn current_minor_mode_maps() -> LispObject {
+    // maps is initialized to null to silence the borrow checker. If this is
+    // not done then it will not compile since maps is not initialized
+    let mut maps: *mut LispObject = ptr::null_mut();
+    unsafe {
+        let num_of_maps = current_minor_maps(ptr::null_mut(), &mut maps);
```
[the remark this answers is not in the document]
sehnsucht13: Hi, sorry for not getting back to this in so long. I have been quite busy.

The reason why `maps` is `&mut maps` is because when building remacs without it, the compiler warns me that the expected type is `*mut *mut lisp::LispObject`. In other words, it requires a double pointer to be passed to the function. Simply passing `maps` will result in a compilation error when cargo builds the project. Is there any other way of passing a pointer to a pointer in rust or have I misunderstood something critical. Sorry if this is a dumb question, I am still quite new to rust and the emacs codebase as a whole. Thanks in advance!

rust_src/src/keymap.rs
```diff
@@ -773,4 +773,15 @@ pub fn apropos_internal(regexp: LispStringRef, predicate: LispObject) -> LispObj
...
+        Flist(num_of_maps, maps)
```
shaleh: We usually try not to use 'F' functions in Rust. But I am torn over whether `list([num_of_maps, maps])` is preferable.

shaleh:
@sehnsucht13 are there tests in the tree for this? Could we add some?

rust_src/src/keymap.rs
agraven: I'd personally prefer it.

sehnsucht13:
Hi, I am sorry but I am quite busy at the moment with school and I don't have time to complete this pull request(come up with and write tests). I will come back to it in 3 or 4 days. This is just a quick message to let everybody know that this has not been abandoned.

rust_src/src/keymap.rs
shaleh: Actually, I think this might be better written `list!(num_of_maps, maps)`.

agraven:
`list!()` seems to be passed a raw pointer, which it I don't think it supports. See [line 1613 in the log](https://travis-ci.org/remacs/remacs/jobs/635146345#L1613). Also, the `Flist`import is no longer used.

agraven:
As `Flist` takes a size and pointer argument I think it fits the return values of `current_minor_maps` better. Constructing a slice from raw parts seems unnecessary here.

shaleh:
Add a comment

agraven:
Do you mean add a comment that explains this to the code?

PR #1520, opened 2019-06-12
Yet more Clippy

nickdrozd:
There are a few changes packed into this one.

- Cut unnecessary negations (mostly shuffling if/else blocks)
- Use isize::min_value() instead of std::isize::MIN, etc
    - https://github.com/rust-lang/rust-clippy/issues/2380
- Rename some shadowing variables
- Cut some unnecessary references
- Use <struct>::default instead of Default::default
- Update inclusive range syntax
- Cut open import
- Self
- dyn trait


nickdrozd:
I've been collecting these for a while, like a bird building a nest.

rust_src/src/editfns.rs
```diff
@@ -458,15 +458,13 @@ pub fn byte_to_string(byte: EmacsInt) -> LispObject {
 /// Return the first character in STRING.
 #[lisp_fn]
 pub fn string_to_char(string: LispStringRef) -> EmacsInt {
```
shaleh: Good stuff

rust_src/src/filelock.rs
```diff
@@ -25,7 +25,6 @@ use crate::{
     remacs_sys::{Qnil, Qstringp, Qt},
     threads::ThreadState,
 };
```
shaleh: Double win. No more asterisk and using Self.

rust_src/src/fns.rs
```diff
@@ -591,10 +591,10 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
         let mut new = unsafe { make_uninit_bool_vector(nbits).force_bool_vector() };
         new.as_mut_slice().copy_from_slice(boolvec.as_slice());
         new.into()
```
shaleh: Nice.

rust_src/src/fonts.rs
```diff
@@ -117,7 +117,7 @@ impl LispFontObjectRef {
         unsafe { font_add_log(c_str.as_ptr(), self.into(), result.into()) }
     }
```
shaleh: `_frame` was used here because when `window-system` is not enabled the `frame` argument is unused and clippy rightly complains.

rust_src/src/frames.rs
```diff
@@ -633,12 +633,12 @@ pub fn frame_focus(frame: LispFrameLiveOrSelected) -> LispObject {
 ///
 /// If there is no window system support, this function does nothing.
 #[lisp_fn(min = "1", name = "x-focus-frame", c_name = "x_focus_frame")]
```
shaleh: Same explanation as above.

rust_src/src/multibyte.rs
```diff
@@ -165,16 +166,17 @@ impl Codepoint {
     }
 
     // Equivalent to UNIBYTE_TO_CHAR
-    pub fn unibyte_to_char(self) -> Codepoint {
-        match self.is_ascii() {
-            true => self,
-            false => Codepoint::from_raw(self.0 as u8),
+    pub fn unibyte_to_char(self) -> Self {
+        if self.is_ascii() {
```
shaleh: What is wrong with using `match` on the bool like before? Is it poor Rust style or was this a personal choice?

rust_src/src/windows.rs
```diff
@@ -95,7 +95,7 @@ impl LispWindowRef {
 
         if mode_line_height >= 0 {
             mode_line_height
```
shaleh: Why did you change from `!=` to `< 0`? The C code used:
```
#define CURRENT_MODE_LINE_HEIGHT(W)					\
  (W->mode_line_height >= 0						\
   ? W->mode_line_height						\
   : (W->mode_line_height						\
      = (MATRIX_MODE_LINE_HEIGHT (W->current_matrix)			\
	 ? MATRIX_MODE_LINE_HEIGHT (W->current_matrix)			\
	 : estimate_mode_line_height					\
	     (XFRAME (W->frame), CURRENT_MODE_LINE_FACE_ID (W)))))
```

Which means `MATRIX_MODE_LINE_HEIGHT` is only checked for being not 0. -1 and 1 are both true values as far as `if` is concerned. Only `0` is false.

nickdrozd: I don't know, I must have misread something.

rust_src/src/multibyte.rs
nickdrozd: Clippy flags it. Not important of course, but matching on bools does seem like a strange style.

rust_src/src/fonts.rs
nickdrozd: Clippy was complaining about that when I ran it locally, but I guess it just doesn't handle this situation very well.

rust_src/src/windows.rs
shaleh: I read it twice myself. The variable checked flips from the if to the else if which adds to the confusion.

nickdrozd:
@shaleh I updated `rust-toolchain`, uncovering a few more flagged items.

rust_src/src/windows.rs
nickdrozd: This is fixed btw (that is, the change was removed entirely).

rust_src/src/marker.rs
```diff
@@ -561,17 +561,20 @@ fn set_marker_internal_else(
...
             .map_or(false, |m| m.buffer() == Some(buf))
     {
-        bytepos = buf.charpos_to_bytepos(charpos);
+        buf.charpos_to_bytepos(charpos)
     } else {
-        let beg = buf.buffer_beg_byte(restricted);
-        let end = buf.buffer_end_byte(restricted);
-        bytepos = clip_to_bounds(beg, bytepos as EmacsInt, end);
-    }
+        clip_to_bounds(
```
shaleh: A benefit to the temporary variables is they make it easier to inspect values in a debugger.

PR #1522, opened 2019-06-12
revert remacs version to 26.2

ngortheone:
@shaleh 

Closes:
https://github.com/remacs/remacs/issues/1521

Partially fixes
https://github.com/remacs/remacs/issues/1513

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,7 +23,7 @@ dnl  along with GNU Emacs.  If not, see <https://www.gnu.org/licenses/>.
 
 AC_PREREQ(2.65)
 dnl Note this is parsed by (at least) make-dist and lisp/cedet/ede/emacs.el.
-AC_INIT(GNU Emacs, 27.0.50, bug-gnu-emacs@gnu.org, , https://www.gnu.org/software/emacs/)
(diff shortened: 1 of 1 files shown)


shaleh:
I would like to see us revert the changelog and other bits back too. Which will also help with the merges from emacs.

ngortheone:
Surprisingly this tiny change break a lot of things..

ngortheone:
What "other bits" have to be reverted in order to achieve parity with 26.2?

shaleh:
I *think* that is all. Changelog, some docs. I suspect everything else is all ok.

An interesting thing to do would be to look at the difference between https://github.com/emacs-mirror/emacs/tree/emacs-26 and remacs tree.

ngortheone:
I have a suspicion that someone else broke tests... Previous commit fails `make check` too

ngortheone:
I did take a look at the difference between remacs and https://github.com/emacs-mirror/emacs/tree/emacs-26 
And the difference seems huge (maybe to my untrained eye). It looks like we missed a whole lot of patches. See my rant on gitter..

Issue #1496, opened 2019-05-13
Port compose-string-internal


brotzeit:
```c
DEFUN ("compose-string-internal", Fcompose_string_internal,
       Scompose_string_internal, 3, 5, 0,
       doc: /* Internal use only.

Compose text between indices START and END of STRING, where
START and END are treated as in `substring'.  Optional 4th
and 5th arguments are COMPONENTS and MODIFICATION-FUNC
for the composition.  See `compose-string' for more details.  */)
  (Lisp_Object string, Lisp_Object start, Lisp_Object end,
   Lisp_Object components, Lisp_Object modification_func)
{
  ptrdiff_t from, to;

  CHECK_STRING (string);
  validate_subarray (string, start, end, SCHARS (string), &from, &to);
  compose_text (from, to, components, modification_func, string);
  return string;
}
```

baBounch:
I'd like to port this.

shaleh:
See our Gitter channel if you have questions.

A6GibKm:
This should be closed.

shaleh:
Merged.

PR #1525, opened 2019-07-03
Compose string internal

baBounch:
#1496

### DIFF
--- a/rust_src/src/composite.rs
+++ b/rust_src/src/composite.rs
@@ -1,0 +1,38 @@
+//! Composite sequence support.
+
+use libc::ptrdiff_t;
+
+use remacs_macros::lisp_fn;
+
+use crate::{
+    lisp::LispObject,
+    remacs_sys::{compose_text, validate_subarray, CHECK_STRING, SCHARS},
+};
+
+/// Internal use only.
+///
+/// Compose text between indices START and END of STRING, where
+/// START and END are treated as in `substring'.  Optional 4th
+/// and 5th arguments are COMPONENTS and MODIFICATION-FUNC
+/// for the composition.  See `compose-string' for more details.
+#[lisp_fn(min = "3")]
+pub fn compose_string_internal(
+    string: LispObject,
+    start: LispObject,
+    end: LispObject,
+    components: LispObject,
+    modification_func: LispObject,
+) -> LispObject {
+    let mut from: ptrdiff_t = 0;
+    let mut to: ptrdiff_t = 0;
+
+    unsafe {
+        CHECK_STRING(string);
+        validate_subarray(string, start, end, SCHARS(string), &mut from, &mut to);
+        compose_text(from, to, components, modification_func, string);
+    }
+
+    return string;
+}
+
+include!(concat!(env!("OUT_DIR"), "/composite_exports.rs"));
--- a/rust_src/src/composite.rs
+++ b/rust_src/src/composite.rs
@@ -32,7 +32,7 @@ pub fn compose_string_internal(
         compose_text(from, to, components, modification_func, string);
     }
 
-    return string;
+    string
 }
 
 include!(concat!(env!("OUT_DIR"), "/composite_exports.rs"));
--- a/rust_src/src/composite.rs
+++ b/rust_src/src/composite.rs
@@ -1,14 +1,12 @@
 //! Composite sequence support.
 
-use std::convert::TryInto;
-
 use remacs_macros::lisp_fn;
 
 use crate::{
     fns::validate_subarray_rust,
     lisp::LispObject,
     multibyte::LispStringRef,
-    remacs_sys::{EmacsInt, Fcons, Fput_text_property, Qcomposition},
+    remacs_sys::{EmacsInt, Fput_text_property, Qcomposition},
 };
 
 /// Internal use only.
@@ -19,24 +17,16 @@ use crate::{
 /// for the composition.  See `compose-string' for more details.
 #[lisp_fn(min = "3")]
 pub fn compose_string_internal(
-    string: LispObject,
-    start: LispObject,
-    end: LispObject,
+    string: LispStringRef,
+    start: EmacsInt,
+    end: EmacsInt,
     components: LispObject,
     modification_func: LispObject,
 ) -> LispObject {
-    let string = LispStringRef::from(string);
-
     let (from, to) =
-        validate_subarray_rust(LispObject::from(string), start, end, string.len_chars());
+        validate_subarray_rust(string.into(), start.into(), end.into(), string.len_chars());
 
-    compose_text_rust(
-        from,
-        to,
-        components,
-        modification_func,
-        LispObject::from(string),
-    );
+    compose_text_rust(from, to, components, modification_func, string);
 
     LispObject::from(string)
 }
@@ -51,20 +41,20 @@ pub fn compose_text_rust(
     end: EmacsInt,
     components: LispObject,
     modification_function: LispObject,
-    string: LispObject,
+    string: LispStringRef,
 ) {
-    unsafe {
-        let prop = Fcons(
-            Fcons(LispObject::from(end - start), components),
-            modification_function,
-        );
+    let prop = (
+        (LispObject::from(end - start), components),
+        modification_function,
+    );
 
+    unsafe {
         Fput_text_property(
-            LispObject::from(start),
-            LispObject::from(end),
+            start.into(),
+            end.into(),
             Qcomposition,
-            prop,
-            string,
+            prop.into(),
+            string.into(),
         );
     }
 }
@@ -78,11 +68,11 @@ pub extern "C" fn compose_text(
     string: LispObject,
 ) {
     compose_text_rust(
-        start.try_into().unwrap(),
-        end.try_into().unwrap(),
+        start as EmacsInt,
+        end as EmacsInt,
         components,
         modification_function,
-        string,
+        string.into(),
     );
 }
 
--- a/rust_src/src/composite.rs
+++ b/rust_src/src/composite.rs
@@ -18,17 +18,16 @@ use crate::{
 #[lisp_fn(min = "3")]
 pub fn compose_string_internal(
     string: LispStringRef,
-    start: EmacsInt,
-    end: EmacsInt,
+    start: Option<EmacsInt>,
+    end: Option<EmacsInt>,
     components: LispObject,
-    modification_func: LispObject,
-) -> LispObject {
-    let (from, to) =
-        validate_subarray_rust(string.into(), start.into(), end.into(), string.len_chars());
+    modification_function: LispObject,
+) -> LispStringRef {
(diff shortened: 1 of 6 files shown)


test/rust_src/src/composite-tests.el
```diff
@@ -0,0 +1,16 @@
+;;; composite-tests.el --- Tests for composite.rs
+
+;;; Code:
+
+(require 'ert)
+
```
shaleh: Tests. Huzzah!

rust_src/src/composite.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use libc::ptrdiff_t;
+
+use remacs_macros::lisp_fn;
+
+use crate::{
+    lisp::LispObject,
+    remacs_sys::{compose_text, validate_subarray, CHECK_STRING, SCHARS},
+};
+
```
shaleh: Normally, the Remacs style is to use LispString and EmacsInt instead of LispObject. However, since all of the calls are still in C there is little value to using the new names at the moment.

Both `validate_subarray` and `compose_text` appear to be straight forward to port. Would you be up to giving them a go? Then this function could get a proper Rust port.

For `validate_subarray`, make a Rust function with the following signature:

`pub fn validate_subarray_rust(array: LispObject, from: EmacsInt, to: EmacsInt, size: isize) -> (EmacsInt, EmacsInt)`
then make one with the C signature:
```
#[no_mangle]
pub extern "C" fn validate_subarray(array: LispObject, from: LispObject, to: LispObject, size: isize, new_from: &mut LispObject, new_to: &mut LispObject) // notice there is no return this time
```

The idea being we make the Rust functions behave in an idiomatically Rust fashion and wrap them to the keep the original C working.

baBounch: sure, and thank you for the advice.

baBounch: The original function would turn nil into 0, is it okay if I make from and to LispObjects, and then convert them to EmacsInt inside the function to check for nil?

rust_src/src/composite.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Internal use only.
+///
+/// Compose text between indices START and END of STRING, where
+/// START and END are treated as in `substring'.  Optional 4th
+/// and 5th arguments are COMPONENTS and MODIFICATION-FUNC
+/// for the composition.  See `compose-string' for more details.
+#[lisp_fn(min = "3")]
+pub fn compose_string_internal(
+    string: LispObject,
```
shaleh: Because you are now using `LispObject::from` you can put those types right here into the function signature. The `lisp_fn` machinery will run `foo.into()` on each parameter.

```
pub fn compose_string_internal(
    string: LispStringRef,
    start: EmacsInt,
    end: EmacsInt,
    components: LispObject,
    modification_func: LispObject
) -> LispObject
```

rust_src/src/composite.rs
```diff
@@ -0,0 +1,89 @@
...
+    start: LispObject,
+    end: LispObject,
+    components: LispObject,
+    modification_func: LispObject,
+) -> LispObject {
+    let string = LispStringRef::from(string);
```
shaleh: Now need for this line.

rust_src/src/composite.rs
```diff
@@ -0,0 +1,89 @@
...
+/// has COMPONENTS and MODIFICATION-FUNC.
+///
+/// If STRING is non-nil, then operate on characters contained between
+/// indices START and END in STRING.
+pub fn compose_text_rust(
+    start: EmacsInt,
+    end: EmacsInt,
+    components: LispObject,
+    modification_function: LispObject,
+    string: LispObject,
```
shaleh: use a `LispStringRef`

rust_src/src/composite.rs
```diff
@@ -0,0 +1,89 @@
...
+) {
+    unsafe {
```
shaleh: Instead of `Fcons` you can use Rust tuples here (I am fairly certain...).
`let prop = ((LispObject::from(end - start), components), modification_function);` and then call `prop.into()` below.

rust_src/src/composite.rs
```diff
@@ -0,0 +1,89 @@
...
+        let prop = Fcons(
+            Fcons(LispObject::from(end - start), components),
+            modification_function,
+        );
+
+        Fput_text_property(
+            LispObject::from(start),
```
shaleh: We usually use `into`
```suggestion
            start.into(),
```

rust_src/src/composite.rs
```diff
@@ -0,0 +1,89 @@
...
+            LispObject::from(end),
```
shaleh:
```suggestion
            end.into(),
```

rust_src/src/composite.rs
```diff
@@ -0,0 +1,89 @@
...
+#[no_mangle]
+pub extern "C" fn compose_text(
+    start: isize,
+    end: isize,
+    components: LispObject,
+    modification_function: LispObject,
+    string: LispObject,
+) {
+    compose_text_rust(
+        start.try_into().unwrap(),
```
shaleh: Why the `try_into`? `isize` should cast to `EmacsInt` without fail.

rust_src/src/fns.rs
```diff
@@ -598,4 +599,55 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
...
+/// Count negative values backwards from the end.
+/// Set *IFROM and *ITO to the two indexes used.
+pub fn validate_subarray_rust(
+    array: LispObject,
+    from: LispObject,
+    to: LispObject,
+    size: isize,
+) -> (EmacsInt, EmacsInt) {
+    // convert from, to and size to Emacs Int
+    let int_size = EmacsInt::from(LispObject::from(size));
```
shaleh: No need to turn `size` into a LispObject before casting it to `EmacsInt`.

rust_src/src/fns.rs
```diff
@@ -598,4 +599,55 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
...
+    let mut int_from = from.map_or(0, EmacsInt::from);
```
shaleh: perfect.

rust_src/src/fns.rs
```diff
@@ -598,4 +599,55 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
...
+        int_from += int_size;
+    }
+    if int_to < 0 {
+        int_to += int_size;
+    }
+
+    // check if from is less than to, or if from or to are out of range.
+    if !(0 <= int_from && int_from <= int_to && int_to <= int_size) {
+        unsafe {
+            args_out_of_range_3(array, LispObject::from(int_from), LispObject::from(int_to));
```
shaleh: Whereever the C code needed these fixed argument number macros remacs usually has a Rust macro. You can find them in rust_src/src/eval_macros.rs. In this case you want `args_out_of_range!()`.

rust_src/src/fns.rs
```diff
@@ -598,4 +599,55 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
...
+    from: LispObject,
+    to: LispObject,
+    size: isize,
+    new_from: *mut isize,
+    new_to: *mut isize,
+) {
+    let (f, t) = validate_subarray_rust(array, from, to, size);
+
+    unsafe {
+        *new_from = f.try_into().unwrap();
```
shaleh: No need for `try_into` the cast will never fail.

rust_src/src/fns.rs
```diff
@@ -598,4 +599,55 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
...
+    // make negative numbers reverse array
+    if int_from < 0 {
+        int_from += int_size;
+    }
+    if int_to < 0 {
+        int_to += int_size;
+    }
+
+    // check if from is less than to, or if from or to are out of range.
+    if !(0 <= int_from && int_from <= int_to && int_to <= int_size) {
```
shaleh: This is a great place for DeMorgen's theorem.
```suggestion
    if (0 > int_from || int_from > int_to || int_to > int_size) {
```

rust_src/src/composite.rs
```diff
@@ -0,0 +1,79 @@
...
+#[lisp_fn(min = "3")]
+pub fn compose_string_internal(
+    string: LispStringRef,
+    start: EmacsInt,
+    end: EmacsInt,
+    components: LispObject,
+    modification_func: LispObject,
+) -> LispObject {
+    let (from, to) =
+        validate_subarray_rust(string.into(), start.into(), end.into(), string.len_chars());
```
shaleh: Is there a reason that `validate_subarray_rust` can't take `start` and `end` as `EmacsInt`?

rust_src/src/composite.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    compose_text_rust(from, to, components, modification_func, string);
+
+    LispObject::from(string)
```
shaleh: You can simply return the string and make the return type `LispStringRef`. The `lisp_fn` macro handles the types coming and and going out.

rust_src/src/composite.rs
```diff
@@ -0,0 +1,79 @@
...
+///
+/// If STRING is non-nil, then operate on characters contained between
+/// indices START and END in STRING.
+pub fn compose_text_rust(
+    start: EmacsInt,
+    end: EmacsInt,
+    components: LispObject,
+    modification_function: LispObject,
+    string: LispStringRef,
+) {
```
shaleh: That reads really nicely!

rust_src/src/fns.rs
```diff
@@ -598,4 +598,53 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
...
+
+#[no_mangle]
+pub extern "C" fn validate_subarray(
+    array: LispObject,
+    from: LispObject,
+    to: LispObject,
+    size: isize,
+    new_from: *mut isize,
+    new_to: *mut isize,
+) {
```
shaleh: Perfect.

rust_src/src/fns.rs
```diff
@@ -598,4 +598,53 @@ pub fn copy_sequence(mut arg: LispObject) -> LispObject {
...
+    from: Option<EmacsInt>,
+    to: Option<EmacsInt>,
+    size: isize,
+) -> (EmacsInt, EmacsInt) {
+    // change from, to and size to EmacsInt
+    let int_size = size as EmacsInt;
+    let mut int_from = from.unwrap_or(0);
+    let mut int_to = to.unwrap_or(int_size);
+
+    // make negative numbers reverse array
```
shaleh: This comment is not clear to me.

```
// Negative indexes count from the end of the array.
// Adding size to them should turn them into
// equivalent positive indexes.
```

baBounch: That is better, thank you. I just pushed the changes.

agraven:
Since the last request from @shaleh has been addressed I'm merging this.

PR #1526, opened 2019-07-11
Port make-frame-visible.

rbartlensky:
This PR hopefully ports `make-frame-visible` to Rust. I am not including `make-frame-invisible` into this PR as it I need to port many functions for that, making the PR too big to review. I am not sure how to create unit tests for this function.


shaleh:
```
error: variable does not need to be mutable
1863   --> src/frames.rs:713:9
1864    |
1865713 |     let mut frame_ref = window_frame_live_or_selected(frame);
1866    |         ----^^^^^^^^^
1867    |         |
1868    |         help: remove this `mut`
```

rust_src/src/frames.rs
```diff
@@ -683,6 +685,43 @@ pub fn frame_face_alist(frame: LispFrameLiveOrSelected) -> LispObject {
     frame_ref.face_alist
 }
 
+/// Update the display_time slot of the buffers shown in WINDOW and all its descendants.
+fn make_frame_visible_1(mut window: LispWindowRef) {
+    if window.is_null() {
+        return;
+    }
+    loop {
```
shaleh: Instead of a loop, let's implement an Iterator for LispWindowRef that knows how to walk the `next` window attribute. Look at LispSymbolRef's iterator as an example.

rbartlensky: Yup, you are right. I made this change in my latest commit.

rbartlensky:
> ```
> error: variable does not need to be mutable
> 1863   --> src/frames.rs:713:9
> 1864    |
> 1865713 |     let mut frame_ref = window_frame_live_or_selected(frame);
> 1866    |         ----^^^^^^^^^
> 1867    |         |
> 1868    |         help: remove this `mut`
> ```

I also included a fix for this in my last commit, however, I am not sure it is the best way to handle it...

rust_src/src/frames.rs
```diff
@@ -683,6 +685,39 @@ pub fn frame_face_alist(frame: LispFrameLiveOrSelected) -> LispObject {
...
+    #[cfg(feature = "window-system")]
+    {
+        let mut frame_ref: LispFrameRef = frame.into();
+        if frame_ref.is_gui_window() {
+            unsafe {
+                x_make_frame_visible(frame_ref.as_mut());
+            }
+        }
+    }
+    let frame_ref: LispFrameRef = frame.into();
```
shaleh: Maybe if we did this at the top? Then there would not be a reason to repeat it.

rbartlensky: The thing is that the `cfg` block needs a mutable pointer in the call to `x_make_window_visible`. However, when that `cfg` block is not compiled, the `mut` of `frame_ref` is unnecessary. I have a new commit to solve this issue without calling `into` multiple times.

rbartlensky:
@shaleh  I guess this is ready for another review. However, I am not sure why `make check` doesn't work...

yellowsquid:
Resurrected this here (#1573).

agraven:
Closing since #1573 replaces this PR

PR #1529, opened 2019-07-24
WIP keymap functions

sehnsucht13:
I have done some rough work on porting some keymap functions but I would love some feedback on the work so far. Right now, none of them have been tested but I am wondering if I am on the right track in terms of it being a faithful port(good rust practices) and any other general tips. I am quite new to rust so I am still trying to get comfortable with it.

There are some comments in the code which indicate what I would like to have looked at. One of the most troublesome functions is the `_copy_keymap_item` which is a mess of `unwrap` and `as_cons()`. I am hoping that there is a cleaner way or at the very least know that this is the most optimal way currently. Thanks in advance.

### DIFF
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -2,8 +2,9 @@
 
 use std;
 use std::ptr;
+use std::convert::TryInto;
 
-use libc::c_void;
+use libc::{c_void, ptrdiff_t};
 
 use remacs_macros::lisp_fn;
 
@@ -21,16 +22,18 @@ use crate::{
     lists::{LispCons, LispConsCircularChecks, LispConsEndChecks},
     multibyte::LispStringRef,
     obarray::intern,
+    marker::marker_position,
     remacs_sys::{
         access_keymap, apropos_accum, apropos_accumulate, apropos_predicate, call2,
         copy_keymap_item, describe_vector, list2, make_save_funcptr_ptr_obj, map_char_table,
         map_keymap_call, map_keymap_char_table_item, map_keymap_function_t, map_keymap_item,
-        map_obarray, maybe_quit, menu_item_eval_property, safe_call1, specbind,
+        map_obarray, maybe_quit, menu_item_eval_property, safe_call1,
+    specbind, 
     },
     remacs_sys::{char_bits, current_global_map as _current_global_map, globals, EmacsInt},
     remacs_sys::{
         Fcommand_remapping, Fcons, Fcurrent_active_maps, Fevent_convert_list, Flength,
-        Fmake_char_table, Fset_char_table_range, Fterpri,
+        Fmake_char_table, Fset_char_table_range, Fterpri, 
     },
     remacs_sys::{
         QCfilter, Qautoload, Qkeymap, Qkeymap_canonicalize, Qkeymapp, Qmenu_item, Qmouse_click,
@@ -162,12 +165,15 @@ pub extern "C" fn _get_keyelt(object: LispObject, autoload: bool) -> LispObject 
     }
 }
 
+// TODO: Fix this mess with all the unwrap
+// TODO: Avoid using FCONS?
 #[no_mangle]
 pub extern "C" fn _copy_keymap_item(elt: LispObject) -> LispObject {
     if elt.is_not_cons() {
         elt;
     }
 
+    // Turn elt into a cons cell in order to use its car
     let mut elt = elt.as_cons().unwrap();
     let mut res = elt;
     let mut tem = elt;
@@ -229,6 +235,28 @@ pub extern "C" fn _preferred_sequence_p(seq: LispObject) -> i64 {
     }
     // Make compiler shut up for now
     0 as i64
+}
+
+/* Return the offset of POSITION, a click position, in the style of
+   the respective argument of Fkey_binding.  */
+pub extern "C" fn _click_position(position: LispObject) -> ptrdiff_t {
+    let mut curr_pos: EmacsInt;
+
+    if position.is_integer(){
+       curr_pos = position.into();
+    }
+    else if position.is_marker(){
+        curr_pos = marker_position(position).try_into().unwrap();
+    }
+    else{
+        curr_pos = ThreadState::current_buffer_unchecked().get_pt().try_into().unwrap();
+    };
+
+    if ThreadState::current_buffer_unchecked().beg() <=
+        curr_pos.try_into().unwrap() && curr_pos as ptrdiff_t <= ThreadState::current_buffer_unchecked().get_zv(){
+        return curr_pos as ptrdiff_t;
+    }
+    args_out_of_range!(ThreadState::current_buffer_unchecked(), position);
 }
 
 // Which keymaps are reverse-stored in the cache.
--- a/rust_src/src/keymap.rs
+++ b/rust_src/src/keymap.rs
@@ -25,12 +25,12 @@ use crate::{
         access_keymap, apropos_accum, apropos_accumulate, apropos_predicate, copy_keymap_item,
         describe_vector, make_save_funcptr_ptr_obj, map_char_table, map_keymap_call,
         map_keymap_char_table_item, map_keymap_function_t, map_keymap_item, map_obarray,
-        maybe_quit, specbind, call2, safe_call1, list2, menu_item_eval_property,
+        maybe_quit, specbind, call2, safe_call1, list2, menu_item_eval_property, 
     },
     remacs_sys::{char_bits, current_global_map as _current_global_map, globals, EmacsInt,},
     remacs_sys::{
         Fcommand_remapping, Fcurrent_active_maps, Fevent_convert_list, Fmake_char_table,
-        Fset_char_table_range, Fterpri, Fcons, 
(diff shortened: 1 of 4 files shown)


rust_src/src/keymap.rs
```diff
@@ -19,22 +20,24 @@ use crate::{
     lisp::LispObject,
     lists::{nth, setcdr},
     lists::{LispCons, LispConsCircularChecks, LispConsEndChecks},
+    marker::marker_position,
     multibyte::LispStringRef,
     obarray::intern,
     remacs_sys::{
```
shaleh: Things like `call2` and `list2` are almost always the wrong thing in the Rust code. There is usually a function or a macro to mimic the C code's needs.

rust_src/src/keymap.rs
```diff
@@ -19,22 +20,24 @@ use crate::{
...
-        maybe_quit, specbind,
+        access_keymap, apropos_accum, apropos_accumulate, apropos_predicate, call2,
+        copy_keymap_item, describe_vector, list2, make_save_funcptr_ptr_obj, map_char_table,
+        map_keymap_call, map_keymap_char_table_item, map_keymap_function_t, map_keymap_item,
+        map_obarray, maybe_quit, menu_item_eval_property, safe_call1, specbind,
     },
-    remacs_sys::{char_bits, current_global_map as _current_global_map, globals, EmacsInt},
+    remacs_sys::{char_bits, current_global_map as _current_global_map,
+    globals, EmacsInt, PURE_P},
     remacs_sys::{
```
shaleh: Whenever you look to import an `F` function check if it has been re-implemented in Rust and call the Rust native function instead. `Flength` for instance. `Fcons` can usually be replicated by making a tuple and calling `into`.

rust_src/src/keymap.rs
```diff
@@ -62,6 +65,193 @@ pub extern "C" fn set_where_is_cache(val: LispObject) {
...
+Also if OBJECT has a menu string as the first element,
+remove that.  Also remove a menu help string as second element.
+
+If AUTOLOAD, load autoloadable keymaps
+that are referred to with indirection.
+
+This can GC because menu_item_eval_property calls Feval.  */
+#[no_mangle]
+pub extern "C" fn _get_keyelt(object: LispObject, autoload: bool) -> LispObject {
+    loop {
```
shaleh:
```suggestion
    while let Some(cons) = object.as_cons() {
```

Does that read better? Then if the "if not cons" and "as_cons" below are not needed.

shaleh: Actually, reading below, it appears we could go a little further and instead do:

`while let Some(car, cdr) = object.into() {`

However, I would suggest we find a better name for `car` and `cdr` here.

rust_src/src/keymap.rs
```diff
@@ -62,6 +65,193 @@ pub extern "C" fn set_where_is_cache(val: LispObject) {
...
+
+        // Convert the lisp object to a cons-cell in order to manipulate it
+        let consObject = object.as_cons().unwrap();
+
+        // Local mutable copy of object since we cannot modify the object
+        // variable passed to the function
+        let mut localObject = object;
+
+        if consObject.car().eq(Qmenu_item) {
+            if consObject.cdr().is_cons() {
```
shaleh:
```suggestion
            if let Some(blah, blah) = consObject.cdr().into() {
```

Or use a match

```
match consObject.cdr().into() {
    Some(blah, blah) => ......
    None => break localObject;
}
```

rust_src/src/keymap.rs
```diff
@@ -62,6 +65,193 @@ pub extern "C" fn set_where_is_cache(val: LispObject) {
...
+        }
+    }
+}
+
+// TODO: Fix this mess with all the unwrap but how?
+// TODO: Avoid using FCONS?
+#[no_mangle]
+pub extern "C" fn _copy_keymap_item(elt: LispObject) -> LispObject {
+    if elt.is_not_cons() {
+        elt;
```
shaleh: ? Are you trying to return `elt` here? If you the return keyword is needed.

rust_src/src/keymap.rs
```diff
@@ -62,6 +65,193 @@ pub extern "C" fn set_where_is_cache(val: LispObject) {
...
+        else if consObject.cdr().is_string() {
+            localObject = consObject.into();
+        } else {
+            break localObject;
+        }
+    }
+}
+
+// TODO: Fix this mess with all the unwrap but how?
+// TODO: Avoid using FCONS?
```
shaleh: Read my guidance earlier and suggestions above. Should make this code more manageable.

rust_src/src/keymap.rs
```diff
@@ -62,6 +65,193 @@ pub extern "C" fn set_where_is_cache(val: LispObject) {
...
+    let curr_pos: EmacsInt;
+
+    if position.is_integer() {
+        curr_pos = position.into();
+    } else if position.is_marker() {
+        curr_pos = marker_position(position).try_into().unwrap();
+    } else {
+        curr_pos = ThreadState::current_buffer_unchecked()
+            .get_pt()
+            .try_into()
```
shaleh: try_into without actually handling failures is something we would like to avoid. This can lead to Remacs crashing unexpectedly.

sehnsucht13:
Thanks a lot for the review! I will work on making the changes!

shaleh:
@sehnsucht13 I avoided reviewing the actual port and commented on your mechanics. That seemed more helpful at this stage.

For C functions, if no C code calls them feel empowered to refactor them as needed.

sehnsucht13:
Thanks again. I appreciate you taking the time to do this!

sehnsucht13:
In the function `_copy_keymap_item`, there is a large and unsightly(in my opinion) usage of `obj.as_cons().unwrap()`. Should we add a function that is called `as_cons_unsafe()` which does this under the hood in order to make the code more readable or would that be a dangerous choice?

shaleh:
There is a definition for `From<LIspObject> for (LispObject, LispObject)`. Every Cons cell is really a tuple internally of (car, cdr). A Lisp list looks like `(car, (car, (car, (car, cdr))))`.

sehnsucht13:
I implemented the changes mentioned. 
- I have gotten rid of `try_into` and it turns out that a simple cast to EmacsInt will do the trick. I have no idea what I was thinking at the time since we are just comparing a ptrdiff_t with a usize.
- I have rewritten the two functions `get_keyelt` and `copy_keymap_item` to use the `while let` and `if let` for (hopefully) easier readability than before. The namings(head1,2...) are things which still need to be changed to something a bit more meaningful.

There is only one problem and that has to do with the `copy_keymap_item` function. The original C function uses 3 temporary variables called `tem`, `elt` and `res`. The `res` variable contains the value returned from the function but `elt` and `res` seem to be used to simply iterate over the cons cells. I have eliminated the use of `tem` since it was used to destructure each lisp object into a cons cell(I do that with if let). On the other hand, I am not too sure about elt itself and what its doing. Is it just a helper for `tem` or am I missing something here? Thanks in advance

sehnsucht13:
Another question as well. What do we do about testing the rust code that has been ported? Some of it is accessible from emacs itself so `ert` can be used but what about functions which are not? Can we write unit tests for those?

shaleh:
At the moment we cannot test Rust functions if they use ANY C types. Which basically means we cannot unit test.

I have this PR https://github.com/remacs/remacs/pull/1511 which needs someone's time. It makes it possible to run tests from C that are mostly implemented as Rust tests.

sehnsucht13:
Build fails. I will take a look at it tommorow and fix it up

sehnsucht13:
Just a bit of an update on this. I have not forgotten about it but I have school finals for summer semester coming up so this will have to wait for a bit.

rust_src/src/keymap.rs
```diff
@@ -62,6 +63,177 @@ pub extern "C" fn set_where_is_cache(val: LispObject) {
...
+                menu-item's definition to get the real definition to
+                use.  */
+                for val in object_iter {
+                    if val.eq(QCfilter) && autoload {
+                        // Get the next object in the list
+                        let mut filter = val.as_cons().unwrap().cdr();
+
+                        unsafe {
+                            filter = list!(filter, list!(Qquote, object));
+                            object = menu_item_eval_property(filter);
```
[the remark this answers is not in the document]
shaleh: I think because `object` is redefined within the while look on line 110. The value returned on 155 is the original value.

rust_src/src/keymap.rs
```diff
@@ -62,6 +63,177 @@ pub extern "C" fn set_where_is_cache(val: LispObject) {
...
+            return object;
+        }
+    }
+
+    object
+}
+
+#[no_mangle]
+pub extern "C" fn _copy_keymap_item(elt: LispObject) -> LispObject {
+    if let Some((head, tail)) = elt.into() {
```
shaleh: I would rewrite this for clarity.

a)

```
match elt.into() {
    None => { return elt; }
    Some((head, tail)) => {
        ...
    }
}
```
b)
```
let cons: Option<(LispObject, LispObject)> = elt.into();
if cons.is_none() {
    return elt;
}
let (head, tail) = cons;
.....
```

PR #1530, opened 2019-07-31
RFC: Add macro documentation

sehnsucht13:
I have been thinking. One of the things which sometimes slows me down is finding the equivalents of some C macros such as `XINT` and others. This might also be a hurdle for anybody who joins the project. Should we begin documenting what the equivalent C macro is for some of the functions which are implemented for `LispObject`, `LispBufferRef` and so on?

We can perhaps document them like as follows(the first line is a bit redundant so perhaps those should be left out):
```rust
    /// Check if object is a cons cell
    /// Equivalent to CONSP macro
    pub fn is_cons(self) -> bool {
        self.get_type() == Lisp_Type::Lisp_Cons
    }
```

If this is done then calling the command `cargo doc --open --document-private-items` and then going to the documentation for `LispObject`(as an example) and searching for `CONSP` will bring the user directly to the`is_cons` function. This will quite possibly speed up the onboarding of new contributors and also make it easier to refresh one's mind when porting over C code.

I have a couple of commits here which will show exactly what type of documentation I am talking about. Its a bit tedious to do but if approved, I have no problem going over the structures and documenting this.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -114,7 +114,9 @@ pub const BUF_BYTES_MAX: ptrdiff_t = buf_bytes_max();
 pub type LispBufferRef = ExternalPtr<Lisp_Buffer>;
 pub type LispOverlayRef = ExternalPtr<Lisp_Overlay>;
 
+/// Reference to a lisp buffer
 impl LispBufferRef {
+    /// Create a new buffer with name NAME
     pub fn create_new(name: LispStringRef) -> Self {
         if name.is_empty() {
             error!("Empty string for buffer name is not allowed");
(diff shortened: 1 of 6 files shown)


rust_src/src/buffers.rs
```diff
@@ -199,14 +201,16 @@ impl LispBufferRef {
 
         buffer.into()
     }
+    
     pub fn is_read_only(self) -> bool {
         self.read_only_.into()
     }
-
+    /// Equivalent to BEG macro
```
shaleh: newline here too right?

We have resisted putting direct "this matches blah in C" in remacs because we will have to yank these comments eventually. As we have less and less C they are less and less helpful.

rust_src/src/buffers.rs
```diff
@@ -890,11 +896,14 @@ impl LispBufferRef {
...
 
 impl LispObject {
+    /// Check if object is a buffer
+    /// Equivalent to BUFFERP
     pub fn is_buffer(self) -> bool {
         self.as_vectorlike()
             .map_or(false, |v| v.is_pseudovector(pvec_type::PVEC_BUFFER))
     }
-
+    /// Convert object to a buffer reference
```
shaleh: newline

sehnsucht13:
I see that makes more sense. I will close this pull request for now. I will start up a new wiki page and detail the c <-> rust equivalences there. I will work on adding documentation as I go along with porting then. Thanks!

PR #1534, opened 2019-08-06
Port string_char_to_byte and string_byte_to_char

agraven:
This is a more or less direct port of the C code. I experimented with using if expressions directly in the assignments of `best_above` and friends, but the two conditionals created enough visual noise to reduce readability.

I moved the statics for caching over to rust, but I wasn't sure if I should use `lazy_static` or not since they don't need to be initialized at runtime/first use.

I also added some convenience functions for using `char_head_p`, `multibyte_length_by_head` and `multibyte_char_at` on strings, and I'm not entirely sure how the `multibyte_char_at` one should look with regards to (absence of) error handling.

The PR is based on `the lisp_string_width` PR as it's the only place in the rust code the index conversion functions are used, and additionally I want to remove all invocations of C code from `lisp_string_width` to enable wring rust tests.

On the subject of tests, none are in this PR yet; I intend to add some ASAP.


rust_src/src/multibyte.rs
```diff
@@ -348,13 +357,214 @@ impl LispStringRef {
...
+        let mut best_above_byte = self.len_bytes() as usize;
+
+        // Return if contents are unibyte.
+        if best_above == best_above_byte {
+            return char_index;
+        }
+
+        // Use cache operating on the cached string?
+        unsafe {
+            if LispObject::from(self) == string_index_cache {
```
shaleh: Why store the LispObject instead of a `Option<LispStringRef>`?

rust_src/src/multibyte.rs
```diff
@@ -348,13 +357,214 @@ impl LispStringRef {
...
+
+            byte
+        } else {
+            // Target is backwards.
+            let mut byte = best_above_byte;
+
+            while best_above > char_index {
+                byte -= 1;
+                while !self.is_head_at(byte) {
+                    byte -= 1;
```
shaleh: Why is it backwards one but forwards `head_len(byte)`?

rust_src/src/multibyte.rs
```diff
@@ -348,13 +357,214 @@ impl LispStringRef {
...
+
+        // Update cache
+        unsafe {
+            string_index_cache = self.into();
+            bytepos_cache = byte_index;
+            charpos_cache = char_index;
+        }
+        byte_index
+    }
+
```
shaleh: It certainly feels like there is a way to right a generic function here that use the char or byte type to do the right thing.

rust_src/src/multibyte.rs
```diff
@@ -403,6 +613,26 @@ impl LispStringRef {
...
+    /// Checks whether `index` is the beginnign of a character sequence (also
+    /// counts single byte characters).
+    pub fn is_head_at(self, index: usize) -> bool {
+        char_head_p(self.as_slice()[index])
+    }
+
+    /// Gets the codepoint starting at the specified index, and its length.
+    ///
+    /// # Panics
```
shaleh: Is there a way to have a version that would use an Option or a Result?

rust_src/src/multibyte.rs
```diff
@@ -1088,3 +1339,43 @@ pub unsafe extern "C" fn str_to_unibyte(
...
+}
+
+#[no_mangle]
+pub extern "C" fn lisp_string_width(
+    string: LispObject,
+    precision: isize,
+    nchars: *mut isize,
+    nbytes: *mut isize,
+) -> isize {
+    let precision = if precision < 0 {
```
shaleh: A potential alternative is `let precision = precision.try_into<usize>().ok()`

rust_src/src/multibyte.rs
agraven: The difference is that there's two while loops. When moving backwards, you can't use the head length to preemptively check how big a jump you need to make, so you move backwards byte by byte until you hit a head.

rust_src/src/multibyte.rs
agraven: Definitely, you just return None if the index isn't a head or if the slice isn't long enough to hold the expected number of bytes.

rust_src/src/multibyte.rs
```diff
@@ -88,6 +93,10 @@ pub const MAX_MULTIBYTE_LENGTH: usize = 5;
 
 const BYTE8_OFFSET: u32 = 0x3F_FF00;
 
+static mut string_index_cache: LispObject = Qnil;
```
shaleh: This is marked staticpro in the C source. Is that still relevant here?

agraven: I actually don't know, but I don't imagine it is because as far as I can tell you only use the cached LispObject if you make a second call on the same string, and if you do that it has to be alive anyway, but maybe I'm misunderstanding what staticpro does.

shaleh: My understanding it is means "protected static" and prevents the items from being garbage collected.

agraven: In that case I think my reasoning for not needing it applies. In case it doesn't, where should I invoke `staticpro`? I've grepped around the codebase and can't find an obvious place to put it.

shaleh: What you are missing is a Rust version of `syms_of_fns`. Looks at the Rust syms_of_cmds. For a while we had syms_of_cmds_rust (or was it rust_syms_of_cmds...).

agraven:
As the lisp_string_width PR has been merged, I plan on finishing this PR sometime soon

rust_src/src/multibyte.rs
agraven: I only realized this now, but from what I can tell storing a `LispObject` is necessary for making `staticpro` work, as the pointer supplied to it has to point to a static variable, and storing an `ExternalPtr` but supplying a `*mut LispObject` to `staticpro` would require converting to a tagged pointer/LispObject at the time of invocation. This would mean that the variable only lasts the length of the funtion call. This is a by-product of a pointer to a LispObject essentially being two layers of pointers.

rust_src/src/multibyte.rs
agraven: It was quite a while ago so it's understandable if you don't remember, but I don't quite understand what you mean/meant, could you elaborate?

shaleh: I was referring to the if/else block that leads into this. It looks like something using the known type could help. But maybe not.

agraven: I don't think so

rust_src/src/multibyte.rs
agraven: I've added a `rust_syms_of_multibyte` with the staticpro invocation and am calling it from `syms_of_fns`. Since the original function was in fns but the new one in multibyte this seemed like the best available option. Let me know if you want me to change names/locations around.

rust_src/src/multibyte.rs
agraven: A checked version has been added as `char_and_len_safe`.

src/fns.c
```diff
@@ -2742,6 +2607,8 @@ that disables the use of a file dialog, regardless of the value of
 this variable.  */);
   use_file_dialog = 1;
 
+  rust_syms_of_multibyte();
```
shaleh: This is a little unorthodox, but you don't really have anywhere else to put it.

agraven:
Unless anyone has any additional feedback, I think this PR is ready for merge

PR #1535, opened 2019-08-10
Port set-window-scroll-bars

### LINKED ISSUE #1459: Port set-window-scroll-bars
```c
DEFUN ("set-window-scroll-bars", Fset_window_scroll_bars,
       Sset_window_scroll_bars, 1, 5, 0,
       doc: /* Set width and type of scroll bars of window WINDOW.
WINDOW must be a live window and defaults to the selected one.

Second parameter WIDTH specifies the pixel width for the vertical scroll
bar.  If WIDTH is nil, use the scroll bar width of WINDOW's frame.
Third parameter VERTICAL-TYPE specifies the type of the vertical scroll
bar: left, right, nil or t where nil means to not display a vertical
scroll bar on WINDOW and t means to use WINDOW frame's vertical scroll
bar type.

Fourth parameter HEIGHT specifies the pixel height for the horizontal
scroll bar.  If HEIGHT is nil, use the scroll bar height of WINDOW's
frame.  Fifth parameter HORIZONTAL-TYPE specifies the type of the
horizontal scroll bar: bottom, nil, or t where nil means to not display
a horizontal scroll bar on WINDOW and t means to use WINDOW frame's
horizontal scroll bar type.

Return t if scroll bars were actually changed and nil otherwise.  */)
  (Lisp_Object window, Lisp_Object width, Lisp_Object vertical_type,
   Lisp_Object height, Lisp_Object horizontal_type)
{
  struct window *w
    = set_window_scroll_bars (decode_live_window (window),
			      width, vertical_type, height, horizontal_type);
  return w ? (apply_window_adjustment (w), Qt) : Qnil;
}
```

clample:
Resolves #1459

The implementation is very similar to the [porting of set-window-fringes](https://github.com/remacs/remacs/pull/1367)


test/rust_src/src/windows-tests.el
```diff
@@ -189,3 +189,21 @@
 ;; frames inside tests (see https://github.com/remacs/remacs/issues/1429).
 (ert-deftest window-pixel-height-before-size-change ()
   (window-pixel-height-before-size-change))
+
+;; TODO: find a way to ignore COLUMNS, LINES. Otherwise it fails when run by hand.
```
clample: It looks like the columns and lines returned by `window-scroll-bars` will vary. It is calculated based on width and height of the scroll bar. For example, the test fails if I open emacs and run the test locally. 

I'm not sure the best way to make this test lest fragile.

rust_src/src/windows.rs
```diff
@@ -2017,4 +2017,49 @@ pub fn window_pixel_height_before_size_change(window: LispWindowValidOrSelected)
...
+    let updated_window = unsafe {
+        set_window_scroll_bars(
+            window.as_mut(),
+            width,
+            vertical_type,
+            height,
+            horizontal_type,
+        )
+    };
+
```
shaleh: Perhaps flip the following logic?

```
if updated_window.is_null() {
    false
} else {
    unsafe { apply_window_adjustment(updated_window) };
    true
}
```

test/rust_src/src/windows-tests.el
shaleh: hmmmm

rust_src/src/windows.rs
clample: Good point, this is more readable to me than `!updated_window.is_null()`.

I updated the PR

shaleh:
This all looks good. Now to figure out how to test it properly....

clample:
The failure is from `regex-tests-PTESTS`. I think that it might be unrelated, but I'm not sure what the root cause is.

test/rust_src/src/windows-tests.el
clample: So I believe I kind of found the reason why the column value varied. When Remacs is run without a window system (in the terminal), there is no scroll bar and columns is always 0. 

I haven't pinned down the exact line where this happens, but it seems to be how it works.

To work around it - I check [window-system](https://www.gnu.org/software/emacs/manual/html_node/elisp/Window-Systems.html) in the test.

shaleh:
@clample this one needs a rebase.

rust_src/src/windows.rs
```diff
@@ -2017,4 +2017,53 @@ pub fn window_pixel_height_before_size_change(window: LispWindowValidOrSelected)
...
+)]
+pub fn set_window_scroll_bars_lisp(
+    window: LispWindowLiveOrSelected,
+    width: LispObject,
+    vertical_type: LispObject,
+    height: LispObject,
+    horizontal_type: LispObject,
+) -> bool {
+    let mut window: LispWindowRef = window.into();
+    let updated_window = unsafe {
```
denis631: @clample In the original implementation, the window is accessed via `decode_live_window` which checks for selected window being nil first, if yes -> use it instead of the passed window, otherwise check the passed window for being live (otherwise throwing a "wrong type argument exception") and then use it.

Do you think this may be the case here?
reference: window.c line 264

src/window.c
```diff
@@ -5773,36 +5770,6 @@ set_window_scroll_bars (struct window *w, Lisp_Object width,
...
-horizontal scroll bar: bottom, nil, or t where nil means to not display
-a horizontal scroll bar on WINDOW and t means to use WINDOW frame's
-horizontal scroll bar type.
-
-Return t if scroll bars were actually changed and nil otherwise.  */)
-  (Lisp_Object window, Lisp_Object width, Lisp_Object vertical_type,
-   Lisp_Object height, Lisp_Object horizontal_type)
-{
-  struct window *w
-    = set_window_scroll_bars (decode_live_window (window),
```
denis631: was referencing this `decode_live_window` call

rust_src/src/windows.rs
shaleh: @denis631 `LispWindowLiveOrSelected` is meant to encapsulate the logic of `decode_live_window`.

denis631: Ohhh so the `#lisp_fn` macro does all the necessary transformations? It this case mapping `LispObject` to `LispWindowLiveOrSelected`? I see. Nice. Sorry for false negative

shaleh: @denis631 correct. The `lisp_fn` macro usually works out to be:
```
pub fn new_defun(obj1: LispObject, obj2: LispObject, .......) -> LispObject {
    let result = original_defun(obj1.into(), obj2.into(), ....);
    result.into()
}
```

Which is super simple. All of the error handling, checking, and the like occurs in the `From` implementations. The `lisp_fn` macro gets more complicated when the lisp function is of variable arguments. These collect the args into a Vector.

Our convention is that where a Lisp function takes optional parameters we use the `Option` type. In the C code this is represented with a Nil value. Essentially C's LispObject is really an `Option<LispValue>`.

PR #1537, opened 2019-08-13
Port window-scroll-bar-width and window-scroll-bar-height

clample:
Hello!

I've added a port for `window-scroll-bar-width` and `window-scroll-bar-height`. Most of the logic was already ported, which made it more straightforward.


clample:
Thank you for the review @shaleh!

PR #1538, opened 2019-08-14
Port window-scroll-bars

clample:


rust_src/src/windows.rs
```diff
@@ -2017,4 +2059,34 @@ pub fn window_pixel_height_before_size_change(window: LispWindowValidOrSelected)
...
+        Some(window.scroll_bar_width)
+    } else {
+        None
+    };
+    let height = if window.scroll_bar_height >= 0 {
+        Some(window.scroll_bar_height)
+    } else {
+        None
+    };
+    list!(
```
shaleh: Why did the original use `Fcons` instead of `list6()`?

clample: I believe that they only have up to `list5`: https://github.com/remacs/remacs/blob/fb2e5d73bbd221568e9fbeaa2e6d169a422455de/src/lisp.h#L3706-L3711

With `list!`, I think the rust version really shines

shaleh: Agreed. `list!` is so much nicer. I suspected that was the answer about list5, but I did not have the chance to check.

shaleh:
What is left to complete for this PR?

clample:
@shaleh sorry for abandoning this one for a while. If I remember correctly, I wanted to merge https://github.com/remacs/remacs/pull/1535 first as it had some tests using window scroll bars that I thought would be nice to have additional coverage.

That PR had some issues I was unable to figure out, though.

Issue #1418, opened 2019-04-06
Port last-nonminibuffer-frame


brotzeit:
```c
DEFUN ("last-nonminibuffer-frame", Flast_nonminibuf_frame,
       Slast_nonminibuf_frame, 0, 0, 0,
       doc: /* Return last non-minibuffer frame selected. */)
  (void)
{
  Lisp_Object frame = Qnil;

  if (last_nonminibuf_frame)
    XSETFRAME (frame, last_nonminibuf_frame);

  return frame;
}
```

denis631:
@brotzeit, @shaleh can we close the issue and update the board?

brotzeit:
Sure, thanks! I invited you to the project.

PR #1550, opened 2019-11-06
[Frame] migrate static variable last_non_minibuffer_frame from C to Rust

denis631:
Original function to port, which is defined in the ticket/issue #1418 `last-nonminibuffer-frame` is accessing a global static variable (which afaik is not accessible from Rust, cause it's sort of fileprivate)

This variable is used in other functions as well, so in order to port this function, I decided to firstly, port the static variable itself.
All accesses to it in C are done via function calling (get/set) to Rust

### DIFF
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -729,29 +729,39 @@ pub fn make_frame_invisible(frame: LispFrameLiveOrSelected, force: bool) {
 
 #[lisp_fn]
 pub fn last_nonminibuffer_frame() -> LispObject {
-    let frame = get_last_nonminibuffer_frame();
-
-    if frame.is_null() {
-        Qnil
-    } else {
-        frame.into()
+    unsafe {
+        last_non_minibuffer_frame.into()
     }
 }
 
-/// A frame which is not just a mini-buffer, or NULL if there are no such
-/// frames. This is usually the most recent such frame that was selected.
-static mut last_non_minibuffer_frame: LispFrameRef = LispFrameRef::new(std::ptr::null_mut());
+/// A frame to a guaranteed existing frame which is not just a mini-buffer,
+/// or None if there are no such frames.
+/// This is usually the most recent such frame that was selected.
+static mut last_non_minibuffer_frame: Option<LispFrameRef> = None;
 
 /// Return current last non-minibuffer frame reference
+/// or a pointer to null in case there are no such frames
 #[no_mangle]
 pub extern "C" fn get_last_nonminibuffer_frame() -> LispFrameRef {
-    unsafe { last_non_minibuffer_frame }
+    unsafe {
+        match last_non_minibuffer_frame {
+            Some(frame) => frame,
+            None => LispFrameRef::new(std::ptr::null_mut()),
+        }
+    }
 }
 
 /// Set current last non-minibuffer frame reference
 #[no_mangle]
 pub extern "C" fn set_last_nonminibuffer_frame(frame: LispFrameRef) {
-    unsafe { last_non_minibuffer_frame = frame }
+    unsafe {
+        // frame reference may be null, since it is called from C
+        if frame.is_null() {
+            last_non_minibuffer_frame = None
+        } else {
+            last_non_minibuffer_frame = Some(frame)
+        }
+    }
 }
 
 /// Return the value of frame parameter PROP in frame FRAME.
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -729,9 +729,7 @@ pub fn make_frame_invisible(frame: LispFrameLiveOrSelected, force: bool) {
 
 #[lisp_fn]
 pub fn last_nonminibuffer_frame() -> LispObject {
-    unsafe {
-        last_non_minibuffer_frame.into()
-    }
+    unsafe { last_non_minibuffer_frame.into() }
 }
 
 /// A frame to a guaranteed existing frame which is not just a mini-buffer,
--- a/rust_src/src/frames.rs
+++ b/rust_src/src/frames.rs
@@ -727,6 +727,17 @@ pub fn make_frame_invisible(frame: LispFrameLiveOrSelected, force: bool) {
     }
 }
 
+#[lisp_fn(min = "0")]
+pub fn last_nonminibuffer_frame() -> LispObject {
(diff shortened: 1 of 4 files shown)


rust_src/src/frames.rs
```diff
@@ -727,6 +727,33 @@ pub fn make_frame_invisible(frame: LispFrameLiveOrSelected, force: bool) {
     }
 }
 
+#[lisp_fn(min = "0")]
```
shaleh: No need for the min=, `lisp_fn` handles checking the number of arguments. Use `min` when a function can take less parameters than defined like say this function:
`pub fn takes_optional_arguments(first: LispObject, second: LispObject, third: LispObject)`

That should fail if not given 3 parameters. But `min=1` would let `second` and `third` be omitted and they will be sent as `Qnil`.

rust_src/src/frames.rs
```diff
@@ -727,6 +727,33 @@ pub fn make_frame_invisible(frame: LispFrameLiveOrSelected, force: bool) {
...
+pub fn last_nonminibuffer_frame() -> LispObject {
+    let frame = get_last_nonminibuffer_frame();
+
+    if frame.is_null() {
```
shaleh: I wonder if this exposes a bug/failure in the current code. The `impl` for LispObject could check for NULL and return Qnil. Otherwise do we silently transform the null into a bogus pointer?

rust_src/src/frames.rs
denis631: I see. Thanks ;)

rust_src/src/frames.rs
denis631: Yes. I assume you are right and current `from` implementation is wrong.

In theory the function, which I implemented should be a one-liner `get_last_nonminibuffer_frame().into()`, but since current `from` implementation doesn't work this way (idk why, maybe it's a bug), I couldn't do it like that.

I think of making a new PR, where I fix the `from` method and if it's gonna be merged I will polish this one. What do you think?

shaleh: > I think of making a new PR, where I fix the from method and if it's gonna be merged I will polish this one. What do you think?

this is a solid plan. Maybe there are reasons for the current implementation. Maybe not. If nothing else, we can improve the comments in the code.

denis631: TL;DR I propose altering my API instead and change `the last_non_minibuffer_frame` from `LispFrameRef` to `Option<LispFrameRef>`

The extern functions, which are exposed to C indeed can pass null reference hidden under LispFrameRef (e.g. frame.c line: 1838), but our Rust code should always point to the non-null frame (can we make a wrapper around this and make our Rust function receive Option<LispFrameRef>?)
i.e. in the setter, which is called from C, I will make an extra check, in order to check for nil
---
I looked at the code one more time and I think the reason why this was done this way is the assumption that if we have a `LispFrameRef` instance, then we assume that the ref is always pointing to a frame, meaning, it can not contain a pointer to `null`

According to types though, it can be point to null, since `LispFrameRef`is a type alias of `ExternalPtr<Lisp_Frame>` and `ExternalPtr` has an `is_null` method

But by checking the API, like for instance `LispObject::as_frame` returns an `Option<LispFrameRef>` and not a `LispFrameRef` which then points to null.

rust_src/src/frames.rs
```diff
@@ -727,6 +727,43 @@ pub fn make_frame_invisible(frame: LispFrameLiveOrSelected, force: bool) {
     }
 }
 
+#[lisp_fn]
+pub fn last_nonminibuffer_frame() -> LispObject {
+    unsafe {
+        last_non_minibuffer_frame.into()
+    }
+}
```
denis631: The conversion is guaranteed by
```
impl<T> From<Option<T>> for LispObject
where
    LispObject: From<T>,
{
    fn from(v: Option<T>) -> Self {
        match v {
            None => Qnil,
            Some(v) => v.into(),
        }
    }
}
```

which is defined in lisp.rs

src/frame.c
```diff
@@ -1850,17 +1832,18 @@ delete_frame (Lisp_Object frame, Lisp_Object force)
 
   /* If we've deleted the last_nonminibuf_frame, then try to find
      another one.  */
+  struct frame *last_nonminibuf_frame = get_last_nonminibuffer_frame();
   if (f == last_nonminibuf_frame)
     {
-      last_nonminibuf_frame = 0;
+      set_last_nonminibuffer_frame(0);
```
denis631: this is one example where the passed frame reference may be null

rust_src/src/frames.rs
```diff
@@ -727,6 +727,43 @@ pub fn make_frame_invisible(frame: LispFrameLiveOrSelected, force: bool) {
...
+pub extern "C" fn set_last_nonminibuffer_frame(frame: LispFrameRef) {
+    unsafe {
+        // frame reference may be null, since it is called from C
+        if frame.is_null() {
+            last_non_minibuffer_frame = None
+        } else {
+            last_non_minibuffer_frame = Some(frame)
+        }
+    }
+}
```
denis631: This is may feel a little weird, because optional ref is sort of `Option^2`, because `Option` may be `Some`/`None` and also reference may point to null (`None`) or a real frame (`Some`)

Should we add a doc comment above LispFrameRef declaration stating that LispFrameRef is guaranteed to point to a real frame. Guaranteed by developers, not by the type system. This is due to C interop and this will be fixed later?

shaleh: Why use an `Option` to define `last_non_minibuffer_frame`? If LispFrameRef is a pointer we can simply check it for NULL right?

rust_src/src/frames.rs
shaleh: Yeah, this is something none of us have been happy with. We should refactor the handling of LispFrameRef. That isn't a requirement for this PR though.

My recollection is there are edge cases where NULL is in fact possible where it acts like a None value. LispFrameRefs will be C pointers until we fully port them over to Rust, which will be a while.

shaleh:
This looks good. Barring any other concerns we can approve this when the comment is added as mentioned earlier. 

Solid work @denis631

rust_src/src/frames.rs
denis631: I think we should move away from Frame pointers and we indeed can do that, but for this, we should migrate all code that accesses frame to Rust, afaiu.

Issue #1493, opened 2019-05-13
Port invisible-p


brotzeit:
```c
DEFUN ("invisible-p", Finvisible_p, Sinvisible_p, 1, 1, 0,
       doc: /* Non-nil if text properties at POS cause text there to be currently invisible.
POS should be a marker or a buffer position; the value of the `invisible'
property at that position in the current buffer is examined.
POS can also be the actual value of the `invisible' text or overlay
property of the text of interest, in which case the value itself is
examined.

The non-nil value returned can be t for currently invisible text that is
entirely hidden on display, or some other non-nil, non-t value if the
text is replaced by an ellipsis.

Note that whether text with `invisible' property is actually hidden on
display may depend on `buffer-invisibility-spec', which see.  */)
  (Lisp_Object pos)
{
  Lisp_Object prop
    = (NATNUMP (pos) || MARKERP (pos)
       ? Fget_char_property (pos, Qinvisible, Qnil)
       : pos);
  int invis = TEXT_PROP_MEANS_INVISIBLE (prop);
  return (invis == 0 ? Qnil
	  : invis == 1 ? Qt
	  : make_number (invis));
}
```

ngortheone:
Hi @brotzeit , I will try to port this one. 
Since this will be my first I have a bunch of questions:

What should I do with `NATNUMP`, `MARKERP` and `TEXT_PROP_MEANS_INVISIBLE` macros?
Are there ported rust alternatives? 

I found this in `src/numbers.rs` that looks like something similar 
```rust
    pub fn is_natnum(self) -> bool {
        self.as_fixnum().map_or(false, |i| i >= 0)
    }
```

Same for `MARKERP`

src/marker.rs
```rust
    pub fn is_marker(self) -> bool {
...
```

I have not found anything that looks like `TEXT_PROP_MEANS_INVISIBLE`

ngortheone:
What file should this go to?

shaleh:
We tend to use the same file name as the original C. If you look for other functions ported from the same file it should be clear.

https://github.com/remacs/remacs/wiki/Porting-Cookbook is our attempt at helping new people to the code. Please point out where this is insufficient or out of date.

brotzeit:
The cookbook is out of date, but it should still be helpful.

ngortheone:
@brotzeit @shaleh 

I was trying to find answers in Porting Cookbook as well as other Wiki pages but here is what remains unclear to me:

[...]

PR #1552, opened 2019-11-16
Port invisible-p

jamesnvc:
<strike>Currently not fully functional, but still working on it.</strike>
This seems to be working now & I've added some explicit tests.
It's been a while since I've written Rust, so hopefully my code is reasonably idiomatic!
 Addresses #1493


rust_src/src/xdisp.rs
```diff
@@ -37,4 +41,83 @@ pub fn trace_redisplay(arg: Option<InteractiveNumericPrefix>) {
     }
 }
 
+fn invisible_prop(propval: LispObject, list: LispCons) -> EmacsInt {
+    for tem in list.iter_cars(LispConsEndChecks::off, LispConsCircularChecks::off) {
+        if propval.eq(tem) {
+            return 1;
+        }
+        if tem.is_cons() && propval.eq(tem.as_cons().unwrap().car()) {
+            if tem.as_cons().unwrap().cdr().is_nil() {
```
db48x: You might find this to be a bit more readable:
```
let (car, cdr) = tem.into();
if car.eq(propval) {
    return if cdr.is_nil() { 1 } else { 2 }
}
```

jamesnvc: Oh, that is quite a bit nicer, thank you!

rust_src/src/xdisp.rs
```diff
@@ -37,4 +41,81 @@ pub fn trace_redisplay(arg: Option<InteractiveNumericPrefix>) {
...
+        if prop.is_nil() {
+            0
+        } else {
+            1
+        }
+    } else {
+        match cur_buf.invisibility_spec_.as_cons() {
+            Some(cons) => invisible_prop(prop, cons),
+            None => 0,
+        }
```
jamesnvc:
```rust
cur_buf.invisibility_spec_.as_cons().map_or(0, |cons| invisible_prop(prop, cons))
```

rust_src/src/xdisp.rs
```diff
@@ -37,4 +41,81 @@ pub fn trace_redisplay(arg: Option<InteractiveNumericPrefix>) {
...
+                }
+                if tem.is_cons() {
+                    let (car, cdr) = tem.into();
+                    if propelt.eq(car) {
+                        return if cdr.is_nil() { 1 } else { 2 };
+                    }
+                }
+            }
+        }
+    }
```
jamesnvc: Factor out identical loop into lambda?

rust_src/src/xdisp.rs
```diff
@@ -3,10 +3,14 @@
 use remacs_macros::lisp_fn;
 
 use crate::lisp::LispObject;
```
shaleh: This belongs in the group with lists and remacs_sys below.

rust_src/src/xdisp.rs
```diff
@@ -37,4 +41,80 @@ pub fn trace_redisplay(arg: Option<InteractiveNumericPrefix>) {
...
+                return val;
+            }
+        }
+    }
+
+    0
+}
+
+// originally the macro TEXT_PROP_MEANS_INVISIBLE from intervals.h
+fn text_prop_means_invisible(prop: LispObject) -> EmacsInt {
```
shaleh: Is there a reason to keep the C version and this version?

jamesnvc: Yes, `TEXT_PROP_MEANS_INVISIBLE `is used in a bunch of other places, in xdisp.c and keyboard.c. Presumably this could be moved into intervals.rs, but I'm not sure if Rust is okay with that sort of circular dependency (since xdisp.rs will need to load it from intervals.rs, but the function calls `invisible_prop` from xdisp.rs)?

shaleh: What you could do is write a small Rust function which exposes this function to C using the expected C name. This way when there are no more callers the C version can be removed and only the Rust version remains.

Yes, Rust handles that kind of circular depends just fine. I wish Python did!

jamesnvc: Is it possible to have the Rust function called in place of the C macro, or will I need to change the C macro to just call a function that's declared `extern` & implemented in Rust?

db48x: Macros are resolved long before linking the program, so to change what a macro does you will need to change the definition of the macro :)

jamesnvc: 👍 cool, that's what I figured, but I wasn't sure if there was some sort of Deep Rust Magic that would let it replace a macro :grin:

shaleh: @db48x true, but in this case, does it _need_ to be a macro?

We can keep two copies of the logic, no worries there. It would be my preference to avoid that when it makes sense.

jamesnvc: Dumb question -- I've added an `intervals.rs` file in `rust_src/src`, but when I try to add `use crate::intervals::text_prop_means_invisible` to xdisp.rs, it can't find the crate. What do I need to do to add the new file to the project?

jamesnvc: ah, okay; figured out I need to add it to src/lib.rs

src/intervals.h
```diff
@@ -223,11 +223,9 @@ set_interval_plist (INTERVAL i, Lisp_Object plist)
 /* If PROP is the `invisible' property of a character,
    this is 1 if the character should be treated as invisible,
    and 2 if it is invisible but with an ellipsis.  */
-
+extern int text_prop_means_invisible(Lisp_Object); // defined in intervals.rs
```
shaleh: Do `int` and `EmacsInt` always agree? I seem to recall this is not the case across platforms.

jamesnvc: Oops; changed to be EMACS_INT. In this case, the return value is always 0, 1, or 2, but in any case, the types probably should match.

shaleh: We have heard complaints in the past from people compiling on ARM that when the types don't match Rust will refuse to compile. Even though logically the types are the same.

jamesnvc:
What are the next steps I should take to more this forward?

PR #1553, opened 2019-11-20
Update to nightly 2019-11-17

shaleh:
`#[allow(unions_with_drop_fields)]` is no longer needed.
`#![allow(clippy::missing_safety_doc)]` is now needed because Clippy expects all uses of `unsafe` to be documented.
`#![feature(ptr_cast)]` is no longer needed.
`[first, rest..]` is no longer valid syntax. An '@' is needed.


shaleh:
@nickdrozd I asked in the Rust Gitter channel, the needed change is an '@'. I updated the PR to reflect the much simpler change.

nickdrozd:
Even better.

shaleh:
Let's try it again....

agraven:
The build failure is complaining about an unneeded `ptr_cast` feature flag, but it's been removed from the branch's source tree. Does the CI need to be redone or what's going on?

shaleh:
Let's see how this one goes.

PR #1555, opened 2019-12-12
Impl Not for LispObject; cut is_nil and is_not_nil

nickdrozd:


shaleh:
I need to scan this line by line, but overall solid. Thanks for taking the effort.

rust_src/src/alloc.rs
```diff
@@ -57,7 +57,7 @@ pub fn bool_vector(args: &mut [LispObject]) -> LispObject {
     let vector = unsafe { make_uninit_bool_vector(args.len() as EmacsInt) };
 
     for (i, arg) in args.iter().enumerate() {
-        unsafe { bool_vector_set(vector, i as EmacsInt, arg.is_not_nil()) }
+        unsafe { bool_vector_set(vector, i as EmacsInt, !!*arg) }
```
harryfei: IMHO, `!!*arg` is so confusing and unreadable.

shaleh:
```
diff --git a/rust_src/src/alloc.rs b/rust_src/src/alloc.rs
index 9ce09f8aad..ff099d9ffa 100644
--- a/rust_src/src/alloc.rs
+++ b/rust_src/src/alloc.rs
@@ -57,7 +57,7 @@ pub fn bool_vector(args: &mut [LispObject]) -> LispObject {
     let vector = unsafe { make_uninit_bool_vector(args.len() as EmacsInt) };

     for (i, arg) in args.iter().enumerate() {
-        unsafe { bool_vector_set(vector, i as EmacsInt, !!*arg) }
+        unsafe { bool_vector_set(vector, i as EmacsInt, bool!(*arg)) }
     }

     vector
diff --git a/rust_src/src/eval_macros.rs b/rust_src/src/eval_macros.rs
index 8e09736d9d..f24d507dab 100644
--- a/rust_src/src/eval_macros.rs
+++ b/rust_src/src/eval_macros.rs
@@ -7,6 +7,10 @@
  * the unsafe is not used.
  */

+macro_rules! bool {
+    ($symbol:expr) => { !!$symbol }
+}
+
 /// Macro to generate an error with a list from any number of arguments.
 /// Replaces xsignal0, etc. in the C layer.
 ///
```

That is one alternative. Then we have code like:

```
if bool!(some_lisp_val) {
   // do something
}
```

nickdrozd:
@harryfei @brotzeit @agraven This change should address your complaint.

agraven:
I'd still prefer something akin to the macro shaleh proposed, while the automatic dereferencing is definitely an improvement, `!!` is still not a very intuitive symbol, I personally have to stop and consciously think about what it means every time I see it.

shaleh:
What do you say Nick? Is `bool!()` concise enough? Do we stick with `is_nil()`? I am not sold on the double bang having seen how this reads. But this is a group effort so my vote is not the one that matters. I know that '!!' is somewhat common in Javascript circles. So maybe this for me is generational.

nickdrozd:
I don't know, it's hard to get enthusiastic about any of these options. I just wish we could automatically coerce things to booleans in a boolean context...is that so much to ask? Why is it that coercion is fine with negation, but not otherwise?

shaleh:
I hear ya @nickdrozd. 1000% agreed.

db48x:
Yea, I feel like `if` should have a trait the way `for` does. Maybe we should write a proposal. Maybe there already is one? I haven't looked, but it seems like a fairly obvious one.

Issue #1208, opened 2019-01-04
Validate our usage of `transmute`


shaleh:
https://doc.rust-lang.org/std/mem/fn.transmute.html lays out cases where it should be used and alternatives to use instead. We should see if we can reduce our usage.

PR #1556, opened 2019-12-12
Remove most uses of transmute

carlosmn:
Almost every single transmute is used to cast between versions of `ExternalPtr<T>` and we can instead get away with casting the pointers.

I wondered about using some `From<ExternalPtr<U>> for ExternalPointer<T>` but these are behind a type check so I thought maybe it shouldn't be *that* easy to convert between them.

This handles most of #1208 although there are a couple of uses which aren't nearly as clear-cut which I've left out of this PR.

### DIFF
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -232,7 +232,8 @@ impl LispBufferRef {
     }
 
     pub fn markers(self) -> Option<LispMarkerRef> {
-        unsafe { (*self.text).markers.as_ref().map(|m| mem::transmute(m)) }
+        let markers = unsafe { (*self.text).markers };
+        LispMarkerRef::from_ptr(markers as *mut _)
     }
 
     pub fn mark_active(self) -> LispObject {
@@ -720,11 +721,11 @@ impl LispBufferRef {
     }
 
     pub fn overlays_before(self) -> Option<LispOverlayRef> {
-        unsafe { self.overlays_before.as_ref().map(|m| mem::transmute(m)) }
+        LispOverlayRef::from_ptr(self.overlays_before as *mut _)
     }
 
     pub fn overlays_after(self) -> Option<LispOverlayRef> {
-        unsafe { self.overlays_after.as_ref().map(|m| mem::transmute(m)) }
+        LispOverlayRef::from_ptr(self.overlays_after as *mut _)
     }
 
     pub fn as_live(self) -> Option<Self> {
@@ -954,7 +955,7 @@ impl From<LispObject> for Option<LispOverlayRef> {
 impl LispMiscRef {
     pub fn as_overlay(self) -> Option<LispOverlayRef> {
         if self.get_type() == Lisp_Misc_Type::Lisp_Misc_Overlay {
-            unsafe { Some(mem::transmute(self)) }
+            Some(LispOverlayRef::new(self.as_ptr() as *mut _))
         } else {
             None
         }
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -233,7 +233,7 @@ impl LispBufferRef {
 
     pub fn markers(self) -> Option<LispMarkerRef> {
         let markers = unsafe { (*self.text).markers };
-        LispMarkerRef::from_ptr(markers as *mut _)
+        ExternalPtr::from_ptr(markers.cast())
     }
 
     pub fn mark_active(self) -> LispObject {
@@ -721,11 +721,11 @@ impl LispBufferRef {
     }
 
     pub fn overlays_before(self) -> Option<LispOverlayRef> {
-        LispOverlayRef::from_ptr(self.overlays_before as *mut _)
+        ExternalPtr::from_ptr(self.overlays_before.cast())
     }
 
     pub fn overlays_after(self) -> Option<LispOverlayRef> {
-        LispOverlayRef::from_ptr(self.overlays_after as *mut _)
+        ExternalPtr::from_ptr(self.overlays_after.cast())
     }
 
     pub fn as_live(self) -> Option<Self> {
--- a/rust_src/src/buffers.rs
+++ b/rust_src/src/buffers.rs
@@ -955,7 +955,7 @@ impl From<LispObject> for Option<LispOverlayRef> {
 impl LispMiscRef {
     pub fn as_overlay(self) -> Option<LispOverlayRef> {
         if self.get_type() == Lisp_Misc_Type::Lisp_Misc_Overlay {
-            Some(LispOverlayRef::new(self.as_ptr() as *mut _))
+            unsafe { Some(mem::transmute(self)) }
         } else {
             None
         }
--- a/rust_src/src/fonts.rs
+++ b/rust_src/src/fonts.rs
@@ -1,10 +1,9 @@
 //! font support
 
+use std::ffi::CString;
 use std::ptr;
 
 use remacs_macros::lisp_fn;
-
-use std::{ffi::CString, mem};
 
 use crate::{
     data, fns,
@@ -161,7 +160,7 @@ impl From<LispObject> for Option<LispFontObjectRef> {
     fn from(o: LispObject) -> Self {
         o.as_vectorlike().and_then(|v| {
             if v.is_pseudovector(pvec_type::PVEC_FONT) && o.is_font_object() {
-                Some(unsafe { mem::transmute(o) })
+                Some(LispFontObjectRef::new(v.as_ptr() as *mut _))
             } else {
                 None
             }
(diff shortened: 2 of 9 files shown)


carlosmn:
Clippy is complaining about

```
error: casting from `*const remacs_sys::Lisp_Misc_Any` to a more-strictly-aligned pointer (`*mut remacs_sys::Lisp_Overlay`) (4 < 8 bytes)
   --> src/buffers.rs:958:38
    |
958 |             Some(LispOverlayRef::new(self.as_ptr() as *mut _))
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: #[deny(clippy::cast_ptr_alignment)] on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_ptr_alignment
```

which I think means that the `transmute` would be just as bad?

db48x:
Yes, that alignment mismatch is a problem.

`Lisp_Misc_Overlay` is just the first couple of fields for a bunch of types, enough to discriminate between them and for the garbage collector to work on them without having to know all the details. They'll all need to have the same alignment, which I suppose just happens to work in C. We'll need to tell bindgen to use an alignment of 8 when it generates bindings for that struct.

carlosmn:
The type itself is indeed 4 bytes and we have https://github.com/remacs/remacs/blob/1def345e4335b7a2efcd8e5e0bab62f0f9c1ee01/rust_src/src/lisp.rs#L798-L802 which asserts this.

I wonder if there are any systems in common use that do care about this.

carlosmn:
The original type in `lisp.h` is

```C
struct Lisp_Misc_Any            /* Supertype of all Misc types.  */
{
  ENUM_BF (Lisp_Misc_Type) type : 16;           /* = Lisp_Misc_??? */
  bool_bf gcmarkbit : 1;
  unsigned spacer : 15;
};
```

which is not something that reliably works even cross-compiler really, but apparently is good enough for us.

This makes bindgen tell rustc to align this to 4 bytes via the following generated code

```rust
#[repr(C)]
#[repr(align(4))]
#[derive(Debug, Copy, Clone)]
pub struct Lisp_Misc_Any {
    pub _bitfield_1: __BindgenBitfieldUnit<[u8; 4usize], u16>,
}
```

Removing the alignment attribute brings it down to 2, which is probably due to the `u16` in there. Manually setting the alignment to 8 makes clippy happy but I don't know how to tell bindgen to generate this or whether the alignment is related to the size of registers/pointers (`linux32 cargo  clippy` is happy but I've no idea how much rustc cares about this).

carlosmn:
Huh, the signbit test is failing in release mode but works in debug.

carlosmn:
It turns out that rust gives us a function to ask whether the sign bit is set (`f64::is_sign_negative`) so we can just use that instead of dealing with bits and IEEE representation ourselves.

I've also brought back the transmute where changing it to a cast gave us errors due to alignment in order to unblock the other changes.

rust_src/src/buffers.rs
```diff
@@ -232,7 +232,8 @@ impl LispBufferRef {
     }
 
     pub fn markers(self) -> Option<LispMarkerRef> {
-        unsafe { (*self.text).markers.as_ref().map(|m| mem::transmute(m)) }
+        let markers = unsafe { (*self.text).markers };
+        LispMarkerRef::from_ptr(markers as *mut _)
```
shaleh: Could we add a macro to reduce the noise required for these conversions? Bonus points for a way to discover the type.
```
macro_rules! SOME_GREAT_NAME {
    ($kind:expr, $symbol:expr) => {
        $kind::from_ptr($symbol as *mut _)
    }
}

SOME_GREAT_NAME!(LispMarkerRef, markers)
```

rust_src/src/buffers.rs
```diff
@@ -954,7 +955,7 @@ impl From<LispObject> for Option<LispOverlayRef> {
 impl LispMiscRef {
     pub fn as_overlay(self) -> Option<LispOverlayRef> {
         if self.get_type() == Lisp_Misc_Type::Lisp_Misc_Overlay {
-            unsafe { Some(mem::transmute(self)) }
+            Some(unsafe { mem::transmute(self) })
```
shaleh: Why is this one safe?

rust_src/src/fonts.rs
```diff
@@ -4,7 +4,7 @@ use std::ptr;
 
 use remacs_macros::lisp_fn;
 
-use std::{ffi::CString, mem};
+use std::ffi::CString;
```
shaleh: should be above 'ptr' ^^^

rust_src/src/fonts.rs
```diff
@@ -161,7 +161,7 @@ impl From<LispObject> for Option<LispFontObjectRef> {
     fn from(o: LispObject) -> Self {
         o.as_vectorlike().and_then(|v| {
             if v.is_pseudovector(pvec_type::PVEC_FONT) && o.is_font_object() {
-                Some(unsafe { mem::transmute(o) })
+                Some(LispFontObjectRef::new(v.as_ptr() as *mut _))
```
shaleh: Similar to my macro question/suggestion earlier. Is there a way to have "PtrRef from C pointer version"?

rust_src/src/libm.rs
```diff
@@ -14,8 +13,7 @@ mod sys {
 
 /// Return the sign bit of the float.
 pub fn signbit(x: f64) -> bool {
-    let bits: u64 = unsafe { mem::transmute(x) };
-    bits >> 63 != 0
+    x.is_sign_negative()
```
shaleh: Higher level language for the win!

shaleh:
@carlosmn this PR looks good. I'd love for us to find a way to reduce the visual clutter. My macro suggestion is one potential solution but I am open to others.

rust_src/src/buffers.rs
carlosmn: It's not, I just re-added the `unsafe` inside the `Some` as a stylistic choice when I added the commit explicitly bringing it back so the commit explains why, but it does look odd in the diff so I can put it back to how it was at the start.

rust_src/src/buffers.rs
carlosmn: rust 1.38 introduced a `cast` method for pointers which would make the `as *mut _` slightly nicer (the impl is basically just that but generic).

With that this would look like `LispMarkerRef::from_ptr(markers.cast())`. I suppose we can follow that name and have something like `markers.cast()` perform this itself as these are all `ExternalPtr<T>`

carlosmn: I managed to implement something like this for different `ExternalPtr<T>` which I think looks pretty neat, but this call and a few others are specifically arbitrary pointers which gets a bit trickier. I apparently cannot implement `impl<T> From<*mut c_void> for Option<ExternalPtr<T>>` because `Option` isn't ours (I think that's what the compilers is telling me).

I can wrap it and implement `impl<T> From<*mut c_void> for MaybeExternalPtr<T>` but now I have one type I don't really want. Maybe this will work in combination with a macro or maybe `Deref`?

carlosmn:
As this is getting a bit of a bigger issue, let's bring the discussion around reducing visual clutter to the top level.

With the addition of a `cast()` method  and a `From<*mut T> for ExternalPointer<T>` we can remove most of the type syntax as the compiler can just look at the function signatures.

But the ones where we return `Option<YadaYadaRef>` is a bit harder as it won't let me implement `From<*mut T> for ExternalPointer<T>`. I can work around the orphan rules by wrapping option, but at that point we have a new type we don't actually want.

We can let the compiler do more type-figuring work by using the raw pointer's `cast()` method and using `ExternalPtr::from_ptr()` which makes the compiler figure out which types we mean. This is better, but still more verbose than a type-less `transmute` that we started with.

The way that works is that it's a function so there is no type name to prefix it. We can write something like

```rust
pub fn external_ptr_from_ptr<T>(p: *mut T) -> Option<ExternalPtr<T>> {
    ExternalPtr::<T>::from_ptr(p.cast())
}
```

which is what a `From` would look like if we were allowed to implement it. I don't know if calling such a function is actually better than what we have now with `ExternalPtr::from_ptr(v.cast())`.

carlosmn:
One other thing we could do here is to create our own trait to work do the work a `From<*mut T> for Option<ExternalPtr<T>>` would do. That could enable something like `self.overlays_before.into_externalptr()` but there the gains seems significantly smaller as we're moving away from easily recognisable traits.

I also just realised that my `ExternalPtr<T>::cast() -> ExternalPtr<U>` should be doable in a `From` so we don't reinvent patterns.

carlosmn:
I've fixed up the change to where the ffi is in `fonts.rs` and I think this should be good enough for now.

This PR introduces a couple of changes that should make moving between types easier, or at least make the code less verbose with types, but that should be for another PR if we're to call this done at some point.

Issue #1549, opened 2019-11-03
I had to run autogen.sh twice


joonazan:
On the first run it installed a compiler and on the second run it made the configure file. This is not how the README says it works.

shaleh:
That is definitely odd. Can you tell us more about your system?

joonazan:
That was on a somewhat outdated Ubuntu on a Chromebook. I was not able to reproduce on my desktop Arch Linux.

shaleh:
Thank you for the update. I am inclined to close this if we cannot reproduce it. The autogen.sh script is made for us by the autotools suite. The only fix I can think of would be to better define a dependency inside configure.ac. But we would need to know which dependency needs help.

cnasc:
I just had the same thing happen to me. System details:
```
x86_64 Linux 5.3.15-138.current
AMD Ryzen 5 2400G with Radeon Vega Graphics @ 8x 3.6GHz
```

One key piece is that I am working from inside the docker image, not sure if @joonazan did the same.

Here's my actual session:
```
~/code/remacs $ sudo docker-compose exec remacs ./autogen.sh
Checking Rust toolchain install ...
info: syncing channel updates for 'nightly-2019-07-18-x86_64-unknown-linux-gnu'
info: latest update on 2019-07-18, rust version 1.38.0-nightly (bc2e84ca0 2019-07-17)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
Remacs currently requires Rust toolchain version nightly-2019-07-18.
```

I thought that at this point I would be good to run `configure`, but:

```
~/code/remacs $ sudo docker-compose exec remacs ./configure --enable-rust-debug
OCI runtime exec failed: exec failed: container_linux.go:346: starting container process caused "exec: \"./configure\": stat ./configure: no such file or directory": unknown
```

After verifying that the file actually didn't exist, I tried running `autogen` again and got the following:

```
~/code/remacs $ sudo docker-compose exec remacs ./autogen.sh
Checking Rust toolchain install ...
Your system has the required Rust toolchain installed for building Remacs.
Checking whether you have the necessary tools...
[...]

PR #1560, opened 2020-01-06
Detect when rustup upgrades the local Rust version.

shaleh:
autogen.sh was failing because the return value of rustup was not
understood which lead to an assertion.

Simplify the rustup and rust version checks by using `rustup toolchain list` which does not try to update the system.

Closes: #1549

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -189,7 +189,7 @@ if [ -n $rust_toolchain_vers_path ] ; then
     retval=$?
 
     case $retval in
-        0) echo "Your system has the required Rust toolchain installed for building Remacs." ;;
+        0|3) echo "Your system has the required Rust toolchain installed for building Remacs." ;;
         1) echo >&2 "Remacs currently requires Rust toolchain version $remacs_version."
 	   echo >&2 "Run 'rustup install $remacs_version'."
 	   exit 1 ;;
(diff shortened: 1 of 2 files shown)


autogen.sh
```diff
@@ -189,7 +189,7 @@ if [ -n $rust_toolchain_vers_path ] ; then
     retval=$?
 
     case $retval in
-        0) echo "Your system has the required Rust toolchain installed for building Remacs." ;;
+        0|3) echo "Your system has the required Rust toolchain installed for building Remacs." ;;
```
shaleh: I was a bit hasty with assuming 3 was safe. Looking deeper.

autogen.sh
```diff
@@ -149,32 +149,17 @@ echo "Checking Rust toolchain install ..."
...
-    fi
-
-    echo $rustup_active_version | grep 'directory override'  >/dev/null && return 2
-
-    return 3
+    # rustup show will automatically install things without asking and provides no control over this.
+    # Instead, let's use 'toolchain list' which is idempotent and makes no changes.
+    # When cargo begins to compile the rust-toolchain file will be honored so no further work is needed.
+    rustup toolchain list | grep -q $remacs_version
+    return $retval
```
agraven: `retval` doesn't seem to be set anywhere in the function body, is this intentional?

shaleh: carp. Nope, I lost it when I simplified the code.

PR #1561, opened 2020-01-12
Use nightly 2020-01-08

shaleh:
This is the most recent one with a working clippy.


rust_src/src/base64.rs
```diff
@@ -202,18 +202,18 @@ fn test_multibyte_base64_decode_1() {
 #[test]
 fn test_linewrap_base64_decode_1() {
     let input1 = "
```
shaleh: The change here is not obvious. The original code used Windows ^M newlines.

rust_src/build.rs
```diff
@@ -77,10 +77,7 @@ impl ModuleInfo {
             let tmp = path_as_str(mod_path.file_name()).to_string();
             let path = mod_path.join("mod.rs");
             if path.is_file() {
-                return Some(ModuleInfo {
-                    path: path,
-                    name: tmp,
-                });
+                return Some(ModuleInfo { path, name: tmp });
```
agraven: Might as well rename `tmp` to `name` as well so it has a semantically meaningful and can be shortened too, if that's not outside the scope of this PR

shaleh: That sounds reasonable, let me look at it.

PR #1566, opened 2020-02-26
Port keybindings and supports.

shaleh:
Move initially defined keybindings to Rust.

Use 'AsRef' Trait to make keybinding definition more hygienic.
Along with 'KeyChar' the need for casting is contained within
'initially_defined_key'.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5466,8 +5466,14 @@ fi || AC_MSG_ERROR(['src/epaths.h' could not be made.])
 
 
 CARGO_DEFAULT_FEATURES=""
+if test "$HAVE_DBUS" = "yes"; then
+    CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"use-dbus\", "
+fi
 if test "$HAVE_LIBXML2" = "yes"; then
     CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"use-xml2\", "
+fi
+if test "$USE_FILE_NOTIFY" = "yes"; then
+    CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"use-file-notify\", "
 fi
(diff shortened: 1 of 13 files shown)


configure.ac
```diff
@@ -5466,9 +5466,15 @@ fi || AC_MSG_ERROR(['src/epaths.h' could not be made.])
 
 
 CARGO_DEFAULT_FEATURES=""
```
shaleh: These were added to allow all of the keybindings to be ported.

rust_src/src/buffers.rs
```diff
@@ -1756,17 +1758,6 @@ pub extern "C" fn reset_buffer_local_variables(mut buffer: LispBufferRef, includ
     buffer.reset_local_variables(include_permanent)
 }
```
shaleh: This code was moved to the end to make it easier to find.

rust_src/src/buffers.rs
```diff
@@ -1982,4 +1973,23 @@ pub fn byte_char_debug_check(b: LispBufferRef, charpos: isize, bytepos: isize) {
     }
 }
```
shaleh: `KeyChar` here is saving us from casting between chars, u8, and EmacsInt.

rust_src/src/keyboard.rs
```diff
@@ -262,37 +264,6 @@ pub fn recursive_edit() {
     }
 }
```
shaleh: Solely a move for easier managing.

src/buffer.c
```diff
@@ -5278,12 +5278,14 @@ Functions running this hook are, `get-buffer-create',
   Fput (intern_c_string ("erase-buffer"), Qdisabled, Qt);
 }
 
+#ifdef IGNORE_RUST_PORT
```
shaleh: This is never defined. But it makes merges from emacs easier. No more missing functions.

PR #1567, opened 2020-02-26
Update Travis CI configuration.

shaleh:
Builds are not working. Let's resolve it.


shaleh:
I would like to see the GitHub Actions CI take over and we drop Travis. What needs to be done there is we need to skip the failing tests for now. Once we have completed the master refresh we can see if the problems go away or if we need to dig in deeper.

PR #1568, opened 2020-02-27
Transition to GitHub Actions for CI.

shaleh:
Add the following as CI running with GitHub Actions.

rustfmt is run first essentially as a quick smoke test. If the PR is poorly formatted we can stop quickly.

clippy is then run. This acts as a code sanity and compilation check for the actual tests will be run. 

Finally, each platform runs their tests. This only happens if both rustfmt and clippy succeed.


db48x:
Looks like the Clippy test is running autogen, configure, etc first. Is that necessary, or can it be skipped?

shaleh:
Clippy has to be able to build the project. Which requires make-docfile, globals.h, etc etc.

agraven:
Closing and re-opening to regenerate expired build logs

agraven:
The linux no gui builds fails with a few errors like
```
frame.c:79:1: error: no previous prototype for ‘fget_internal_border_width’ [-Werror=missing-prototypes]
 fget_internal_border_width(const struct frame *f)
```
which I guess means `frame.h` isn't being loaded properly, which is strange, because none of the other build have this issue. I'll look into it some more

db48x:
I see that `fget_internal_border_width` is inside an `#ifdef HAVE_WINDOW_SYSTEM`; perhaps the call to it should be similarly ifdef'd?

agraven:
It's specifically the set of functions from line 1555 to 1558. Strange. I'm gonna mess around on my local system to see if I can reproduce and fix it.

agraven:
Part of the issue seems to be that `-Wmissing-prototype` is elevated from warning to error in CI, but not otherwise.

db48x:
I think it's fine for it to be an error, as long as it is consistent between normal and CI builds. To fix the error, I'm pretty sure we can just put an ifdef around the definitions of these functions, just like there's one around the declarations.

agraven:
That's what I ended up doing and it worked. Now I just can't figure out why the mac build is failing. I've looked at the build log but don't see where the error is

PR #1569, opened 2020-03-08
Quantum leap to Emacs 27 branch.

shaleh:
Let's make the move to emacs-27. If I knew how to make this a smaller PR I would do it. Suggestions welcomed.

All of the C source is now the emacs-27 source. Instead of deleting code
all code is wrapped in 'ifdef IGNORE_RUST_PORT'. This should help in two
ways:
1. Easier merges since there are no deletions to conflict with.
1. Easier to track changes since the diffs should be smaller and will show us what changed.

My hope is these changes will make the port go more smoothly. Now we can easily see what parts of the source changed so we can compare them to  the Rust code. If no changes happened in the ifdef'ed code then there is likely not a lot to worry about.

I wrapped each function or code block individually in ifdefs on purpose. This will make it easier to read the code and see that a chunk is ported. Scrolling up and down a thousand lines or more to find the beginning of an ifdef was not something I wanted people to have to do.

Fallout:
 - reverts the removal of MSDOS. This was making merges too painful. I 100% support the idea here but the reality is not working out.
 - reverts some of the only GTK work for similar reasons. This is currently a bit messy.

Current status:
Does not build. I have made it to `eval.rs`, working alphabetically. There is more code to comment out.

TODO:
 - finish identifying ported code
 - walk the Rust code looking for places Emacs has deviated.
 - re-examine the Rust ported code for bugs, style, etc.
 - fix make-docfile to handle the ifdef change.
 - resolve expected issues with pdumper
 - resolve problems caused by Emacs supporting bignum. The number support in general needs updating as a result.
 - double check lisp code. There are surely hiccups here.


shaleh:
I had purposefully let this code get behind the emacs-27 branch. After I made the PR I ran `git merge emacs/emacs-27`. Totally painless. One conflict -- the versioning in configure.ac.

brotzeit:
Sounds good. I think keeping the C code is the only possible way.

harryfei:
How can we help you to complete this PR?

PR #1570, opened 2020-03-16
Replace clip_to_bounds with num::clamp

shaleh:


shaleh:
@db48x That was my thinking.

PR #1573, opened 2020-04-24
Port `make-frame-visible`.

yellowsquid:
Resurrect PR (#1526) and port `make-frame-visible`. Fixed the failing tests and made it more rusty.

### DIFF
--- a/rust_src/src/frame.rs
+++ b/rust_src/src/frame.rs
@@ -5,6 +5,7 @@ use libc::c_int;
 use remacs_macros::lisp_fn;
 
 use crate::{
+    buffers::LispBufferRef,
     lisp::{ExternalPtr, LispObject},
     lists::assq,
     lists::{LispConsCircularChecks, LispConsEndChecks},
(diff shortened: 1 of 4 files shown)


shaleh:
Looks good on first read. With CI being busted currently one of us needs to try it out locally.

rust_src/src/windows.rs
```diff
@@ -636,6 +642,23 @@ impl LispObject {
...
+}
+
+impl Iterator for LispWindowIter {
+    type Item = LispWindowRef;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.current {
+            None => None,
+            Some(window) => {
+                self.current = window.next.as_valid_window();
```
agraven, line 656: The window is not checked for validity in the original `make_frame_visible_1`, so you can just call `.as_window()`

rust_src/src/frame.rs
```diff
@@ -691,6 +695,36 @@ pub fn frame_face_alist(frame: LispFrameLiveOrSelected) -> LispObject {
     frame_ref.face_alist
 }
 
+/// Update the display_time slot of the buffers shown in WINDOW and all its descendants.
+fn make_frame_visible_1(window: LispWindowRef) {
```
agraven, line 699: This function should preferably be a closure inside that `make_frame_visible`

PR #1577, opened 2020-05-25
Modifies Dockerfiles for current Rust toolchain

polaris64:
This commit changes all Dockerfiles so that the nightly-2020-04-07 Rust toolchain is used since this was changed in 5b59c0f77113a1330853cff077a08f00c8f7fc88. Without this change, building via a Docker container fails due to the Rust toolchain incompatibility.

### DIFF
--- a/docker/fedora/Dockerfile
+++ b/docker/fedora/Dockerfile
@@ -25,6 +25,6 @@ ENV PATH "/root/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sb
 RUN curl https://sh.rustup.rs -o rustup.sh && \
     sh rustup.sh \
         --default-host x86_64-unknown-linux-gnu \
-        --default-toolchain nightly-2019-02-27 -y && \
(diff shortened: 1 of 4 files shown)


shaleh:
@polaris64 thanks for this.

How could we add a CI check to catch this in the future?

polaris64:
We could perhaps modify the Dockerfiles so that the `rust-toolchain` file is first copied into the image and then read and used in the calls to `rustup`?

shaleh:
mmmm

What if we had a check which grepped the Dockerfiles for the Rust version and compared it to the rust-toolchain file? That seems really simple and fast. A PR which updates the toolchain file without updating Docker would fail but any other PR would be ok.

shaleh:
Either way, this does not block your work @polaris64 I will merge it in.

polaris64:
No problem, glad to help :)

I don't know a great deal about CI systems so I'm probably not the best person to try modifying the configuration, but if you'd like me to make that other proposed change to the Dockerfiles then let me know and I'd be happy to create another PR.

PR #1579, opened 2020-06-24
Wrap lines in README.md

tarsius:
Because that makes it easier to read that document inside of Emacs instead of in a browser.


ZelphirKaltstahl:
In Emacs you can just display long lines wrapped. Is this really a valid concern for reading inside Emacs?

tarsius:
Yes one can do that but that tends to still result in lines that have to many characters for comfortable reading, forcing either slow reading or manual resizing of the window. It is especially weird if a file contains manually filled paragraphs as well as paragraphs that are placed on a single huge line.

db48x:
There are modes that can make reading long lines of prose as comfortable as you want, without needing to change the window size. Still, it's a good idea; we can just wrap it on some convenient column number so that nobody needs to bother with modes.

PR #1580, opened 2020-08-03
Rewrite multibyte_length with slice patterns

agraven:
Since advanced slice patterns have been stablized, I've been meaning to rewrite this function using them, because I have a very hard time keeping its logic in its current form straight in my head. I also added comments clarifying exactly which bits are being checked for what.

I seem to have made some kind of logic error though, because the build fails, and I can't figure out where the error is.

### DIFF
--- a/rust_src/src/multibyte.rs
+++ b/rust_src/src/multibyte.rs
@@ -854,31 +854,30 @@ pub unsafe extern "C" fn str_to_multibyte(
 /// Same as `MULTIBYTE_LENGTH` macro in C.
 #[allow(clippy::if_same_then_else)]
 fn multibyte_length(slice: &[c_uchar], allow_encoded_raw: bool) -> Option<usize> {
-    let len = slice.len();
-    if len < 1 {
-        None
-    } else if slice[0] & 0x80 == 0 {
-        Some(1)
-    } else if len < 2 || slice[1] & 0xC0 != 0x80 {
-        None
-    } else if !allow_encoded_raw && slice[0] & 0xFE == 0xC0 {
-        None
-    } else if slice[0] & 0xE0 == 0xC0 {
-        Some(2)
-    } else if len < 3 || slice[2] & 0xC0 != 0x80 {
-        None
-    } else if slice[0] & 0xF0 == 0xE0 {
-        Some(3)
-    } else if len < 4 || slice[3] & 0xC0 != 0x80 {
-        None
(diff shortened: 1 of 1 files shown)


rust_src/src/multibyte.rs
```diff
@@ -854,31 +854,30 @@ pub unsafe extern "C" fn str_to_multibyte(
...
-    } else if len < 5 || slice[4] & 0xC0 != 0x80 {
-        None
-    } else if slice[0] == 0xF8 && slice[1] & 0xF0 == 0x80 {
-        Some(5)
-    } else {
-        None
+    // The bits a byte "starts" with are the most significant ones in these comments
+    match slice {
+        // true if a starts with 0
+        &[a] if a & 0x80 == 0 => Some(1),
```
db48x, line 860: Don't all of the match patterns have to have wildcards at the end? I think this one should be `&[a, ..]`, for example.

agraven: That's true, the original code checks for failure conditions rather than success conditions and that has tripped me up a couple of times.

rust_src/src/multibyte.rs
```diff
@@ -854,31 +854,30 @@ pub unsafe extern "C" fn str_to_multibyte(
...
+        // true if a starts with 11110, b, c, and d start with 10
+        &[a, b, c, d, ..] if a & 0xF8 == 0xF0 && (b & 0xC0) | (c & 0xC0) | (d & 0xC0) == 0x80 => {
+            Some(4)
+        }
+        // true if a is 1111 1000, b starts with 1000, c, d, and e starts with 10
+        &[a, b, c, d, e, ..]
+            if a == 0xF8 && b & 0xF0 == 0x80 && (c & 0xC0) | (d & 0xC0) | (e & 0xC0) == 0x80 =>
+        {
+            Some(5)
+        }
```
FredTheDino, line 879: I do hate to just burst in like this, but you've changed the behavior here. This code won't return the same result for
```a=0xF8, b=0x80. c=0x00, d=0x00, e=0xC0```, your code will return length 5, and the old code length None.

I haven't verified this so feel free to ignore me, but I do have a hunch!

Happy coding! :D

agraven: No I think you're right, the old code required the b to be a continuation byte at an earlier point in the code, but since constraints don't cascade in this version I'll have to add the extra check. Thanks for catching this!

PR #1583, opened 2020-08-24
Port char-width

### LINKED ISSUE #1151: Port char-width
```c
DEFUN ("char-width", Fchar_width, Schar_width, 1, 1, 0,
       doc: /* Return width of CHAR when displayed in the current buffer.
The width is measured by how many columns it occupies on the screen.
Tab is taken to occupy `tab-width' columns.
usage: (char-width CHAR)  */)
  (Lisp_Object ch)
{
  int c;
  ptrdiff_t width;

  CHECK_CHARACTER (ch);
  c = XINT (ch);
  width = char_width (c, buffer_display_table ());
  return make_number (width);
}
```

A6GibKm:
Closes #1151.

The code works but to run I have to manually remove `char-width` from `bindings.rs` to avoid

```
error[E0428]: the name `char_width` is defined multiple times
     --> src/../generated/bindings.rs:28333:5
      |
28333 |     pub fn char_width(arg1: ::libc::c_int, arg2: *mut Lisp_Char_Table) -> isize;
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `char_width` redefined here
      | 
     ::: src/remacs_sys.rs:92:5
      |
92    |     pub fn char_width (c: libc::c_int, dp: *mut Lisp_Char_Table) -> ptrdiff_t;
      |     -------------------------------------------------------------------------- previous definition of the value `char_width` here
      |
      = note: `char_width` must be defined only once in the value namespace of this module
```

I could use the definition in the bindings but I am not sure on what is the difference between `ptrdiff_t` and `isize` and if the use of `c_int` is fine.

### DIFF
--- a/rust_src/src/character.rs
+++ b/rust_src/src/character.rs
@@ -3,6 +3,7 @@
 use libc::{c_uchar, ptrdiff_t};
 
 use remacs_macros::lisp_fn;
+use crate::remacs_sys::{char_width, buffer_display_table};
 
 use crate::{
     lisp::LispObject,
@@ -24,6 +25,18 @@ impl LispObject {
 /// Same as the `CHAR_HEAD_P` macro.
 pub const fn char_head_p(byte: c_uchar) -> bool {
     (byte & 0xC0) != 0x80
+}
+
+/// Return width of CHAR when displayed in the current buffer.
+/// The width is measured by how many columns it occupies on the screen.
+/// Tab is taken to occupy `tab-width' columns.
+/// usage: (char-width CHAR)
+#[lisp_fn(c_name = "char_width", name = "char-width")]
+pub fn char_width_lisp(ch: LispObject) -> EmacsInt {
+    // CHECK_CHARACTER(ch);
+    let c: EmacsInt = ch.into();
+    let width: ptrdiff_t = unsafe { char_width(c as i32, buffer_display_table()) };
+    width as EmacsInt
 }
 
 /// Decrement the buffer byte position `POS_BYTE` of the current buffer
--- a/rust_src/src/remacs_sys.rs
+++ b/rust_src/src/remacs_sys.rs
@@ -89,6 +89,7 @@ extern "C" {
         depth: EmacsInt,
         sexpflag: bool,
     ) -> LispObject;
+    pub fn char_width (c: libc::c_int, dp: *mut Lisp_Char_Table) -> ptrdiff_t;
     pub fn read_minibuf(
         map: Lisp_Object,
         initial: Lisp_Object,
--- a/src/character.c
+++ b/src/character.c
@@ -95,22 +95,6 @@ char_width (int c, struct Lisp_Char_Table *dp)
 }
 
 
-DEFUN ("char-width", Fchar_width, Schar_width, 1, 1, 0,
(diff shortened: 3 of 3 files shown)


rust_src/src/character.rs
```diff
@@ -26,6 +27,18 @@ pub const fn char_head_p(byte: c_uchar) -> bool {
     (byte & 0xC0) != 0x80
 }
 
+/// Return width of CHAR when displayed in the current buffer.
+/// The width is measured by how many columns it occupies on the screen.
+/// Tab is taken to occupy `tab-width' columns.
+/// usage: (char-width CHAR)
+#[lisp_fn(c_name = "char_width", name = "char-width")]
+pub fn char_width_lisp(ch: LispObject) -> EmacsInt {
```
db48x, line 35: The `lisp_fn` macro can check and convert arguments for us if you let it. Specify that the input type is an `EmacsInt` here and it will convert it from `LispObject` for you, and bail out with an error if someone passes in anything but a number.

rust_src/src/remacs_sys.rs
```diff
@@ -89,6 +89,7 @@ extern "C" {
         depth: EmacsInt,
         sexpflag: bool,
     ) -> LispObject;
+    pub fn char_width (c: libc::c_int, dp: *mut Lisp_Char_Table) -> ptrdiff_t;
```
db48x, line 92: Use the declaration from bindings.rs, since it is automatically generated and will thus keep up with any changes to the C function. It'll probably never change until we remove it, but that's a small amount of future labor saved.

db48x (changes requested):
This is good work, but it can be improved slightly

rust_src/src/lisp.rs
```diff
@@ -351,6 +351,19 @@ impl From<LispObject> for bool {
     }
 }
 
+impl From<LispObject> for char {
```
db48x, line 354: I just remembered: we already did this but not for the `char` type directly. We occasionally have to use values outside the normal unicode range, so we have a `Codepoint` type in multibyte.rs to use instead.

rust_src/src/character.rs
```diff
@@ -32,10 +32,9 @@ pub const fn char_head_p(byte: c_uchar) -> bool {
 /// Tab is taken to occupy `tab-width' columns.
 /// usage: (char-width CHAR)
 #[lisp_fn(c_name = "char_width", name = "char-width")]
-pub fn char_width_lisp(ch: EmacsInt) -> EmacsInt {
-    // CHECK_CHARACTER(ch);
-    let width: isize = unsafe { char_width(ch as i32, buffer_display_table()) };
-    width as EmacsInt
+pub fn char_width_lisp(ch: char) -> isize {
```
db48x, line 35: So instead you can use the `Codepoint` type for the input here.
```suggestion
pub fn char_width_lisp(ch: Codepoint) -> isize {
```

db48x (changes requested):

A6GibKm:
I am currently running `make check`. Will post when ready.

db48x (approved):
Now it's perfect!

PR #1584, opened 2020-08-24
fix deceptive readme

Phate6660:
Clang is required, so it should be stated as such.

Building remacs fails due to:

```
thread 'main' panicked at 'Unable to find libclang: "couldn\'t find any valid shared libraries matching: [\'libclang.so\', \'libclang-*.so\', \'libclang.so.*\'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"', /home/valley/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.46.0/src/lib.rs:1652:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ more Emacs-y.
    IMPORTANT: Whenever the toolchain updates, you have to reinstall
    rustfmt manually.
 
-2. You will need a C compiler and toolchain. On Linux, you can do
+2. You will need Clang and a toolchain. On Linux, you can do
    something like:
 
         apt install build-essential automake clang libclang-dev


Phate6660:
All right, I admit this wasn't the best way start a PR.

But still, it DOES require Clang due to bindgen, so it would be really nice if the requirement was stated as such.

I wouldn't've made this PR in the first place if I would've known beforehand that Clang was required. 
I wouldn't've bothered to even try using the software at all.

Phate6660:
Ok, can someone tell me what I have to do to get people to at least interact with this PR?
Do I need to change the wording of something? Do I sound too temperamental?
Seriously, this PR has been open since Aug 24th and not even a single comment or anything that shows that someone has even looked at this.

README.md
```diff
@@ -139,7 +139,7 @@ more Emacs-y.
    IMPORTANT: Whenever the toolchain updates, you have to reinstall
    rustfmt manually.
 
-2. You will need a C compiler and toolchain. On Linux, you can do
+2. You will need Clang and a toolchain. On Linux, you can do
```
db48x, line 142: Wouldn't it be better to say that it requires both GCC and Clang? It is possible to build Remacs entirely with Clang, but the instructions in this readme don't give all of the details needed to do that.

db48x (commented):
I must have missed this when you first posted it, so apparently all you needed to do to get a response was kvetch.

README.md
Phate6660: I agree, and thank you for responding. I'll edit the README again and push the commit, let me know what you think when I do.

And I apologize for the late response. I always happen to be away from my PC like right when people respond. :/

db48x: No need to apologize; this is an inherently asynchronous medium.

db48x (approved):

PR #1585, opened 2020-08-24
Port string

A6GibKm:


A6GibKm:
Fails with 
```
Loading international/mule-cmds...
Loading case-table...
Loading /var/home/deathwish/.opt/remacs/lisp/international/charprop.el (source)...
Loading international/characters...
Wrong type argument: stringp, 8262
make[1]: *** [Makefile:807: bootstrap-emacs] Error 255
```

A6GibKm:
Similar to `char_to_string`.

test/rust_src/src/fns-tests.el
```diff
@@ -85,6 +85,12 @@
     (should (equal [nil nil nil nil nil t t t t t] (vconcat (reverse A))))
     (should (equal A (reverse (reverse A))))))
 
+(ert-deftest test-reverse-bool-vector ()
```
db48x, line 88: This seems like an odd name for this test, and it duplicates an existing name :)

db48x (changes requested):
Mostly perfect!

test/rust_src/src/fns-tests.el
A6GibKm: Yes. I wonder what I was thinking when I came up with the name.

rust_src/src/character.rs
```diff
@@ -101,4 +101,20 @@ pub fn char_resolve_modifiers(character: LispObject) -> EmacsInt {
     char_resolve_modifier_mask(character.into())
 }
 
+/// Concatenate all the argument characters and make the result a string.
+/// usage: (string &rest CHARACTERS)
+#[lisp_fn]
+pub fn string(args: &[LispObject]) -> LispObject {
+    let n = args.len();
+    let mut buffer = vec![0_u8; n * MAX_MULTIBYTE_LENGTH];
```
tshepang, line 109:
```suggestion
    let mut buffer = vec![0_u8; args.len() * MAX_MULTIBYTE_LENGTH];
```
did this because I see nowhere else `n` is used

A6GibKm: `n` is used in `make_string_from_bytes`.

